PR #11, opened 2018-02-05
Respect protocol

khuezy:
I was encountering a problem w/ the nextjs app deployed on `https`. I did not provide the `SERVER_URL` in the `.env` file (I like to let the library handle that w/ the respective context of the environment).

When I tried to oauth using Google credentials, it would take me to the `http://...` protocol instead of the `https`.

Heroku is on `https`, and the redirect was taking me to the `http://.../callback`

This PR fixes the issue, but I'm not sure if it's the correct thing to do.

Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -263,7 +263,8 @@ module.exports = (nextApp, {
     }
 
     const token = uuid()
-    const url = (serverUrl || `http://${req.headers.host}`) + `${pathPrefix}/email/signin/${token}`
+    const http = req.secure ? 'https' : 'http'
+    const url = (serverUrl || `${http}://${req.headers.host}`) + `${pathPrefix}/email/signin/${token}`
 
     // Create verification token save it to database
     functions.find({ email: email })
--- a/index.js
+++ b/index.js
@@ -113,6 +113,14 @@ module.exports = (nextApp, {
     expressApp.use(lusca.csrf())
   }
 
+   // Respect protocol
+   expressApp.use(function(req, res, next) {
+    if (req.headers["x-forwarded-proto"] === "https") {
+      req.connection.encrypted = true;
+    }
+    next();
+  });
+
   /**
    * With sessions configured we need to configure Passport and trigger
    * passport.initialize() before we add any other routes.
--- a/index.js
+++ b/index.js
@@ -271,8 +271,7 @@ module.exports = (nextApp, {
     }
 
     const token = uuid()
-    const http = req.secure ? 'https' : 'http'
-    const url = (serverUrl || `${http}://${req.headers.host}`) + `${pathPrefix}/email/signin/${token}`
+    const url = (serverUrl || `${req.protocol}://${req.headers.host}`) + `${pathPrefix}/email/signin/${token}`
 
     // Create verification token save it to database
     functions.find({ email: email })


iaincollins:
Hey thanks!

It's a good idea to respect the the protocol for the email verification URL. I completely agree it's preferable to have little configuration that needs to be set as possible.

1. For Email

After looking at the [Express documentation](http://expressjs.com/en/api.html#req.protocol), I think we can even simplify it down to just changing line 266 to use `req.protocol`:

    const url = (serverUrl || `${req.protocol}://${req.headers.host}`) + `${pathPrefix}/email/signin/${token}` 

Apparently Express checks `X-Forwarded-Proto` and sets this automatically.

Please do let me know if you spot any problems with doing it that way.

2. For oAuth

I don't think this change actually impacts oAuth, as far I can see?

I think actually Google and other oAuth service just use whatever  protocol / hostname / port is configured as the callback URL in their control panels, so maybe it need changing there.

I use now.sh for an example site at https://nextjs-starter.now.sh and from looking at the requests it never drops down to HTTP as far as I can see, but maybe I'm wrong!

khuezy:
Thanks for the response!

Good call on `req.protocol`.

In my Google Credentials dashboard, I have the correct `https` callback, but it redirects to `http`. When I added the `X-Forwarded-Proto` middleware, it works as intended.

khuezy:
The example site works on `now.sh`, might be something up w/ Heroku?

By the way, I get an internal error when trying to delete an account.

iaincollins:
Was just looking at Express and X-Forward-Proto, I think I get it now, thanks!

It turns out for req.protocol / req.secure to work in express `expressApp.set('trust proxy', 1)` needs to be set, and basically does just what you've done.

I incorrectly assumed that was the default behaviour in Express (I happened across it while looking at the [Express Session documentation](https://github.com/expressjs/session)). Maybe we should that by default.

Thanks for letting me know about the error on account deletion! I'm seeing that too, I must have broken something on that site - will take a look!

khuezy:
Awesome, thanks!

Please refactor to `expressApp.set('trust proxy', 1)` if it's cleaner.

iaincollins:
Thanks for the patch! I might refactor to use `expressApp.set('trust proxy', 1)` in place of the if/else but only once I've actually tried it out and replicated to see it's a suitable replacement!

It overlaps a bit with some also wanting to have cookies be secure only by default, so will test both behaviours together (I'll give it a go on Heroku as well, and do it without setting SERVER_URL to shake it down).

iaincollins:
I've just rolled this out on NPM as 1.4.0 (with a small bit of refactoring and exposing it as an option).

If you have any issues with it please let me know!

khuezy:
Sweet! Thanks for the quick fix :)

iaincollins:
Your welcome! And thanks to the heads up about not being able to delete your account.

I traced it back to the example usage app having an out of date version of next-auth. It turns out this was because it had an old shrinkwrap file, because I'd accidentally published an NPM shrinkwrap file in next-auth, which was breaking the updating of shrinkwrap files in apps that used it.

I've fixed this in next-auth too now, as of v1.4.1. Glad you spotted that issue or I'd not had noticed that glaring bug.

PR #28, opened 2018-04-13
add csrf options

relifeted:
add csrf options

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,7 @@ const uuid = require('uuid/v4')
 module.exports = (nextApp, {
   bodyParser = true,
   csrf = true,
+  csrfOptions = {},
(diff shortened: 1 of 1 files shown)


relifeted:
actually, i am not using next-auth in my project, cause i need make a lot changes to fits requirements, i am using this to check CSRF in my project:

https://github.com/expressjs/csurf

it may fit with your design?

iaincollins:
Thank you for the feedback and yes that's super relevant!

I stil want to refactor this further in future so that it uses double submit cookie instead of requiring a session (and csurf looks like it might be a good fit) but for now I have refactored it as above (so it can be `true` (default), `false` (off) or an object with options).

PR #29, opened 2018-04-19
Respect strategyOptions.callbackUrl

joshuar500:
Checks to see if the user has added their own `callbackUrl` in `strategyOptions`.


iaincollins:
Thanks!

PR #33, opened 2018-06-13
fix broken link to .env.default because it has been renamed to .env.example

9oelM:
It's just a broken link that I'm trying to fix. That's all.


iaincollins:
Thanks!

Issue #47, opened 2018-09-05
Typescript Types?


NickBolles:
Are there any typescript typings or any plans to make them?

If not I might whip some up.

iaincollins:
There are not, but this would be very welcome.

If you could provide an example (even if it's just a set of instructions here!) of how to check they work and are correct to make maintain easier that would be great!

NickBolles:
Sounds good. I'll submit a PR an a bit. 

I don't know of any good way to check that they work with a JavaScript project besides manually looking over them. But they just need to be updated whenever a module, function or object signature changes.

falleco:
Hey! I tried to use the TS types you provided, but found some problems using it in my projects.
I created #71 to apply the changes I needed, and also provided a typescript example.

As I am no typescript pro, criticism is welcome.

iaincollins:
Being tracked in #220

lluia:
Yup just the last part of publishing the types missing, I'll try to give it a push 🙏🏻

PR #50, opened 2018-09-06
Add types

NickBolles:
Added a quick pass at typescript types. I'm not sure everything is 100% because I haven't used next-auth much yet, so let me know if anything looks off. 

Also added the sideEffects: false key, as [defined by Webpacke](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) to help treeshaking, especially for next client builds.


NickBolles:
Also closes #47

PR #51, opened 2018-09-07
case sensitive providerName in passport.authenticate

aol-nnov:
Strategy name in Passoport is not enforced to be all lower case (at least I haven't found that rule), and it's up to strategy author to pick a name.
So, there's a chance that strategy with mixed case name [exists](https://github.com/exlinc/keycloak-passport/blob/master/index.js#L23).
Without this change such strategies won't work with next-auth.

### DIFF
--- a/src/passport-strategies.js
+++ b/src/passport-strategies.js
@@ -277,11 +277,11 @@ module.exports = ({
     providerOptions
   }) => {
     // Route to start sign in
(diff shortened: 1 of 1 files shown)


iaincollins:
I was about to have to revert this to release but @sponte raised a PR which fixed the issue I was having with it - this is now live in 1.12.1. If you have any further problems with Keycloak support please let me know!

PR #52, opened 2018-09-27
Added options for bodyParser.json() and bodyParser.urlencoded()

jenssogaard:
By default bodyParser has a limit of 100kb which in some cases needs to be increased. I have added two options, one for both bodyParser.json and bodyParser.urlencoded. 

I have tested the implementation in a project where I do file upload and and everything passes. Scenarios:
- with default options
- with limit set and file uploaded is below limit
- with limit set and file uploaded is above limit

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,9 @@ const uuid = require('uuid/v4')
 
 module.exports = (nextApp, {
   bodyParser = true,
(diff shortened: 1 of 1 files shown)


iaincollins:
Oh awesome thank you! Merging now and will try and push out in a new version at the weekend.

(I need to update the docs as not options like this that have been added are not yet documented).

PR #58, opened 2018-11-12
Exposing params in the getProfile call

sponte:
passport-oauth2 calls back with extra parameters if specified - this is based on the number of arguments passed to the callback function when defining the strategy (https://github.com/jaredhanson/passport-oauth2/blob/master/lib/strategy.js). This change exposes that extra parameter. This is useful if users want to store id_token for instance which is available in the _params object.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,8 +202,8 @@ The example configuration provided is for Mongo DB. By defining the behaviour in
 #### Required
 
 * find({id,email,emailToken,provider})
-* insert(user, oAuthProfile)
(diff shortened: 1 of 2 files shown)


iaincollins:
Oh no, I still haven't merged this! Will take a look this week.

iaincollins:
Oh wow, passing `providerName` to passport.use() when passing the strategy fixed a problem I was having. Nice! Thank you. This is now live in 1.12.1.

PR #62, opened 2018-12-04
Release 1.12.1; improve compatibility, fix session save bug; update dependancies

iaincollins:
* `sessionResave` now defaults to `true`, which ensures sessions always rotate properly.

This can be disabled for special use cases - it does not work well with some Express Session Stores, but for most people it should be fine and should rotate sessions correctly. If set to `false` user sessions are likely to expire prematurely. It is strongly recommended you do not change this from the default setting.

As a side effect, this will case all sessions - including anonymous sessions created when users have connected but not logged in yet - to be saved as a session in the store. This is because the current Cross Site Request Forgery option associates a token with a session in the browser.

If you want to avoid creating anonymous sessions,  set `csrf` to `false` to disable CSRF protection;  sessions will then only be created when a user signs in. A future update should include the option to implement CSRF using the Double Submit Cookie method so it can be enabled without causing this side effect.

* All dependancies updated, including the example to include latest release of Next.js and React.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,8 +202,8 @@ The example configuration provided is for Mongo DB. By defining the behaviour in
 #### Required
 
 * find({id,email,emailToken,provider})
-* insert(user, oAuthProfile)
-* update(user, oAuthProfile)
+* insert(user, oAuthProfile, providerParams)
+* update(user, oAuthProfile, providerParams)
 * remove(id)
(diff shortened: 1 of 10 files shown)


iaincollins (commented):
The Git commit history is a mess. Oops.

PR #63, opened 2018-12-14
Include cookie name

ghost:
Support for an optional cookie name, since the default `connect.sid` from [ `express-session`](https://www.npmjs.com/package/express-session) may not be recommended or loved by everyone


iaincollins:
Ooh neato, good idea - thanks for this!

I'm hoping to give next-auth some attention this week before the holidays.

PR #64, opened 2018-12-15
`trustedIdentity` flag in provider config

### LINKED ISSUE #61: Unable to login through passport strategy if local account exists.
Assume we have a local account for `user@server.com` email. All attempts to sign in using external identity providers (through a passport strategy) that have `user@server.com` as email in profile info are rejected because of this line:

https://github.com/iaincollins/next-auth/blob/61eedb26b0d447ddba497f8c141947878c981f52/src/passport-strategies.js#L237

It's probably reasonable (according to the comment before this line), but IMHO instead of just rejecting successful sign-in in this case, it's better to redirect to another page with a prompt to enter user's passport to confirm the intent.

If it makes sense, I would try to prepare a PR to resolve it this way (or probably the other way you might suggest) since it's a blocker for me.

drudv:
It should resolve #61

### DIFF
--- a/src/passport-strategies.js
+++ b/src/passport-strategies.js
@@ -76,6 +76,37 @@ module.exports = ({
         // profile object with just { id, name, email } properties.
         let profile = getProfile(_profile)
 
+        const linkToExistingAccount = user => {
+          const nextUser = Object.assign({}, user)
+
+          // If we don't already have a name for the user, use value the
+          // name value specfied in their profile on the remote service.
+          nextUser.name = user.name || profile.name
+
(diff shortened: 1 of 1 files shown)


iaincollins:
Oh wow this looks 💯 - thank you!

As I still haven't rigged up automated testing - which is super embarrassing at this point I'm going to merge but not update the published NPM version with this, or another recent PR, till I've done that; hopefully later this week.

I'd like to be sure the flows are right and would like to take a look at updating the examples and documentation, and maybe including some default sign in views to simplify usage, as a bridge to an even simpler version.

PR #84, opened 2019-09-24
Update libraries.

### LINKED ISSUE #83: Importing next-auth/client gets regeneratorRuntime is not defined
import NextAuth from "next-auth/client";    

I've tried adding in import "regenerator-runtime/runtime"; 

 Tried .babelrc

{
  "presets": ["next/babel"],
  "plugins": [
    [
      "@babel/plugin-transform-runtime",
      {
        "regenerator": true
      }
    ]
  ]
}

xanderim:
should fix #83


iaincollins:
Hi @xanderim 

Thanks for the PR! Sorry it took so long to get back to you.

Things were on pause for a while but [version 2.0](#99) should be out very soon.

I'm going to close this PR, as 2.0 is a complete re-write and the fix no longer applies (and lack of test suite in 1.0 means I can test open PRs easily).

This has been super useful though as - thanks to feedback like this - the new version avoids the pitfall of a custom babel config that this required.

PR #110, opened 2020-05-10
feat(client): Add optional provider for using useSession with context api

Fumler:
Users might want to use the `useSession()` hook in several components, and with the current hook implementation that would mean doing one request to the api per use. This change intends to add an alternate hook which uses context instead.

So you would still have the `useSession()` for a simple "go and get session" thing, but then also one which requires you to wrap your components in a provider but allows you to use `useGlobalSession()` anywhere without doing multiple requests.

Not sure if this is something you want but figured I'd at least make a PR :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,23 +71,17 @@ export default () => {
 
 *This is all the code you need to add support for signing in to a project!*
 
-If you want to avoid multiple requests being made from the `useSession()` hook being used in several components, there is also `useGlobalSession()` in combination with `<NextAuth.SessionProvider>` which uses [React Context](https://reactjs.org/docs/context.html)
+Using `useSession()` like this will work perfectly fine, but keep in mind this will do a network request to fetch the session in each component you use it. If you want to only do one network request you can wrap your component tree in our `Provider`, which will make `useSession()` use [React Context](https://reactjs.org/docs/context.html) instead.
 
+You can use this `Provider` on a specific page if you want, or for all pages by overriding your `_app.js` file in Next.js [as documented here](https://nextjs.org/docs/advanced-features/custom-app)
+Example `pages/_app.js`:
 ```javascript
-const Welcome = () => {
-  const [session, loading] = NextAuth.useGlobalSession()
-
-  return <>
-    {loading && <p>Checking session…</p>}
-    {!loading && session && <p>Welcome {session.user.name || session.user.email}.</p>}
-    {!loading && !session && <p><a href="/api/auth/signin">Sign in here</p>}
-  </>
-}
-
-export default () => {
-  <NextAuth.SessionProvider>
-    <Welcome />
-  </NextAuth.SessionProvider>
+export default ({ Component, pageProps }) => {
+  return (
+    <NextAuth.Provider>
+      <Component {...pageProps} />
+    </NextAuth.Provider>
+  )
 }
 ```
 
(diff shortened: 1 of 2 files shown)


iaincollins:
Oh awesome, this is great, thank you!

I really wanted to add support for context , but wasn't sure of the best way to do expose it. This looks much simpler than what I was expecting to have to need to do, which is excellent.

I'm happy to merge but before I do, as the names are exposed as the API wanted to run past a few things with you to get your view on them:

1. What do you think of calling the provider `NextAuth.Provider` vs `NextAuth.SessionProvider`?

I am asking as am really not sure what good convention is here. I've seen one or two examples with `.Provider` but I am not sure if that is a common pattern and which is better. Only asking now to avoid any confusion caused by changing it later. :-)

2. Do you think there is any way we could merge `useGlobalSession` and `useSession`?

I ask it seems like it would make for a simpler API to just provide a single hook, but I am not sure if that's even possible without passing in an extra param - can it detect how it is being called? 

Again, I'm not sure what is best practice convention wise here; I haven't looked at many external examples.

Would appreciate your thoughts! Thanks again for the PR!

iaincollins:
PS: Sorry for any bike shedding! Only asking questions about names, etc as keeping the API surface simple seems like it's a high value thing to do and I genuinely am not sure the simplest sensible way to do this.

For some background, I was thinking of just having `useSession()` behave like the NextAuth v1 client and use localStorage for caching. The provider approach is obviously much more standard and is nicer, so I'd MUCH rather do it this way!

I would welcome thoughts on if providing a HoC would be valuable. I'm aware that using `_app.js` inhibits some of the Next.js optimisations so had wanted to try and avoid steering folks in that direction, but I don't know if HoC cause any problems / if they were still a good idea (do people prefer provider pattern?).

Fumler:
> What do you think of calling the provider NextAuth.Provider vs NextAuth.SessionProvider?
This seems fine to me, I can't think of any special conventions here so shouldn't be a problem

> Do you think there is any way we could merge useGlobalSession and useSession?
I'm not sure if it's possible to detect how it's being called, but can maybe check somehow if the context provider exists. I'll take a look.

> I'm aware that using _app.js inhibits some of the Next.js optimisations
I don't think this is the case anymore. I think I remember at some point they said having `_app.js` would disable some features, but it seems like that has changed to only happen if your `_app.js` file is using `getInitialProps` -> https://nextjs.org/docs/advanced-features/custom-app

iaincollins:
> sure if it's possible to detect how it's being called, but can maybe check somehow if the context provider exists. I'll take a look.

Oh thanks! 

> I don't think this is the case anymore. I think I remember at some point they said having _app.js would disable some features, but it seems like that has changed to only happen if your _app.js file is using getInitialProps 

Oh, awesome! I'm glad to be wrong about that, thanks for the info.

Fumler:
Was easy enough to have one hook do both things depending on if you have a `Provider` or not. Can you check the code again @iaincollins and let me know if this is better?

iaincollins:
Oh that is brilliant, thank you!

This is perfect, will merge it in now and try updating the example with it.

It occurs to me I should probably have made the method signature`useSession(session, options)` instead of `useSession(session, options)`, as it's unlikely most folk will ever want to pass options (and probably most folk won't care about passing server side session object over frankly), but if there are options it's likely they will want to set more than one.

If you have any ideas on that, like should options perhaps be provider props, would welcome your thoughts.

iaincollins:
Update: I've just pushed this to `next-auth@beta` (beta 16)

I've not had a chance to try the provider out yet, but looking forward to that!

PR #111, opened 2020-05-10
Added Mixer, Discord, Slack and Reddit (partially)

LoriKarikari:
## Description
I've added Mixer, Discord, Slack, and Reddit providers.
The OAuth flow for Reddit is working but I can't get the profile.
It keeps throwing 401 unauthorized and I'm not really sure why yet.
Might look at it later or you can give it a shot (it's probably something small).

I also added a random value to the state parameter because Reddit doesn't work without one... This can be a temporary solution for now. While writing this it crosses my mind that I was maybe also able to add the state parameter to the config object itself. You can edit it however you like.

## Type of change
 New feature (non-breaking change which adds functionality)


iaincollins:
Oh wow, this looks amazing!

Good catch on the state parameter. I hadn't added that as I don't fully understand how it is used by providers and wasn't sure if it's a thing that might need to be tracked for some of them.

It's not currently documented, but an `options.secret` value is currently auto-generated if not specified and is used as a salt for hashes (etc). We could pass in the 'options' param to the function there (or even just the secret) if need? If this works though then even better!

Wonder what is up with that Reddit profile call. Maybe it wants an extra parameter or something in the header…

I'm not sure why rebasing is disabled for me on this PR, but I can squash and merge this in as if you want, or I can leave it for a bit if you'd like to keep adding stuff to it. I don't expect I'll be touching the oAuth stuff or the providers (as, other than the issue you've addressed, I think it's working okay) so happy either way!

LoriKarikari:
You can merge it if you want it's up to you. I'll add some more providers the coming week and take a look at that Reddit profile error. Maybe you can leave this one open until I've added more providers?

iaincollins:
I've sent you an invite to be a maintainer so you can merge them in when you want without having to jump through hoops if you like! 😀

No preferences on other providers - Facebook would be nice maybe? - these are great! I didn't know there was Slack oAuth integration.

iaincollins:
^ Sorry about introducing conflicts, I think running `npm run lint` should magically resolve them.

If not, am happy to take the versions in your repo and discard changes currently in my next-auth-2 branch; it looks like it's just conflicting over formatting (and only actual meaningful code changes are in your branch).

This was slightly annoying of me to do (sorry) but I've found using the linter really helpful, as it found a load of unnecessary promise cruft I'd written in the adapter, and was able to remove about 50 lines from a ~250 line file.

iaincollins:
Oh I just realised with the GitHub UI I can amend changes in your commit (somehow, magically) to fix the issues. :D If you don't mind, I would love to merge this in ASAP, even if not complete!

Is that okay with you?

I have other changes I would like to make in providers, and I would hate for it to get out of date :-)

LoriKarikari:
Hi, didn't have time to continue with this yet. You can merge it. I'll add more providers and  some docs  soon 😊

iaincollins:
Awesome, thank you!

PR #121, opened 2020-05-17
Update docs and website dir structure

iaincollins:
# Changes

* Now has 'www' directory at root level for the website (was 'docs').
* The 'docs' directory now only contains Markdown docs.
* Docusarus config looks in '../docs' for the docs.

This is deployed with now.sh to https://next-auth-docs.now.sh

@LoriKarikari Have added the new provider docs you made to the nav menu!

@dvanoni Thanks for doing this and for the research into options, this looks like a great setup!  

Appreciate being able to use `next-auth-docs.now.sh`, so thanks for that too. I've used Vercel + now.sh to deploy and works great.

I tried setting up a team, but yeah can see it will cost me per user too, and I don't think we need that, as nice as it would be, so just used my personal account for now.

PS: Happy to consider moving this repo to a GitHub org at some point if we want more flexibility around permissions.


iaincollins:
Update: This worked fine locally, but links seem to not work fine deployed to https://next-auth-docs.now.sh now so I think I messed something up, going to investigate :-)

(I thought it just wasn't working on the preview for some other reason.)

iaincollins:
I have had to move the `docs` folder back to `www/docs`.

Having it in a separate folder works fine locally, but with now.sh when you deploy a subdirectory it seems like it doesn't check out any other files outside that directory (that's a shame but I can live with that).

The docs are deployed just fine to https://next-auth-docs-igdjx6er7.now.sh

The `next-auth-docs.now.sh` domain is showing the old deployment (e.g. https://next-auth-docs.now.sh/getting-started doesn't work), I've checked the dashboard and the settings are right; I think that's an issue with Vercel. Maybe it will fix itself.

@ndom91 RE: `next-auth.js.org` - Just had a look I think we need to make the DNS for it point at `alias.zeit.co` to work with the way now.sh works. :-)

ndom91:
Ah yeah that makes sense. If we point it at alias.zeit.co we still have to setup 'next-auth.js.org' as a custom domain in the now.sh dashboard I think. 

Otherwise, good stuff. Glad to see it working :)

PR #122, opened 2020-05-17
Add providers

LoriKarikari:
Okay, I'm adding some more providers. I already pushed and will update it with commits so everyone can stay up-to-date or if you want certain providers added I can merge it.
I've currently added Facebook but wasn't able to test it locally because off the callback URL problem. I logged the profile response and was able to get the user's info so I think it should work. @iaincollins maybe you can test this on the example site?

### DIFF
--- a/src/providers/facebook.js
+++ b/src/providers/facebook.js
@@ -1,1 +1,21 @@
-// @TODO Facebook provider
+export default (options) => {
+  return {
+    id: 'facebook',
+    name: 'Facebook',
+    type: 'oauth',
+    version: '2.0',
+    scope: 'email',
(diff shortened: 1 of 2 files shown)


iaincollins:
@LoriKarikari Thanks will do!

(No idea what that fake build error is about, will see if I can make the bot not do that.)

iaincollins:
@LoriKarikari This looks great. Will push it out as a new beta and (hopefully later tonight) will try it out on the example site and will post back how I got on!

iaincollins:
@LoriKarikari Facebook worked first time :D http://next-auth-example.now.sh

PR #129, opened 2020-05-18
Docs - Add search

ndom91:
Added an offline search component. It rebuilds the index on each build. 

The other option built into docusaurus is Algolia, but you'd need to register an account, etc. 

Oh and in the `docusaurus.config.js`, I currently commented it out, but theres a config option you can uncomment that adds a type of announcement bar along the top for time sensitive stuff, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,4 +116,4 @@ export default ({ Component, pageProps }) => {
 }
 ```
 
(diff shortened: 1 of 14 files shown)


iaincollins (approved):
Awesome! Thanks for the info about the announcement bar, that might be useful!

PR #132, opened 2020-05-18
Add session expiry logic

iaincollins:
* By default, sessions are 30 day 'rolling sessions' and the timestamp for when they expire is extended when they are accessed to keep them alive.
* When sessions expire (ie after 30 days of inactivity), session object returns empty (as if there is no session) and users must sign in in again.
* Cleaning up old sessions from the database is not currently handled by the default adapter, but I do intend to add some logic to do this (added @TODO).
* The session expiry date can be changed by passing a custom updateSession() callback handler function in the options to the default adapter.

Using a custom `updateSession()` method with the default adapter, it is possible to specify other behaviour:

e.g.

* Disable rolling sessions (e.g. force a new login every X days).
* Create a session expiry date far into the future on initial sign in, so that they effectively never expire.
* Set a decently long max expiry time (e.g. 90+ days) but only actually update the session expiry time if the current expiry time is < 30 days; so that sessions stay valid for 30 days (and at most 90 days of inactivity) so that idle sessions are valid for at least 30 days (and maybe longer) but you don't need to write to your session database as often (useful if slow/expensive).

Note: Adapter options are passed as second option to the default adapter (the first option being the DB connection details). This is probably confusing and might be a design mistake.

```js
const adapter = Adapter.Default({ /* database object * /}, {
  updateSession: async (session, isNewSession) => {
    // 1st arg is the current session (or null) so it's easy to check current
    // expiry date, get user specific info, etc.
    // 2nd arg is true if this is a brand new session.
    //
    // Function should return an ISO date (e.g. toISOString) or false/null to
    // prevent an update from being applied; but should always return a session
    // if isNewSession is set or the sign in will fail.
  }
})
```

Relying on on Adapter options is a little obtuse / confusing and so I'm considering it an 'advanced option' right now. In future, we might change how session expiry dates and behaviour is set to make it easier.

Note: There are some other updates in this PR, that's just from the linter and some improvements to formatting of contributing guide.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,13 +8,13 @@ Please see the [Code of Conduct](CODE_OF_CONDUCT.md) and follow any templates co
 
 ## Pull Requests
 
-* Pull Requests should be raised for larger changes.
-* Ideally, lots of small changes should be bundled up in to a larger PR.
-* The latest changes are always in `master`
-* Merging (and pushing merge commits to `master`) is disabled in this repo
-* Rebasing is prefered to keep a clean commit history.
-* Pushing directly to master should be reserved for minor updates / one off fixes.
-* Pull Requests are great for batching up changes and making them easy to review and for tagging people so they can track changes, but they do not require approval from other commiters (commiters may raise and merge their own requests).
+* The latest changes are always in `master` 
+* Pull Requests should be raised for larger changes
+* Pull Requests do not need approval before merging for those with contributor access (it's just helpful to have them to track changes)
+* Rebasing in Pull Requests is prefered to keep a clean commit history (see below)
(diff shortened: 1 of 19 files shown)


iaincollins:
I've pushed this change out as `next-auth@2.0.0-beta.37`.

It includes the new Facebook profile from @LoriKarikari too, will test it on the example site later!

iaincollins:
BTW, as there are no Acceptance Tests for this yet, this is an example of what you see in a session (and in the cookie) with the new version:

Screenshot 2020-05-18 at 12 44 57

I went with a default of 30 days for expiry for now.

Wanted to get this feature in and a similar (shorter) expiry to verification emails as they are blocking features, and can come back and make it easier to configure later.

Definitely want to revisit this, to make it easier to customise in future. :-) It is still configurable in this release, just awkward to configure and I want to make it easier.

PR #133, opened 2020-05-19
Refactor adapter logic and add email verification 

iaincollins:
There is a lot more going on in this PR than I would like to do at once, but it's all connected.

## Changes in this PR

* Refactored adapter, with less redundant logic
* Removed logic from models
* Added email verification expiry support (defaults to 24 hours)
* Refactored session expiry handling and unified it with how email expiry works
* Default session expiry is still 30 days
* Now only updates expiry for a session at most once every 24 hours by default, to reduce writes to database
* Email verification max age, session max age and how often sessions are updated (to reduce database writes) are all simple options now
* Invalid sessionTokens are now deleted from the client
* Email verfication messages are now deleted once used (or when expired)
* Debug output is now an option (set `debug: true` to enable)
* Removed confusing options / callback from default adapter (except for passing in custom models/schemas)
* Adapter can now access all next-auth options, to make configuration easier

## New options

After implementing session expiry (and looking back at v1 for inspiration) it was clearer how both it and email expiry could be flexible and easier to customise.

How to use the new options, along with the default values that get set for them, is shown below.

```js
const options = {
  site: process.env.SITE,
  providers: [ /* providers */ ],
  adapter: Adapters.Default({ /* adapter options */ }),
  sessionMaxAge: 30*60*60*1000, // Sessions expire after 30 days of being idle
  sessionUpdateAge: 24*60*60*1000, // Sessions updated only if session is greater than this
  verificationMaxAge: 24*60*60*1000, // Verification (email sign in) links expire after 24 hours
  debug: false // Set to true to enable debug messages to be displayed
}

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "2.0.0-beta.39",
+  "version": "2.0.0-beta.40",
   "description": "An authentication library for Next.js",
   "repository": "https://github.com/iaincollins/next-auth.git",
   "author": "Iain Collins <me@iaincollins.com>",
--- a/src/adapters/typeorm/index.js
+++ b/src/adapters/typeorm/index.js
@@ -5,23 +5,8 @@ import { createHash } from 'crypto'
 import { CreateUserError } from '../../lib/errors'
 import Models from './models'
 
-const Adapter = (config, options) => {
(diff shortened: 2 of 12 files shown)


iaincollins:
Published to NPM as `next-auth@2.0.0-beta.40`

PR #134, opened 2020-05-19
Fix parameter params

mebrein:


iaincollins (approved):

iaincollins:
Thanks! I changed this in the app and probably forgot to mention it to anyone.

(I will probably rename this option again in future as appreciate it's not obvious why it's named that.)

mebrein:
Thanks Iain, I am working on an integration with IdentityServer 4. And I just logged in for the first time with this identity provider. Awesome!!!!

PR #142, opened 2020-05-21
Customise website theme and homepage

iaincollins:
I love this documentation site! <3

The example server usage on the homepage is something I'm thinking of switching to the syntax for the default way to configure the database (i.e. just set a database string or object and don't have to think about 'Adapters' except for if you are actually doing your own database support).

I'm also thinking of adding in JWT sessions, where sessions are not stored in a database at all in the longer term, which should make things even simpler for some people, but I think that is further down the road.

The reason why adapters were created in the first place was to allow people to avoid including code in production they would not need - e.g. if they wanted to use something other than TypeORM - but the size for the function is fairly small (about 2.8 MB) so perhaps this is not a major concern. I'm going to think it over though and might do something in between.

I'll update with some actual documentation soon. :-)

### DIFF
--- a/www/docusaurus.config.js
+++ b/www/docusaurus.config.js
@@ -1,6 +1,6 @@
 module.exports = {
   title: 'NextAuth',
-  tagline: 'Serverless Authentication, written from scratch for Next.js',
+  tagline: 'Serverless authentication for Next.js',
   url: 'https://next-auth.js.org',
   baseUrl: '/',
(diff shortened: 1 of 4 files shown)


LoriKarikari:
+1 for adding passwordless 😂😊

iaincollins:
I've not forgotten about renaming it in the providers either, just holding off a bit on renaming it in code while I think of implications of names for things!

PR #149, opened 2020-05-22
Update document site

iaincollins:
This is a work in progress, but wanted to give folks a heads up!

* Branding is **NextAuth.js** to reduce confusion with other software.
NB: Similar to Next.js, it remains `NextAuth` in React (and next-auth in URLs/file systems.

* Have started to update docs, but still in progress!
* Have updated project example, is much simpler now we have a documentation site. :D

Will start smashing bugs this weekend, after have worked on the docs for a bit more!


ndom91:
Oh man this looks so much better! I love the npm / github icons in the menu bar as well. I looked into doing this but couldnt find an easy way to do it. Great job ;)

LoriKarikari (approved):
Looks great! Nice work! But the authentications page are we keeping that one? I thought you wanted to avoid having to add configurations for all social providers?

ndom91 (approved):

iaincollins:
> Oh man this looks so much better! I love the npm / github icons in the menu bar as well. I looked into doing this but couldnt find an easy way to do it. Great job ;)

Thanks for creating the site! This is really great and so much easier to update!

I go the icons from looking round other Docusaurus sites!

> Looks great! Nice work! But the authentications page are we keeping that one? I thought you wanted to avoid having to add configurations for all social providers?

Oh totally, I really like the way you did it! I should remove that page.

Thank you both for taking a look! Will follow up with smaller updates.

PR #154, opened 2020-05-23
changed options to params

LoriKarikari:
Updated `options` to `params` in some provider's config.


iaincollins:
@LoriKarikari I'm sorry I broke this :-( Thank you for fixing it! I renamed it to match the arguments taken by an oAuth library that I've been using and I forgot about it when I committed the fix. Thanks also for cleaning up all those other tickets!

LoriKarikari:
@iaincollins haha no worries!

PR #155, opened 2020-05-24
Fix docs typos: directory is called “pages”

lachlanjc:


LoriKarikari:
Thanks for the fix @lachlanjc!

PR #164, opened 2020-05-26
Add custom pages

iaincollins:
Now supports 'pages' option, which can be any URL.

If specified, these replace the built in pages.

Example usage:

```javascript
pages: {
  signin: 'https://example.com/signin',
  signout: 'https://example.com/signout',
  checkEmail: 'https://example.com/check-email',
  error: 'https://example.com/error'
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "2.0.0-beta.47",
+  "version": "2.0.0-beta.48",
   "description": "An authentication library for Next.js",
   "repository": "https://github.com/iaincollins/next-auth.git",
   "author": "Iain Collins <me@iaincollins.com>",
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -176,6 +176,7 @@ export default async (req, res, userSuppliedOptions) => {
       sessionUpdateAge: 24 * 60 * 60 * 1000, // Sessions updated only if session is greater than this value (0 = always, 24*60*60*1000 = every 24 hours)
       verificationMaxAge: 24 * 60 * 60 * 1000, // Email/passwordless links expire after 24 hours
       debug: false, // Enable debug messages to be displayed
+      pages: {}, // Custom pages (e.g. sign in, sign out, errors)
       // Custom options override defaults
       ...userSuppliedOptions,
       // These computed settings can values in userSuppliedOptions but override them
@@ -197,6 +198,12 @@ export default async (req, res, userSuppliedOptions) => {
     // Get / Set callback URL based on query param / cookie + validation
     options.callbackUrl = await callbackUrlHandler(req, res, options)
 
+    const redirect = (redirectUrl) => {
+      res.status(302).setHeader('Location', redirectUrl)
+      res.end()
+      return done()
+    }
+
     if (req.method === 'GET') {
       switch (action) {
         case 'providers':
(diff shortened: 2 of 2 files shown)


iaincollins:
Published as `next-auth@2.0.0-beta.48`

LoriKarikari:
This is awesome!

ndom91:
Hmm so I'm trying to set this up to get a feel for it, but I cant seem to get access to any of the provider variables like you do in your default signin page. 

I have the `pages` object in the `[...slug].js` config pointing signin to a specific page. That seems to work. However, the signin page I land at then gets passed no props. Is that how its supposed to work?

iaincollins:
Agree, we need to an example to document this and provide an example.

Created #165 to track documenting pages.

Basically if you had a rest API call to `/api/auth/providers` it will return all configured providers and you could loop over that to display a list of sign in providers and the URLs for sign in buttons to use for them.

v1 had a `NextAuth.providers()` function in the client to make this really easy, and we should probably add something like that (basically identical to the `NextAuth.session()` method, but calls `/providers` instead of `/session`).

PS: If you spot any bugs on this / if it doesn't work as expect would appreciate feedback!

ndom91:
Okay, thanks for the hint. Got it working ;)

I'll put up my example

iaincollins:
BTW I think I've realised a possible future bug (ie not a bug now, but made me think about something that might be come an issue).

The sign up page should take a `?callbackURL=/page/user/was/on` parameter if passed, so it can pass that same option on to the sign up buttons, so that the user is returned to the page they were on before they signed in.

This works right now but is a little messy, so might come back to it.

I am thinking of making a `NextAuth.signin()` and `NextAuth.signout()` functions to handle redirecting to whatever the sign in / sign out pages are, in the client, which could save the currently URL automatically before redirecting.

It could even take an argument, like `NextAuth.signin('google')` so it could be a simple onClick event for a sign in button to make them easy to create.

(This would be totally optional to use, but I assume people quite like an easy way to make pages that would ensure the current URL would be preserved after signing in.)

ndom91:
Ooo yeah I great idea. I actually had this issue in another project thats currently using nextauth v1 still. I ended up writing a cookie I think and checking that or local storage or something convoluted like that haha.

ehrhart:
+1 for `NextAuth.signin('provider')` to redirect to the provider page directly.

Also I think it would be great if we could have a way of signing out without being redirected to the signout page. Now I understand it is currently required because we need a form with a CSRF token, but maybe we could retrieve `csrfToken` (perhaps with a hook like `NextAuth.useSession()`?) so we can build our own form and signout button anywhere (eg. in a navbar).

What do you think?

iaincollins:
Absolutely! I was just thinking about getting round to this over lunch today actually.

While it needs to be a form submission - because to do it securely, it should have a CSRF token as you point out - this can absolutely be done by a `NextAuth.signout()` function!

I've been holding off on this until I had also created a `NextAuth.csrfToken()` method which it could use to fetch the token, though you are right we could also put it in the session object; but thought it might be useful to also expose - e.g. for people making their own email sign up forms.

I expect to add `signin()`, `signout()` and `csrfToken()` functions to the beta later this week. I don't think I'm tracking this is an issue, but I should be.

israelidanny:
It's actually not very clear how to override the custom pages. I'm figuring out that I'm missing the CSRF and I guess it's in the cookies, however - am I expected to manually fetch it from cookies?

israelidanny:
Made it work. Ended up having to request CSRF on the client:
```typescript
  const [csrfToken, setCsrfToken] = useState();

  useEffect(() => {
    async function fetchToken() {
      if (typeof window !== "undefined") {
        const res = await fetch(`/api/auth/csrf`);
        const csrf = await res.json();
        setCsrfToken(csrf.csrfToken);
      }
    }
    fetchToken();
  }, []);
```

And then setting it inside the form:
```typescript
<input type="hidden" value={csrfToken} name="csrfToken" />
```

iaincollins:
Ah yes sorry that isn't documented yet (getting the CSRF Token to do a POST submission for email) but that is exactly right!

It's not intentionally obtuse, I'm just planning to add a `NextAuth.csrfToken()` and maybe actually just a `signin(email)` method similar to version 1, over the weekend and to document it then.

So this won't break but will get easier soon!

PR #168, opened 2020-05-26
Add db uri support

iaincollins:
Can now specify a database as a URI / connection string!

Additional options can be passed as query parameters.

Examples:

```
database: 'mysql://username:password@127.0.0.1:3306/database_name?synchronize=true',
```

```
database: 'postgres://username:password@127.0.0.1:5432/database_name?synchronize=true',
```

```
database: 'mongodb://username:password@127.0.0.1:27017/database_name?synchronize=true',
```

This is not a breaking change as passing an object still works, this just makes database configuration much easier for most use cases.

This matches the behaviour in the documentation - I wanted to make it this simple from the outset (which is is how it's described in the documentation), but took a little while to get to this point.

@TODO Update documentation to include these examples (and update the example site to use this approach too).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "2.0.0-beta.48",
+  "version": "2.0.0-beta.49",
   "description": "An authentication library for Next.js",
   "repository": "https://github.com/iaincollins/next-auth.git",
   "author": "Iain Collins <me@iaincollins.com>",
--- a/src/adapters/typeorm/index.js
+++ b/src/adapters/typeorm/index.js
@@ -6,6 +6,33 @@ import { CreateUserError } from '../../lib/errors'
 import Models from './models'
 
 const Adapter = (config, options = {}) => {
+  // If input is URL string, automatically convert to DB object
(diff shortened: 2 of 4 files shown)


iaincollins:
Published as `next-auth@2.0.0-beta.49`

LoriKarikari:
This already worked? I was connecting to Postgres with a string. Or did you break it 😂

iaincollins:
LOL really? (!) It was supposed to work that way, but I wasn't able to get it to work.

I think it was maybe working by accident! 😂

I'd be interested to know what your config looked like!

LoriKarikari:
It was a few versions ago I don't remember which one exactly. I started using SQLite for quick testing. It was the same as above but without `synchronize=true` and it worked perfectly lol.

LoriKarikari:
That's also why I added it to the docs.

iaincollins:
It is actually supposed to be a feature of TypeORM according to some docs, as are using environment variables, but I could get them to work either. 🤷‍♂️

It's possible I was looking at old docs though - ran into that quite a bit with TypeORM, especially when it came to adding MongoDB support.

PR #170, opened 2020-05-26
Improve database URI handling

iaincollins:
* Fix bug in parser (.query -> search)
* Comments to explain what is going on
* Fallback to TypeORM parser

See comments on https://github.com/iaincollins/next-auth/commit/1e9053d879f1145d470099837d4e90b9b558d4b6


iaincollins:
Published as `2.0.0-beta.50`

PR #171, opened 2020-05-26
Add docs for custom signin page

ndom91:


iaincollins:
Oh awesome, thanks!

I'll try and get round to adding the `NextAuth.providers()` method but this gives us some breathing room, it's great have this actually documented :D

Will leave a couple of inline-suggestions!

www/docs/configuration.md
```diff
@@ -97,4 +97,55 @@ If you want to customize, extend or replace the models, you can do this by using
...
+```jsx title="/pages/auth/signin"
+import React from 'react'
+
+const SignIn = ({ providers }) => {
+  return (
+    <div>
+      {providers && Object.values(providers).map(provider => (
+        <p key={provider.name}>
+          <a href={provider.signinUrl}>
+            <Button id='signin-btn' type='submit' appearance='primary' block>Sign in with {provider.name}</Button>
```
iaincollins, line 126:
```suggestion
            <Button type='submit' appearance='primary' block>Sign in with {provider.name}</Button>
```

Should we drop the ID (or make it `className=`) so the buttons don't all have the same ID?

iaincollins (approved):

www/docs/configuration.md
ndom91: Oh yeah my bad, missed deleting that from copying it out of my project :sweat_smile:

Issue #169, opened 2020-05-26
Added Sign in with apple


geraldnolan:
![image](https://user-images.githubusercontent.com/17180783/82893143-dc460d80-9f50-11ea-89c6-a1be75803afb.png)

I've added **sign in with apple**. Before I create a pull request, would you like this for v2?  
*_The oauth flow is a little bit different, and I wanted to run the changes by you._

A new type was created "**oauth-apple**"  

```
export default (options) => {
    return {
      id: 'apple',
      name: 'Apple',
      type: 'oauth-apple',
      version: '2.0',
      scope: 'name email',
      params: { grant_type: 'authorization_code' },
      accessTokenUrl: 'https://appleid.apple.com/auth/token',
      authorizationUrl: 'https://appleid.apple.com/auth/authorize?response_type=code&id_token&response_mode=form_post',
      profileUrl: null,
      profile: (profile) => {
        return {
          id: profile.sub,
          name: profile.sub,
          email: profile.email
        }
      },
      ...options
    }
  }
```

To get the user profile, you ask the server for a token, and it providers a jwt id_token to decode.  I had to add jwt-decode as a dependency. 

Also, you need to add response_mode=form_post if you pass in scope parameters according to apple documentation. The callback has to be a POST and not a GET

```
case 'callback':
            if (provider && options.providers[provider]) {
              callback(req, res, options, done)
            } else {
              res.status(400).end(`Error: HTTP POST is not supported for ${url}`)
              return done()
            }
            break
```

The secret has to be encoded as a jwt using your AuthKey_Q6YSW63G.k8 file downloaded from apple developer portal. 

**AppleClientSecret.js** would have to be added to the Developer project to create the jwt secret at runtime. Allowing the dev to have more control on the expiration/settings of the jwt.

The **profileUrl** is set to null.  They do not have an endpoint to get the profile.

Additional documentation would need to be made..

iaincollins:
Oh, outstanding! Yes please, I would love a PR for that!

I appreciate the oAuth code is messy right now. I had to patch in a hacky work around some issues in the `node-oauth` when a change to the Twitch API broke Twitch provider (about a week after we created it! 🤦‍♂️), and I haven't got round to refactoring it to clean it up since.

PR #174, opened 2020-05-27
Sign in with apple

geraldnolan:
This PR changes:

- Adds ability to sign in with apple  #169 

To test you must secure your local development server. (**next-auth-example**).  Apple doesn't allow for localhost in domains or subdomains.

I refactored the code and removed the option "oauth-apple"

How to:
[https://medium.com/@anMagpie/secure-your-local-development-server-with-https-next-js-81ac6b8b3d68](https://medium.com/@anMagpie/secure-your-local-development-server-with-https-next-js-81ac6b8b3d68)

Edit your **hostfile** and create a fake domain: 

127.0.0.1 dev.iaincollins.io

This documentation could be simplified but for the most part I followed: (Creating Apple Services IDs)
[https://developer.okta.com/blog/2019/06/04/what-the-heck-is-sign-in-with-apple](https://developer.okta.com/blog/2019/06/04/what-the-heck-is-sign-in-with-apple)

The tricky part after that is creating the secret.  The two things that tripped me up were:

**TeamID** and the **KeyID**.

The TeamID is located on the top right after logging in.
The KeyID is located after you create the Key look for before you download the k8 file. 

```
let privateKey=`-----BEGIN PRIVATE KEY-----
-----END PRIVATE KEY-----`;

let exp = Math.floor(Date.now() / 1000) + ( 86400 * 180 ); // Make it expire within 6 months
  let keyid = "INSERT_KEYID"; //KEYID

  const claims = {
    iss: "INSERT_TEAMID", //TeamID
    exp,
    sub: process.env.APPLE_ID
  };
  // Sign the claims using the private key
  var clientSecret = jwt.sign(claims, privateKey, {
    algorithm: 'ES256',
    keyid
  });

 console.log(clientSecret);

```

Let me know if you have any questions

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "2.0.0-beta.44",
+  "version": "2.0.0-beta.50",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -4436,6 +4436,11 @@
         "array-includes": "^3.0.3",
         "object.assign": "^4.1.0"
       }
+    },
+    "jwt-decode": {
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/jwt-decode/-/jwt-decode-2.2.0.tgz",
+      "integrity": "sha1-fYa9VmefWM5qhHBKZX3TkruoGnk="
     },
     "kind-of": {
       "version": "6.0.3",
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   ],
   "license": "ISC",
   "dependencies": {
+    "jwt-decode": "^2.2.0",
     "nodemailer": "^6.4.6",
     "oauth": "^0.9.15",
     "preact": "^10.4.1",
--- a/src/providers/apple.js
+++ b/src/providers/apple.js
@@ -2,7 +2,7 @@ export default (options) => {
     return {
       id: 'apple',
       name: 'Apple',
-      type: 'oauth-apple',
+      type: 'oauth',
       version: '2.0',
       scope: 'name email',
       params: { grant_type: 'authorization_code' },
--- a/src/providers/apple.js
+++ b/src/providers/apple.js
@@ -1,0 +1,22 @@
+export default (options) => {
+    return {
+      id: 'apple',
+      name: 'Apple',
+      type: 'oauth-apple',
+      version: '2.0',
+      scope: 'name email',
+      params: { grant_type: 'authorization_code' },
+      accessTokenUrl: 'https://appleid.apple.com/auth/token',
+      authorizationUrl: 'https://appleid.apple.com/auth/authorize?response_type=code&id_token&response_mode=form_post',
+      profileUrl: null,
+      profile: (profile) => {
+        return {
+          id: profile.sub,
+          name: profile.name == null ? profile.sub : profile.name,
+          email: profile.email
+        }
+      },
+      ...options
+    }
+  }
+  
--- a/src/providers/apple.js
+++ b/src/providers/apple.js
@@ -1,22 +1,21 @@
 export default (options) => {
-    return {
-      id: 'apple',
-      name: 'Apple',
-      type: 'oauth',
-      version: '2.0',
-      scope: 'name email',
-      params: { grant_type: 'authorization_code' },
-      accessTokenUrl: 'https://appleid.apple.com/auth/token',
-      authorizationUrl: 'https://appleid.apple.com/auth/authorize?response_type=code&id_token&response_mode=form_post',
-      profileUrl: null,
-      profile: (profile) => {
-        return {
-          id: profile.sub,
-          name: profile.name == null ? profile.sub : profile.name,
-          email: profile.email
-        }
-      },
-      ...options
-    }
+  return {
+    id: 'apple',
+    name: 'Apple',
+    type: 'oauth',
+    version: '2.0',
+    scope: 'name email',
+    params: { grant_type: 'authorization_code' },
+    accessTokenUrl: 'https://appleid.apple.com/auth/token',
+    authorizationUrl: 'https://appleid.apple.com/auth/authorize?response_type=code&id_token&response_mode=form_post',
+    profileUrl: null,
+    profile: (profile) => {
+      return {
+        id: profile.sub,
+        name: profile.name == null ? profile.sub : profile.name,
+        email: profile.email
+      }
+    },
+    ...options
   }
-  
+}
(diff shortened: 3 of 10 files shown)


iaincollins:
@geraldnolan Oh wow, this looks amazing!

I'm taking a small break and won't be able to try it out myself Thursday or Friday but this looks great, and I can't see anything I'd do differently. 

Thank for for wading through that sketchy oAuth code to do this, the existing code really needs refactoring as it has been hacked around a but and isn't the easiest thing to work with (without dumping console.log statements all over the place) 🙈.

Thank you for the detailed instructions! I'll let you know how I get on!

@LoriKarikari Reading this, am thinking of changing my mind and that maybe having a more detailed document page per provider *is* a good idea for *some* providers after all? I think the table is perfect for 99% of cases, but looking at this case, I'm grateful for the extra info!

LoriKarikari:
@iaincollins yeah something like this needs detailed instructions for sure. But the other 99% is just register your app, get your tokens and add your callback url 😂 so i wouldn't document those

iaincollins:
Thanks again @geraldnolan !

_tl;dr I can't test this end to end right now, but I'm going to merge it in because it looks good as far as I can tell!_

## Testing

* I followed the documentation and those links and the extra info you provided was great!
* Unfortunately, my ADC account has expired, so I can't create an application in ADC
* However, I checked out the branch and I do see the expected behaviour when I use Safari, even though the journey cannot complete (because I didn't provide valid credentials to the provider).

  *It seems you need a valid Client ID to even this prompt in other browsers as Apple fallback to displaying a webpage that checks it; only Safari has a built in native dialog).*

Screenshot 2020-05-27 at 14 02 22
Screenshot 2020-05-27 at 14 02 37

* I also tested out the changes don't cause any other problems with other providers.

  I think the approach of hardcoding some logic here specific to Apple is fine in this context.

  We can always refactor it a bit `src/server/lib/oauth/callback.js` is cleaned up in future. The comments there are very helpful. Including `jsw-decode` is good too, as I am going to pull in a lib for that anyway when I come to add JWT token support, so I'm sure it won't just be for the Apple provider.

I'm going to merge this in now, and will update the documentation for this provider, and re-use the information you have provided here when I do that.

## Questions

* Do you still specify `clientId` and `clientSecret` when using the Apple provider?
* `clientId` should be the value in ADC for `client_id` - is that correct?
* `clientSecret` only needs to be generated once ever, using the example code - is that correct?
* What is `process.env.APPLE_ID` in that context (is that an email address?)

Thank you!

iaincollins (approved):
Approved: See comments and questions in conversation of #174 (above)

geraldnolan:
@iaincollins I thought you were taking a break? :smiley:

**Do you still specify clientId and clientSecret when using the Apple provider?**

Yes, In the case I generated secret outside the app and placed in the env file.  We may want more sophistication and help the user generate the clientSecret.  I didn't put that code in **next-auth** cause I thought the end user wanted some control?

```
Providers.Apple({
      clientSecret: process.env.APPLE_SECRET
    }),
```

**clientId should be the value in ADC for client_id - is that correct?**

From the example appove it should be: (env_file) 

```
TWITTER_ID=
TWITTER_SECRET=
GITHUB_ID=
GITHUB_SECRET=
TWITCH_ID=
TWITCH_SECRET=
APPLE_ID=dev.iaincollins.io
APPLE_SECRET=**GENERATED_JWT_TOKEN_FROM_CODE_ABOVE**
```

** its the servicesid from the developer portal

**clientSecret only needs to be generated once ever, using the example code - is that correct?**

The clientSecret will need to generated once the token expires.  I set it to 6 months. (Maximum time for Apple)

**What is process.env.APPLE_ID in that context (is that an email address?)**
from env file: dev.iaincollins.io

Hopefully I answered those in ways that makes sense. Shoot me another message if I need to clarify anything else.

iaincollins:
Thank you, that's very clear!

I just have one follow up question!

RE: Secret generation

I'm considering writing a little command line function to come with the library, to help with things like setting up database schemas the first time and checking the current schema, and was thinking the code snippet you provided might be useful to in that context

e.g. so folks do something like this:

 `npx next-auth secret apple {APPLE_ID} {KEY_ID} {TEAM_ID}`

But that lead me to wonder, do you think it would be appropriate to consider generate the secret automatically at run time instead, to make this easier?

e.g. The behaviour could be refactored to allow `clientSecret` to be an object (other than a string) and running a callback on it if defined.

e.g. it could look like this in the front end:

```javascript
Providers.Apple({
  clientSecret: { 
  }
}),
```

And in `src/providers/apple.js` it could have a pre-defined clientSecret function like this:

```javascript
  ...
  clientSecret: ({appleId, keyId, teamId, privateKey}) => {
      const claims = {
        iss: teamId,
        exp: Math.floor(Date.now() / 1000) + ( 86400 * 180 ), // 6 months
        sub: appleId
      }
    
      return jwt.sign(claims, privateKey, {
        algorithm: 'ES256',
        keyid
      })
  })
```

* Alternatively, client secret in the client could look something like this, and the function in `src/providers/apple.js` could be called something else, like `clientSecretCallback`:

```javascript
Providers.Apple({
  clientSecret: { 
    appleId: process.env.APPLE_ID,
    privateKey: process.env.APPLE_PRIVATE_KEY
  }
}),
```

Do you think an approach like this would make sense, or not really?

If it's computationally expensive, we could probably have it invoked once at app start up and then re-used on subsequent calls (as even a Serverless function stays memory resident and doesn't run from scratch every time it's called).

iaincollins:
> @iaincollins I thought you were taking a break? 😃

Also: LOL 🙈

geraldnolan:
```
clientSecret: ({appleId, keyId, teamId, privateKey}) => {
      const claims = {
        iss: teamId,
        exp: Math.floor(Date.now() / 1000) + ( 86400 * 180 ), // 6 months
        sub: appleId
      }
    
      return jwt.sign(claims, privateKey, {
        algorithm: 'ES256',
        keyid
      })
  })
```

@iaincollins I like your thinking there.  You will have less issues being raised.

One issue I found is trying to put the private key in as env variable.  It's a little bit tricky to put it in a single line.

process.env.APPLE_PRIVATE_KEY  might have to be a location path on disk?

The exp claim should be configurable as well.

iaincollins:
@geraldnolan Thank you, that's very helpful!

I agree, I wondered about the difficulty of putting a key in an environment variable too.

At least with `now.sh` command line tool - and with the GitHub secrets UI - multiline environment variables are fairly easy, as the former can read them from stdin and the GitHub secrets UI lets you paste them, but it's much more alkward in an `.env` file, so maybe the docs ought to show how to use it with `fs.readFileSync()` as well for easier local development.

I really appreciate this; this is a wonderful feature to have!

If I do any refactoring of it as above I'll test it as best I can, but will tag you in the update so you can see if/when it goes out so you have a heads up!

PR #182, opened 2020-05-30
Doc Fix - Provider component import path

jibingeo:


iaincollins:
Thanks! This is a dumb typo that will confuse a lot of folks I'm sure.

A big update to the docs is coming!

PR #184, opened 2020-05-31
Doc Fix - clientSecret is required for Auth0

jibingeo:
Seems like `clientSecret` is mandatory for Auth0 Provider.


iaincollins:
Oh, huh! Thank you!

I haven't tried the OAuth provider myself, but I'll assume that's a typo in the docs and merge this in now (we can go back and change it later if wrong).

The docs are currently being updated, and we should have more explicit examples for each provider soon.

PR #189, opened 2020-06-01
Update documentation

### LINKED ISSUE #176: Refactor and document Apple provider
The new Apple provider is great, but it is slightly trickier provider to work with than most and there is some scope for a small amount of refactoring to make it easier to use.

We should also pull in the very helpful documentation in #174 and add it to the website.

Suggestion for refactoring the provider, so it can be an object (as well as string):

```javascript
Providers.Apple({
  clientSecret: { 
    appleId: process.env.APPLE_ID,
  }
})
```

In `src/providers/apple.js` we could add an option like this:

```javascript
clientSecretCallback: ({appleId, keyId, teamId, privateKey}) => {
  return jwt.sign({
      iss: teamId,
      exp: Math.floor(Date.now() / 1000) + ( 86400 * 180 ), // 6 months
      sub: appleId
  },
  privateKey,
  {
      algorithm: 'ES256',
      keyid
    })
}
```

Refactoring the OAuth callback behaviour to support this doesn't seem like it would be too hard. 

_Originally posted by @iaincollins in https://github.com/iaincollins/next-auth/pull/174#issuecomment-634801822_

iaincollins:
* Documented new features in the client
* Expends options documentation (new documented features!)
* Adds documentation for Apple provider (Resolves #176)

@geraldnolan I would be grateful for any feedback on the update to the Apple provider!

I've incorporated your guide into https://next-auth.js.org/providers/apple – I left in both behaviours for passing secrets, in case generating them on the fly is not something people want to do.

This release has been published as `next-auth@2.0.0-beta.53`.

I think this release is a reasonable candidate to cut for a v2.0 release.

I appreciate we could continue to work on it for a while and do lots of improvements and features, but I think if we release 2.0 we will quickly know what people think is missing or doesn't work well.

However, if the feedback is that things don't work well in this release and there are serious bugs, certainly I'd want to fix them first. Would love to know what people think.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "2.0.0-beta.52",
+  "version": "2.0.0-beta.53",
   "description": "An authentication library for Next.js",
   "repository": "https://github.com/iaincollins/next-auth.git",
   "author": "Iain Collins <me@iaincollins.com>",
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -238,10 +238,10 @@ export default async (req, res, userSuppliedOptions) => {
             return done()
           }
           break
-        case 'check-email':
-          if (options.pages.checkEmail) { return redirect(options.pages.checkEmail) }
+        case 'verify-request':
+          if (options.pages.verifyRequest) { return redirect(options.pages.verifyRequest) }
 
-          pages.render(req, res, 'check-email', { site }, done)
+          pages.render(req, res, 'verify-request', { site }, done)
           break
         case 'error':
           if (options.pages.error) { return redirect(`${options.pages.error}${options.pages.error.includes('?') ? '&' : '?'}error=${error}`) }
--- a/src/server/pages/index.js
+++ b/src/server/pages/index.js
@@ -2,7 +2,7 @@ import fs from 'fs'
 import path from 'path'
 import signin from './signin'
 import signout from './signout'
-import checkEmail from './check-email'
+import verifyRequest from './verify-request'
 import error from './error'
 
 // Future releases will support customization (via inline or external CSS)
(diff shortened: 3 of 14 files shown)


geraldnolan:
@iaincollins I wanted to get these changes to you before you did a release.  Once I get a little bit more time I will add a pull request.  I started adding additional documentation to help the user set up the process.  
https://github.com/geraldnolan/next-auth/commit/155c3d07aa3acbe78b89bc9ea30477b767527513


**UnhandledPromiseRejectionWarning: ReferenceError: keyid is not defined** 

I tried updating the ENV variable from APPLE_KEY_ID=XXXXXX to APPLE_KEY_ID="XXXX" it didn't fix the issue. The only work around that seems to fix the bug if you set a local variable like:

```js
      clientSecretCallback: async ({ appleId, keyId, teamId, privateKey }) => {

      let keyid = keyId;
      const response = jwt.sign(
        {
          iss: teamId,
          exp: Math.floor(Date.now() / 1000) + (86400 * 180), // 6 months
          sub: appleId
        },
        privateKey.replace(/\\n/g, '\n'),
        {
          algorithm: 'ES256',
          keyid
        }
      )
      return Promise.resolve(response)
    },

```
You might have noticed that I also added a **replace(/\\n/g, '\n')** function for those that put privatekey in as ENV Variable.


Below threw a bug in MySQL.  You can't have text as unique.  (https://stackoverflow.com/questions/14033378/make-text-column-as-unique-key)

```js
export const VerificationRequestSchema = {
  name: 'VerificationRequest',
  target: VerificationRequest,
  columns: {
    id: {
      primary: true,
      type: 'int',
      generated: true
    },
    identifer: {
      type: 'text'
    },
    token: {
      type: 'text',
      unique: true
    },
    expires: {
      type: 'timestamp'
    }
  }
}

```

PR #191, opened 2020-06-01
Improve styling on built-in pages

iaincollins:
Improved font usage and button apperance.


iaincollins:
Published as `next-auth@2.0.0-beta.55`

PR #193, opened 2020-06-01
Hotfix for bugs in VerificationRequest Schema and Apple Provider

### LINKED ISSUE #192: Bugs in Beta 56
Reported by @geraldnolan 

@iaincollins I wanted to get these changes to you before you did a release.  Once I get a little bit more time I will add a pull request.  I started adding additional documentation to help the user set up the process.  

https://github.com/geraldnolan/next-auth/commit/155c3d07aa3acbe78b89bc9ea30477b767527513

**UnhandledPromiseRejectionWarning: ReferenceError: keyid is not defined** 

I tried updating the ENV variable from APPLE_KEY_ID=XXXXXX to APPLE_KEY_ID="XXXX" it didn't fix the issue. The only work around that seems to fix the bug if you set a local variable like:

```js
      clientSecretCallback: async ({ appleId, keyId, teamId, privateKey }) => {

      let keyid = keyId;
      const response = jwt.sign(
        {
          iss: teamId,
          exp: Math.floor(Date.now() / 1000) + (86400 * 180), // 6 months
          sub: appleId
        },
        privateKey.replace(/\\n/g, '\n'),
        {
          algorithm: 'ES256',
          keyid
        }
      )
      return Promise.resolve(response)
    },

```
You might have noticed that I also added a **replace(/\\n/g, '\n')** function for those that put privatekey in as ENV Variable.

Below threw a bug in MySQL.  You can't have text as unique.  (https://stackoverflow.com/questions/14033378/make-text-column-as-unique-key)

```js
export const VerificationRequestSchema = {
  name: 'VerificationRequest',
  target: VerificationRequest,
  columns: {
    id: {
      primary: true,
      type: 'int',
      generated: true
    },
    identifer: {
      type: 'text'
    },
    token: {
      type: 'text',
      unique: true
    },
    expires: {
      type: 'timestamp'
    }
  }
}

```

_Originally posted by @geraldnolan in https://github.com/iaincollins/next-auth/pull/189#issuecomment-636857129_

iaincollins:
Resolves #192  (both issues) raised by @geraldnolan in #189

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "2.0.0-beta.56",
+  "version": "2.0.0-beta.57",
   "description": "An authentication library for Next.js",
   "repository": "https://github.com/iaincollins/next-auth.git",
   "author": "Iain Collins <me@iaincollins.com>",
--- a/src/adapters/typeorm/models/verification-request.js
+++ b/src/adapters/typeorm/models/verification-request.js
@@ -19,7 +19,7 @@ export const VerificationRequestSchema = {
       type: 'text'
     },
     token: {
-      type: 'text',
+      type: 'varchar',
       unique: true
     },
     expires: {
--- a/src/providers/apple.js
+++ b/src/providers/apple.js
@@ -37,7 +37,7 @@ export default (options) => {
         privateKey,
         {
           algorithm: 'ES256',
-          keyId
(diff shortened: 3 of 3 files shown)


iaincollins:
@geraldnolan Thanks for raising this! 

Published as `next-auth@2.0.0-beta.57`

PR #195, opened 2020-06-03
Add JWT session support

iaincollins:
* Now has` jwt` and `jwtSecret` options
* Set `jwt: true` to use JWT instead of DB for session
* Enable `debug: true` to log JWT to console if you want to see what it contains
* Includes updated documentation to show how to use it


iaincollins:
Published as `next-auth@2.0.0-beta.58`

PR #199, opened 2020-06-03
Fix bug linking accounts when using JWT with Mongo

### LINKED ISSUE #198: Mongodb does not work with Email auth
Hello,
Thanks for this great library, I'm on version 2.0.0-beta.59 and it seems like Mongodb does not work with Email auth.

After clicking on the link being sent by email I get this error :
`EMAIL_CALLBACK_ERROR TypeError: Cannot read property 'id' of undefined
    at next-auth/dist/server/lib/callback-handler.js:1:1381
`
It seems related to this line,
https://github.com/iaincollins/next-auth/blob/50b9743bb6a659c6b3061f07b7f5db1ede00ff04/src/server/lib/callback-handler.js#L68

Shouldn't it be "user._id" when the database is mongodb ?

iaincollins:
Resolves #198

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "2.0.0-beta.52",
+  "version": "2.0.0-beta.60",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


iaincollins:
Published as `next-auth@2.0.0-beta.60`

PR #201, opened 2020-06-03
Credentials

positonic:
@iaincollins the idea is to pass in a function that calls the api and returns a token like this example, which hardcodes a return object:

const options = {
  site: process.env.SITE,
  providers: [
  ],
  database: process.env.DATABASE_URL,
  doLogin: function (username, password) {
    return {
      success: true,
      token:
        'thisisatoken.therearemanylikeit.butthisoneismine',
      user: {
        firstName: 'Jaxxx',
        lastName: null,
        email: 'jaxxx@gmail.com'
      }
    }
  }

See my question on line 95 of src/server/routes/signin.js below, setting the cookie is usually done in callback, but I'm not sure I need to use callback.js for credentials?

Just another idea, but I thought it could be powerful to be able to pass in a class or functions that with each function of the typeorm adapter.

However I'm far more excited to see review the new jwt functionality you just dropped and see if/how that fits instead or as a part for me.

Let me know if you think this pull request is in the right direction, and if so what you would do differently, as well as any other feedback... :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "2.0.0-beta.60",
+  "version": "2.0.0-beta.59",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/css/index.css
+++ b/src/css/index.css
@@ -22,7 +22,8 @@ form {
   padding: 0;
 }
 
-input[type="text"] {
+input[type='text'],
+input[type='password'] {
   box-sizing: border-box;
   display: block;
   width: 100%;
@@ -90,8 +91,8 @@ a.site {
 }
 
 .page {
-  position: absolute; 
-	width: 100%;
+  position: absolute;
+  width: 100%;
   height: 100%;
   display: table;
   margin: 0;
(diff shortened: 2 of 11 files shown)


iaincollins:
Thanks! Yeah I'd really like to discuss the options for this - those are all the right questions I think and was thinking similar things (and my thinking on that has evolved over time, as features have gone in).

I'm probably going to take a small break but will follow up tomorrow or Friday, would really like to discuss this and write up options to see if they work for folks!

jamespfarrell:
Ok great I'm happy to let this sit for now, and discuss this with you, as I have 2 use cases I'm working for this at the moment.

iaincollins:
Hey just to let you know I'd like to get back in touch regarding this at some point this week!

I just got done with support for [Callbacks](https://next-auth.js.org/configuration/callbacks) and have an idea of how I think credentials based auth could work really well for almost any scenario using the new `signin()` callback.

I'm thinking of providing some examples, but if the credentials flow supported that it would be possible to sign in using any backend system to validate credentials.

It's now possible to sign in with NextAuth.js without a database (using JWT) and I think these two things could work really well together.

iaincollins:
Hi there!

I've dropped #262 as a bonus feature. It's limited, but might be useful.

https://next-auth.js.org/providers/credentials

If it's too limited we can review and see what else we can do in future to improve on it!

PR #204, opened 2020-06-04
Add session usage example to client.md

TwoAbove:
Added example of using session for api routes


LoriKarikari:
Thanks @TwoAbove!

PR #219, opened 2020-06-06
Simplified Auth0 config

LoriKarikari:
Simplified the Auth0 config and made it more consistent with other providers that also need a (sub)domain like Okta. Also updated the docs for this.


iaincollins:
Oh wow neato, I didn't think of that - that much nicer!

PR #71, opened 2019-01-25
Fixing missing comma and typescript proper support

falleco:
Hello!

I am using this library for a while now, never had any problems with it, but recently I started to migrate my codebase to typescript and realized that the typings that were added here doesn't provide enough info to be used in a typescript project.

This PR changes:
- I fixed a missing comma at index.js...
- Rewrote the index.d.ts and client.d.ts so that the already existing types would be used outside the project, also there were some missing information/invalid types there
- Provided an example in typescript, that also worked as a test to the changes. Basically, it is the example repository in ts

Since I am no a typescript pro, I did only enough to compile/execute my project (and the example) with correct typing. If you see any problem with the PR I would be glad to make the necessary changes and resubmit.

PR #220, opened 2020-06-06
Typescript support

0ubbe:
## Summary 

Following the effort @falleco started on #71, this PR aims to bring updated types for [v2](https://github.com/iaincollins/next-auth/tree/master). 

### 🗺&nbsp; Roadmap 

- [x] Write types for [**`Server`**](https://next-auth.js.org/getting-started/example)
- [x] Write types for [**`Server/Providers`**](https://next-auth.js.org/options/providers)
- [x] Write types for [**`Server/Adapters`**](https://next-auth.js.org/options/advanced-options#adapter)
- [x] Write types for[ **`Client`**](https://next-auth.js.org/getting-started/client)
- [x] Try to compile types against existing project
- [ ] Add TS example project
- [ ] Add TS documentation

### 🚨 &nbsp; Disclaimer

I've been working with TS for a while, but this is my first time typing a library so feel free to comment extensively on what can be improved or might be incorrect. 🙏🏻 

I'm also not familiar with `typeorm` or the inner workings of both `adapters` and `provider` so @iaincollins your guiding light 💡 here much appreciated 🤞🏻


iaincollins:
Thanks @lluia, I really appreciated you picking up the work from @falleco for v1! I left the old PR both as reference for myself, and hoping someone might be able to help with it.

I just wanted to say it will probably be a few days before I can pick this up again, but it's really appreciated and it is something I would like to have.

One thing it would be really nice to have some developer tests or a script that we could run with an `npm run` command to help indicate what is missing types so we can improve that over time.

We already have linting with eslint, using StandardJS (`npm run lint`), if using a linter plugin is helpful there? Any recommendations on how we could do something like that would be really welcome. My reason for asking is that I would want to be sure we maintain them easily and they don't break over time when we change something and forget to update types.

shamoons:
Looking forward to this

lluia:
@iaincollins after some investigation I decided to go with [`dtslint`](https://github.com/microsoft/dtslint) for linting the declaration files.

It seems is what most libraries are using 🧐 ( _given is my first try doing this happy to move to other suggestions..._ )

Currently, it can't compile due to an error on `next` types ( which we depend on )
```
$ npx dtslint types
Error: At /Users/lluis.agusti/Code/next-auth/node_modules/next/types/index.d.ts:{"line":14,"character":6}: 'ts-ignore' is forbidden.
    at /Users/lluis.agusti/Code/next-auth/node_modules/dtslint/bin/index.js:198:19
    at fulfilled (/Users/lluis.agusti/Code/next-auth/node_modules/dtslint/bin/index.js:6:58) 
```
I'm looking on how to bypass that 🤔 🤞🏻

iaincollins:
> @iaincollins after some investigation I decided to go with [`dtslint`](https://github.com/microsoft/dtslint) for linting the declaration files.
> 
> It seems is what most libraries are using 🧐 ( _given is my first try doing this happy to move to other suggestions..._ )
> 
> Currently, it can't compile due to an error on `next` types ( which we depend on )
> 
> ```
> $ npx dtslint types
> Error: At /Users/lluis.agusti/Code/next-auth/node_modules/next/types/index.d.ts:{"line":14,"character":6}: 'ts-ignore' is forbidden.
>     at /Users/lluis.agusti/Code/next-auth/node_modules/dtslint/bin/index.js:198:19
>     at Generator.next (<anonymous>)
>     at fulfilled (/Users/lluis.agusti/Code/next-auth/node_modules/dtslint/bin/index.js:6:58) 
> ```
> 
> I'm looking on how to bypass that 🤔 🤞🏻

Thanks! This seems like a good direction!

This is not an area I'm familiar with myself, but if we can get one or two examples working I'd be very happy to pitch in and try and improve (and maintain) coverage.

lluia:
@iaincollins I opened an issue on **`dtslint`** to see if I can get help with the error above 💭
https://github.com/microsoft/dtslint/issues/297

I was thinking instead from writing the declaration files manually 🤔  , we could move the library to **TS** and let the compiler generate the types for us ( might be more scalable ). It'll involve though some work though 😆

arnodel:
I tried to start adding types to for the `client/` directory but could only make it work by adding a `client.d.ts` file to the root of the repo (next to `client.js`).  How can I put these definitions within the `types` directory instead?

lluia:
@arnodel I think the trick is to declare the module in this way:
```ts
declare module 'next-auth/client' {
  // type declarations here
}
```
in this way, it won't matter where declaration files are located ( [as per TS docs](https://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html#pitfalls-of-namespaces-and-modules) ) and hence we can make it live inside of `types/` ( _keeping the project root cleaner in this way_ ) 💅🏻

iaincollins:
^  Thank you to everyone working on this, I'm not commenting because I don't have anything to add to the convo but this is very much appreciated. 🙂

LoriKarikari:
Not into TypeScript @iaincollins?

iaincollins:
@LoriKarikari  I prefer strongly typed languages in general - C# is my all time favourite language - but I'm not personally a fan of TypeScript in practice, as I've run into too many issues with it (e.g. edge cases, compatibility issues, etc).

If pushed I would say I prefer Flow, however, I get that TypeScript is the most popular way of using types in JavaScript and a lot of folks get value out of it and I'm happy to support it.

I'm looking forward to seeing Deno evolve though! I think having better out of the box for types (fast and native) will make it a very different developer experience.

arnodel:
Thanks @lluia for the example.  However it doesn't appear to work for me with the new change - I cannot import from `next-auth` or `next-auth/client`.  I have attached screenshots from my IDE (vscode).  I must admit that although I know typescript well enough, the type declarations for javascript node modules are mysterious to me - so it is possible that I am under some basic misapprehension.

Screenshot 2020-06-15 at 19 57 49

Screenshot 2020-06-15 at 19 57 59

Previously those imports were resolved (with `types/index.d.ts` not containing a module declaration and `client.d.ts` at the module root).

lluia:
@arnodel where are you trying to import those types? Currently they're not published hence not consumable on a project.

We'reusing TDSLint for now to help us write the declaration files which you can do through `yarn lint:types`. 🤔

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

arnodel:
@lluia I am importing from a project that uses `next-auth` :).   So let me check that I understand: currently the type declaration files are not exported, later they will be put in `dist/`, alongside the `js` files that they provide type declarations for?  I would find it useful to have that in place as I am currently using it.   Currently I am using the a workaround which consists of addings the following `compilerOptions` to the `tsconfig.json` of the project that uses `next-auth`:

```
   "baseUrl": ".",
    "paths": {
      "next-auth/*": ["node_modules/next-auth/types/*"]
    }
```

That doesn't seem a good long term solution to me however (no other package that I depend on requires that).

Also, when running `yarn lint:types` I got an error: 

```
Error: /Users/adelobelle/personal/next-auth/types/client.d.ts:2:1
ERROR: 2:1  no-single-declare-module  File has only 1 module declaration — write it as an external module. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-single-declare-module.md
```

The [link](https://github.com/Microsoft/dtslint/blob/master/docs/no-single-declare-module.md) says that `declare module` should "typically be avoided" - and to me it makes senses if the files are going to end up in `dist/` as I describe above.

So perhaps it would be better to rewrite the contents of `index.d.ts` and `client.d.ts` without the `declare module` statements?  I have made it work by changing the `compilerOptions/paths` configutation of `types/tsconfig.json` slightly - see https://github.com/arnodel/next-auth/commit/3260060a2f68ea1fc1152cd4661205151041c624

lluia:
@arnodel good catch, I think as you suggest using `compilerOptions/paths` for now seems a better approach than  `declare module` 💯  . 

To clarify: for now, the `types/*` directory helps us keep the types in one place and lint them through [`dtslint`](https://github.com/microsoft/dtslint/). Later, once we're happy with the types, we will need to copy what's in `types/*.d.ts` to `dist/` at build time, so we have them published on NPM and clients of this library can consume them.

If you need them straight away for a project you're working I'd suggest the best is to use [`npm link`](https://docs.npmjs.com/cli/link) with this branch 🤔

arnodel:
@lluia personally I wouldn't mind having the type definition files in the `src/` directory.  I feel like having them close to the js files makes the link more obvious, but I don't actually know what is generally done.  The dependencies I use seem to either be written in Typescript or have a separate `@types` package.

I also think it would be good to publish those types as soon as possible.  IMHO as this is a fork / PR, there is no need to worry about exposing a non-stable interface to unsuspecting users, but it allows people to use it easily (knowing the risks).

In my opinion lack of typescript definitions is likely to put off a fair number of potential users, so providing something usable early would be very valuable.

lluia:
@arnodel the main purpose of `types/` is for [`dtslint`](https://github.com/Microsoft/dtslint#add-types-for-a-library-not-on-definitelytyped) to work, however, we can always move it to [`@types`](https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package) 🤷🏻‍♂️ , either way, the workflow is similar. 

About publishing early versions, I don't think is a good idea to publish types half-way or not working. In case you need them ASAP, I suggest writing your  **declaration file** in your app or `npm link` meanwhile 🤔

lluia:
@iaincollins @arnodel I guess moving them to [`@types`](https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package) feels more stable: if we submit a PR there with the types for this package then is gonna be reviewed by people with a lot of experience writing `*.d.ts` files and can advise us further. 

I'm going to move what's being done here there for now, and then close then this PR ? 🤔

lluia:
## Update 🍉

@iaincollins @arnodel 

I added types for what I think covers most of the library ( 👀 _last commit_ ), the two APIs that are not fully typed are:
-  [**adapter**](https://next-auth.js.org/schemas/adapters) ( _typed with generic objects_ )
```ts
interface GenericObject {
  [key: string]: any;
}

type Adapter = (config: ConnectionOptions) => Promise<GenericObject>;

interface Adapters {
  Default: Adapter;
  TypeORM: {
    Adapter: Adapter;
    Models: GenericObject;
  };
}
```
- [**callbacks**](https://next-auth.js.org/configuration/callbacks) ( _typed with generic objects_ )
```ts
interface Callbacks {
  signin(profile: GenericObject, account: GenericObject, metadata: GenericObject): Promise<void>;
  redirect(url: string, baseUrl: string): Promise<string>;
  session(session: GenericObject, token: GenericObject): Promise<GenericObject>;
  jwt(token: GenericObject, oAuthProfile: GenericObject): Promise<GenericObject>;
}
```
Both these API accept tons of different configurations, I think is fair to leave them loosely typed for the time being and people interested in having more strongly typed can open news PRs.

## What's Next 💭

- [ ] Investigate whether to add these types as part of the build process or have them moved to `DefinitivelyTyped`.  
- [ ] Publish a **beta version** of the library with the types on it
- [ ] Create a [TS example repo](https://github.com/iaincollins/next-auth-example) consuming the beta and verifying all is fine

types/client/index.d.ts
```diff
@@ -1,10 +1,64 @@
...
+}
+
+interface GenericObject {
+  [key: string]: any;
 }
 
 type UseSessionResult = [Session, boolean];
 
 declare function useSession(): UseSessionResult;
+declare function getSession(context: NextContext): Session | null;
```
arnodel, line 33: These functions return promises

lluia: Good catch 👍🏻 , I actually need to export as well the declarations for the aliases as well ( `session`, `crsfToken` and `providers` ) 💆🏻‍♂️

LoriKarikari:
Nice work so far @lluia!

bamorim:
Hi @lluia is there anything I can do to help you with this?
Thanks for this PR!

types/providers/index.d.ts
```diff
@@ -0,0 +1,206 @@
+// Minimum TypeScript Version: 3.8
+
+interface Providers {
+  Email: Email;
+  Credentials: Credentials;
+  Apple: Apple;
+  Twitter: Twitter;
+  Facebook: Facebook;
+  Github: Github;
```
bamorim, line 9: The name of the provider exported is GitHub and not Github.

types/index.d.ts
```diff
@@ -0,0 +1,109 @@
...
+/// <reference types="node" />
+import type { ConnectionOptions } from 'typeorm';
+import type { IncomingMessage, ServerResponse } from 'http';
+import type { PossibleProviders } from './providers';
+import Adapter from './adapters';
+
+interface InitOptions {
+  site: string;
+  providers: PossibleProviders[];
+  database: ConnectionOptions | string;
```
bamorim, line 11: Is this a required option? From the [docs](https://next-auth.js.org/getting-started/example#add-api-route) they say it is optional (only required to persist accounts)

lluia: Good catch 👏🏻  , @iaincollins could you confirm this? 🙏🏻

lluia:
@bamorim I think all types are ready ( some can be improved though, look for `TODO` in the changed files... ).

What's missing now is [to write the tests](https://github.com/DefinitelyTyped/DefinitelyTyped#testing) and find the best way [to publish them](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html). I think ideally we would move this to [`DefinitivelyTyped`](https://github.com/DefinitelyTyped/DefinitelyTyped). 

Regarding publishing, we need to make sure we structure the files correctly given this library exposes four modules 📐
```
'next-auth' // server 
'next-auth/client' 
'next-auth/providers' 
'next-auth/adapters' 
```
As I understand the published types should then have this structure 💭
```
inde.d.ts // types for server
client.d.ts
providers.d.ts
adapters.d.ts
```
However, we need to play with the compiler to make sure it picks them up correctly. I'll try to give a push mid-week ( Wednesday ) but, any help is welcome especially if you have more experience typing libraries 👍🏻

harrysolovay:
What're your thoughts on migrating the codebase to TypeScript, instead of trying to mirror the types (and keep them in sync)?

I'm more than happy to PR, if you @iaincollins decide this is desirable.

lluia:
@harrysolovay I suggested this [here also](https://github.com/iaincollins/next-auth/pull/220#issuecomment-644270011), I think as an MVP we can publish the manual types on **DefinitivelyTyped** and in parallel open an issue and discuss how desirable ( and how much effort ) would be to move the source files to TS.

We need help [testing the types written on this PR](https://github.com/DefinitelyTyped/DefinitelyTyped#testing), so maybe it would a good place to start with your contributions 👍🏻

iaincollins:
I agree with @lluia that the best steps are to see how far we get here, and to introduce tests to ensure we maintain coverage as we go. Getting to that point itself is not easy - though hopefully it is less of a moving target now 2.0 is out of beta - but it's premature to suggest we can handle migrating the entire codebase until we can at least get to that point.

Personally, I think that point has best trade off of effort vs benefit, and I am very happy to commit to support exporting types for API surfaces as I think it has clear benefits for users of the library and we can generally define well what request/responses we return (except in the case of raw objects from OAuth providers).

---

### On moving the project to TypeScript

Migrating the actual project TypeScript should only happen after there are goals with defined benefits and understanding the tradeoffs - and crucially that means _accepting that there are tradeoffs_ as well as benefits.

Certainly it's obvious what the benefits to users of the module are to have type definitions, but it would have been practically impossible to release this project if the whole thing were written in TypeScript (by any meaningful definition) because it would have multiplied the complexity and taken additional weeks or months to get to the same point, and I don't have time for that.

For sure using TS would almost certainly mean less bugs, and means spending less time on squashing them and the idea is that pays back dividends in future, but shipping with bugs you can fix later is better than never shipping.

If we had a larger / well funded / dedicated team that was up for supporting this in the long haul, I'd be open to taking a different view (as the cost/benefit changes); but the prospect of having someone start a re-write, lose interest and for us to reduce the number of people willing/able to collaborate on it while having increased LoC and overhead of maintenance is not appealing. Re-writes can really kill momentum on open source projects, because people have little incentive to push through the slog to get to working on the bits they feel deliver value.

### Why would it be such a PITA on this project?

The main difficultly in this particular project that request and response objects from the services it integrates with are so varied and so conditionally and ambiguous that it's really hard to wrangle request/responses and you'd end up having to use any/object types all over the place or expect people to write a bunch of type definitions for each provider (ugh!) *and* provide a way for those type definitions to be conditional (as the same API endpoints return different responses under different situations).

### How do we get to a better place?

I would suggest we get type definitions in (as @lluia suggests) and focus on supporting the tests that @JeffersonBledsoe has been working on, and getting to the point where can safely make more radical changes with confidence.

Once we have these two things - with a CD/CI pipeline that requires both pass before changes can be merged - we could revisit things then and decide on an order we might want to tackle them if we did want to start refactoring things in TS.

I would favour a gradual approach at that point rather than a big bang. I suspect it would take a long time to make the transition (maybe months, maybe a year or more) and that is something that would need to be factored in. Any time spent on this is time spent not adding new features or working on bugs, so the project may seem to stagnate, which is again another reason for caution.

I appreciate one dedicated person could do a rewrite in much shorter time, but it would still be several weeks worth of work and would involve a lot of tedious regression testing. It only took me ~6 weeks working in spare time (evenings, weekends and lunch breaks) to build v2  (with help from the other core team members), so it's possible, but not fun and I expect would take longer as you'd have to face up against challenges with database integration and provider definitions to implement types in any meaningful way (i.e. without falling back to any/object in key places and hoping for the best).

Personally, if someone asked me how long it would take me working on it **full time** to migrate the entire project to TS I'd estimate at least as long as it took to write in the first place and that it would probably take longer than that.

LoriKarikari:
@iaincollins what about gradually transitioning to TS? All the JS keeps working anyway so it's not a complete rewrite from scratch. For me, it's fine if it just stays JS btw just giving my 2 cents.

harrysolovay:
@iaincollins I completely agree with everything you've said. I've created a draft nonetheless, so we can start hammering out specifics. I don't expect it to be merge-ready all too soon... but it's a starting point. More testing is gonna be critical to ensuring a smooth transition.

lluia:
### Update

@iaincollins @harrysolovay @LoriKarikari I opened a new PR on **DefinitivelyTyped**: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45775 with all the types updated.

### Tests
This [test file covers the types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4e7f90108a74a137361eef3c47c0a08017d433c0/types/next-auth/next-auth-tests.ts) for the exported API of **next-auth** ( cc. @iaincollins ).

Hopefully, I didn't miss anything, but if so we can always publish and keep on iterating.

### Owners
I marked myself and @iaincollins  as owners of the future `@types/next-auth` package, @harrysolovay let me know if you want to be added as well ( _or anybody else too_ ). Once published, we need to sync on how to communicate best API changes on this package so types can be always up to date.

### Follow up
I'll follow up there and leave this open for reference 👍🏻 ( _we're almost there!_ 🎊 )

bamorim:
Great work @lluia! :100:  Sorry for not being that present. I'll be working on my Next.js project next weekend and then I can spare some time helping. (I'm on a really tight schedule between multiple side projects, so sorry about that). 

@iaincollins I totally get your point. Better ship it and make it work, and then, once you have a stable enough library then migrating to TS is easier than prototyping in a lot of cases. It might not even be worth, as you said. But new versions of TS keep adding niceties to the type system like `unknown` added in 3.0 and the better recursive type alias added in 3.7. What I mean is that I feel like TS is trying to evolve to make those cases easier to represent. My point is that waiting may make this job actually easier just because TS will release new features.

samsamson33:
@lluia what's happening with the typings now? I see that the CI is failing on DefinitelyTyped... is there anything I can do help out?

lluia:
@samsamson33 I need to have a look, hopefully, this week  👀 🤞🏻 , I think the compiler is trying to [parse the types from `next`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6b75d575a9b35acfdb908be605f8fc1c8af99c4d/types/next-auth/index.d.ts#L9) and failing:
```
##[error]node_modules/next/dist/next-server/lib/router/router.d.ts(72,21): error TS2304: Cannot find name 'PopStateEvent'.
```
A workaround, for the time being, is to not relly on `next` types and type those manually. I'll give it a go 👨🏻‍💻 ( _feel free to suggest any improvements or changes you see fit_ 💯  )

samsamson33:
@lluia could not finding `PopStateEvent` possibly be because you don't have `"dom"` in your tsconfig's `lib`? (Sorry if I'm way off the mark here)

lluia:
@samsamson33 ah good point 💥 , let me push it to the PR and see if it makes the error go away 🤞🏻

NickBolles:
@iaincollins @lluia @samsamson33 @bamorim and others. awesome work on this! (I was disappointed that [my original work on typings](https://github.com/nextauthjs/next-auth/commit/ba555a83f926e19d3f5f85364baae0723fc24c9d) weren't updated for this release, so I came looking)

I spent an hour or two trying out converting next auth to typescript. I just have to finish typing the options and a few other small issues (and taking some types from this PR). @iaincollins would you be more open to accepting a PR like this? Most of it is utilizing type inference or `any`. but at the very least it's up to date and can be improved incrementally with (hopefully) no impact on existing users (js or TS), I've added minimal interfaces where it makes sense, and fixed a few type errors where necessary. I'd be curious to hear your objections and drawbacks of typescript (beyond initial development, because I think everyone agrees with that point)

NickBolles:
@iaincollins @lluia @samsamson33 @bamorim See the PR I just submitted for a working migration to typescript. I'm sure there's more work to be done, so please check it out and provide feedback!

We might end up abandoning it, because of reasons explained earlier in this chain, but I thought I'd remove the "too much work to transition" hurdle. (because really its not hard... we could have stopped at just a rename and add a bunch of `any`s, but I wanted to do it right).

iaincollins:
I favour @lluia approach in this PR, which is incremental and gives a chance to test things out as a team.

With regard to TypeScript, supporting pull request is my priority right now.

If anyone is looking for a way to help out to support a move to TypeScript (beyond helping out with this PR) the most useful thing right now would be to help out with the end to end tests, which have sadly been on hold for too long.

lluia:
## Update 🗓

So the types are finally ready on **DefinitivelyTyped** 💆🏻‍♂️ 🎊 🤞🏻 💥
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45775

I'll need now a couple of reviews (maybe @iaincollins and @NickBolles ?) to be able to  merge it 💚

I'm aware they're less than a perfect but I think is a good start and we can get them out and iterate later.

I think the key for them to grow strong is to sync well whenever changes happen in this library. As a first step making `CHANGELOG.md` would be good, but here's [more 🥗 for thought](https://github.com/DefinitelyTyped/DefinitelyTyped#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library).

## Why it took so long ⏲

I was caught forever with a [cryptic error](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/6045bad877426c236e055e7ce10d0b28d30f3ad7/checks?check_suite_id=998312490) related to [`'source-map'`](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/23649) that went away once I removed `next` types from the dependencies. This, however, means I'm typing **`NextApiRequest`** and **`NextApiResponse`** manually which seems a bit brittle and I'm eager to see how it plays once the types are released... 

The fact also that I lost my job due to COVID and been actively looking for a new one since a few weeks made a bit challenging keeping up with open-source 😅

types/index.d.ts
```diff
@@ -0,0 +1,109 @@
+// Minimum TypeScript Version: 3.8
+/// <reference types="node" />
+import type { ConnectionOptions } from 'typeorm';
+import type { IncomingMessage, ServerResponse } from 'http';
+import type { PossibleProviders } from './providers';
+import Adapter from './adapters';
+
+interface InitOptions {
+  site: string;
```
m5r, line 9: I can't find any information about this option from the [docs](https://next-auth.js.org/configuration/options), what is it for ?

types/index.d.ts
```diff
@@ -0,0 +1,109 @@
...
+  providers: PossibleProviders[];
```
m5r, line 10: Each of these providers are typed to return `void` when actually they return an object ([email provider implementation](https://github.com/nextauthjs/next-auth/blob/main/src/providers/email.js#L5))

In the current state, the following example from the [docs](https://next-auth.js.org/getting-started/example#add-api-route) generates the following error

```typescript
import type { NextApiRequest, NextApiResponse } from "next";
import Providers from "next-auth/providers";

const options = {
	providers: [
		Providers.GitHub({
			clientId: process.env.GITHUB_ID!,
			clientSecret: process.env.GITHUB_SECRET!,
		}),
	],
};

export default (req: NextApiRequest, res: NextApiResponse) => NextAuth(req, res, options)

```

```
Error:(14, 82) TS2345: Argument of type '{ providers: void[]; }' is not assignable to parameter of type 'InitOptions'.
  Types of property 'providers' are incompatible.
    Type 'void[]' is not assignable to type 'PossibleProviders[]'.
      Type 'void' is not assignable to type 'PossibleProviders'.
```

types/providers/index.d.ts
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Auth0
+ */
+type Auth0 = (options: ProviderAuth0Options) => void;
+
+interface ProviderAuth0Options {
+  clientId: string;
+  clientSecret: string;
+  subdomain: string;
```
m5r, line 149: From the [docs example](https://next-auth.js.org/providers/auth0#example) this field should be named `domain`

m5r (changes requested):

lluia:
@m5r thanks for your comments 💯 , I'll try to get them addressed soon. Could you next time comment on the [**DefinitivelyTyped** PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45775)? The work is carried on there and this one is just open for reference.

types/index.d.ts
lluia: This was already updated in the [actual types PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45775/files) 👍🏽

types/providers/index.d.ts
lluia: This was already updated in the actual types PR 👍🏽

types/index.d.ts
lluia: Good catch, I see this options have been simplified quite a lot on **v3**, I'll update on the [types PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45775/files)

lluia:
Finally [`@types/next-auth`](https://www.npmjs.com/package/@types/next-auth) is published, we can continue to raise PRs in [**DefinitivelyTyped**](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45775/files) to improve them 💚 💯

LoriKarikari:
Great work @lluia!

iaincollins:
@lluia Thank you and congratulations! <3

What are your thoughts on merging this PR into NextAuth.js also, so we can work towards better support for types natively?

alxyuu:
are there any plans to add types for `next-auth/jwt` as well?

lluia:
@iaincollins I think the types on this PR are a bit outdated, so in that case would be best to directly copy what we have on **DefinitivelyTyped** 🤔 

 I'd say for now lets see how `@types/next-auth` plays out, there's people making PRs on it  🎊 , too bad I'm on holidays this week 🏖️ with limited access to internet, I wanted to find time to create the **example TS repo** 🤞🏽

JuanM04:
I could do an exact copy of the example repo but in TypeScript. Just create an empty repo in the organization and I'll PR it

JuanM04:
> are there any plans to add types for `next-auth/jwt` as well?

@alxyuu, I did them. The issue is that I need some types from `jose`, and to do that, [this PR](https://github.com/microsoft/DefinitelyTyped-tools/pull/101) has to be approved. Then, I will push the types for JWT!

lluia:
@JuanM04 [here you go](https://github.com/nextauthjs/next-auth-tyepscript-example) if you have time 👍🏽 , would be cool to keep the file structure same as the [default example repo](https://github.com/nextauthjs/next-auth-example).

In case you don't find time don't worry! I'll try to PR on it once I'm back from hols ⛱ next week ( @iaincollins )

zenflow:
@lluia Thanks for this contribution, and more importantly, enjoy your holidays! 😃

JuanM04:
Thank you, @lluia! Enjoy your holidays!

JuanM04:
I finished the repo, but some types are wrong. I [opened a PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/47310/) fixing them

iaincollins:
@lluia Thanks for your feedback! Sounds good to me. Have a great holiday! :-)

kwent:
Thank you. Should be added in the documentation as well :) with the command `npm install @types/next-auth --save-dev`

lucascherifi:
great !! thx

PR #221, opened 2020-06-06
Add new OAuth providers

LoriKarikari:
I'm adding some additional OAuth providers. 
- [x]   Cognito
- [x] Battle.net (no email)


iaincollins:
Had a thought about supporting no email providers.

One thing we could do *really* easily would be to make it an option and if email is set as not required then autofill the email field on a user account with a dummy field (e.g. `{uid}@localhost`).

This isn't the most elegant solution but it would be really quick and simple to check for this option and would allow sign in without an email address if you didn't care about it.

Like with version 1 people would still need to watch out for this, but if wasn't the default it seems like safe behaviour and a quick win?

LoriKarikari:
That seems like a nice solution if it's an option you can enable. I was thinking that the 'no email'providers could be used for linking only.

tobilg:
Thanks for the work on Cognito! I’d suggest to also enable setting the scopes via options, and supporting custom auth domains as well...

LoriKarikari:
Thanks for the feedback @tobilg. Do you use a custom domain with Cognito? I can't test that as I don't use AWS, I just created a trial account to test this out.

LoriKarikari:
And sorry for the delay on this one, currently a bit busy but will try to finish this asap or merge the ones that are finished.

iaincollins:
> Thanks for the work on Cognito! I’d suggest to also enable setting the scopes via options, and supporting custom auth domains as well...

Just to clarify - as it's perhaps not obvious and I'm not sure it's explictly mentioned anywhere in the documentation yet - all the providers have the '...options' spread, so you can actually override any of the properties (e.g. scope, domain, URLs, etc) just by passing them as options when specifying the provider.

LoriKarikari:
Does it completely override the scopes or does it add them to the existing scopes? Because I can imagine that people will forget the necessary scopes and the authentication doesn't work anymore.

iaincollins:
> Does it completely override the scopes or does it add them to the existing scopes? Because I can imagine that people will forget the necessary scopes and the authentication doesn't work anymore.

Oh yeah it completely overrides them.

I think that is something I am okay with personally, because I think the extra complexity required to support merging them is probably not worth it and seems like it would just make things more confusing.

Improving the documentation could maybe make what the default values are more transparent - maybe if auto-generated some of the documentation for each provider?

I'd like to do something like that but haven't thought too much about it yet; am aware we'd still like to be able to have custom docs (e.g. for Apple), but I think it would be great if there could be some boilerplate at the start of each provider documentation page

e.g. that list lists all the properties, etc.

src/providers/cognito.js
```diff
@@ -0,0 +1,23 @@
...
+  return {
+    id: 'cognito',
+    name: 'Cognito',
+    type: 'oauth',
+    version: '2.0',
+    scope: 'openid profile email',
+    params: { grant_type: 'authorization_code' },
+    accessTokenUrl: `https://${domain}.auth.${region}.amazoncognito.com/oauth2/token`,
+    authorizationUrl: `https://${domain}.auth.${region}.amazoncognito.com/oauth2/authorize?response_type=code`,
+    profileUrl: `https://${domain}.auth.${region}.amazoncognito.com/oauth2/userInfo`,
```
iaincollins, line 12: Hey there! Really excited about Amazon Cognito support :D

I was wondering if we should do all domain configuration on providers a bit like this (below).

I was thinking it might be easier for users to take the same approach for each provider, and often is easier to copy/paste the complete domain from the admin UI for the provider when setting it up?

```suggestion
  return {
    id: 'cognito',
    name: 'Cognito',
    type: 'oauth',
    version: '2.0',
    scope: 'openid profile email',
    params: { grant_type: 'authorization_code' },
    accessTokenUrl: `https://${options.domain}/oauth2/token`,
    authorizationUrl: `https://${options.domain}/oauth2/authorize?response_type=code`,
    profileUrl: `https://${options.domain}/oauth2/userInfo`,
```

LoriKarikari: Ooh yeah, I can change that if you want. I'm currently fixing some rebase mess from switching from master to main 😅

LoriKarikari: This probably also solves the custom domain issue I guess?

iaincollins: Ah lol I feel your pain!

I did that and had to check StackOverflow to undo the mess I made. 

You might not need this, but if it helps, this is what I ended up doing:

1. `git reset --soft HEAD~1 ` to undo the last checkout, but restoring the changes to the working directory.

   You can do multiple commits here instead of just the last one, but I didn't have many commits in the branch and found it easier to restore and stash each commit separately and then replay them, rather than checking out the last 2-3 commits at once, so I could review the changes more easily.
2. Saving the files it spit out in a stash.
3.  Repeating that for each commit I'd done, saving each commit as a new stash, checking how many I had left to go with `git log`.

Then once I'd saved all the commits as stashes , I did `git reset --hard origin/main` and then applied the stashes back (and committed each one) in order.

It's possible to generate a patch file too I think, but I've only done that a couple of times so found this easier.

LoriKarikari:
I've temporarily removed LinkedIn and Stripe so I could merge the working providers.

LoriKarikari:
@iaincollins I just noticed the formatting is all over the place. Maybe we should add eslint and prettier?

iaincollins:
@LoriKarikari +1

We actually have `npm run lint` and `npm run lint:fix` already, but they aren't configured to run as tests or on commit (so you have to run them manually right now).

We could automate them before commit with Husky if you like?

I'd actually quite like that but didn't want to annoy folks with it until we needed it!

romMidnight:
@LoriKarikari Any chance you could add Linkedin provider? I am trying adding one right now, but can not get around fetching email and profile image separately as Linkedin does not return that information with GET /me, probably fetching them with await in "profile" async function would be a solution, if only we had access to accessToken inside profile function to set an auth header.

LoriKarikari:
I was planning to add q LinkedIn provider but scrapped it because I don't have a LinkedIn account to log in and test it out. The email address isn't required anymore so it can be skipped (if you don't need it).

romMidnight:
email is not required but it's nice to have, especially when you want to link other oauth accounts with user, does having accessToken parameter as profile function args make sense? that way we could fetch additional profile information by including token in auth header (email & profile image in case of linkedin) @iaincollins

iaincollins:
Ugh, it looks like Linked In have a lousy API - it works differently to every other implementation. They seem to have hacked on scope support for this but only with a completely different URL to pull phone number and email addresses.

We have the access token available so we could pass it in here. 

If it's captured as a feature request with a provider config supplied that folks can use then I'll look at the best way to pass it in.

Suppling accessToken a second argument to the callback function is the obvious approach, but will want to see what other use cases coming up we might have to see if that's the best approach.

PR #222, opened 2020-06-06
WIP: Console Errors with URLs

ndom91:
Testing overriding `console.error()` to append a line with links to the docs after every named error.


iaincollins (approved):
Sorry don't have time to reply but looks cool!

ndom91:
Okay so this very first prototype generates the following type of output on `console.errors`:

image

I'm sure theres a more elegant way of doing this rather than overriding `console.error` and having to import that iife into each area of the library, but this definitely works for now. 

I also added an error page with all of THESE_TYPES_OF_ERRORS that I could find, available as hash on the end of the URL.

iaincollins (changes requested):
Hey, so been thinking about this and I think we probably do need to make a library that exports a function that handles logging / formatting logs (and import that lib into wherever we want to use it).

The reason for that is that we probably need to avoid overwriting `console.error` in the library in case it causes problems for folks using the library / importing functions in their own code.

What if we tried it in one or two places to start with, and went back and updated other functions later / as we go - so we don't have to do them all at once in a big PR?

(I am totally fine if we did want to do more, just thinking it might be easier to try it with a few first?)

If I am wrong about any of these assumptions do let me know. :-)

ndom91:
Yeah I think overwriting console.error is kinda extreme too. 

You mean just writing a custom error function that basically console.errors the info passed into it + the link (for now)?

iaincollins:
Yeah, I think a function is probably best - I love the idea of one that looks like what you have and handles and spirts out a link to the documentation error page!

ndom91:
Okay so what do you think, should we make a whole nother external library that we import or do you want to just make our own function that we use in these few places?

ndom91:
Okay so current state of things is that the `console.error` overwrite is now replaced with a custom function, and the docs line gets bolded in the console for extra style points ;) haha

image

iaincollins:
Sorry the big update I pushed out (my last big PR for 2.0) has caused conflicts. :-(

The PR kind of got away from me, but it nails a bunch of issues we were getting asked about.

This looks good! I think it would be great if `consoleErr` was actually something like `log` or `logger` (not sure of a good name?) and had `debug` and `error` functions, and could even overridden using options for those who want to use a logging service.

e.g.

```js
logger: {
  debug: (({args}) => { console.log({...args}) },
  error: (({args}) => { console.error({...args}) }
}
```

But I would be very happy to go with the current approach and refactor later. :-)

ndom91:
Hey no problem, so I refactored the function to call it `logger` and made the primary method `error` what we had previously. I added a `debug` dummy method for now, you can obviously refactor that / use it as you see fit.

I'll fix up the merge conflicts real quick and get this merge ready. We can iterate on it then. 

What do you think?

ndom91:
Okay so this looks to be merge ready now from my side. Good to go whenever your ready :)

iaincollins:
Thanks, this looks great, please go for it!

iaincollins (approved):

PR #223, opened 2020-06-07
Refactor JWT, Sessions and add allowSignin() method

### LINKED ISSUE #57: Add example to secure some urls
How should one go about securing some urls? I'm guessing if one requires authentication, it means we want to protect about unauthenticated parties.

Ideally, when rendering server-side, if a url is protected it should be redirected to an auth page `/auth`, and when calling a protected api from browser, the server should return some error.

I would be happy to submit a PR for docs/code/whatever, just need some pointers.

### LINKED ISSUE #78: Question: Adding attributes to props.session.user
Hello,

I was asking this in a closed topic but i don't know how visible it is.
I wonder what is the best approach to getting some information about the user from a 3rd party web api and injecting it in the current user. insert a fetch in the insert function?
insert: (user, oAuthProfile)...

i.e. i'd like to implement roles and authorization. and i'd like to keep the role in the same session.

### LINKED ISSUE #203: Remove error - NextAuth requires a 'database' or 'adapter' option to be specified.
> Error:
>  NextAuth requires a 'database' or 'adapter' option to be specified.
>  See documentation for details https://next-auth.js.org

Now that jwt is to be default

### LINKED ISSUE #207: Getting provider id from providers route
Using the [providers()](https://next-auth.js.org/getting-started/client#providers) method currently returns an object with:

```js
{
  name,
  type,
  signinUrl,
  callbackUrl
}
```

However, the object is missing the `id` property. This property is needed for matching a provider with a user account's `providerId` field.

Would it be possible to add the provider id to the `providers()` method?

### LINKED ISSUE #215: Encrypt JWT after signing
Add [crypto-js](https://www.npmjs.com/package/crypto-js) and use `crypto-js/aes` to encode JSON Web Tokens after they have been signed, and to decode them before they are verified.

The initial implementation will use a symmetric algorithm, and the `jwtSecret` will be used to encrypt and to sign the token. This allows for a simple and reasonably secure implementation that can be expanded later to support other authentication mechanisms, like asymmetric algorithms, in a backwards compatible way.

## What is in the JWT currently?

* ID
* Name
* Email
* AccessToken
* OAuth Provider details (Account ID, AccessToken, RefreshToken)

### What should be private?

The only "private" information currently "exposed" to a client is the database user ID (e.g. an `int` if using an SQL DB or an `Object ID` if using MongoDB). It can be desirable not to expose this, though it's rarely considered actually "secret".

The `RefreshToken` and `AccessToken` for the service the user used belong to the user, and it's okay for them to know them.

### Who can read this data?

Only the server, the browser and the user can read this data.

The cookie is HTTP only 

### Why encrypt the JWT?

Currently, the JWT is signed but not encrypted.

It should be encrypted to avoid inadvertently exposing data, especially if we add functions to allow people to customise what is stored in the JWT.

### Considerations

It would be sensible to move to the methods for setting (encrypting + signing) and getting (decrypting + verifying) the JWT into functions like `set()` and `get()` methods, and to allow those functions to be defined as options server side.

Example method signatures:

* `async setJWT({ secret, user, provider, options })`
* `async getJWT({ secret, user, provider, options })`

The current `jwt: true` and `jwtSecrets` option could change to be an object.

e.g.

```js
jwt: {
  enabled: true,
  secret: '',
  set: async ({ secret, token }) => { return encryptedToken },
  get: async ({ secret, encryptedToken }) => { return token },
}
```

The `token` would be populated with the current contents by default, but people could add whatever they need to it, and choose whatever encryption and description system they want.

#### Consider similar changes to session options

If we are changing this, it seems sensible to similarly change the `sessionMaxAge` and `sessionUpdateAge` options too, and to add a `get` method to support returning a custom session object when using a database based session.

```js
session: {
  maxAge: 30*24*60*60*1000,
  updateAge: 24*60*60*1000,
  get: async ({ user, session }) => { return customSessionObject }
}
```

This would keep the behaviour similar.

iaincollins:
# Change log

## Database

- [x] Databases are now optional - useful with OAuth + JWT if you only need access control
- [x] Updated documentation and added example code for custom database adapters

## JWT

- [x] JWT option is now an object that groups JWT related options together (was a boolean)
- [X] Refactored JWT lib and add AES encryption / decryption as well as signing / verification
- [x] Allows JWT encode/decode methods to be overridden as options
- [x] Contents of JWT can easily customised - without needing to use custom encode/decode
- [x] Exported JWT methods so they can be called from custom API routes
- [x] Updated documentation for new JWT options

## Sessions

- [x] All session options (eg. `maxAge`, `updateAge`) now grouped under single `session` option
- [x] Using JWT for sessions is now enabled from session object (`session.jwt: true`)
- [x] All options involving time now use seconds (instead of milliseconds) for consistency
- [x] Added option to customise the Session object that is returned from `/api/auth/session`
- [x] Update documentation for new Session options

## Other improvements

- [x] Added `allowSignin()` option to control what users / accounts are allowed to sign in
- [x] Refactored `callbackUrlHandler()` - this option  is now called `allowCallbackUrl()` 
- [x] Minor improvements to NextAuth.js client API methods
- [x] Minor to NextAuth.js API routes
- [x] Minor improvements to built-in error pages
- [x] Refactored database models
   All tables now include a `created` column for each row which contains the `datetime` of when the row (e.g. User / Account / Session) was created.
  Additionally, sessions now use the name 'expiry' for the expiry `datetime` value for consistency with other models.

---

## Open issues resolved by this update

* Resolves #215
* Resolves #207
* Resolves #203
* Resolves #78
* Resolves #57 (with JWT example)


iaincollins:
This work is almost complete - it grew in scope slightly as it increased, to ensure the API is as consistent as possible and to help keep the code maintainable.

It's only missing one small bit of functionality and some documentation now.

The new `jwt` and `session` options are now consistant, and look like this:

```js
session: {
  maxAge: 30 * 24 * 60 * 60,
  updateAge: 24 * 60 * 60
},
jwt: {
  enabled: false,
  secret: '', // Auto generated if not specified
  maxAge: 30 * 24 * 60 * 60,
  // Custom encode/decode functions for signing + encryption can be specified
  // if you want to override what is in the JWT or how it is signed.
  
  // encode: async ({ secret, key, token, maxAge }) => {},
  // decode: async ({ secret, key, token, maxAge }) => {},
}
```

* For consistancy they now all use **seconds** instead of milliseconds.
* If `jwt.enabled` is set to `true`, the values in `session` are not used.
* If `jwt.enabled` is not set to `true`, the options in `jwt` are not used.

----

The built-in JWT encode/decode functions look like this:

```js
import jwt from 'jsonwebtoken'
import CryptoJS from 'crypto-js'


const encode = async ({ secret, key = secret, token, maxAge }) => {
  const signedToken = await sign({ secret, token, maxAge })
  const encryptedToken = await encrypt({ key, token: signedToken })
  return encryptedToken
}

const decode = async ({ secret, key = secret, token, maxAge }) => {
  const decryptedToken = await decrypt({ key, token })
  const verifiedToken = await verify({ secret, token: decryptedToken, maxAge })
  return verifiedToken
}


const sign = async ({ secret, token, maxAge }) => {
  // If maxAge is specified, strip existing exp value from token if set, as only
  // one of them should be specified on a JWT
  if (maxAge && token.exp) { delete token.exp }

  const signedToken = jwt.sign(token, secret, { expiresIn: maxAge })
  return signedToken
}

const verify = async ({ secret, token, maxAge }) => {
  const verifiedToken = jwt.verify(token, secret, { maxAge })
  return Promise.resolve(verifiedToken)
}

const encrypt = async ({ key, token }) => {
  const encryptedToken = CryptoJS.AES.encrypt(token, key).toString()
  return encryptedToken
}

const decrypt = async ({ key, token }) => {
  const decryptedBytes = CryptoJS.AES.decrypt(token, key)
  const decryptedToken = decryptedBytes.toString(CryptoJS.enc.Utf8)
  return decryptedToken
}
```

iaincollins:
Note: I've deleted my earlier comment, as this PR has evolved expanded over time and don't want anyone to confused by the previous summary.

This is now a much bigger update than I had planned, but it's done, and only missing documentation (plus any bug fixes for issues that get spotted).

It realistically pushes the release date for v2.0 back slightly to the features time to settle, but it addresses a wide range of issues and adds supports for many commonly requested features and I think most people will feel is well worth the wait.

Highlights:

* A database can now be optional if using OAuth for Sign in - if no database is specified, JWT sessions will be enabled automatically
* New `allowSignin(user, account)` option allows you to control who can sign in with a callback function, even if you are not using a database
* New refactored `session` and `jwt` options
* New `session.get(session)` option allows you to customise what is returned from `/api/auth/session`
* New `jwt.set(token)` option allows you to customise what is stored in the JWT
* New encode/decode functions for JWT which can be used in custom API routes
* New `allowCallbackURL(url)` option
* Verification `maxAge` is now defined on providers that use verification (e.g. email provider) so is now a provider-specific option
* JWT now both signed and encrypted. Every aspect of it can be customised.

----

Examples of new options:

```js
session: {
  // Use JWT for session - can be used with or without user database
  jwt: false, 
  
  // Seconds - How long until an idle session expires and is no longer valid
  maxAge: 30 * 24 * 60 * 60, 
  
  // Seconds - How frequently to write to the database to extend a session
  // Set to 0 to always update database. Note: This option is ignored if JWT.
  updateAge: 24 * 60 * 60 
  
  // Easily add custom properties to session (response from `/api/auth/session`)
  get: async (session) => {
    session.customSessionProperty = "ABC123"
    return session
  }
}
```

```js
jwt: {
  // Easily add custom to the JWT; updated every time it is accessed
  set: async (token) => { 
    token.customJwtProperty = "ABC123"
    return token
  }
  
  // secret: 'my-secret-123', // Secet auto-generated if not specified
    
  // Custom encode/decode functions for signing + encryption can be specified
  // if you want to override what is in the JWT or how it is signed.
  // encode: async ({ secret, key, token, maxAge }) => {},
  // decode: async ({ secret, key, token, maxAge }) => {},
}
```

```js
allowSignin: async (user, account) => {
  // Return true if user / account is not allowed to sign in
  // or false if an access denied message should be displayed
  return true
},
```

----

The built-in JWT encode/decode functions look like this:

```js
import jwt from 'jsonwebtoken'
import CryptoJS from 'crypto-js'

const encode = async ({ secret, key = secret, token = {}, maxAge }) => {
  // If maxAge is set remove any existing created/expiry dates and replace them
  if (maxAge) {
    if (token.iat) { delete token.iat }
    if (token.exp) { delete token.exp }
  }
  const signedToken = jwt.sign(token, secret, { expiresIn: maxAge })
  const encryptedToken = CryptoJS.AES.encrypt(signedToken, key).toString()
  return encryptedToken
}

const decode = async ({ secret, key = secret, token, maxAge }) => {
  if (!token) return null
  const decryptedBytes = CryptoJS.AES.decrypt(token, key)
  const decryptedToken = decryptedBytes.toString(CryptoJS.enc.Utf8)
  const verifiedToken = jwt.verify(decryptedToken, secret, { maxAge })
  return verifiedToken
}

export default {
  encode,
  decode
}
```

You can use the built-in JWT decode/encode function from a custom API route, like this:

```js
import jwt from 'next-auth/jwt'

export default async (req, res) =>  {
  // Load encrypted (and signed) token from cookie
  const useSecureCookiePrefix = process.env.NODE_ENV === 'production'
  const cookieName = `${ useSecureCookiePrefix ? '__Secure-' : '' }next-auth.session-token`
  const sessionToken = req.cookies[cookieName]

  if (!sessionToken) { return res.end(`No JWT token`) }

  try {
      // Decrypt and verify token
    const token = await jwt.decode({ secret: 'my-secret-123', token: sessionToken })
    res.end(`Contents of JWT:\n${JSON.stringify(token, null, 2)}`)
  } catch (error) {
    console.error(error)
    res.end('Unable to decode JWT')
  }
}
```

iaincollins:
Published as `next-auth@2.0.0-beta.67`

PR #241, opened 2020-06-10
Make email address optional

### LINKED ISSUE #131: Support for signing in without an email address
## Overview

Currently v2 requires an email address on on oAuth profile to be able to sign in.

An issue with this is some providers, notably Twitter, are little different to most oAuth providers in that they don't return the email address unless you explicitly enable the option.

In the case of Twitter, this is the  option "**Request email address from users**" permission:

Screenshot 2020-05-18 at 08 58 12

### Context 

v1 copes oAuth providers that didn't return an email address by creating a dummy email address that looked something like `{profileId}-{providerName}@localhost.localdomain` (see [this line](https://github.com/iaincollins/next-auth/blob/v1/src/passport-strategies.js#L122) and [these lines](https://github.com/iaincollins/next-auth/blob/v1/src/passport-strategies.js#L90) from v1). 

This ensured the accounts had a unique, well formed (but not real) email address associated with them.

It looks for email addresses like this and replace them with the users real email address if they ever link their account with with another service that did have an email address.

*NB: Account linking is technically possible in v2 (you just sign in, then sign in again with a different service) but it's not clear yet, and you can't unlink accounts.*

Users could also specify a valid email address if they were signed in; this is not something that is supported in v2 yet (when it is, it will work differently by requesting users confirm the email address before it is changed in the DB).

### What should we do?

The easiest solution is to tell people to enable the option to return the email address in Twitter.

We could put a note about this in the docs.

Some other things we could do:

* We could improve the error messaging - eg return a specific error page that explains what is wrong (this might be the easiest thing; and we could even do this first, and change it later).

* We could try to do the same behaviour as v1 and support creating accounts without an email address. This is a great end user experience but requires some hackery like the above code I've linked to in v1 to implement and might be confusing for developers to have users with weird address in their user table if they forget to enable this option on sites like Twitter.

* Perhaps we could just put special logic in the Twitter `profile()` handler, to return a dummy email address (e.g. `{profileId}:{providerName}@localhost.localdomain`)?

iaincollins:
I think we have a working solution to allows signing in without an email address that would resolve #131 and work for everybody.

* Tested with MySQL, Postgres and MongoDB
* Tested with no database
* Tested with multiple combinations of OAuth accounts from a range of providers, with and without email addresses
* `allowSignin()` is now also passed the full raw oAuthProfile object on sign in (see docs)
* `jwt.set()` is now also passed the full raw oAuthProfile object after first sign in (see docs)

Combined with the above, we no longer need to cater for error handling for signing in without an email address as this can be done in all scenarios by using the `allowSignin()` handler and a custom error page for anyone that wants to do that (although I suspect that will be an edge case).

Additionally, now that these functions exist and the full oAuthProfile data passed to them, we don't need to have constraints on providers like caring if a provider things an email address is verified or not, as that can and should be checked in the `allowSignin()` callback.

Note: It's still possible the name of the `allowSignin()` callback (which was introduced in an earlier release) will change before v2 release so that it will align with future callback methods.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "2.0.0-beta.68",
+  "version": "2.0.0-beta.70",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -18,8 +18,8 @@
     "publish:canary": "npm publish --tag canary",
     "lint": "standard",
     "lint:fix": "standard --fix",
-    "test:db:start": "docker-compose -f test/docker/docker-compose.yml up",
-    "test:db:stop": "docker-compose -f test/docker/docker-compose.yml down"
+    "test:db:run": "docker-compose -f test/docker/docker-compose.yml up",
+    "test:db:clean": "docker-compose -f test/docker/docker-compose.yml down"
   },
   "files": [
     "dist",
--- a/src/adapters/typeorm/index.js
+++ b/src/adapters/typeorm/index.js
@@ -174,7 +174,7 @@ const Adapter = (config, options = {}) => {
 
     // Display debug output if debug option enabled
     // @TODO Refactor logger so is passed in appOptions
-    function _debug (debugCode, ...args) {
+    function debugMessage (debugCode, ...args) {
       if (appOptions.debug) {
         logger.debug(debugCode, ...args)
       }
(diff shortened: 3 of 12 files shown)


iaincollins:
I had to do a follow up fix, but I am happy this works as intended now - after multiple runs of manual testing.

I'm going to merge this in now but I am not going to publish it just yet - to bundle multiple potentially breaking changes up into a one release I'm going to work on a couple of things before publish a new release:

1. Some minor work on the models first, including adding `created at` / `updated at` value to all rows and a `last active` and `email verified` properties to user model.

I'll also be [documenting how to extend them](https://typeorm.io/#/separating-entity-definition) and the database structure.

Mostly this is just adding to documenting the models and schemas.

2. I'm going to define the events / callback system

This will mean a breaking change to `allowSignin()` and `allowCallbackURL()` and the `session.get` and `jwt.set` methods, but will provide a consistent API for all events / callbacks.

The option will probably look something like this:

```js
events: {
  signin,
  signout,
  createUser,
  updateUser,
  deleteUser,
  linkAccount,
  unlinkAccount,
  redirect,
  getSession,
  setJwt
}
```

Issue #244, opened 2020-06-11
getJWT throws an error


lnikell:
I'm using the current master version and nextjs 9.4.4. I tried to use getJwt method as described in the documentation and got an error that can't get '__Secure-next-auth.session-token' of undefined. During the investigation I found a bug that in the source code:
```
const getJwt = async ({ req, secret, cookieName }) => {
  if (!req || !secret) throw new Error("Must pass { req, secret } to getJWT()");
  const secureCookieName = "__Secure-next-auth.session-token";
  const insecureCookieName = "next-auth.session-token";
  const cookieValue = cookieName
    ? req.cookies[cookieName]
    : req.cookies[insecureCookieName];

  if (!cookieValue) {
    return null;
  }
```
req.cookies used however cookies stored in req.headers.cookie object

iaincollins:
The function `req.cookies` middleware is [built-in to Next.js](https://nextjs.org/docs/api-routes/api-middlewares) and is also is used internally by NextAuth.js. If the `getJwt()` function isn't working it would be helpful to provide an example of how you are using it.

iaincollins:
I've just [checked the source](https://github.com/iaincollins/next-auth/blob/master/src/lib/jwt.js) and that source code above doesn't match the source of the `getJwt()` method in master, or any version of it from source control so I'm not sure where it comes from.

This is the only version of the `getJwt()` function in the history for the master branch:

```js
const getJwt = async ({ req, secret, cookieName }) => {
  if (!req || !secret) throw new Error('Must pass { req, secret } to getJWT()')

  const secureCookieName = '__Secure-next-auth.session-token'
  const insecureCookieName = 'next-auth.session-token'
  const cookieValue = cookieName ? req.cookies[cookieName] : req.cookies[secureCookieName] || req.cookies[insecureCookieName]

  if (!cookieValue) { return null }

  try {
    return await decode({ secret, token: cookieValue })
  } catch (error) {
    return null
  }
}
```

iaincollins:
The the live demo at [next-auth-example.now.sh](https://next-auth-example.now.sh/) has been updated to use this method as a practical example of how to implement the usage as shown in the documentation.

Once signed in, you can view the decrypted JWT for your session here:
[next-auth-example.now.sh/api/jwt-example](https://next-auth-example.now.sh/api/jwt-example)

[...]

Issue #243, opened 2020-06-11
How to get contents of encrypted JWT token?


lnikell:
What could be the best way of getting encrypted JWT token on the server-side and client-side?
There could be a need to pass JWT token further down to another API service.
Currently, I found the way to sign JWT decrypted token again using session config in [...nextauth].js file
```
  session: {
    jwt: true,
    get: async (session, decryptedToken) => {
      session.token = await jwt.encode({
        secret:  "" ,
        token: decryptedToken,
        maxAge: 24 * 60 * 60,
      });
      return session;
    },
  },
```

iaincollins:
The `session.get()` callback allows you to return custom data to the front end, such as from a database or encrypted in a JWT, but you cannot modify the JWT from it.

The callback `jwt.set()` allows you to read and to modify the JWT and is also called when a session is checked. Any changes made to the token there are saved in the encrypted JWT.

The `session.get()` callback does not currently let you modify the JWT or resign it but this might change in future if is merged with behaviour of `jwt.set()`.

Note: Both these callbacks may be renamed in future, but the functionality will remain the same.

You can also read a JWT from an API route, like this:

```js
import jwt from 'next-auth/jwt'

const secret = process.env.JWT_SECRET

export default async (req, res) =>  {
  // Automatically decrypts and verifies JWT
  const token = await jwt.getJwt({ req, secret })
  res.end(JSON.stringify(token, null, 2))
}
```

Note: The secret needs to match the secret set for `jwt.secret`.

lnikell:
@iaincollins I have a question and proposal. 
I guess it's a common practice to send JWT tokens further down to different API services. Currently in the session as far as I understand you store signed JWT then encrypted by AES algorithm using again secret.
I guess you do it for security reasons to don't store token by itself on client side that could be used for access as well. So I think it could be good to add extra options to retrieve original encrypted token from the server-side.  I will send you PR for this.

Pixelatex:
I found this issue while searching for a way to get the original un-decrypted JWT-token.

getJWT is not an actual function and was probably renamed? There is no reference on it anywhere.

I'm posting my solution here for others that arrive to this issue via searching:

```
  const token = await jwt
    .getToken({ req, secret, raw: true })
    .catch((e) => console.error(e))
  res.send(JSON.stringify(token, null, 2))
```
So yes, simply adding the `raw` option does the trick.
[...]

Issue #247, opened 2020-06-12
New Callbacks / Events system


iaincollins:
**Summary of proposed feature**

We should have more complete and consistent events / callback system.

**Purpose of proposed feature**

It helps addresses issues raised in #100 and other issues people have raised with regard to auditing and customisation of behaviour, by providing a consistent approach for the entire application for all events / callbacks.

**Detail of the proposed**

It is proposed to remove the current methods:

* `allowSignin()`
* `allowCallbackURL()`
* `session.get()`
* `jwt.set()`

And replace them with new methods, as `callbacks` and `events`:

### Callbacks 

Callbacks are asynchronous functions that return a response that can be used to control application. behaviour.

For example, allowing a user to sign in or or not, changing a URL a user is about to redirected to, or altering the contents of  Session object or JSON Web Token before it is returned.
 
```js
callbacks: {
  signin,
  redirect,
  session,
  jwt
}
```

### Events

Events are asynchronous functions that do not return a response.

They are useful for audit logging.

```js
events: {
  signin,
  signout,
  createUser,
  linkAccount,
  session,
  error
}
```

**Potential problems**

This is a breaking change, but the options being removed would have very similar replacements.

**Describe any alternatives you've considered**

N/A

**Additional context**

N/A

iaincollins:
Additional events will be added in future, as features are added to 2.0 (e.g. unlinking accounts, deleting accounts, etc).

shamoons:
Any example of how to actually use these new events?

iaincollins:
They are documented here https://next-auth.js.org/configuration/callbacks

There is an example of using the signin callback in the comments of this issue:
https://github.com/iaincollins/next-auth/issues/266

At some point will get around to creating some examples of how to use them for specific problems that come up a lot (e.g. accessing Refresh or Access Tokens).

shamoons:
What about an example for `events`

iaincollins:
They work the same way, but are much simpler.

This is also covered in the documentation. 

https://next-auth.js.org/configuration/options#events

iaincollins:
I think we will write some tutorial articles / blog posts in a bit, that show how to use the features more clearly in practice, probably after launch, once the API is final :-)

PR #249, opened 2020-06-12
Added function for getting original JWT

lnikell:
That PR includes the following changes:
- `getJwt` function renamed to `getJwtPayload`. It was confusing that getJwt returns token payload rather than token by itself that what I expect to see from the name. 
- `getJwt `function returns signed token that could be used for the purpose of forwarding it to other API service
- Removed async from `getJwt` function since there is no need for this. Actually I checked and there is no need for async declaration for all function inside `src/lib/jwt.js`
- Added `.prettierrc` config, it will allow us to have similar formatting rules for all contributors.

Related issues: #244 #243

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,3 @@
+package.json
+package-lock.json
+dist
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,7 @@
+{
+  "endOfLine": "lf",
+  "semi": false,
+  "singleQuote": true,
+  "tabWidth": 2,
+  "trailingComma": "es5"
+}
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -1,7 +1,14 @@
 // fetch() is built in to Next.js 9.4 (you can use a polyfill if using an older version)
 /* global fetch:false */
-import { useState, useEffect, useContext, createContext, createElement } from 'react'
+import {
+  useState,
+  useEffect,
+  useContext,
+  createContext,
+  createElement
+} from 'react'
 import logger from '../lib/logger'
+import { parseCookies } from '../utils'
 
 // Note: In calls to fetch() from universal methods, all cookies are passed
 // through from the browser, when the server makes the HTTP request, so that
@@ -62,7 +69,9 @@ const useSessionData = (session) => {
       logger.error('CLIENT_USE_SESSION_ERROR', error)
     }
   }
-  useEffect(() => { _getSession() }, [])
+  useEffect(() => {
+    _getSession()
+  }, [])
   return [data, loading]
 }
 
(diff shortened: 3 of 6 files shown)


iaincollins:
Thanks for the PR, I appreciate it!

This isn't direction we are taking that is [on the roadmap](https://github.com/iaincollins/next-auth/milestone/1), specifically we have plans for the [event / callback model](https://github.com/iaincollins/next-auth/issues/247) to align the various use cases for callbacks and events.

If people want to pass data between the JWT and the client the way for them do to this is to use the session callback / event handler.

This is currently `session.get()`. In the new model it will be the `session()` handler. It will also be possible to define the initial JWT contents in the `signin()` handler (e.g. populate it with data returned from the OAuth provider on sign in).

The intention is to try and create a more consistent API than we have currently, and - and I appreciate this is not yet outline anywhere yet - is to honour features such as custom encode/decode routines (which are undocumented and only partially supported, but are going to be critical for some use case).

RE: Accessing encrypted tokens / passing them through 

From a Next.js route it's only one line to read the encrypted JWT, so I don't have any plans to provide a method for that as it's pretty straightforward.

e.g.

```js
return encryptedJwt = req.cookies['__Secure-next-auth.session-token'] || req.cookies['next-auth.session-token]
```

The client is Universal / Isomorphic, and actually does something like this to pass through the raw cookies from the client when fetching a session when the `getSession()` function is run server side:

e.g. you can do something like this from any server side route to pass through the JWT cookie:

```js
const fetchRes = fetch('https://example.com/path/to/service', { headers: { cookie: req.headers.cookie } })
```

I think this is something I think we should have in the documentation we add for JWT, which I intend to do once the refactoring of the API is complete.

lnikell:
Hi @iaincollins, thank you for the feedback. 
> RE: Accessing encrypted tokens / passing them through
I guess there is a misunderstanding. The goal is to get signed JWT since it could be verified by other API services. Using 
```
req.cookies['__Secure-next-auth.session-token']
```
That example will give you encrypted with AES JWT. So generally speaking currently in the code we sign JWT and that process actually also is the encryption process, and then we encrypt it again with AES, which is done I guess for security reasons. That AES encrypted token you can't pass anywhere else, and for example, for services like Hasura, it won't work since they support JWT and not some JWT encrypted with something else.

iaincollins:
> I guess there is a misunderstanding. The goal is to get signed JWT since it could be verified by other API services.

Okies, that makes sense, thanks for the explanation!

The original version of the JWT function (while it was in development) exported separate sign, verify, encrypt and decrypt functions to support this, before they were combined in to encrypt and decrypt to keep things simpler for everyone.

For now people can specify custom encrypt and decrypt functions on the `jwt` object (though I am not sure this is documented yet!…), or just decode the JWT in the normal way, but I'll keep this use case in mind when I look at #247 so that this is possible!

PR #260, opened 2020-06-14
Add callbacks and events

### LINKED ISSUE #247: New Callbacks / Events system

iaincollins:
This PR resolves #247

## Breaking Changes

The following callbacks have been removed and replaced:

* `allowSignin()`
* `allowCallbackURL()`
* `session.get()`
* `jwt.set()`

### New Callbacks

The following new callbacks have been created, you can specify asynchronous functions for any of these callbacks, by specifying them in the `callbacks` option.

```js
callbacks: {
  signin,
  redirect,
  session,
  jwt
}
```

### New Events

Events are asynchronous functions that do not return a response. You can specify asynchronous functions for any of these events, by specifying them in the `events ` option.

Events are useful for audit logging.

```js
events: {
  signin,
  signout,
  createUser,
  linkAccount,
  session,
  error
}
```

## Callbacks

Built in callback functions and descriptions of when they are invoked, their parameters and their responses.

```js
/**
 * Use the signin callback to control if a user is allowed to sign in or not.
 *
 * This is triggered before sign in flow completes, so the user profile may be
 * a user object (with an ID) or it may be just their name and email address,
 * depending on the sign in flow and if they have an account already.
 *
 * When using email sign in, this method is triggered both when the user
 * requests to sign in and again when they activate the link in the sign in
 * email.
 *
 * @param  {object} profile  User profile (e.g. user id, name, email)
 * @param  {object} account  Account used to sign in (e.g. OAuth account)
 * @param  {object} metadata Provider specific metadata (e.g. OAuth Profile)
 * @return {boolean|object}  Return `true` (or a modified JWT) to allow sign in
 *                           Return `false` to deny access
 */
const signin = async (profile, account, metadata) => {
  const isAllowedToSignIn = true
  if (isAllowedToSignIn) {
    return Promise.resolve(true)
  } else {
    return Promise.resolve(false)
  }
}

/**
 * Redirect is called anytime the user is redirected on signin or signout.
 * By default, for security, only Callback URLs on the same URL as the site
 * are allowed, you can use this callback to customise that behaviour.
 *
 * @param  {string} url      URL provided as callback URL by the client
 * @param  {string} baseUrl  Default base URL of site (can be used as fallback)
 * @return {string}          URL the client will be redirect to
 */
const redirect = async (url, baseUrl) => {
  return url.startsWith(baseUrl)
    ? Promise.resolve(url)
    : Promise.resolve(baseUrl)
}

/**
 * The session callback is called whenever a session is checked.
 * e.g. `getSession()`, `useSession()`, `/api/auth/session` (etc)
 *
 * @param  {object} session  Session object
 * @param  {object} token    JSON Web Token
 * @return {object}          Session that will be returned to the client 
 */
const session = async (session) => {
  return Promise.resolve(session)
}

/**
 * This callback is called whenever a JSON Web Token is created / updated.
 * e.g. On sign in, `getSession()`, `useSession()`, `/api/auth/session` (etc)
 * 
 * On initial sign in, the raw oAuthProfile is passed if the user is signing in 
 * with an OAuth provider. It is not avalible on subsequent calls. You can
 * take advantage of this to persist additional data you need to in the JWT.
 *
 * @param  {object} token         Decrypted JSON Web Token
 * @param  {object} oAuthProfile  OAuth profile - only available on sign in
 * @return {object}               JSON Web Token that will be saved
 */
const jwt = async (token, oAuthProfile) => {
  return Promise.resolve(token)
}

export default {
  signin,
  redirect,
  session,
  jwt
}
```

## Events

Built in event handlers descriptions of when they are invoked.

```js
const signin = async (message) => {
  // Event triggered on successful sign in
 }

const signout = async (message) => { 
  // Event triggered on signout
}

const createUser = async (message) => {
  // Event triggered on user creation
}

const linkAccount = async (message) => {
  // Event triggered when an account is linked to a user
}

const session = async (message) => {
  // Event triggered when a session is active
}

const error = async (message) => {
  // @TODO Event triggered when something goes wrong in an authentication flow
  // This event may be fired multiple times when an error occurs
}

export default {
  signin,
  signout,
  createUser,
  linkAccount,
  session,
  error
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "2.0.0-beta.74",
+  "version": "2.0.0-beta.75",
   "description": "An authentication library for Next.js",
   "repository": "https://github.com/iaincollins/next-auth.git",
   "author": "Iain Collins <me@iaincollins.com>",
--- a/src/adapters/typeorm/index.js
+++ b/src/adapters/typeorm/index.js
@@ -320,7 +320,7 @@ const Adapter = (typeOrmConfig, options = {}) => {
       debugMessage('GET_VERIFICATION_REQUEST', identifer, token)
       try {
         // Hash token provided with secret before trying to match it with datbase
-        // @TODO Use bcrypt function here instead of simple salted hash
+        // @TODO Use bcrypt instead of salted SHA-256 hash for token
         const hashedToken = createHash('sha256').update(`${token}${secret}`).digest('hex')
         const verificationRequest = await connection.getRepository(VerificationRequest).findOne({ identifer, token: hashedToken })
 
--- a/src/lib/logger.js
+++ b/src/lib/logger.js
@@ -11,10 +11,12 @@ const logger = {
     }
   },
   debug: (debugCode, ...text) => {
-    console.log(
-      `[next-auth][debug][${debugCode}]`,
-      text
-    )
+    if (process && process.env && process.env._NEXT_AUTH_DEBUG) {
+      console.log(
+        `[next-auth][debug][${debugCode}]`,
+        text
+      )
+    }
   }
 }
 
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -3,6 +3,8 @@ import jwt from '../lib/jwt'
 import cookie from './lib/cookie'
 import callbackUrlHandler from './lib/callback-url-handler'
 import parseProviders from './lib/providers'
+import events from './lib/events'
+import callbacks from './lib/callbacks'
 import providers from './routes/providers'
 import signin from './routes/signin'
 import signout from './routes/signout'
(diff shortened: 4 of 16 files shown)


iaincollins:
Published as `next-auth@2.0.0-beta.75`

PR #262, opened 2020-06-15
Add limited support for credentials sign in

iaincollins:
The functionality in this update  is limited but working and the feature is fully documented.

You can use any credentials and backend or crypto token and you can easily persist data in the encrypted JSON Web Token.

The constraints of this implementation are that users authenticated with a Credentials based provider are not persisted to the database and JSON Web Tokens need to be enabled.

I expect we will review this in future and see how we might want to extend or improve on it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "2.0.0-beta.76",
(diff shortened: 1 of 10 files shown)


iaincollins:
Published as `next-auth@2.0.0-beta.77`

PR #267, opened 2020-06-15
Rename "Identifer" to "Identifier"

JeffersonBledsoe:
As far as I can tell, the parameter "identifer" is a typo which seems to have stuck and ended up in the documentation. This PR renames all instances of "identifer" to "identifier".


LoriKarikari:
thanks for the PR @JeffersonBledsoe

PR #272, opened 2020-06-16
Fix documentation

laeckerv:
Somehow the example wasn't correct. The example imported `session` instead of `getSession` from `next-auth/client`.

I also adapted the other imports to be consistent.


iaincollins:
Thanks!

It was like this NexAuth.js as client exports `session`, `providers` and `csrfToken` (those were the original method names) but these often clash with variable names people use to store the result of them if they are using deconstruction to get the function names, so now it also exports them as `getSession()`, `getProviders()` and `getCsrfToken()`.

Issue #254, opened 2020-06-13
Request For Comments - Database Schema


iaincollins:
Please take a look at the database schema documentation and leave feedback here.

**Of primary interest is the feedback on types used for reach property.**

We are not too interested in bike shedding over property names, except in cases where there is a strong supporting citation (e.g. document from a standard bodies); no personal preferences please - but if you can point to common conventions or standards and include citations, please do!

Please bear in mind the models are intended to work with any ANSI SQL database (as well as Document databases, like MongoDB, with some transformation); so features that tend to be database specific (such as relationship mapping) are out of scope.

You can find more documentation about the schema at https://next-auth.js.org/schemas/models

iaincollins:
Note: The conversation on here is locked to avoid people leaving feedback in more than one place, please leave it on the pull request!

LoriKarikari:
In DBs they usually use `created_at` and `updated_at` or their camelCase variants for NoSQL instead of `created` and `updated`.  Don't want to nitpick over names as you said but it's nice to have some consistency with the rest of your database.

iaincollins:
Thanks for looking and feeding back! Yeah, I was torn over that one…

I've always used `created_on` / `updated_on` as that used to be really common, but it does *seem* like `created_at` is more popular, at least these days?

I didn't have any hard evidence and could find any anything like StandardJS but for SQL naming conventions, so I tried to avoid it by not using either `_on` or `_at`.

I think I saw on my feed that you also work on a database related tool, so I'm guessing you are more in touch with what is the norm these days than I am!

If we think `_at` is better and nobody contests that happy to change it!

LoriKarikari:
Also why the several `not null` `default null`? This is the same as just saying that the field can be nullable as you add a default value that is null anyway. Or is there any reason for this?

iaincollins:
Oh! I think that is a misleading artifact of how have abstracted the metadata for the types.

I was planning to update the docs with the actual SQL used to create the tables (including indexes) at some point when I figure that out.

It's easy to get the structure with MySQL but it's harder with Postgres as it doesn't have a `DESCRIBE` command in SQL. It does have \d in the command line app, but that is harder to access from an integration test.

I'm was thinking if seeing if I could capture the SQL TypeORM generates and using that for documentation and in the tests; it would be more complete and more useful.

iaincollins:
[...]

Issue #66, opened 2018-12-17
Automate testing; add code coverage


iaincollins:
There isn't any formal testing and at this point it's a defect in the project and should be a priority.

This issue is something of a blocker for #226 (automating publication of new releases).

The focus should probably be on having end-to-end acceptances tests at first, as those have more value than unit tests. We don't need to test everything, but we do need some sort of smoketest. Puppeteer seems like the best candidate for this.

## Options

* Mocha + Chai
* Jest + Enzyme (great for testing React and Node.js projects in the same repo)
* Puppeteer

## Scope

There is already a docker compose suite that spins up MySQL, Postgres and MongoDB instances.

* I think it would be sensible to test all those databases with two OAuth providers plus the email  sign in flow (even if the email flow has to be partly mocked).
* We should also test the same flows with no database configured.
* Ideally we would also test with a database and JWT enabled (but if the above two flows work, this is likely to work also).
* We should also test the database schema for each of those databases; a snapshot test for those is probably sufficient, just so schema changes are flagged.

## Feedback

As this is a topic that attracts a lot of bike-shedding I'm going to come out and say that we are really only looking for feedback from either existing contributors or from people interested in helping with authoring tests.

JeffersonBledsoe:
Just wanted to chime in, I'd be up for helping with writing some E2E tests! I've yet to use next-auth, but plan to use it on a couple of personal projects. Would a fresh look on things be beneficial for test writing, or would you rather somebody with deeper knowledge of next-auth?

Could I also recommend taking a look at [Cypress](https://www.cypress.io/) as another potential option for the E2E tests. It's API is just as nice as Puppeteer, but I've found it to be more stable and has easier to write best-practise tests (mostly due to great documentation and a good community around it)

iaincollins:
> Just wanted to chime in, I'd be up for helping with writing some E2E tests! I've yet to use next-auth, but plan to use it on a couple of personal projects. Would a fresh look on things be beneficial for test writing, or would you rather somebody with deeper knowledge of next-auth?

That would be fantastic - a fresh pair of eyes would be really helpful!

I'm very happy to collaborate on this, to help where existing knowledge of next-auth is useful.

I would have started with tests but from a practical perspective I really needed to have a working solution for a couple of projects (as v1 isn't viable for serverless) and wouldn't have been able to do that as quickly; but I'd really like that to be the focus now, to give us confidence when working on changes post-release.

[...]

PR #274, opened 2020-06-16
Rename created and updated properties on models

iaincollins:
Based on changes suggested in #254

* Renames `created` property to `created_at` (`createdAt` on MongoDB) on all models
* Renames `updated` property to `updated_at` (`updated At` on MongoDB) on all models
* Improves schema tests and related documentation


iaincollins:
## What I did

I thought I would document what I did as this "process" (which is a strong word for it…) is not documented anywhere right now! I wouldn't mind refactoring the database tests as part of this PR.

There is potentially a wider question of how annoying the build process is and if automatic magnification is annoying that you and other contributors might have a view on.

### Start test databases

To get started, I loaded `testdocker/docker-compose.yml`, which includes Docker instances of MySQL, Postgres and MongoDB - this obviously Docker, but doesn't require anything else.

**Note: They all listen locally on their default ports, so will clash if you are already running any of these databases on your system. This is something I could address in the tests by running them from inside a Docker container.**

The Docker images have hard coded auth credentials (`nextauth:password`) and include a database called `nextauth`.

1. Reset databases with `npm run test:db:clean`
2. Start databases with `npm run test:db:run`
3. Check everything is working with `npm run test` (should not error)
4. Start `npm run watch` in the background - to build from source, and trigger new builds when we start refactoring the models

I use the commands to run/clean up the databases during manual testing too (as it's a lot easier to test against different databases that way).

### Start test databases

1. Update schemas in `test/schemas/*` to expect `created_at` and `updated_at`
2. Run `npm run test` - it should error and print out the schema it found; you can compare this against the one it was expecting in `test/schemas/*`

### Modify schemas

1. Change the schemas in `src/adapters/typeorm/models` - this was a very simple and small change.
2. . Run `npm run test` to run the tests - I say tests, but right now they are just node scripts that exit with success or fail states.

The test actually failed on me on first run as the column order on MySQL had changed between updates, but i was able to see that easily from the diff and update the fixture in `test/schemas`.

The whole process is a bit janky but it's helpful to have some level of automation. I am happy to work on improving the process in this PR, including improving the data structure and quality of the fixtures.

iaincollins:
@LoriKarikari Based on feedback I've removed the 'default' value when the field is not nullable so it's more clear if a field actually has a default value or not.

Note: I don't actually think hand rolling something like this is a great idea, I was just out of other better ways to do this in the short term. I think the new test stuff that is coming will result in something much better (and we can uses some of this to help get there I think) 

RE: @JeffersonBledsoe / #66

PR #282, opened 2020-06-18
Fix typo in auth0 provider config

dnbkr:
auth0 was configured with `respponse_type` but should be `response_type`


iaincollins:
Thanks for the fix! Will go out in the next release.

PR #286, opened 2020-06-19
Gitlab Provider

SabariVig:
Added Gitlab Provider


iaincollins:
Thanks @SabariVig - appreciate the documentation and fixes to the broken URLs too!

For consistency, we will probably use the name `GitLab` (vs `Gitlab`) when it's published.

I hope it will go out in an update later today.

SabariVig:
Hey There I just realized that gitlab does nit show up in the dropdown list in the documentation website. I went through the code cannot figure out why its causing the issue

iaincollins:
Ah, good catch! Forgot about that. 

It looks like we need to add GitLab, Box and Yandex to `www/sidebars.js`.

PR #291, opened 2020-06-20
Added Providers To Sidebar

SabariVig:


iaincollins:
Thanks for doing this!

Not sure why the build has failed with that error, I tested it locally and it's fine so merging it in (and will check it deploys okay).

PR #292, opened 2020-06-21
Add `email verified` timestamp; refactor timestamps

iaincollins:
## Changes in this pull request

* The `email_verified` / `emailVerified` property is now populated when a sign in link is used.
* Some timestamp related bugs were highlighted in testing and prompted some refactoring.
* All timestamps are now stored constantly as UTC in MySQL, Postgres and MongoDB.
* For MySQL, a default timezone of UTC (`Z` for Zulu time) is now set in the adapter.
* For Postgres, a transform to use the time `timestamp with time zone` (instead of `timestamp without time zone`) is used to address this.
* For MongoDB, refactoring to pass timestamps as JavaScript date objects was enough to ensure all timestamps are stored in UTC as `ISODate()` objects.
* Using JavaScript data objects - rather than ISO strings - also resolved issues with other databases.
* Documentation and tests updated.

## Context

Supporting timezones across different databases turns out to be more work than anticipated, as it's not entirely handled by TypeORM.

There is quite some history with times zone handling with TypeORM and Postgres in particular - which seems to have been made worse by quirky Postgres driver behaviour - and I had to sift through a lot of issues and a lot of debate (and quite a lot of bad and/or outdated advice) to try and work out the best strategy.

After some extensive testing, these changes seems to work regardless of the server, database or client timezones, but this is a complex topic and issues related to it may come up again in future.

## Considerations

In future, we may well get bugs reported related to time zone handling, if not with the database, with the regard to cookie expiry times, session time reported by `getSession()`, or timestamps in JSON Web Tokens.

I've tried to ensure ISO timestamps with UTC values are used everywhere that is appropriate, but it's not easy to anticipate all scenarios.

If we do see issues related to timezone support in future, I think we should tread very carefully and not rush to make changes. This is an area I'm looking forward to writing some unit and integration tests for.

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -1,7 +1,6 @@
 {
   "presets": [
     ["@babel/preset-env", { "targets": { "esmodules": true } } ],
-    ["minify", { "builtIns": false } ]
   ],
   "comments": false,
   "overrides": [
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "2.0.0-beta.82",
+  "version": "2.0.0-beta.83",
   "description": "An authentication library for Next.js",
   "repository": "https://github.com/iaincollins/next-auth.git",
   "author": "Iain Collins <me@iaincollins.com>",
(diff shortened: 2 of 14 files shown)


iaincollins:
Published as `next-auth@2.0.0-beta.83`.

If you think you see any conceptual problems with this Pull Request, please comment on this PR or (even better) raise as an issue / bug report and link back to this PR from the issue if seems relevant.

PR #294, opened 2020-06-21
Improve client session syncing

### LINKED ISSUE #252: Support Provider Session Changes
Right now, it is possible to attach additional properties to the Session object by overriding the `session.get` option. This feature can be used to join data from other tables in the database.
If the session is attached to the global context using the `Provider` component under the main App, it is impossible to change it later.
Therefore, if the underlying value of the property changes, there is no way to force a re-fetch of the Session object from the server. The object in the client will be stale.

There should be some way to demand a session re-fetch. For example, the `useSession` hook could return a function that would trigger a re-fetch if called.

iaincollins:
* Session state is now synced across Windows and Tabs.
* Adds `NEXTAUTH_CLIENT_MAXAGE` option (set via env var in `next.config.js`) to tell the client to sync the session state with the server every `X` seconds (defaults to `0`, which is never).

Resolves #252 

NextAuth-Sync

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "2.0.0-beta.84",
+  "version": "2.0.0-beta.85",
   "description": "An authentication library for Next.js",
   "repository": "https://github.com/iaincollins/next-auth.git",
   "author": "Iain Collins <me@iaincollins.com>",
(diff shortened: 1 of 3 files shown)


iaincollins:
Published as `next-auth@2.0.0-beta.85`

PR #302, opened 2020-06-22
feat(adapter): Add opinionated prisma adapter

Fumler:
Tried it locally with a Google provider and it works with sessions and all that. I'm not sure about the new providers like `email`, so would appreciate if someone could test that.

As for adding the adapter, I think this might be a bit too opinionated, but I figured as long as there is documentation on how people can set this up it should be fine.

Happy for any feedback.

### DIFF
--- a/src/adapters/index.js
+++ b/src/adapters/index.js
@@ -1,6 +1,8 @@
 import TypeORM from './typeorm'
+import Prisma from './prisma'
 
 export default {
   Default: TypeORM.Adapter,
-  TypeORM
+  TypeORM,
+  Prisma
 }
--- a/src/adapters/prisma/index.js
+++ b/src/adapters/prisma/index.js
@@ -1,0 +1,325 @@
+import { createHash } from 'crypto'
+
+import { CreateUserError } from '../../lib/errors'
+import logger from '../../lib/logger'
+
+const Adapter = (prismaConfig, options = {}) => {
+  const { prisma, modelMapping = {
+    User: 'user',
+    Account: 'account',
+    Session: 'session',
+    VerificationRequest: 'verificationRequest'
+  }} = prismaConfig
+
+  const { User, Account, Session, VerificationRequest } = modelMapping
+
+  async function getAdapter (appOptions) {
+    function debugMessage(debugCode, ...args) {
+      if (appOptions && appOptions.debug) {
+        logger.debug(`PRISMA_${debugCode}`, ...args)
+      }
+    }
+
+    if (appOptions && (!appOptions.session || !appOptions.session.maxAge)) {
+      debugMessage('GET_ADAPTER', 'Session expiry not configured (defaulting to 30 days')
+    }
+    const defaultSessionMaxAge = 30 * 24 * 60 * 60 * 1000
+    const sessionMaxAge = (appOptions && appOptions.session && appOptions.session.maxAge)
+      ? appOptions.session.maxAge * 1000
+      : defaultSessionMaxAge
+    const sessionUpdateAge = (appOptions && appOptions.session && appOptions.session.updateAge)
+      ? appOptions.session.updateAge * 1000
+      : 0
+
+    async function createUser (profile) {
+      debugMessage('CREATE_USER', profile)
+      try {
+        return prisma[User].create({
+          data: {
+            name: profile.name,
+            email: profile.email,
+            image: profile.image,
+            emailVerified: profile.emailVerified && Object.prototype.toString.call(profile.emailVerified) === '[object Date]' ? profile.emailVerified.toISOString() : null
+          }
+        })
+      } catch (error) {
+        logger.error('CREATE_USER_ERROR', error)
+        return Promise.reject(new CreateUserError(error))
+      }
+    }
+
+    async function getUser (id) {
+      debugMessage('GET_USER', id)
+      try {
+        return prisma[User].findOne({ where: { id } })
+      } catch (error) {
+        logger.error('GET_USER_BY_ID_ERROR', error)
+        return Promise.reject(new Error('GET_USER_BY_ID_ERROR', error))
+      }
+    }
+
+    async function getUserByEmail (email) {
+      debugMessage('GET_USER_BY_EMAIL', email)
+      try {
+        if (!email) { return Promise.resolve(null) }
+        return prisma[User].findOne({ where: { email } })
+      } catch (error) {
+        logger.error('GET_USER_BY_EMAIL_ERROR', error)
+        return Promise.reject(new Error('GET_USER_BY_EMAIL_ERROR', error))
+      }
+    }
+
+    async function getUserByProviderAccountId (providerId, providerAccountId) {
+      debugMessage('GET_USER_BY_PROVIDER_ACCOUNT_ID', providerId, providerAccountId)
+      try {
+        return prisma[Account].findOne({ where: { providerAccountId } })[User]()
+      } catch (error) {
+        logger.error('GET_USER_BY_PROVIDER_ACCOUNT_ID_ERROR', error)
+        return Promise.reject(new Error('GET_USER_BY_PROVIDER_ACCOUNT_ID_ERROR', error))
+      }
+    }
+
+    async function updateUser (user) {
+      debugMessage('UPDATE_USER', user)
+      try {
+        const { id, ...rest } = user
+        return prisma[User].update({ where: { id }, data: rest })
+      } catch (error) {
+        logger.error('UPDATE_USER_ERROR', error)
+        return Promise.reject(new Error('UPDATE_USER_ERROR', error))
+      }
+      
+    }
+
+    async function deleteUser (id) {
+      debugMessage('DELETE_USER', userId)
+      try {
+        return prisma[User].delete({ where: { id } })
+      } catch (error) {
+        logger.error('DELETE_USER_ERROR', error)
+        return Promise.reject(new Error('DELETE_USER_ERROR', error))
+      }
+    }
+
+    async function linkAccount (userId, providerId, providerType, providerAccountId, refreshToken, accessToken, accessTokenExpires) {
+      debugMessage('LINK_ACCOUNT', userId, providerId, providerType, providerAccountId, refreshToken, accessToken, accessTokenExpires)
+      try {
+        return prisma[Account].create({
+          data: {
+            accessToken,
+            refreshToken,
+            compoundId: createHash('sha256').update(`${providerId}:${providerAccountId}`).digest('hex'),
+            providerAccountId,
+            providerId,
+            providerType,
+            accessTokenExpires,
+            [User]: {
+              connect: {
+                id: userId
+              }
+            }
+          }
+        })
+      } catch (error) {
+        logger.error('LINK_ACCOUNT_ERROR', error)
+        return Promise.reject(new Error('LINK_ACCOUNT_ERROR', error))
+      }
+    }
+
+    async function unlinkAccount (userId, providerId, providerAccountId) {
+      debugMessage('UNLINK_ACCOUNT', userId, providerId, providerAccountId)
+      try {
+        return prisma[Account].delete({ where: { providerAccountId } })
+      } catch (error) {
+        logger.error('UNLINK_ACCOUNT_ERROR', error)
+        return Promise.reject(new Error('UNLINK_ACCOUNT_ERROR', error))
+      }
+    }
+
+    async function createSession (user) {
+      debugMessage('CREATE_SESSION', user)
+      try {
+        let expires = null
+        if (sessionMaxAge) {
+          const dateExpires = new Date()
+          dateExpires.setTime(dateExpires.getTime() + sessionMaxAge)
+          expires = dateExpires.toISOString()
+        }
+
+        return prisma[Session].create({
+          data: {
+            expires,
+            [User]: {
+              connect: {
+                id: user.id
+              }
+            }
+          }
+        })
+      } catch (error) {
+        logger.error('CREATE_SESSION_ERROR', error)
+        return Promise.reject(new Error('CREATE_SESSION_ERROR', error))
+      }
+    }
+
+    async function getSession (sessionToken) {
+      debugMessage('GET_SESSION', sessionToken)
+      try {
+        const session = await prisma[Session].findOne({ where: { sessionToken }})
+
+        // Check session has not expired (do not return it if it has)
+        if (session && session.expires && new Date() > session.expires) {
+          await prisma[Session].delete({ where: { sessionToken } })
+          return null
+        }
+
+        return session
+      } catch (error) {
+        logger.error('GET_SESSION_ERROR', error)
+        return Promise.reject(new Error('GET_SESSION_ERROR', error))
+      }
+    }
+
+    async function updateSession (session, force) {
+      debugMessage('UPDATE_SESSION', session)
+      try {
+        if (sessionMaxAge && (sessionUpdateAge || sessionUpdateAge === 0) && session.expires) {
+          // Calculate last updated date, to throttle write updates to database
+          // Formula: ({expiry date} - sessionMaxAge) + sessionUpdateAge
+          //     e.g. ({expiry date} - 30 days) + 1 hour
+          //
+          // Default for sessionMaxAge is 30 days.
+          // Default for sessionUpdateAge is 1 hour.
+          const dateSessionIsDueToBeUpdated = new Date(session.expires)
+          dateSessionIsDueToBeUpdated.setTime(dateSessionIsDueToBeUpdated.getTime() - sessionMaxAge)
+          dateSessionIsDueToBeUpdated.setTime(dateSessionIsDueToBeUpdated.getTime() + sessionUpdateAge)
+
+          // Trigger update of session expiry date and write to database, only
+          // if the session was last updated more than {sessionUpdateAge} ago
+          if (new Date() > dateSessionIsDueToBeUpdated) {
+            const newExpiryDate = new Date()
+            newExpiryDate.setTime(newExpiryDate.getTime() + sessionMaxAge)
+            session.expires = newExpiryDate
+          } else if (!force) {
+            return null
+          }
+        } else {
+          // If session MaxAge, session UpdateAge or session.expires are
+          // missing then don't even try to save changes, unless force is set.
+          if (!force) { return null }
+        }
+
+        const { id, ...rest } = session
+        return prisma[Session].update({ where: { id }, data: rest })
+      } catch (error) {
+        logger.error('UPDATE_SESSION_ERROR', error)
+        return Promise.reject(new Error('UPDATE_SESSION_ERROR', error))
+      }
+    }
+
+    async function deleteSession (sessionToken) {
+      debugMessage('DELETE_SESSION', sessionToken)
+      try {
+        return prisma[Session].delete({ where: { sessionToken } })
+      } catch (error) {
+        logger.error('DELETE_SESSION_ERROR', error)
+        return Promise.reject(new Error('DELETE_SESSION_ERROR', error))
+      }
+    }
+
+    async function createVerificationRequest (identifier, url, token, secret, provider) {
+      debugMessage('CREATE_VERIFICATION_REQUEST', identifier)
+      try {
+        const { site } = appOptions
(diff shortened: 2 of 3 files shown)


iaincollins:
Thank you, this is wonderful and I'm really happy to include it!

It will probably be a few days before I check it out (I'd like to take a short break XD) but I love this.

If, while testing it and learning how it works myself, would you mind if I pushed some updates to the branch for minor refactoring?

I'm not actually sure if that's something I want to do yet (or even if it makes sense with Prisma), just thinking about any email related fixes as required, and/or how I might make the models exportable / customisable if that makes sense to do.

I'm very happy to push them to another branch and raise a PR if you think it might be wiser to discuss them first (given I am not familiar with Prisma) - please do let me know if you think that might be best.

Fumler:
Feel free to push any changes you wish directly

brunocrosier:
Hey, this is awesome work @Fumler !!

I noticed one bug though. When trying this with GitHub (might happen with other auth providers too), I encountered an error where the signup / signin process would fail in Prisma because the `type` of `providerAccountId` is set as a `string` in `schema.prisma`, but it is an `int` when provided by GitHub

I simply wrapped the `providerAccountId` with `String()` in the `prisma.js` file that you wrote, and everything worked great 

```javascript
import { createHash } from "crypto"

import { CreateUserError } from "next-auth/dist/lib/errors"
import logger from "next-auth/dist/lib/logger"

const Adapter = (prismaConfig, options = {}) => {
  const {
    prisma,
    modelMapping = {
      User: "user",
      Account: "account",
      Session: "session",
      VerificationRequest: "verificationRequest",
    },
  } = prismaConfig

  const { User, Account, Session, VerificationRequest } = modelMapping

  async function getAdapter(appOptions) {
    function debugMessage(debugCode, ...args) {
      if (appOptions && appOptions.debug) {
        logger.debug(`PRISMA_${debugCode}`, ...args)
      }
    }

    if (appOptions && (!appOptions.session || !appOptions.session.maxAge)) {
      debugMessage(
        "GET_ADAPTER",
        "Session expiry not configured (defaulting to 30 days"
      )
    }
    const defaultSessionMaxAge = 30 * 24 * 60 * 60 * 1000
    const sessionMaxAge =
      appOptions && appOptions.session && appOptions.session.maxAge
        ? appOptions.session.maxAge * 1000
        : defaultSessionMaxAge
    const sessionUpdateAge =
      appOptions && appOptions.session && appOptions.session.updateAge
        ? appOptions.session.updateAge * 1000
        : 0

    async function createUser(profile) {
      debugMessage("CREATE_USER", profile)
      try {
        return prisma[User].create({
          data: {
            name: profile.name,
            email: profile.email,
            image: profile.image,
            emailVerified:
              profile.emailVerified &&
              Object.prototype.toString.call(profile.emailVerified) ===
                "[object Date]"
                ? profile.emailVerified.toISOString()
                : null,
          },
        })
      } catch (error) {
        logger.error("CREATE_USER_ERROR", error)
        return Promise.reject(new CreateUserError(error))
      }
    }

    async function getUser(id) {
      debugMessage("GET_USER", id)
      try {
        return prisma[User].findOne({ where: { id } })
      } catch (error) {
        logger.error("GET_USER_BY_ID_ERROR", error)
        return Promise.reject(new Error("GET_USER_BY_ID_ERROR", error))
      }
    }

    async function getUserByEmail(email) {
      debugMessage("GET_USER_BY_EMAIL", email)
      try {
        if (!email) {
          return Promise.resolve(null)
        }
        return prisma[User].findOne({ where: { email } })
      } catch (error) {
        logger.error("GET_USER_BY_EMAIL_ERROR", error)
        return Promise.reject(new Error("GET_USER_BY_EMAIL_ERROR", error))
      }
    }

    async function getUserByProviderAccountId(providerId, providerAccountId) {
      debugMessage(
        "GET_USER_BY_PROVIDER_ACCOUNT_ID",
        providerId,
        providerAccountId
      )
      try {
        return prisma[Account].findOne({
          where: { providerAccountId: String(providerAccountId) },
        })[User]()
      } catch (error) {
        logger.error("GET_USER_BY_PROVIDER_ACCOUNT_ID_ERROR", error)
        return Promise.reject(
          new Error("GET_USER_BY_PROVIDER_ACCOUNT_ID_ERROR", error)
        )
      }
    }

    async function updateUser(user) {
      debugMessage("UPDATE_USER", user)
      try {
        const { id, ...rest } = user
        return prisma[User].update({ where: { id }, data: rest })
      } catch (error) {
        logger.error("UPDATE_USER_ERROR", error)
        return Promise.reject(new Error("UPDATE_USER_ERROR", error))
      }
    }

    async function deleteUser(id) {
      debugMessage("DELETE_USER", userId)
      try {
        return prisma[User].delete({ where: { id } })
      } catch (error) {
        logger.error("DELETE_USER_ERROR", error)
        return Promise.reject(new Error("DELETE_USER_ERROR", error))
      }
    }

    async function linkAccount(
      userId,
      providerId,
      providerType,
      providerAccountId,
      refreshToken,
      accessToken,
      accessTokenExpires
    ) {
      debugMessage(
        "LINK_ACCOUNT",
        userId,
        providerId,
        providerType,
        providerAccountId,
        refreshToken,
        accessToken,
        accessTokenExpires
      )
      try {
        return prisma[Account].create({
          data: {
            accessToken,
            refreshToken,
            compoundId: createHash("sha256")
              .update(`${providerId}:${providerAccountId}`)
              .digest("hex"),
            providerAccountId: String(providerAccountId),
            providerId,
            providerType,
            accessTokenExpires,
            [User]: {
              connect: {
                id: userId,
              },
            },
          },
        })
      } catch (error) {
        logger.error("LINK_ACCOUNT_ERROR", error)
        return Promise.reject(new Error("LINK_ACCOUNT_ERROR", error))
      }
    }

    async function unlinkAccount(userId, providerId, providerAccountId) {
      debugMessage("UNLINK_ACCOUNT", userId, providerId, providerAccountId)
      try {
        return prisma[Account].delete({
          where: { providerAccountId: String(providerAccountId) },
        })
      } catch (error) {
        logger.error("UNLINK_ACCOUNT_ERROR", error)
        return Promise.reject(new Error("UNLINK_ACCOUNT_ERROR", error))
      }
    }

    async function createSession(user) {
      debugMessage("CREATE_SESSION", user)
      try {
        let expires = null
        if (sessionMaxAge) {
          const dateExpires = new Date()
          dateExpires.setTime(dateExpires.getTime() + sessionMaxAge)
          expires = dateExpires.toISOString()
        }

        return prisma[Session].create({
          data: {
            expires,
            [User]: {
              connect: {
                id: user.id,
              },
            },
          },
        })
      } catch (error) {
        logger.error("CREATE_SESSION_ERROR", error)
        return Promise.reject(new Error("CREATE_SESSION_ERROR", error))
      }
    }

    async function getSession(sessionToken) {
      debugMessage("GET_SESSION", sessionToken)
      try {
        const session = await prisma[Session].findOne({
          where: { sessionToken },
        })

        // Check session has not expired (do not return it if it has)
        if (session && session.expires && new Date() > session.expires) {
          await prisma[Session].delete({ where: { sessionToken } })
          return null
        }

        return session
      } catch (error) {
        logger.error("GET_SESSION_ERROR", error)
        return Promise.reject(new Error("GET_SESSION_ERROR", error))
      }
    }

    async function updateSession(session, force) {
      debugMessage("UPDATE_SESSION", session)
      try {
        if (
          sessionMaxAge &&
          (sessionUpdateAge || sessionUpdateAge === 0) &&
          session.expires
        ) {
          // Calculate last updated date, to throttle write updates to database
          // Formula: ({expiry date} - sessionMaxAge) + sessionUpdateAge
          //     e.g. ({expiry date} - 30 days) + 1 hour
          //
          // Default for sessionMaxAge is 30 days.
          // Default for sessionUpdateAge is 1 hour.
          const dateSessionIsDueToBeUpdated = new Date(session.expires)
          dateSessionIsDueToBeUpdated.setTime(
            dateSessionIsDueToBeUpdated.getTime() - sessionMaxAge
          )
          dateSessionIsDueToBeUpdated.setTime(
            dateSessionIsDueToBeUpdated.getTime() + sessionUpdateAge
          )

          // Trigger update of session expiry date and write to database, only
          // if the session was last updated more than {sessionUpdateAge} ago
          if (new Date() > dateSessionIsDueToBeUpdated) {
            const newExpiryDate = new Date()
            newExpiryDate.setTime(newExpiryDate.getTime() + sessionMaxAge)
            session.expires = newExpiryDate
          } else if (!force) {
            return null
          }
        } else {
          // If session MaxAge, session UpdateAge or session.expires are
          // missing then don't even try to save changes, unless force is set.
          if (!force) {
            return null
          }
        }

        const { id, ...rest } = session
        return prisma[Session].update({ where: { id }, data: rest })
      } catch (error) {
        logger.error("UPDATE_SESSION_ERROR", error)
        return Promise.reject(new Error("UPDATE_SESSION_ERROR", error))
      }
    }

    async function deleteSession(sessionToken) {
      debugMessage("DELETE_SESSION", sessionToken)
      try {
        return prisma[Session].delete({ where: { sessionToken } })
      } catch (error) {
        logger.error("DELETE_SESSION_ERROR", error)
        return Promise.reject(new Error("DELETE_SESSION_ERROR", error))
      }
    }

    async function createVerificationRequest(
      identifier,
      url,
      token,
      secret,
      provider
    ) {
      debugMessage("CREATE_VERIFICATION_REQUEST", identifier)
      try {
        const { site } = appOptions
        const { sendVerificationRequest, maxAge } = provider

        // Store hashed token (using secret as salt) so that tokens cannot be exploited
        // even if the contents of the database is compromised.
        // @TODO Use bcrypt function here instead of simple salted hash
        const hashedToken = createHash("sha256")
          .update(`${token}${secret}`)
          .digest("hex")

        let expires = null
        if (maxAge) {
          const dateExpires = new Date()
          dateExpires.setTime(dateExpires.getTime() + maxAge * 1000)
          expires = dateExpires.toISOString()
        }

        // Save to database
        const verificationRequest = await prisma[VerificationRequest].create({
          data: {
            identifier,
            token: hashedToken,
            expires,
          },
        })

        // With the verificationCallback on a provider, you can send an email, or queue
        // an email to be sent, or perform some other action (e.g. send a text message)
        await sendVerificationRequest({
          identifier,
          url,
          token,
          site,
          provider,
        })

        return verificationRequest
      } catch (error) {
        logger.error("CREATE_VERIFICATION_REQUEST_ERROR", error)
        return Promise.reject(
          new Error("CREATE_VERIFICATION_REQUEST_ERROR", error)
        )
      }
    }

    async function getVerificationRequest(identifier, token, secret, provider) {
      debugMessage("GET_VERIFICATION_REQUEST", identifier, token)
      try {
        // Hash token provided with secret before trying to match it with database
        // @TODO Use bcrypt instead of salted SHA-256 hash for token
        const hashedToken = createHash("sha256")
          .update(`${token}${secret}`)
          .digest("hex")
        const verificationRequest = await prisma[VerificationRequest].findOne({
          where: { token: hashedToken },
        })

        if (
          verificationRequest &&
          verificationRequest.expires &&
          new Date() > verificationRequest.expires
        ) {
          // Delete verification entry so it cannot be used again
          await prisma[VerificationRequest].delete({
            where: { token: hashedToken },
          })
          return null
        }

        return verificationRequest
      } catch (error) {
        logger.error("GET_VERIFICATION_REQUEST_ERROR", error)
        return Promise.reject(
          new Error("GET_VERIFICATION_REQUEST_ERROR", error)
        )
      }
    }

    async function deleteVerificationRequest(
      identifier,
      token,
      secret,
      provider
    ) {
      debugMessage("DELETE_VERIFICATION", identifier, token)
      try {
        // Delete verification entry so it cannot be used again
        const hashedToken = createHash("sha256")
          .update(`${token}${secret}`)
          .digest("hex")
        await prisma[VerificationRequest].delete({
          where: { token: hashedToken },
        })
      } catch (error) {
        logger.error("DELETE_VERIFICATION_REQUEST_ERROR", error)
        return Promise.reject(
          new Error("DELETE_VERIFICATION_REQUEST_ERROR", error)
        )
      }
    }

    return Promise.resolve({
      createUser,
      getUser,
      getUserByEmail,
      getUserByProviderAccountId,
      updateUser,
      deleteUser,
      linkAccount,
      unlinkAccount,
      createSession,
      getSession,
      updateSession,
      deleteSession,
      createVerificationRequest,
      getVerificationRequest,
      deleteVerificationRequest,
    })
  }

  return {
    getAdapter,
  }
}

export default {
  Adapter,
}

```

Fumler:
Ah! Good find @brunocrosier. I was only testing with Google, and I think it's a number there as well but is treated as a string.

brunocrosier:
Great, thanks again @Fumler ! Another small thing that I noticed.. In the `schema.prisma` file that you suggested, the `email` field on the `User` model is set as required, whereas I think [it should be nullable](https://next-auth.js.org/schemas/postgres) 

I haven't encountered any bugs as a result of it yet, but maybe it should be set to nullable just in case?

iaincollins:
Thanks for testing this!

> Ah! Good find @brunocrosier. I was only testing with Google, and I think it's a number there as well but is treated as a string.

Yeah IIRC this is usually a number but needs to be treated like a string (as the existing built in models do) as some providers return alphanumeric values for it. Using `String()` is perfect.

iaincollins:
> Great, thanks again @Fumler ! Another small thing that I noticed.. In the `schema.prisma` file that you suggested, the `email` field on the `User` model is set as required, whereas I think [it should be nullable](https://next-auth.js.org/schemas/postgres)
> 
> I haven't encountered any bugs as a result of it yet, but maybe it should be set to nullable just in case?

+1  Yeah the email should be nullable as of v2.0 release.

In earlier beta releases of v2 email was required - I think that's when @Fumler wrote the original version of the Prisma adapter.

In the end, we found so many providers that don't always return a value for it (e.g. privacy settings) that code, and SQL and MongoDB logic was updated to allow for it to be not specified as requiring it was just too painful.

(NB: It is still possible to configure it to still request a user uses an account with some sort of email using the `signin()` callback to reject sign in attempts from an account that doesn't have one, though you'd also want to use a custom error page with it.)

Fumler:
This should be good to go @iaincollins so feel free to merge when you want.

iaincollins:
> This should be good to go @iaincollins so feel free to merge when you want.

Thank you, this looks really great! I hope to get time to try it out tomorrow morning!

phillip055:
Awesome guys, excited for this to get merged

brunocrosier:
Hey folks so just wanted to chime in and say I've noticed a potential problem for Next.js and prisma in general which could affect this. 

The issue is that basically while in dev mode, next creates several connections to the db via prisma which can cause the db to crash. 

Im on mobile now but I will post a slight modification to prisma adapter file in order to prevent this. It might be a good idea to mention the problem and workaround in the docs also, as it will affect any server side rendered pages that use the session

If somebody else wants to do this in the meantime, the code basically just checks whether we are in Dev mode via process.env.NODE_ENV and then only creates a new prisma client if it is not already in the global scope (if global.prisma doesn't exist)

iaincollins:
@brunocrosier Thanks for that! Getting connection handling right was tricky with the TypeORM driver too.

Would appreciate any suggestions for code changes if you have them, but your notes are really helpful.


@Fumler I'm thinking of making a v3 branch and then merging this branch and branch in my PR #379 into it - so we can do integration testing and merge the docs changes in without impacting the live site. Let me know if you have any objection!

Kerumen:
What's @brunocrosier is mentioning is something similar to [this](https://github.com/prisma/prisma-client-js/issues/228#issuecomment-618433162).

I've done this in my app to prevent Prisma crashes in dev mode.

brunocrosier:
That's exactly right ! We just need to do that for the prisma adapter file and optionally add it to the docs

iaincollins:
@Kerumen @brunocrosier Thanks both! I'll merge this in and tag you in on any changes if you are up for reviewing.

We actually do something similar for TypeORM connections, as otherwise we'd just get people raising issues when they ran into the same problems over and over, so it seems like something we should cater for with Prisma too but appreciate I don't have hands on experience with Prisma yet so will try it out first.

iaincollins:
Merging into v3 release to avoid this getting stale, and so we can merge can fix forward for release.

Will link back to this PR from the one we raise from the v3 branch once we have it.

Thanks everyone! <3

yareyaredesuyo:
I tried prisma adapter [exmaple](https://github.com/iaincollins/next-auth/tree/docs/examples/examples/with-prisma). It works fine! but changed scheme settings a bit.

Set id as Int type (same as default TypeORM adapter).

```prisma
model User {
  id            Int       @default(autoincrement()) @id
  email         String    @unique
  name          String?
  emailVerified DateTime?
  image         String?
  sessions      Session[]
  accounts      Account[]
  createdAt     DateTime  @default(now())
  updatedAt     DateTime  @updatedAt
}

model Session {
  id                 Int       @default(autoincrement()) @id
  userId             Int
  user               User      @relation(fields: [userId], references: [id])
  accessToken        String    @default(cuid()) @unique
  sessionToken       String    @default(cuid()) @unique
  accessTokenExpires DateTime?
  expires            DateTime?
  createdAt          DateTime  @default(now())
  updatedAt          DateTime  @updatedAt
}

model Account {
  id                 Int       @default(autoincrement()) @id
  compoundId         String    @unique
  userId             Int
  user               User      @relation(fields: [userId], references: [id])
  providerId         String
  providerType       String
  providerAccountId  String    @unique  # in TypeORM adapter, there is no @unique constraint
  refreshToken       String?
  accessToken        String?
  accessTokenExpires DateTime?
  createdAt          DateTime  @default(now())
  updatedAt          DateTime  @updatedAt
}

model VerificationRequest {
  id         Int      @default(autoincrement()) @id
  identifier String
  token      String   @unique
  expires    DateTime
  createdAt  DateTime @default(now())
  updatedAt  DateTime @updatedAt
}
```

And I found that, current prisma exmaple scheme set `Account/providerAccountId` field to unique constraint.
But I think it should not set unique ( for compatible with default adapter scheme).

Current prisma adapter implementation which use `providerAccountId` for retrieving data, using `findOne` for unique constraint.

```js
    async function getUserByProviderAccountId (providerId, providerAccountId) {
      debugMessage('GET_USER_BY_PROVIDER_ACCOUNT_ID', providerId, providerAccountId)
      try {
        return prisma[Account].findOne({ where: { providerAccountId: `${providerAccountId}` } })[User]() // currently not using bot providerId and providerAccountId
      } catch (error) {
        logger.error('GET_USER_BY_PROVIDER_ACCOUNT_ID_ERROR', error)
        return Promise.reject(new Error('GET_USER_BY_PROVIDER_ACCOUNT_ID_ERROR', error))
      }
    }
```

How about change like this? This does not require providerAccountId's unique constraint.

```js
    async function getUserByProviderAccountId (providerId, providerAccountId) {
      debugMessage('GET_USER_BY_PROVIDER_ACCOUNT_ID', providerId, providerAccountId)
      try {
        return prisma[Account].findOne({ where: { compoundId: `${createHash('sha256').update(`${providerId}:${providerAccountId}`).digest('hex')}` } })[User]()
      } catch (error) {
        logger.error('GET_USER_BY_PROVIDER_ACCOUNT_ID_ERROR', error)
        return Promise.reject(new Error('GET_USER_BY_PROVIDER_ACCOUNT_ID_ERROR', error))
      }
    }
```

iaincollins:
Thanks @yareyaredesuyo that's a good catch! I didn't spot that. Your suggestion looks good.

Maybe `getUserByProviderAccountId` should really have a name like `getUserByProviderAndProviderAccountId`?

### Context

For anyone wondering what is going on here, the reason the model for accounts has a Compound ID is that Provider Account ID's are (of course) only unique to a provider, but not all databases support a compound key (made up of more than one field) like that so combining the Provider ID (e.g. 'google') and the account ID (e.g. '123') to get something like 'google:123' and then hashing it, creates one we can use as a simple UNIQUE constraint, which should work everywhere.

(Using a hash like this is a common approach to solving that problem in relational databases.)

yareyaredesuyo:
@iaincollins It looks nice idea! Because when retrieving data in `getUserByProviderAccountId`, current TypeORM implementation both using `providerId` and `providerAccountId`.

https://github.com/iaincollins/next-auth/blob/2d80a40adec97640a7a51da5b1be42f23f21f66b/src/adapters/typeorm/index.js#L153-L163

PR #303, opened 2020-06-22
docs(examples): Add basic and with-prisma

Fumler:
I'm a fan of how Next.js has examples in the repo as this really helps convey how to do different things. The `with-prisma` example won't work unless #302 is merged though.

This is only a suggestion, feel free to close if not wanted.


iaincollins:
Thanks @Fumler!

I'm a big fan of the `examples` dir too, although there are actually some caveats when including a lib (which I ran into with v1) so it's maybe worth a discussion - would like to describe them and see what people think. Will get back to this in a bit :-)

I definitely would like to have something like this, so if it doesn't go here, we should have it somewhere! Also I really want to learn more about the Prisma migration stuff! :-)

yareyaredesuyo:
Prisma adapter feature looks like merged ( https://github.com/iaincollins/next-auth/pull/302 ), how about this PR ?

iaincollins:
> Prisma adapter feature looks like merged ( #302 ), how about this PR ?

Things are too busy getting v3 ready to ship to handle issue of examples as well but we will get to it!

kadirselcuk (approved):

.gitignore
```diff
@@ -19,10 +19,11 @@ dist
 .env.development.local
 .env.test.local
 .env.production.local
+*.db
```
kadirselcuk, line 22: I'll leave PR here

kadirselcuk (approved):

Issue #301, opened 2020-06-22
Auth0: invalid_request: Missing required parameter: response_type


vacom:
**Describe the bug**
The Auth0 provider coming from the build-ins provider is giving a  response_type error when trying to authenticate.

**To Reproduce**
Using the default example from the documentation for the Auth0 Provider

```js 
import Providers from `next-auth/providers`
...
providers: [
  Providers.Auth0({
    clientId: process.env.AUTH0_CLIENT_ID,
    clientSecret: process.env.AUTH0_CLIENT_SECRET,
    subdomain: process.env.AUTH0_SUBDOMAIN
  })
}
...
```

**Screenshots or error logs**

```js
{
  "date": "2020-06-22T20:03:04.148Z",
  "type": "f",
  "description": "Missing required parameter: response_type",
  "connection_id": "",
  "user_agent": "Chrome 83.0.4103 / Mac OS X 10.15.5",
  "details": {
    "body": {},
    "qs": {
      "redirect_uri": "http://localhost:3000/api/auth/callback/auth0",
      "scope": "openid email profile",
    },
    "error": {
      "message": "Missing required parameter: response_type",
      "oauthError": "invalid_request",
      "type": "request-error"
    }
  },
  "isMobile": false
}
```

iaincollins:
Hmm we have had a few folks contribute to this but don't have a test for it yet.

If you are able to try providing other options you might be able to suggest a fix. This is what the default options are (you can override any of them): https://github.com/iaincollins/next-auth/blob/main/src/providers/auth0.js

I might need to investigate this one.

iaincollins:
Will check this out tomorrow, but if anyone is using Auth0 and/or able to test it out themselves would appreciate any feedback / findings!

iaincollins:
I checked it out myself and found a few bugs in the provider. I've just fixed them, tested it out to verify and published the fix as `next-auth@2.0.1` so it should work for you now if you update to the latest version.

Note: The options have changed slightly to bring them into line with other providers and the Auth0 control panel itself, specifically `subdomain` is now `domain`.

The changes are reflected in the updated documentation:
https://next-auth.js.org/providers/auth0

vacom:
@iaincollins thanks for the fix, is working. 🚀

PR #304, opened 2020-06-22
Fix Auth0 provider

iaincollins:
Addressed a few issues with the provider as it was reported as not working in #301 and I couldn't get it working either.

Have updated the domain property to 'domain' rather than 'subdomain' to match behaviour of similar providers - I think this makes sense as the copy/paste option from Auth0 dashboard is also the fully qualified domain.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "2.0.0",
+  "version": "2.0.1",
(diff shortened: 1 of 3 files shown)


LoriKarikari:
I think I was tweaking the Auth0 provider on some wrong branch and merged it accidentally 😬

Issue #299, opened 2020-06-22
TypeError: Cannot read property 'csrfToken' of null


ricjohn-acosta:
*Please refer to the [documentation](https://next-auth.js.org/getting-started/introduction), the [example project](https://github.com/iaincollins/next-auth-example) and existing issues before creating a new issue.*

**Your question**
Not sure why I'm getting the following error: `TypeError: Cannot read property 'csrfToken' of null` on the client side and `CLIENT_FETCH_ERROR null/csrf TypeError: Only absolute URLs are supported` on the server side. This error only occurs when the browser is recently opened and then navigating directly to /signup page. The error disappears when the index page is loaded first and then navigating to /signup page

**What are you trying to do**
Trying to implement a custom signup page with an email provider and third party providers

**Relevant code**

signup.js
```
import { csrfToken, signin, providers } from "next-auth/client";
import Button from "@material-ui/core/Button";

export default ({ csrfToken, providers }) => {
  return (
    <>
      <form
        method="post"
        action="/api/auth/signin/email"
        onSubmit={(e) => {
          e.preventDefault();
          signin("email", { email: document.getElementById("email").value });
        }}
      >
        <input name="csrfToken" type="hidden" defaultValue={csrfToken} />
        <label>
          Email address
          <input type="text" id="email" name="email" />
        </label>
        <button type="submit">Sign in with Email</button>
      </form>

      {Object.values(providers).map((provider) =>
        provider.name === "Email" ? null : (
          <p key={provider.name}>
            <Button
              href={`/api/auth/signin/${provider.name.toLowerCase()}?callbackUrl=${
                process.env.NODE_ENV !== "production"
                  ? "http://localhost:3000/account-setup"
                  : process.env.SITE
              }`}
            >
              Sign in with {provider.name}
            </Button>
          </p>
        )
      )}
    </>
  );
};

export async function getServerSideProps(context) {
  return {
    props: {
      csrfToken: await csrfToken(context),
      providers: await providers(context),
    },
  };
}
```

**Documentation feedback**
*Documentation refers to searching through [online documentation](https://next-auth.js.org), code comments and issue history. The example project refers to [next-auth-example](https://github.com/iaincollins/next-auth-example).*

* [ ] Found the documentation helpful
* [X] Found documentation but was incomplete
* [X] Could not find relevant documentation
* [X] Found the example project helpful
* [ ] Did not find the example project helpful

iaincollins:
Thanks for the bug report!

[...]

PR #315, opened 2020-06-23
Fix client method errors in Server Side Rendering

### LINKED ISSUE #313: session() doesn't work well when accessing on iOS
**Describe the bug**
I use session() inside serverSideProps.
But, this error occurs only when accessing on iOS.  Others are OK. - android, macOS.
How should I solve the problem?

```
[next-auth][error][CLIENT_FETCH_ERROR] [
  'null/session',
  TypeError: Only absolute URLs are supported
      at getNodeRequestOptions (/home/okikazuhide/salon-tablet/node_modules/next/dist/compiled/node-fetch/index.js:1:145647)
      at /home/okikazuhide/salon-tablet/node_modules/next/dist/compiled/node-fetch/index.js:1:147021
      at new Promise (<anonymous>)
      at fetch (/home/okikazuhide/salon-tablet/node_modules/next/dist/compiled/node-fetch/index.js:1:146956)
      at /home/okikazuhide/salon-tablet/node_modules/next-auth/dist/client/index.js:276:23
      at asyncGeneratorStep (/home/okikazuhide/salon-tablet/node_modules/next-auth/dist/client/index.js:20:103)
      at _next (/home/okikazuhide/salon-tablet/node_modules/next-auth/dist/client/index.js:22:194)
      at /home/okikazuhide/salon-tablet/node_modules/next-auth/dist/client/index.js:22:364
      at new Promise (<anonymous>)
] 
https://next-auth.js.org/errors#client_fetch_error
```

**To Reproduce**
index.tsx
```
export async function getServerSideProps(context:NextPageContext) { 

  const req = context.req
  const session = await session({ req });

  return {
      props: {
        session: session
      },
    };
```

I follow #299 
next.config.js
```
module.exports = {
  env: {
     NEXTAUTH_SITE: "https://example.com",
  },
};
```

[...nextauth].ts options

```
const options = {
  site: "https://example.com",
  providers: [
    Providers.Email({
      server: process.env.EMAIL_SERVER,
      from: process.env.EMAIL_FROM,
    }),
    Providers.Credentials({
      authorize: async (credentials) => authorizeCredentials(credentials), // following the document.
      credentials: {
        email: { label: "Email", type: "text ", placeholder: "jsmith" },
      },
    }),
  ],
  database: process.env.DATABASE_URL,
  session: {
    jwt: true,
  },
};
```

iaincollins:
Update:

I have reviewed and refactored my original changes in this Pull Request.

New beta published as `next-auth@2.1.0-beta.1`

This should resolve #313 and #299 but changes how configuration is set on the client to resolve an otherwise unfixable problem of hitting server side rendered custom pages before any client routes are called.

## Problem

When a NextAuth.js route is called in the browser (e.g. `/api/auth/signin` or `/api/auth/session`) a 'configuration cookie' is set. When using custom pages the initial page loaded can mean the client methods get called without hitting an API route first, so there is no configuration cookie, and no secure way to get the name of the site.

We could use `req.host` header, but that can be spoofed and you could technically exploit that to get the route to call another server, so we need it to be set explicitly; which is also why you need to set the server URL in the API route.

### New client documentation

To use client methods server side in `getServerSideProp()` or `getInitialProps()` you should add the NextAuth.js  `<Provider>` in `pages/apps.js`

### Example

```js
import { Provider } from 'next-auth/client'

export default ({ Component, pageProps }) => {
  const { session } = pageProps
  return (
    <Provider options={{ site: process.env.SITE }} session={session} >
      <Component {...pageProps} />
    </Provider>
  )
}
```

You should then be able to do this:

```js
import { getSession, getCsrfToken, getProviders } from 'next-auth/client'
const Page = ({session, providers, csrfToken}) => (
  <>
    <main>
      <hr/>
      <pre>{JSON.stringify(providers, null, 2)}</pre>
      <hr/>
      <pre>{JSON.stringify(csrfToken, null, 2)}</pre>
      <hr/>
    </main>
  </>
)

  return {
    props: {
      session: await getSession(context),
      providers: await getProviders(context),
      csrfToken: await getCsrfToken(context)
    }
  }
}

export default Page
```

Additionally, only the `getSession()` method now needs to be passed `context` when called server side, the `getProvider()` and `getCsrfToken()` methods no longer need it (this is not a breaking change, just simpler to use them).


iaincollins:
I've released the change happy this works in production! The docs might need a review though.

I'm going to publish this and the new providers in a release (2.1).

iaincollins:
@LoriKarikari @ndom91 @geraldnolan @Fumler  BTW sorry for rushing this one through last night.

###  Problem
There was a flow that was broken and it was fairly significant defect - using client method server side, without first any calls being made to API routes to set configuration cookies would just fail. This was quite a significant defect with no workaround. In practice it was mostly an issue for people using custom pages and the problem with that flow had gone unnoticed.

I wanted to address it quickly as it already had 3 bug reports about it and it was an embarrassing defect that would trip up a lot of folks (and we'd surely get more reports).

Normally I'd want to spend more time in peer review to consider a change like this, but I didn't feel I could just leave it broken. I ended up drafting a proposal, testing it thoroughly as a beta release, realising it could be better and doing another beta release and another round of testing, before releasing it as 2.1 last night.

### Fix

In the end, I added the options to the `<Provider>` which is perhaps a little weird conceptually but keeps the API surface area small and I think makes some sense for users (e.g. "If you want to do server side stuff, it all goes here").

It also means the client docs lead on nicely to that - e.g. "here are client side methods, oh you want SSR too well just do this" – and no one needs to mess around with also putting stuff in `next.config.js`.

How it actually works is a little hacky; it actually just stuffs config options into a global called `__NEXTAUTH`. This is crude but avoids overhead and complexity (for users and for us) and the result is less code to manage for everyone. Next.js actually does something similar.

### Future

I'm still interested on how we could do client config better if anyone has any ideas.

If you have feedback on this PR I still welcome it and happy to steer in a different direction in future if we conclude there is a different approach that makes more sense.

I appreciate the docs will probably need some work. This approach allows them to be simpler and put information about how to configure client options across the app all in one place, but I appreciate they haven't had anyone review them to see if they make sense to someone other than the person who wrote the feature. :-)

kazuhideoki:
This works for me! Thank you.

> ### Example
> ```js
> import { Provider } from 'next-auth/client'
> 
> export default ({ Component, pageProps }) => {
>   const { session } = pageProps
>   return (
>     <Provider options={{ site: process.env.SITE }} session={session} >
>       <Component {...pageProps} />
>     </Provider>
>   )
> }
> ```
> 
> You should then be able to do this:
> 
> ```js
> import { getSession, getCsrfToken, getProviders } from 'next-auth/client'
> const Page = ({session, providers, csrfToken}) => (
>   <>
>     <main>
>       <pre>{JSON.stringify(session, null, 2)}</pre>
>       <hr/>
>       <pre>{JSON.stringify(providers, null, 2)}</pre>
>       <hr/>
>       <pre>{JSON.stringify(csrfToken, null, 2)}</pre>
>       <hr/>
>     </main>
>   </>
> )
> 
> export async function getServerSideProps(context) {
>   return {
>     props: {
>       session: await getSession(context),
>       providers: await getProviders(context),
>       csrfToken: await getCsrfToken(context)
>     }
>   }
> }
> 
> export default Page
> ```

iaincollins:
@kazuhideoki Thank you for the feedback, much appreciated!

BTW there is a small mistake in my example, technically you can omit the `context` option from all calls except `getSession(context)` (which still needs it, as it needs to pass through the session cookie from the user).

e.g. you can just do this now:

```js
  return {
    props: {
      session: await getSession(context),
      providers: await getProviders(),
      csrfToken: await getCsrfToken()
    }
  }
}
```

This is not actually a problem, just FYI - I removed the need to pass it in 2.1 and forgot to update the example.

Romstar:
> @kazuhideoki Thank you for the feedback, much appreciated!
> 
> BTW there is a small mistake in my example, technically you can omit the `context` option from all calls except `getSession(context)` (which still needs it, as it needs to pass through the session cookie from the user).
> 
> e.g. you can just do this now:
> 
> ```js
> export async function getServerSideProps(context) {
>   return {
>     props: {
>       session: await getSession(context),
>       providers: await getProviders(),
>       csrfToken: await getCsrfToken()
>     }
>   }
> }
> ```
> 
> This is not actually a problem, just FYI - I removed the need to pass it in 2.1 and forgot to update the example.

Still not working on the API routes.

pages/api/news/test.ts

```
import { getSession } from 'next-auth/client';


const handler = async (req, res) => { 
const session = await getSession({ req });
}
```

_app.tsx

```
import { Provider } from 'next-auth/client'
// import './styles.css'

export default ({ Component, pageProps }) => {
  const { session } = pageProps
  return (
    <Provider options={{ site: process.env.SITE }} session={session}>
      <Component {...pageProps} />
    </Provider>
  )
}
```

package.json

```
"dependencies": {
    "@hapi/joi": "^17.1.1",
    "@types/hapi__joi": "^17.1.2",
    "colors": "^1.4.0",
    "next": "9.4.4",
    "next-auth": "^2.1.0",
    "pg": "^8.2.1",
    "react": "16.13.1",
    "react-dom": "16.13.1",
    "sass": "^1.26.8",
    "snoowrap": "^1.21.0",
    "twitter": "^1.7.1"
  },
  "devDependencies": {
    "@types/node": "^14.0.13",
    "@types/react": "^16.9.38",
    "typescript": "^3.9.5"
  }
```

.env

```
# Next-auth
SITE=http://localhost:3000
```
[next-auth][error][CLIENT_FETCH_ERROR] [
  '/api/auth/session',
  TypeError: Only absolute URLs are supported

iaincollins:
API routes need to be handled slightly differently, as they can't inherit config the way pages do from `_app.js` for API routes.

Currently the only documentation for how to access data from API routes related to JSON Web Tokens.

However, you can do this to use the client from an API route, but a better API is under consideration so it is not publicly documented yet.

```js
import { setOptions, getSession } from 'next-auth/client'
setOptions({ site: process.env.SITE })

export default async (req, res) =>  {
  const session = await getSession({ req })
  console.log('session', session)
  res.end()
}
```

Romstar:
> API routes need to be handled slightly differently, as they can't inherit config the way pages do from `_app.js` for API routes.
> 
> Currently the only documentation for how to access data from API routes related to JSON Web Tokens.
> 
> However, you can do this to use the client from an API route, but a better API is under consideration so it is not publicly documented yet.
> 
> ```js
> import { setOptions, getSession } from 'next-auth/client'
> setOptions({ site: process.env.SITE })
> 
> export default async (req, res) =>  {
>   const session = await getSession({ req })
>   console.log('session', session)
>   res.end()
> }
> ```

Thanks, it is working and the error isn't occurring anymore.  I guess the original bug wasn't a duplicate?

iaincollins:
> Thanks, it is working and the error isn't occurring anymore

Thanks for confirming it works for you too!

> I guess the original bug wasn't a duplicate?

It turns out both were impacted by the same problem. :-( It worked in some flows but not others. Basically, it was mostly fine except sometimes if you were using custom pages with server side rendering - but that is quite a lot of folks.

This PR added that extra `setOptions` / `options` method, but I'd like to provide a better API, I'm just not sure what that should look like yet. It seems awkward to have to specify config object in every API route.

One option might be automatically look for and load a file like `nextauth.config.js` file in the root, which the API route could pick up too, so all the config could go there and it would just get loaded automatically when a call is made.

I might open up an issue to discuss this!

PR #318, opened 2020-06-23
Add: npm publish workflow

ndom91:
Simply added the example `npm-publish` github action.

The secret from your npm account wuodl still need to be added to this repo, but otherwise you can merge and it should be good to go ;)

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,0 +1,36 @@
+# Publishes module to registry when a new release is created.
+#
+# The following secrets need to be configured for this workflow:
+#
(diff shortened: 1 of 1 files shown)


iaincollins:
Thanks, was meaning to use this!

iaincollins:
The job failed this time - I have done this a dozen times and it always seems to go wrong first time (user error on my part!) - so published manually. 

It only failed because it tried to run `npm test` by default, which tries to run the database tests which don't work without databases being spun up yet (I'm going to move those tests to all run *in* a container at some point soon).

I updated `npm test` to just run the linter for now (e.g. `npm run lint`).

PR #351, opened 2020-06-28
Fix reading of req in getSession()

### LINKED ISSUE #345: Example with getSession() in _app.js?
**Your question**

All the examples I've seen for `getSession()` is for a specific page, but I'd like to pass the `session` prop at the layout level, via `_app.js`.

What is the proper way of using it in `_app.js` for both client/server side application?

**What are you trying to do**

Apply the `session` prop to all page props without duplicating logic by setting it through `_app.js`.

```
// _app.tsx
// some psuedo-code
import { Provider } from 'react-redux'
import { getSession, Provider as AuthProvider } from 'next-auth/client'

const WrappedApp: FC<CustomProps> & StaticComponents = (props) => {
  const { Component, pageProps, session } = props

  return (
    <Provider store={store}>
      <AuthProvider session={session} options={{ site: 'http://localhost:3000' }}>
            <HeaderNav {...props} />
            <Component {...pageProps} />
      </AuthProvider>
    </Provider>
  )
}

// https://github.com/isaachinman/next-i18next/issues/615
WrappedApp.getInitialProps = async (context) => {
  const appProps = await App.getInitialProps(context)
  const session = await getSession(context)
 
  // ----------> returns null
  console.log(session)

  return {
    ...appProps,
    session
  }
}

export default WrappedApp
```

With the example, I also get despite having the proper `site` value:

```
[next-auth][error][CLIENT_FETCH_ERROR] [
  '/api/auth/session',
  TypeError: Only absolute URLs are supported
```

**Documentation feedback**

theogravity:
Fixes https://github.com/iaincollins/next-auth/issues/345

I'm virtually clueless about the internals of next.js right now, please let me know if this is not the correct approach.

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -25,7 +25,11 @@ const setOptions = ({
 }
 
 // Universal method (client + server)
-const getSession = async ({ req } = {}) => {
+const getSession = async ({ req, ctx } = {}) => {
+  if (!req && ctx.req) {
+    req = ctx.req
+  }
+
   const baseUrl = _baseUrl()
   const options = req ? { headers: { cookie: req.headers.cookie } } : {}
   const session = await _fetchData(`${baseUrl}/session`, options)


iaincollins:
I see what you mean, this would make it easier to handle the context object in `_app.js`  being different.

I'd like to mull this over for a little bit and create a tutorial that covers this and see what issues there are and if there are alternative approaches. I think seems reasonable though.

iaincollins:
Going to add this and pull it in for v3

iaincollins:
Thanks for the PR!

Issue #346, opened 2020-06-27
Auth provider [Apple]


erhankaradeniz:
**Your question**
When using the Apple auth provider there is no name in the session.user.name object. it's filled with a random string.

**What are you trying to do**
Retrieving the name of the user that is logged in.

Is there something that I'm missing to fetch the name of a user. the email is returned correctly as a value. Is this intended behaviour? (if yes, that's fine too, but I'm wondering if I can autofill that for a user after login)

iaincollins:
Hmm interesting, I don't know! @geraldnolan is the expert on that and my be able to answer that.

geraldnolan:
@erhankaradeniz Thanks for kicking the the tires with sign in with apple.  You will have to give us a little bit more time to update this.  Once updated you will have to revoke access to your website using the following link.

https://appleid.apple.com/#!&page=signin

Select Stop using Apple ID.  After we update the code, the Name will be set.

![image](https://user-images.githubusercontent.com/17180783/85925301-c1402380-b897-11ea-9b81-d3fdc9e753d4.png)


@iaincollins haha 👍We are working on getting you access so you can be the expert. 
It turns out there is a bit more coding required to get the user object according to apple documentation.  I thought it would return in the id_token response from Apple. (Which only has the email address).  The random string is just the sub (I placed that there just in case the name was required in someone's DB). The user object must be stored on the first login; otherwise, it will not be returned after that.

I have to modify the callback once more:

```jsx
if (req.method === 'POST') {
        console.log("Req ->", req.params)
        console.log("Body ->", req.body)
        var body = JSON.parse(JSON.stringify(req.body));
        code = body.code;
        user = body.user
      }


Body -> [Object: null prototype] {
  state: 'Random State',
  code: 'Random Code',
  user: '{"name":{"firstName":"Gerald","lastName":"Nolan"},"email":"gerald.nolan@noltech.co"}'
}
```


[Apple Documentation](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms)


A successful response contains the following parameters:

**code**
A single-use authorization code that is valid for five minutes.

**id_token**
A JSON web token containing the user’s identity information.

**state**
The state contained in the Authorize URL.

**user**
A JSON string containing the data requested in the scope property. The returned data is in the following format: { "name": { "firstName": string, "lastName": string }, "email": string }

If an error occurs, the HTTP body contains the following parameters:

**error**
[...]

PR #352, opened 2020-06-28
Improvements to Apple provider

iaincollins:
When using a provider that uses Token ID option (like Apple) a user hitting cancel with no longer cause the app to crash.

Users who do this will now be taken back to the sign in page.

This was already working for other providers that didn't use this option but wasn't supported for providers that did use it.

This addresses one issue raised in #346

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,8 @@ yarn-error.log*
 
 # Docusaurus
 www/build
+
+#VS
+/.vs/slnx.sqlite-journal
+/.vs/slnx.sqlite
+/.vs
--- a/src/providers/apple.js
+++ b/src/providers/apple.js
@@ -13,9 +13,11 @@ export default (options) => {
     profileUrl: null,
     idToken: true,
     profile: (profile) => {
+
+      //The name of the user will only return on first login
       return {
         id: profile.sub,
-        name: profile.name == null ? profile.sub : profile.name,
+        name: profile.user != null ? profile.user.name.firstName + " " + profile.user.name.lastName : null,
         email: profile.email
       }
     },
--- a/src/providers/apple.js
+++ b/src/providers/apple.js
@@ -36,7 +36,9 @@ export default (options) => {
           aud: 'https://appleid.apple.com',
           sub: appleId
         },
-        privateKey,
+        // Automatically convert \\n into \n if found in private key. If the key
+        // is passed in an environment variable \n can get escaped as \\n
+        privateKey.replace(/\\n/g, '\n'), 
         {
           algorithm: 'ES256',
           keyid: keyId
--- a/src/providers/apple.js
+++ b/src/providers/apple.js
@@ -13,11 +13,10 @@ export default (options) => {
     profileUrl: null,
     idToken: true,
     profile: (profile) => {
-
-      //The name of the user will only return on first login
+      // The name of the user will only return on first login
       return {
         id: profile.sub,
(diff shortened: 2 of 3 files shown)


geraldnolan (commented):
@iaincollins The error is happening before this section..

I was about to push code either today or tomorrow since I am getting github 500 related errors today.

The body of the request from apple actually tells you the user cancelled.

I was a little bit unfamiliar on how to send the callback. 

```jsx
if (provider.version && provider.version.startsWith('2.')) {
    if (req.method === 'POST') {

      // Get the CODE from Body
      const body = JSON.parse(JSON.stringify(req.body))

      // @TODO Handle error
      if (body.error) {
        logger.error('OAUTH_CALLBACK_HANDLER_ERROR', body.error, req.body, provider.id, code)
        callback(body.error, null, null, null)
      }
```

geraldnolan:
Also, there is an issue with using apple without a database since the only time you get the user's name is on the first login in. :/

Not sure if you can view the form but here is a link on more information
https://developer.apple.com/forums/thread/119826?page=2

iaincollins:
> Also, there is an issue with using apple without a database since the only time you get the user's name is on the first sign in. :/ 

No worries! I was thinking about adding a boiler plate notice to this for OAuth providers that should be used with a database.

Google actually falls into this category too, as it only issues OAuth Refresh Tokens on first login, unless you force an additional consent screen on every sign in (which is really intended for mobile or desktop apps). 

We can just list that for providers that have that limitation.

iaincollins:
> @iaincollins The error is happening before this section..
> 
> I was about to push code either today or tomorrow since I am getting github 500 related errors today.
> 
> The body of the request from apple actually tells you the user cancelled.

Thanks for looking at this! Oh! I accessed the callback route now it does not crash, and I am returned to the sign up page, as that works for other providers. I guess we need to return sooner?

I've been hoping to avoid provider specific approaches to checking if a flow is cancelled (and just going with checking if there is a valid response or not a proxy method for doing that) to avoid provider specific code in general, as they are all different and there doesn't seem to be a standard for it. :-( Maybe that is not possible in this case.

If we do get an error, returning an empty callback `callback()` is fine and will let the page know to do that. Maybe we should just do that on error, as the user will be redirected to the sign up page (and we can still log the error in the console).

This is not ideal obviously but I am inclined to think we kick refactoring down the road a bit until we have tests and also address an issue I've raised in core chat today.

(BTW not sure if you got the email invite! There should be an invitation to https://github.com/orgs/nextauthjs/teams/core waiting for you in your inbox somewhere :-)

geraldnolan:
@iaincollins Did you get my email from Google Doc?

iaincollins:
> @iaincollins Did you get my email from Google Doc?

Yes, thank you! Sorry I have been crazy busy the last few days (we were having some plumbing / gas work done at home). Will reply later today!

iaincollins:
@geraldnolan Merging this into the v3 branch with pull requests from from @Fumler (Prisma adapter) and myself (CSRF, etc) so we can do integration testing on the changes and push it out as a beta, with a fix forward approach, before any of the branches go stale! Will link back to this PR with the one we raise for v3 branch when we have one.

iaincollins:
Update for wider audience: I've tested this and thanks to the work @geraldnolan did changes seem to work great, but we will do some more testing on it and should be able to add the Apple provider to the live demo soon.

PR #355, opened 2020-06-28
Fix missing closing tag in docs

tpatel:
I couldn't run an example right away, here is the quick fix.


iaincollins:
Thanks!

tpatel:
Thanks for the quick merge and the library, I'm super excited to finally find a solid user auth solution in node ❤️ ❤️ ❤️

PR #357, opened 2020-06-29
Add: testing with cypress docs

ndom91:
Added a documentation page to aid in setting up Cypress to login with NextAuth.js in peoples projects.

This setup is what worked for me, but if I missed anything or did something wrong, just let me know.

CC: @JeffersonBledsoe


iaincollins (commented):
Nice work, thank you!

I've been thinking, I'd like to create a topic for Tutorials / Guides.

Do you think we should put this page under a topic like "Tutorials"?

I've got a few draft guides it would be great to place under that if we had it!

ndom91:
Yeah good idea. I'm sure with time there will be more and more as well, and that could be a good place to put them all together.

iaincollins (approved):
Nice! :D

iaincollins:
I am really happy if you want o merge this in and making it a living document and to update it accordingly if needed with v3 :-)

ndom91:
Yeah no problem. Hopefully it saves others some time getting started testing their NextAuth implementation as well 👍

PR #358, opened 2020-06-29
refactor • line 57 : Replace if by ternary operator.

alamenai:


iaincollins:
Thanks for the PR!

I like ternary operators for simple things but I think we want to discourage short pull requests for style reasons, because of the overhead it adds for the team (we already have a stack or PRs that we need to look at!).

The compiled JavaScript actually ends up normalized wiht babel anyway - and further minified by NPM when the module is published, so I'm okay with the extra lines in the source for additional clarity when it comes to readability.

PR #362, opened 2020-06-29
Change .env variables on example to reflect the actual variable name

vasilenka:
What we actually need to authenticate is written on the Twitter page as `API key` and `API secret key`.
So, the name `process.env.TWITTER_CLIENT_ID` and `process.env.TWITTER_CLIENT_SECRET` is confusing and misleading because there's an `App ID` and that's not what we need.

### DIFF
--- a/www/docs/providers/twitter.md
+++ b/www/docs/providers/twitter.md
@@ -18,8 +18,8 @@ import Providers from `next-auth/providers`
 ...
 providers: [
   Providers.Twitter({
-    clientId: process.env.TWITTER_CLIENT_ID,
-    clientSecret: process.env.TWITTER_CLIENT_SECRET
+    clientId: process.env.TWITTER_CLIENT_API_KEY,
+    clientSecret: process.env.TWITTER_CLIENT_API_SECRET_KEY
   })
 }
 ...


LoriKarikari:
We are using client id and client secret as these are the naming conventions specified in the OAuth spec. This also makes that the docs are consistent across different providers. You are however free to name the variables to your liking. @iaincollins any thoughts on this?

iaincollins:
> We are using client id and client secret as these are the naming conventions specified in the OAuth spec. This also makes that the docs are consistent across different providers. You are however free to name the variables to your liking. @iaincollins any thoughts on this?

Honestly this is how I feel too - though I am sympathetic to the confusion using Twitter causes because it uses different names for these values.

Twitter are annoying and confusing as they have actually changed the names for these over time, but never actually stuck with convention in the RFC (which every other provider does). They have even been inconsistent with their own documentation.

Perhaps we should do something like really spell out in the provider documentation that Twitter uses different names for these properties to other OAuth providers, maybe with a (redacted) screen shot to show the right ones to use?

LoriKarikari:
Yeah, I agree with adding that to the docs. I will update add that info. I'm closing this now.

PR #370, opened 2020-06-30
Update client.md

TomAstley:
Fixed small syntax error on line 219 in the signout example. Added a '('


LoriKarikari:
Thanks @TomAstley!

PR #375, opened 2020-07-01
Pass maxAge to the 'jwt.getJwt' function

arunoda:
With that it can pass that to the `jwt.decode` method. This will be useful, if we are using a custom `maxAge` value.


iaincollins:
Thanks! Agree, this makes sense.

PR #379, opened 2020-07-01
HTTP POST, CSRF and API changes for v3

iaincollins:
## Change Log

* The `NEXTAUTH_URL` environment variable can now be set to configure all pages and routes within the application, this replaces the  `site` configuration option in version 2.0.
* Sign in routes all now use HTTP POST (instead of HTTP GET) and require Cross Site Request Forgery Tokens.
* Cross Site Request Forgery Tokens are now checked on all sign in routes - some POST routes were missing it.
* Compliance with the recommended use of the _state_ object in [RFC 6749](https://tools.ietf.org/html/rfc6749). Client state is now checked and verified on all OAuth 2.0 providers, using a value unique and known only to the client. It can be disabled for OAuth providers who do not support the option using `useState: false`.
* The Client API and the options for Events, Callbacks and Pages are all now camelCase (e.g. 'signin()' is now 'signIn()' and 'signOut' is now 'signOut()' This is a breaking change except in the case of the Client API which supports both for backwards compatibility.  Configuration of Events, Callbacks and Pages should be updated.
* Improved client event handling for reduced network and memory footprint, while improving automatic session state syncing in clients and across tabs/windows.
* Changes include refactored event handler logic and improved `clientMaxAge` behaviour (now more passive / efficient) and new `keepAlive` option. Users with short session expiry times will benefit the most from these changes.
* The documentation and example code has been updated and examples improved and expanded.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "3.0.0-beta.0",
+  "version": "3.0.0-beta.1",
   "description": "An authentication library for Next.js",
   "repository": "https://github.com/iaincollins/next-auth.git",
   "author": "Iain Collins <me@iaincollins.com>",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "3.0.0-beta.4",
+  "version": "3.0.0-beta.6",
   "description": "An authentication library for Next.js",
   "repository": "https://github.com/iaincollins/next-auth.git",
   "author": "Iain Collins <me@iaincollins.com>",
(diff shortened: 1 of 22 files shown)


iaincollins:
I [have made an announcement](https://github.com/iaincollins/next-auth/issues/384) about general availability and after considerable testing am going to merge this in and fix forward and talk to the team about what else we might want to include for release of v3.

iaincollins:
Merging in this horrible monster into the v3 before it gets any bigger :-) We are going to do integration testing in the v3 branch - and release beta updates form there - and fix-forward on issues before release.

PR #397, opened 2020-07-06
feat(jwt): Add support for authorization header

Fumler:
Been toying with using [Nexus](https://github.com/graphql-nexus/nexus) together with [Prisma](https://www.prisma.io/). Nexus is a graphql server you can use together with Next.js, as illustrated here: https://github.com/graphql-nexus/examples/tree/master/with-nextjs-and-vercel-and-plugins-prisma

I wanted to secure the graphql api so I decided to use https://github.com/Camji55/nexus-plugin-jwt-auth which uses bearer tokens by default, but is able to take in a custom `verify` function which is perfect to use with `getJwt` from next-auth.

I am currently using next-auth like this in nexus:
```javascript
use(
  auth({
    appSecret: secret,
    protectedPaths,
    verify: async (req) => {
      try {
        let token
        if (req.headers.authorization && req.headers.authorization.split(' ')[0] === 'Bearer') {
          const encodedToken = req.headers.authorization.split(' ')[1]
          const urlDecodedToken = decodeURIComponent(encodedToken)
          token = await decode({ secret, key: secret, token: urlDecodedToken })
        } else {
          token = await getJwt({ req, secret })
        }

        if (token) {
          return token
        } else {
          return {
            token: null,
          }
        }
      } catch (error) {
        console.error(error)
        return {
          token: null,
        }
      }
    },
  })
)
```

but figured there might be more use cases for people to use bearer tokens and since the extra code was minimal to make it work.

With a modified `getJwt` I can reduce the above code to become:
```javascript
use(
  auth({
    appSecret: secret,
    protectedPaths,
    verify: async (req) => await getJwt({ req, secret }),
  })
)

```

Feel free to close the PR if this isn't something you want.

### DIFF
--- a/src/lib/jwt.js
+++ b/src/lib/jwt.js
@@ -26,12 +26,20 @@ const getJwt = async ({ req, secret, cookieName, maxAge }) => {
 
   const secureCookieName = '__Secure-next-auth.session-token'
   const insecureCookieName = 'next-auth.session-token'
-  const cookieValue = cookieName ? req.cookies[cookieName] : req.cookies[secureCookieName] || req.cookies[insecureCookieName]
+  let token = cookieName ? req.cookies[cookieName] : req.cookies[secureCookieName] || req.cookies[insecureCookieName]
 
-  if (!cookieValue) { return null }
+  if (!token) {
+    // Check authorization header if no cookie is present
+    if (req.headers.authorization && req.headers.authorization.split(' ')[0] === 'Bearer') {
+      const urlEncodedToken = req.headers.authorization.split(' ')[1]
+      token = decodeURIComponent(urlEncodedToken)
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


iaincollins:
@Fumler Hi there!

I realised there was a problem with how the existing helper so refactored it in #415 with these changes in mind. As well as adding new features, I want to make a breaking change to the behaviour in v3 that might catch folks out and I'd rather take advantage of that to rename the feature too.

Would appreciate any feedback! I've published it as `next-auth@3.0.0-beta.17` if you have any time to try it out!

iaincollins:
Thanks again for raising this! Closed as should be resolved by #415.

I'll be pushing out a new `next-auth@beta` release today (and updating the release notes) to include this and better document the breaking changes.

PR #398, opened 2020-07-06
Fix using merged models var instead of user provided models in props

tmayr:
While researching how to use custom models found a tiny bug.

When you specify a `model` prop in your adapter initialization, this prop gets preference used in the spread in L32 instead of the previously built/merged `models` variable.

Putting the spread at the beginning and `models` later will make sure that we overwrite whatever the user custom models are with the merged object from default/user provided.

Happy to fix any wording or whatever else is needed, working on a Tutorial for the usage of custom models also.

### DIFF
--- a/src/adapters/typeorm/index.js
+++ b/src/adapters/typeorm/index.js
@@ -28,7 +28,7 @@ const Adapter = (typeOrmConfig, options = {}) => {
   // anything to do them). This function updates arguments by reference.
   adapterTransform(typeOrmConfigObject, models, options)
 
-  const config = adapterConfig.loadConfig(typeOrmConfigObject, { models, ...options })
+  const config = adapterConfig.loadConfig(typeOrmConfigObject, { ...options, models })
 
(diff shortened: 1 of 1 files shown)


tmayr:
Seems like it's an issue in `v3` too, happy to make another pr against that or move this oen to `v3` of the release is soon

iaincollins:
Doh! Thanks for spotting this! This looks great!

Some tutorials for this would be amazing, we get asked about this a lot! It's an intended feature but not one we have any information that explains how to actually use it.

Heads up: I might be refactoring this file to change how transforms work for custom models so it's easier to use them (e.g. work around timestamp type conversion, as TypeORM doesn't handle that automatically and if you are using a Postgres it means timestamps are wrong as don't get saved as UTC).

https://github.com/iaincollins/next-auth/blob/ace756d0cdcd4552e2520b6a74725f69dfddab90/src/adapters/typeorm/lib/transform.js

PR #401, opened 2020-07-06
Add tutorial on how to use custom typeorm models

### LINKED ISSUE #283: Document how to use custom models
Add documentation for how to use custom models with NextAuth.js

It is possible to use custom models (e.g. for Users, Accounts, Sessions, etc) with NextAuth.js but there are no examples and documentation that cover how to actually do this.

We can add documentation that shows how to add or modify fields, and cover common use cases.

For example to use UUID's instead of `int` types for ID columns in an SQL databases, or to set a different default value for a property, or do expand an existing model to add additional fields or Foreign Key constraints.

**Potential problems**
Some properties on models are required / assumed by the Default Adapter and there will be problems if they are renamed in the model.

For example:

* The `id` property (`_id` in MongoDB) is assumed on all models.
* The `name` property is assumed for the users name.
* The `email` property is assumed in the user model. 

Impact / mitigation:

* If the name property is renamed then the users name will not be returned in the session / JWT. This could be mitigated by adding an additional property to the `session` option that is an array of all properties on the model that should be returned in the session.
  e.g. with a default value of of `['name', 'email']`
  Alternatively, this could be a function on the model (to keep all that logic together).
* If the email property is renamed then email sign in will not work correctly. This could be mitigated by making this field an option.

I would not recommended we encourage people to rename properties, as it's likely to be problematic and there may lead to quite serious bugs that are not obvious at first.

Alternatively , we could not let people pass custom models and instead document how to clone the default adapter do that instead.

**Additional context**
* Provides a way to address an issue raised in #254 (RE: UUID's).

tmayr:
Docs on how to use custom typeorm models, goes only as far as specifying how to pass in a custom model, how to extend/inherit from the default ones.

The rest should be up to your typeorm expertise/goals

Maybe I should add a warning about overriding the default properties as they might break something?

Fixes https://github.com/iaincollins/next-auth/issues/283

### DIFF
--- a/www/docs/tutorials/custom-typeorm-models.md
+++ b/www/docs/tutorials/custom-typeorm-models.md
@@ -1,0 +1,80 @@
+---
+id: custom-typeorm-models
+title: Custom TypeORM Models
+---
+
+## Using Custom TypeORM Models
+
+NextAuth.js provides TypeORM models based on [the default tables](https://next-auth.js.org/schemas/models).
+
+You can pass in models extending the defaults by specifying the `models` property on your `adapter` configuration object as the second parameter.
+
+```js title="api/auth/[...nextauth].js"
+import NextAuth from "next-auth";
+import Providers from "next-auth/providers";
+import Adapters from "next-auth/adapters";
+
+import Models from "../../../models";
+
+const options = {
+  providers: [
+    // your providers
+  ],
+
+  adapter: Adapters.TypeORM.Adapter(
+    {
+      // first argument are your database options
+      database: "your-database-string",
+    },
+    {
+      // second argument are your custom adapter options
+      models: {
+        User: Models.User,
+      },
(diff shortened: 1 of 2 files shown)


iaincollins:
Thank you this is great, I really appreciate it!

I'll merge it in to the v3 branch so that it's published when the fix you made goes out.

www/docs/tutorials/custom-typeorm-models.md
```diff
@@ -0,0 +1,80 @@
...
+
+const options = {
+  providers: [
+    // your providers
+  ],
+
+  adapter: Adapters.TypeORM.Adapter(
+    {
+      // first argument are your database options
+      database: "your-database-string",
```
dineshmatta, line 27: database key will not work, correct key here is url

PR #404, opened 2020-07-07
Fix wrong syntax in Apple provider example

okuryu:
It fixes a syntax error in Apple provider example code.


iaincollins:
Thanks!

Issue #406, opened 2020-07-07
Add a URL/Webhook Provider


tmayr:
**Summary of proposed feature**
This would be a similar strategy Email, the main difference would be that it let's you pick an attribute to identify your users and returns an valid URL to be able to continue the flow, how you send that URL to the user is 100% up to you.

**Purpose of proposed feature**
Main motivation for this is that I need to send Authentication links through a Message Provider and not via email, this entails a few things:

- My method of delivery is 100% custom, more akin to a `Webhook` than anything
- My main identifier is not `credentials`, nor `email`, but the `phoneNumber` of the user, which in this case is a custom attribute, not handled by `next-auth` but extended via models by me, and could be any other attribute if intended
- the URL Provider could in the near future be used by the `email` provider, as the primitive is just generating a URL based on a `attribute` (email) and `value` (user email)

**Detail about proposed feature**
I have a very crude PoC running, but this follows 90% of the same flow as the `email provider` does, except it takes an attribute name and value instead of assuming email, you would use a `attribute:value` format to store it

**Potential problems**
I don't foresee any real issues other than quite a bit of code repetition as the flow is, as stated before, 90% like the email, this is definitely a solvable problem in the future as we could just migrate the `email provider` to use the more generic methods of the `url provider` in the future.

**Describe any alternatives you've considered**
I don't see any alternatives as having an arbitrary unique identifier is currently not supported, as for the approach, I thought of naming it `webhook provider` but since the implementation after getting the URL could be anything, I opted for `URL provider`, open to changing the name to whatever makes more sense.

Obviously, happy to contribute the PR to this if it's something that sounds interesting for the project itself.

iaincollins:
Thanks for raising this! Things are a little busy getting v3 in shape for release, but I would love to pick this one up when that is done.

I had something like this in mind when working on the email verification so tried to keep the mechanism generic but I didn't give the actual flows much thought (beyond maybe assuming it might be useful to store different types of token).

I wonder if adding some sort of callback to providers - or supporting the verification callback used in the email provider - in any provider might be good direction.

I'd be interested in a PoC, but actually just a step by step walk through of what the functionality should be might be really helpful as might be able to take an approach that is as generic as possible so can work for other scenarios.

I think using custom credentials (with a custom model) to trigger some sort of verification flow via a phone number shouldn't actually be technically difficult to get to if we can map out the flow!

tmayr:
[...]

PR #409, opened 2020-07-08
Add webhook provider (Proof of Concept)

tmayr:
This is 101% a proof of concept that needs to be cleaned up, it has my prettier configuration applied so even the review is a bit messy, just focus on the big code blocks which were added more than anything.

There's a lot of repetition as I mentioned, because the `email provider` is 90% similar, so taking on the same flow, I mostly copied/pasted and made it a bit more generic with taking `attribute` and `value` instead of hardcoding email and assuming it.

It also changes how the `identifier` is used, before it was just the `email` value, now its `attribute:value`

For now, `phoneNumber` is hardcoded in some parts, but thinking this would come from the `provider` configuration (say you specify `{ validationAttribute }` or something like that, so you could specify whatever you want.

Will leave some comments on the blocks that are actually behavior change

@iaincollins Take a look whenever you can!, will wait for v3 before polishing this based on whatever comes up.

Refs #406


src/providers/webhook.js
```diff
@@ -0,0 +1,19 @@
+export default (options) => {
+  return {
+    id: "webhook",
+    type: "webhook",
+    name: "Webhook",
+    maxAge: 24 * 60 * 60, // How long link are valid for (default 24h)
+    ...options,
+    sendVerificationRequest: ({
```
tmayr, line 8: The name is "inherited" from the `email` provider, nothing says it couldn't change.

src/server/lib/callback-handler.js
```diff
@@ -203,22 +221,60 @@ export default async (sessionToken, profile, providerAccount, options) => {
...
             session,
             user,
-            isNewUser
+            isNewUser,
+          };
+        }
+      }
+    } else if (providerAccount.type === "webhook") {
+      // If signing in with an email, check if an account with the same email address exists already
+      const userByEmail = profile.phoneNumber
```
tmayr, line 238: `phoneNumber` is hardcoded here, need to find out a way to pass in the attribute to this method, and obviously names are off because of copying/pasting email behavior

src/server/lib/callback-handler.js
```diff
@@ -203,22 +221,60 @@ export default async (sessionToken, profile, providerAccount, options) => {
...
+        ? await getUserByAttribute("phoneNumber", profile.phoneNumber)
```
tmayr, line 239: This is the main change from the email, `getUserByAttribute` let's you specify an attribute name and a value instead of assuming email

src/server/lib/callback-handler.js
```diff
@@ -203,22 +221,60 @@ export default async (sessionToken, profile, providerAccount, options) => {
...
+        if (isSignedIn) {
+          if (user.id !== userByEmail.id && !useJwtSession) {
+            // Delete existing session if they are currently signed in as another user.
+            // This will switch user accounts for the session in cases where the user was
+            // already logged in with a different account.
+            await deleteSession(sessionToken);
           }
         }
+
+        // // Update emailVerified property on the user object
```
tmayr, line 254: Need to figure out how to deal with this, maybe have a `verificationProperty` on the `provider` so we know which field to set to true when verification.

src/server/lib/signin/webhook.js
```diff
@@ -0,0 +1,37 @@
+import { randomBytes } from "crypto";
```
tmayr, line 1: Exact same as `src/server/lib/signin/email.js` but recieves an extra value (`(attribute, value, provider, options)` instead of `(email, provider, options)`)

src/server/lib/signin/webhook.js
```diff
@@ -0,0 +1,37 @@
...
+    // Send email with link containing token (the unhashed version)
+    const url = `${baseUrl}/callback/${encodeURIComponent(
+      provider.id
+    )}?attribute=${attribute}&value=${encodeURIComponent(
+      value
+    )}&token=${encodeURIComponent(token)}`;
+
+    // @TODO Create invite (send secret so can be hashed)
+    await createVerificationRequest(
+      attribute + ":" + value,
```
tmayr, line 24: We change the identifier here to `attribute:value` to keep track of what we're validating in the db

src/server/routes/signin.js
```diff
@@ -74,38 +82,41 @@ export default async (req, res, options, done) => {
...
         `${baseUrl}/verify-request?provider=${encodeURIComponent(
           provider.id
         )}&type=${encodeURIComponent(provider.type)}`
-      )
-    res.end()
-    return done()
+      );
+    res.end();
+    return done();
+  } else if (provider.type === "webhook") {
```
tmayr, line 113: This would read from the body to get the correct info and pass it on, implementation is missing mainly because of this being a PoC, not sure how similar its gonna be to email.

tmayr:
Left tons of comments on the relevant parts, excuse the prettier mess!

src/adapters/typeorm/index.js
```diff
@@ -128,224 +146,338 @@ const Adapter = (typeOrmConfig, options = {}) => {
...
-        logger.error('GET_USER_BY_ID_ERROR', error)
-        return Promise.reject(new Error('GET_USER_BY_ID_ERROR', error))
+        logger.error("GET_USER_BY_EMAIL_ERROR", error);
+        return Promise.reject(new Error("GET_USER_BY_EMAIL_ERROR", error));
       }
     }
 
-    async function getUserByEmail (email) {
-      debugMessage('GET_USER_BY_EMAIL', email)
+    async function getUserByAttribute(attribute, value) {
```
tmayr, line 173: Basically the same as `getUserByEmail` but with 2 params

iaincollins:
Thank you for all the comments on this! <3

iaincollins:
Still working on the backlog, but will circle back to this!

balazsorban44:
Hi @tmayr, would you be interested in getting this up-to-date?

tmayr:
@balazsorban44 Sure, any insights on what has changed/or if theres any extra modifications it would require?

balazsorban44:
Hi there! So after a recent chat with another maintainer of this repo, we discussed how we could expand on the `email` provider to support similar ways of authentication, like SMS as well. I am going to close this PR for now, and soon create an RFC for this new provider, which we probably will call something like `token` provider or something. Your PR has had an influence on me, so thank you!

PR #412, opened 2020-07-09
Use id_str for reading in Twitter user ID

jharding:
Hey there!

I ran into a problem authenticating with a newer account on Twitter. The ID of the account is `948493988897554433`, but that number can't be represented in JavaScript (evaluates to `948493988897554400`). As a workaround, we can read in the string value, `id_str`. See https://developer.twitter.com/en/docs/tweets/data-dictionary/overview/user-object

The additional change came from running `yarn run lint --fix`.


iaincollins:
<3 Thank you for spotting this, for the fix and link to the appropriate documentation! We'll get this out in the next release.

PR #415, opened 2020-07-09
Refactor JWT support

iaincollins:
## Changes

* Adds option to disable encryption
* Easy to add custom helper
* Removed getJWT helper
* Added getToken helper
* Helper does not fallback to trying non-prefixed cookie on HTTPS sites
* Supports bearer tokens in HTTP header on helper #397

## getToken() Example

```js
import jwt from 'next-auth/jwt'

const secret = process.env.JWT_SECRET

  const token = await jwt.getToken({ req, secret })
  console.log('JSON Web Token', token)
  res.end()
}
```

**Update:** This should work as of `next-auth@3.0.0-beta.18`.

(The branching was handled incorrectly and the merge included an extra commit it shouldn't have.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "3.0.0-beta.16",
+  "version": "3.0.0-beta.17",
   "description": "An authentication library for Next.js",
   "repository": "https://github.com/iaincollins/next-auth.git",
   "author": "Iain Collins <me@iaincollins.com>",
--- a/src/lib/jwt.js
+++ b/src/lib/jwt.js
@@ -1,37 +1,97 @@
 import jwt from 'jsonwebtoken'
 import CryptoJS from 'crypto-js'
 
-const encode = async ({ secret, key = secret, token = {}, maxAge }) => {
+const DEFAULT_ENCRYPTION_TYPE = 'AES' // Can be one of [ 'AES', false ]
+
+const encode = async ({
+  secret,
+  key = secret,
+  token = {},
+  maxAge,
+  encryption = DEFAULT_ENCRYPTION_TYPE
+}) => {
   // If maxAge is set remove any existing created/expiry dates and replace them
   if (maxAge) {
     if (token.iat) { delete token.iat }
     if (token.exp) { delete token.exp }
   }
+
   const signedToken = jwt.sign(token, secret, { expiresIn: maxAge })
-  const encryptedToken = CryptoJS.AES.encrypt(signedToken, key).toString()
-  return encryptedToken
+
+  switch (encryption) {
+    case false:
+      return signedToken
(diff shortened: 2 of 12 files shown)


Fumler (approved):

Fumler:
Hmm, using header is working for me with the beta version, but seems to have stopped working with cookies. Looking into it.

iaincollins:
> Hmm, using header is working for me with the beta version, but seems to have stopped working with cookies. Looking into it.

Oh! Sorry I was hoping you would not run into that. I wasn't sure if you ALSO needed the cookie.

I hope this explanation makes some sense… 😬 

That's basically the flaw in the pervious approach I was trying to address, the old method fell back to using an unprefixed cookie - but it probably shouldn't really, as (even though it's highly unlikely and very contrived and impractical scenario involving at least two other flaws) it *could* be used to trick a server into using a token from a client it didn't issue.

The new method only uses the insecure cookie if it is safe to do so - i.e. no `NEXTAUTH_URL` is set *or* `NEXTAUTH_URL` is an HTTP (not HTTPS) URL. However, this means if you try to use the `getToken()` helper in another application, it will only use the insecure cookie.

I anticipated this might be an issue so provided the option to set `secureCookie` directly.

```js
getToken({
  req,
  secret,
  secureCookie: (process.env.NODE_ENV === 'production) ? true : false
}),
```

I'd be interested to know if this makes sense, if it works as expected given the above, and what you (and anyone else in core) thinks of the change.

Fumler:
I'm testing it on `http`, and I don't think that's the issue. Debugging so far is telling me the issue is in the `decode` method. It's complaining about `Malformed UTF-8 data`, but the final value in `encode` and the value passed in to `decode` is the same. So not quite sure what is happening here.

Fumler:
The problem was in my options I only set:
```javascript
jwt: {
  secret: process.env.JWT_SECRET
}
```

but in the changes in this PR it seems the encryption `key` is not set to `secret` when `key` is not set explicitly anymore, so there was a different `key` during encryption vs decryption.

Solution for now is to just change the above config to:
```javascript
jwt: {
  secret: process.env.JWT_SECRET,
  key: process.env.JWT_SECRET
}
```

Not sure if the `key` thing is intended or not, but if not it should use `secret` if `key` is not set explicitly.

iaincollins:
@Fumler Thanks so much for trying this out and for all the feedback!

* I've done some testing and I don't need to specify the key as well (and to confirm, you are right - you shouldn't need to!)

   This might be an issue specific to the beta release I cut, but I can't replicate it with the PR so I think it is good to go.

* RE: The horrible error when a token is invalid.

   Ugh, yeah I think we should display a simpler warning message when there is a problem reading a token. The message that gets logged when a token cannot be validated is horrible and not actually helpful.

PR #424, opened 2020-07-12
fallback if user.name is null

styxlab:
Github user profiles may not have a user name (can be null), but they always have a login name. Login name should be used as default or at least as a fallback method.


LoriKarikari:
Thanks @styxlab

PR #426, opened 2020-07-13
JWE / JWA / JSON Object Signing and Encryption

### LINKED ISSUE #224: Review and update JSON Web Token (JWT) payload
To follow good practice, and reduce payload size, properties in JSON Web Token should follow the JWT claims spec at https://www.iana.org/assignments/jwt/jwt.xhtml#claims

## JWT format

The current contents of a JWT created by next auth looks like this:

```js
{
  user: {
    name: 'Iain Collins',
    email: 'me@iaincollins.com',
    image: 'https://example.com/image.jpg',
    id: 1 // User ID will note be specified if used without a database
  },
  // The account object stores details for the authentication provider account 
  // that was used to sign in. It only contains exactly one account, even the 
  // user is linked to multiple provider accounts in a database.
  account: {
    provider: 'google',
    type: 'oauth',
    id: 3218529,
    refreshToken: 'cc0d32d79145091cd6cd8979f0a6d6b67d490899',
    accessToken: '931400799b4a980715bb55af1bb8e01d92316956',
    accessTokenExpires: null
  },
  isNewUser: true, // Is set to true if is first sign in
  iat: 1591150735, // Issued at
  exp: 4183150735  // Expires in
}
```

If we followed the JWT claims spec we would remap the properties and use namespacing for values not covered in the claims spec, and it might look something like this:

```js
{
  name: 'Iain Collins',
  email: 'me@iaincollins.com',
  picture: 'https://example.com/image.jpg',
  client_id: 1,
  "https://next-auth.js.org": {
    oauth_provider: 'google',
    oauth_client_id: 3218529,
    oauth_rt: 'cc0d32d79145091cd6cd8979f0a6d6b67d490899',
    oauth_at: '931400799b4a980715bb55af1bb8e01d92316956',
    oauth_at_exp: null,
    isNewUser: true,
  },
  iat: 1591150735, // Issued at
  exp: 4183150735  // Expires in
}
```

### Implementation

We could keep the current behaviour as is by default (to avoid a breaking change) and put it behind an option.

We could then make this option the default in a future release - which would be a breaking change.

e.g.

```js
jwt: {
  strict: true, // use claims structure (default: false - i.e. current behaviour)
  namespace: "https://next-auth.js.org", // where non-standard claims will be stored
  encrypt: true // to provide this as a simple option for those who need it for compatibility (default: true)
}
```

### Security

*  It's already possible to set custom encode/decode callbacks that define how both encryption and signing work, but a simple toggle for the default behaviour would make things easier for a lot of folks if they want to support interoperability with other services on the same domain and know that the information does not need to be hidden from the client (e.g. contains access levels, but not private keys, session tokens or passwords).
* JSON Web Tokens (RFC 7519 ) are themselves not encrypted when signed (RFC 7515), merely encoded and easily read - unless RFC 7516 (JWE) is also implemented. The default behaviour for NextAuth.js is - and is expected to remain - signing then encrypting, which is widely regarded as best policy, but how it is encrypted by default is under review.
* The method currently used to encrypt JSON Web Tokens is a symmetric-key algorithm using AES. It is recommended to specify a dedicated key for encryption and a secret for signing. If not specified, these are auto-generated from a hash of the configuration object (which itself contains secret tokens, making it a useful mechanism to predictably generate a secure private token). 
* NextAuth.js is an implementation of [RFC 7519]( https://tools.ietf.org/html/rfc7519) (JWT) and [RFC 7515](https://tools.ietf.org/html/rfc7515) (JWS) and then encrypt the entire message with AES. We do not currently implement [RFC 7516](https://tools.ietf.org/html/rfc7516) (JWE), but it is under consideration; the primary caveat being there are some implementation details it's crucial to get right if we go down this road. It would need to a be a point release as changing this mechanism would invalidate existing sessions, unless backwards compatibility was explicitly support.
* The most complete and commonly used implementation for Node.js seems to be [node-jose](https://www.npmjs.com/package/node-jose) from CISCO 
### Rotation

How the token is rotated may be worth reconsidering for those who want to pass tokens through to other services.

Currently tokens are updated on every access and the `iat` and `exp` are both reset when a token is rotated. This is by design, to allow for short lived tokens without impacting the user experience. 

It is unclear if there is a better way to do this and if the current usage is problematic.

* We could set `auth_time` on initial login to preserve the initial auth time
* We could preserve `iat` and just updating `exp` and `updated_at` - this seems the most obvious option, but would require some extra work due to behaviour of the current JWT library (that is why `iat` is currently reset on update).

### Resources

* https://tools.ietf.org/html/rfc7516 JSON Web Encryption (JWE)
* https://tools.ietf.org/html/rfc7515 JSON Web Signature (JWS)
* https://tools.ietf.org/html/rfc7518 JSON Web Algorithms (JWA)
* https://tools.ietf.org/html/rfc7517 JSON Web Key (JWK)
* https://tools.ietf.org/html/rfc7519 JSON Web Token (JWT)
* https://tools.ietf.org/html/rfc7520 Examples of Protecting Content Using JOSE

iaincollins:
**This was not originally planned as part of the security enhancements in 3.0 but after reviewing the options and underlying issues I decided it was worth looking at and to try and roll in this improvement early.**

_Resolves #224 (Review of JWT Payload)_

## Summary

Following an evaluation of JSON Object Signing and Encryption implementations, [panva/jose](https://github.com/panva/jose) was selected for the NextAuth.js JWE implementation. This library was chosen based on the level of completeness, active maintenance, the API and quality of documentation.

While other options offered equal or greater functionality they did not offer a level of completeness with regard to quality of API (including error handling behaviour) and documentation when it came to implementation.

Other clients that were evaluated:

* [square/js-jose](https://github.com/square/js-jose)
* [kjur/jsrsasign](https://github.com/kjur/jsrsasign)
* [cisco/node-jose](https://github.com/cisco/node-jose)

_The implementation of JOSE replaces the signing-with-encryption (using HMAC and symmetric AES) method used in v2 which was conceptually similar but did not use JSON Object Signing and Encryption and did not have built-in support for asymmetric keys (though custom encoding and decoding functions were possible in v2 and are still supported in v3)._

## Features

* Can use separate keys for signing and encryption and a - both can be either symmetric or asymmetric.
* Enable signing only (without encryption) by setting `jwt: { encryption: false }` - this is useful if you know you are only storing public data and want to reduce the size of the payload (or store additional data in the JWT).
* If JWT is enabled but no keys configured, auto-generates symmetric keys from secret for developer convenience (HS512 for signing, PBES2-HS512+A256KW for encryption), using the secret generated from configured provider credentials.

  Relying on auto-generated keys generates a console warning on startup; auto-generated symmetric keys are not intended for use in production - a shared key is not how PBES2 is intended to be used; ECDH-ES+A256KW is recommended by [RFC 7518](https://tools.ietf.org/html/rfc7518#section-4.1) - but working out of the box with JWE (when JWT is enabled) without additional configuration allows people to get started quickly with encrypted JSON Web Tokens.
* There is scope for improving this by automatically generating public/private keys, but there are challenges there as they would need to be persisted somewhere (e.g. in an environment variable) as the Serverless model means that new keys cannot be automatically generated at start up because they would not be consistent across instances.

  There are some fundamental developer experience challenges with an open source solution vs software as a service model that are not easy to resolve.

## Standards Supported

We now have functional and largely complete implementation of the following standards:

* https://tools.ietf.org/html/rfc7515 JSON Web Signature (JWS)
* https://tools.ietf.org/html/rfc7516 JSON Web Encryption (JWE)
* https://tools.ietf.org/html/rfc7517 JSON Web Key (JWK)
* https://tools.ietf.org/html/rfc7518 JSON Web Algorithms (JWA)
* https://tools.ietf.org/html/rfc7519 JSON Web Token (JWT)

This incorporates support for:

* https://tools.ietf.org/html/rfc7638 JSON Web Key Thumbprint 
* https://tools.ietf.org/html/rfc7797 JSON JWS Unencoded Payload Option
* https://tools.ietf.org/html/rfc8037 CFRG Elliptic Curve ECDH and Signatures 

_See also https://tools.ietf.org/html/rfc7520 Examples of Protecting Content Using JOSE_

## JWT Payload

A decision has been made to significantly reduce the default payload and only use valid claim names in the default payload.

```js
{
  name: 'Iain Collins',
  email: 'me@iaincollins.com',
  picture: 'https://example.com/image.jpg',
  "iat": 1594601838,
  "exp": 1597193838
}
```

* These changes will reduce the default cookie size for everyone.
* The changes to the JWT default payload will be a breaking change for anyone using accessing the JWT directly.
* The `iat` and `exp` is updated on rotation as before, no further changes to this have been deemed necessary.
* Support setting `issuer` to enforce `iss` check.
* Additional properties can be saved in the JWT when a user signs in using the `jwt` callback (provision for this has been improved by providing better metadata the first time the method is invoked, at sign in). How to do this is a frequently asked question and is something we need a tutorial for.

## Known issues

* Auto-generating symmetric keys is convenient but not a secure approach at in production as it will be possible to guess keys given it is easy to trigger regenerating them on demand. We will have work to do to nudge people to make their installations more secure once they are up and running through tutorials and nudging behaviour (e.g. informative and helpful warning messages that link to documentation).
* The existing (FAQs, Options, Callbacks) needs updating before this can be merged.
* We will need tutorials to show people how to generate keys for signing and encrypting and to dive into some of the concepts (e.g. using [JOSE Tools](https://github.com/phish108/node-jose-tools) with commands like `jose newkey -s 256 -t ec -a ECDH-ES+A256KW` to generate suitable keys).
* Will still have `jsonwebtoken` and `jwt-decode` as redundant dependancies (e.g. for validating tokens from Open ID providers and creating tokens for Sign In With Apple). I'm reluctant to refactor where they are used PR, it is probably less risky to make that change in another follow up PR and test those changes discretely.


panva:
If I may recommend, use HKDF (rfc5869, e.g. using https://www.npmjs.com/package/futoin-hkdf or your own implementation) to derive the crypto keys out of a secret when "auto-generated" is used. Using HKDF you can consistently derive a secret of a specific length from your shared secret and avoid using the slow an inefficient `PBES2-HS512+A256KW`. Since you want to do symmetric encryption i'd recommend using `alg: 'dir', enc: 'A256GCM'` instead.

Same for signing, you can use different `info` labels for HKDF to end up with two different secrets, one for signing, other for encryption.

We use this in our new express sdk beta branch now - see [derivation](https://github.com/auth0/express-openid-connect/blob/beta-2/lib/hkdf.js), [JWE session encryption](https://github.com/auth0/express-openid-connect/blob/beta-2/lib/appSession.js).

Bottom line `PBES2-HS512+A256KW` is a terrible and slow default.

iaincollins:
@panva Thank you, your time and this feedback is much appreciated!

I've been on the fence about enabling JWE by default (I really want to nudge people in the direction of generating asymmetric keys) but being able to auto-generate two separate keys in this way as default is much better.

panva:
I don't see the point in using asymmetric JWE encryption when you're both the recipient and issuer. For JWS Signing, i am 100% with you, using ECDSA (for performance and signature sizes) or EdDSA (also for balanced performance, but only when Node >= 12.0.0 is used).

src/lib/jwt.js
```diff
@@ -1,70 +1,144 @@
...
+  if (encryption) {
+    // Encryption Key
+    let encryptionKeyValue = encryptionPublicKey || encryptionPrivateKey
+    if (!encryptionKeyValue) {
+      if (!ENCRYPTION_KEY_WARNING) {
+        logger.warn('JWT_ENCODE_AUTO_GENERATED_ENCRYPTION_KEY')
+        ENCRYPTION_KEY_WARNING = true
+      }
+      const autoGeneratedEncryptionSecret = createHash('sha256').update(secret).digest('hex')
+      encryptionKeyValue = `{"kty":"oct","kid":"nextauth-auto-generated-encryption-key","alg":"${DEFAULT_ENCRYPTION_ALGORITHM}","k":"${autoGeneratedEncryptionSecret}"}`
```
panva, line 63: > `"k":"${autoGeneratedEncryptionSecret}"}`

This is not what `k` should be :) `k` are supposed to be base64url encoded octets.

So, e.g. When you have a 32 byte buffer to be used for `A256GCM` direct encryption just instantiate your key like so

```js
const key = jose.JWK.asKey(buffer, { alg: 'A256GCM', use: 'enc', kid: 'nextauth-auto-generated-encryption-key' }
```

iaincollins: Thanks! Yes this was terrible but 'works' as the hash didn't contain any characters that are not valid for Base64 encoding (and didn't get stored anywhere except, in memory so nothing mangled it). HKDF is a MUCH better solution and I will definitely make that the default method for key generation instead of this hackery.

iaincollins:
It's still WIP but I've [updated this based on feedback](https://github.com/iaincollins/next-auth/blob/c5e4621eaeead3aa5b1d8c008d5d1e74ed8b6c75/src/lib/jwt.js) from @panva

iaincollins:
We still need some tutorials, but I am functionally happy with this implementation (the defaults and key generation are much better), so I'm going to merge it into v3 so we can do integration testing and hopefully shake out any issues. I'm happy to still other comments on this PR!

A significant change over v2 is that it defaults to signing only (rather than encryption) which reduces the cookie size considerably, even with zip compression enabled. This is a tradeoff that may resolve issues folks report with regard to not having enough room to store access tokens keys in the JWT.

PR #434, opened 2020-07-15
feat: Add option for non-http cookie to check existence of session

Fumler:
This PR targets the `v3` branch.

Since the session cookie is http only we are not able to check if a user has a session quickly client side without loading the entire app. Adding this cookie will allow you to do check if the user has a session with document.cookie, which will let us do redirects like in the Vercel dashboard for instance.

Example usage:
https://vercel.com/dashboard (check page source)

They have a script tag in head:
```html
<script>
  if (document.cookie && document.cookie.indexOf('authorization=') === -1) {
    location.replace('/login?next=' + encodeURIComponent(location.pathname))
  }
</script>
```

This change will allow you to very quickly redirect the user if not authenticated, since you do not need to mount an entire React app before you can check for the existence of a session. Gives more speedy redirects for pages that require signin and use static generation + client side api requests for data.

```html
<script>
  if (document.cookie && document.cookie.indexOf('next-auth.has-session=') === -1) {
    location.replace('/api/auth/signin')
  }
</script>
```

It is turned off by default, but you can enable it by configuring the options:
```javascript
export default (req, res) => NextAuth(req, res, { cookies: { hasSession: true } })
```

---

Was considering maybe doing something like middlewares/callbacks and providing `req` and `res` to these so a user could implement this themselves, but that seemed like it would require a code change a lot more places, so figured I'd suggest this change instead.

Any comments are appreciated, and as always, feel free to close the PR if this is not a feature you think fits in the project.


iaincollins:
BTW I still like this idea and am sympathetic to the problem as I've had it before too, just been (over) thinking about the best way to handle it from a cookie security PoV (and API design perspective).

PS: Sorry about the v3 update making the branch stale.

Will try never to do that again >.> (at least not any time soon)

Fumler:
Closing this as I think if something like this should be implemented it should probably be in the alternative method mentioned, by exposing e.g. `req` and `res` in a callback. Not sure of the security implications by doing that, I personally think the users themselves are responsible in a case of exposing that in a callback though.

Related issues: https://github.com/nextauthjs/next-auth/issues/543 https://github.com/nextauthjs/next-auth/issues/469

PR #449, opened 2020-07-20
fix(prisma): Explicitly set fields to use when updating user

Fumler:
Targets v3 branch

Attempt at fixing https://github.com/iaincollins/next-auth/issues/384#issuecomment-660640360


iaincollins:
Oh that was fast. :D I didn't know how to do approach addressing this!

Feel free to merge in any changes to the adapter in the v3 branch you think should be in there!

Issue #320, opened 2020-06-24
is it possible to access the raw JWT?


drewwyatt:
*Please refer to the [documentation](https://next-auth.js.org/getting-started/introduction), the [example project](https://github.com/iaincollins/next-auth-example) and existing issues before creating a new issue.*

**Your question**
Please please _please_ forgive me if I've overlooked this in the docs, but i cannot seem to find it: is it possible to access the raw jwt/idtoken (i.e. the string, not the decryped object)

**What are you trying to do**
I need to pass the token to a separate api.

**Documentation feedback**
*Documentation refers to searching through [online documentation](https://next-auth.js.org), code comments and issue history. The example project refers to [next-auth-example](https://github.com/iaincollins/next-auth-example).*

* [x] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [x] Could not find relevant documentation
* [x] Found the example project helpful
* [ ] Did not find the example project helpful

iaincollins:
Thanks for the really clear question!

The short answer is there isn't a simple function for that. The conversation in #305 is a bit relevant though.

## Under consideration

I am thinking of doing one of two things to help in this use case:

* Supporting `jwt: { encrypt: false }` as an option (where JWT would be signed, but not encrypted)
  
This is how JSON Web Tokens are most typically used, but NextAuth.js tries to go further by default, so that it's safe to also so private data that needs to be secret from the user in the JWT.

* Supporting returning the raw JWT (pre-verification) 

I'm interested in feedback regarding which of these (if either) is most useful (or not).

## Workaround

Right now, you would have to write something like this:
```js
import jwt from 'jsonwebtoken'
import CryptoJS from 'crypto-js'

export default async (req, res) =>  {
  // Get token from cookie (names differ on local vs production)
  const secureCookieName = '__Secure-next-auth.session-token'
  const insecureCookieName = 'next-auth.session-token'
  const encryptedToken = req.cookies[secureCookieName] || req.cookies[insecureCookieName]

  // Decrypt using secret and return as string
  const decryptedBytes = CryptoJS.AES.decrypt(encryptedToken, process.env.SECRET)
  const decryptedToken = decryptedBytes.toString(CryptoJS.enc.Utf8)

  // Still signed but not encrypted (this is what you want)
  console.log(decryptedToken)

  // If you wanted to verify it and get the object back, you would do this step too
  const verifiedToken = jwt.verify(decryptedToken, process.env.SECRET)

  res.end()
}
```

[...]

PR #458, opened 2020-07-21
feat: add verify option to getJwt

tmaximini:
As mentioned in #320 it is useful to access the unverified, raw JWT directly in some cases.
For us, we want to communicate with a GraphQL backend and do the verification there, so we need to send that token.
This PR adds a `verify` option to `getJwt()` that defaults to `true`. But if set to `false`, we can get the raw token value.


iaincollins:
Thanks for raising this PR!

This method is actually deprecated in v3 so I'm going to close this, but we will have something very similar soon!

PR #459, opened 2020-07-21
Use getReposistory instead of getManager

rolandreads:
Both getRepository and getManager were being used, not only is it inconsistent, but getManager does not respect connection name specified in config. 

Found this bug when testing Linkedin custom provider, even though I had 
`
const options = {
  database: {
    type: 'postgres',
    name: 'next-auth-connection',
    url: process.env.DATABASE_URL,
    synchronize: process.env.NODE_ENV !== 'production',
  },
...
}
`

I was still getting this error
`
[next-auth][error][CREATE_USER_ERROR] [
  ConnectionNotFoundError: Connection "default" was not found.
`

### DIFF
--- a/src/adapters/typeorm/index.js
+++ b/src/adapters/typeorm/index.js
@@ -1,4 +1,4 @@
-import { createConnection, getConnection, getManager } from 'typeorm'
+import { createConnection, getConnection } from 'typeorm'
 import { createHash } from 'crypto'
 
 import { CreateUserError } from '../../lib/errors'
@@ -111,7 +111,7 @@ const Adapter = (typeOrmConfig, options = {}) => {
       try {
         // Create user account
         const user = new User(profile.name, profile.email, profile.image, profile.emailVerified)
-        return await getManager().save(user)
+        return await connection.getRepository(User).save(user)
       } catch (error) {
         logger.error('CREATE_USER_ERROR', error)
         return Promise.reject(new CreateUserError(error))
@@ -164,7 +164,7 @@ const Adapter = (typeOrmConfig, options = {}) => {
 
     async function updateUser (user) {
       debugMessage('UPDATE_USER', user)
-      return getManager().save(user)
+      return connection.getRepository(User).save(user)
     }
 
     async function deleteUser (userId) {
@@ -178,7 +178,7 @@ const Adapter = (typeOrmConfig, options = {}) => {
       try {
         // Create provider account linked to user
         const account = new Account(userId, providerId, providerType, providerAccountId, refreshToken, accessToken, accessTokenExpires)
-        return getManager().save(account)
+        return connection.getRepository(Account).save(account)
       } catch (error) {
         logger.error('LINK_ACCOUNT_ERROR', error)
         return Promise.reject(new Error('LINK_ACCOUNT_ERROR', error))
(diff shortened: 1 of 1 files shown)


iaincollins:
Can you explain why a custom connection name is something you need?

I am not averse to supporting custom connection names but this is an enhancement and before we make changes that will need regression testing we'd like to understand the context to gauge how pressing the need for this is.

romMidnight:
When using typeorm myself with nextjs & next-auth, custom connection name is required to avoid conflicts between typeorm and next-auth, as they will both use name 'default'

iaincollins (changes requested):
Looking at the at the [docs for the connection manager](https://typeorm.io/#/working-with-entity-manager) and [the source](https://github.com/typeorm/typeorm/blob/51b2a63d/src/connection/Connection.ts#L321) it seems the simplest way to support connection names is to replace `getManager()` with `connection.manager`.

This is the least code and seems to be the most direct approach. If we want to be consistent, we could refactor the other lines with getRepository calls to work the same way too.

This will need regression testing on MySQL, Postgres and MongoDB and a quick pass on SQLite.

romMidnight:
Even without my changes there are 9 usages of connection.getRepository() in that file, I assume if I am possibly introducing regression with connection.getRepository(), 7 other functions already using getRepo are broken too, and the reason I used connection.getRepository() instead of getConnection('name').manager is that connectionName knowledge would be preferable only in _connect function and also using repositories instead of entity manager is a better practice overall

iaincollins:
> Even without my changes there are 9 usages of connection.getRepository() in that file, I assume if I am possibly introducing regression with connection.getRepository(), 7 other functions already using getRepo are broken too, and the reason I used connection.getRepository() instead of getConnection('name').manager is that connectionName knowledge would be preferable only in _connect function and also using repositories instead of entity manager is a better practice overall

Do you have a citation for that?

I've looked at the docs and at the source for TypeORM and can't see anything to suggest that.

romMidnight:
https://typeorm.io/#/active-record-data-mapper, they do not suggest that using repositories is a better practice, but data mapper approach gives more flexibility and makes integrating with other databases easier, as objects are decoupled from database. We can get benefits of data mapper without losing any relative simplicity or bloating codebase with repository related code.

It's a data mapper vs active record debate
https://www.thoughtfulcode.com/orm-active-record-vs-data-mapper/
https://stackoverflow.com/questions/3828265/is-data-mapper-a-more-modern-trend-than-active-record

iaincollins:
> https://typeorm.io/#/active-record-data-mapper, they do not suggest that using repositories is a better practice, but data mapper approach gives more flexibility and makes integrating with other databases easier, as objects are decoupled from database. We can get benefits of data mapper without losing any relative simplicity or bloating codebase with repository related code.
> 
> It's a data mapper vs active record debate
> https://www.thoughtfulcode.com/orm-active-record-vs-data-mapper/
> https://stackoverflow.com/questions/3828265/is-data-mapper-a-more-modern-trend-than-active-record

None of those links even mention TypeORM.

Having looked at what is going on TypeORM under the hood, I'm going to stick recommending the option that is less code and results in less function invocations unless there is actually concrete reason to do otherwise.

If you don't want to make the changes request that's no problem, someone else can pick up refactoring and testing.

iaincollins:
To avoid the overhead of additional testing, I've rolled in changes to address this as part of wider work in #460.

Other changes include a change to the default connection name to help steer people away from re-using the connection NextAuth.js creates (for those not deploying with Serverless and who are sharing connection handles).

PR #461, opened 2020-07-22
Adapter improvements

iaincollins:
I've been trying out the Prisma adapter and addressing some minor issues - for context, I just did the same for TypeORM adapter in #460.

**I've now done quite a bit of testing with Prisma with both MySQL and Postgres and found it works really well.**

This PR includes some enhancements / fixes to the Prisma adapter, a change to the TypeORM adapter for MySQL databases to improve it (addressing an issue with different timestamp fields using different levels of precision) and includes updated documentation useful for all users that has come off the back of testing.

#### Prisma Adapter

* **Updated `getUserByProviderAccountId` to use `compoundId` when searching for accounts.**
  As different providers might use the same ID, the provider name should also be taken into account when searching by account ID and this achieves that.
* **Updated how `linkAccount` and `createSession` store `userId` as ran into issues with it.**
  Similar to the TypeORM have refactored it in a simpler way that avoids a relationship between entities. I appreciate not using relationships is less elegant and atypical with an RDBMS, but it makes compatibility with different databases much easier and has very little performance impact as it's not invoked often (only on sign in) which is why the TypeORM adapter makes this tradeoff. I think this is something that could be more sophisticated later if/when we split off the adapter code.
* **Updated example schema in docs to show how to use it with Postgres.**
  Now includes mappings for the default table/column names and indexes. I think having parity in the expected database schemas will make it easier to keep the documentation up to date in the long run

#### TypeORM Adapter

* **Timestamps in MySQL all consistently declared with `TIMESTAMP(6)` unless a different level of precision is specified.**
   This is consistent with behaviour for created/updated fields in TypeORM for MySQL, and with timestamp fields in Postgres.

#### Documentation

* [Added database questions to FAQ](https://next-auth-docs-git-feature-prisma-adaptor-refactor.iaincollins.vercel.app/faq#databases)
* [Improved database option documentation](https://next-auth-docs-git-feature-prisma-adaptor-refactor.iaincollins.vercel.app/configuration/databases)
* [Added MySQL SQL schema to documentation](https://next-auth-docs-git-feature-prisma-adaptor-refactor.iaincollins.vercel.app/schemas/mysql)
* [Added Postgres SQL schema to documentation](https://next-auth-docs-git-feature-prisma-adaptor-refactor.iaincollins.vercel.app/schemas/postgres)
* [Improved adapter documentation](https://next-auth-docs-git-feature-prisma-adaptor-refactor.iaincollins.vercel.app/schemas/adapters)

I think there is scope to feature the Prisma adapter more prominently in future, but I think this is quite a big improvement for now, especially around transparency of databases schemas and ease of getting started.

The SQL schemas I've added should match up with the schemas generated by TypeORM for the built in models if it's allowed to generate the schema from the modes (I've done some testing to confirm this). This approach informs some of the default types for MySQL as it seems best to be consistent and transparent about it.

For those who know what they are doing, it's fine to use different types for IDs and for timestamps (etc). as long as you don't use the option to `sync` which the default adapter. As we have more comprehensive documentation (with copy+paste SQL schemas), I have made using the `sync` option less prominent in the documentation.

Note: If using Prisma to generate the SQL from scratch with `npx @prisma/cli migrate save --experimental` it will create slightly different SQL to that generate by TypeORM, but the example schema is completely compatible with the SQL schemas provided for MySQL and Postgres (i.e. the schemas in the documentation are valid for both).

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,13 @@
     "test:db:postgres": "node test/postgres.js",
     "test:db:mongodb": "node test/mongodb.js",
     "db:start": "docker-compose -f test/docker/docker-compose.yml up -d",
+    "db:start:mongo": "docker-compose -f test/docker/mongo.yml up -d",
+    "db:start:mysql": "docker-compose -f test/docker/mysql.yml up -d",
+    "db:start:postgres": "docker-compose -f test/docker/postgres.yml up -d",
     "db:stop": "docker-compose -f test/docker/docker-compose.yml down",
+    "db:stop:mongo": "docker-compose -f test/docker/mongo.yml down",
+    "db:stop:mysql": "docker-compose -f test/docker/mysql.yml down",
+    "db:stop:postgres": "docker-compose -f test/docker/postgres.yml down",
     "prepublishOnly": "npm run build",
     "publish:beta": "npm publish --tag beta",
     "publish:canary": "npm publish --tag canary",
@@ -57,7 +63,8 @@
   "peerOptionalDependencies": {
     "mongodb": "^3.5.9",
     "mysql": "^2.18.1",
-    "pg": "^8.2.1"
+    "pg": "^8.2.1",
+    "@prisma/client": "^2.3.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.8.4",
--- a/src/adapters/prisma/index.js
+++ b/src/adapters/prisma/index.js
@@ -1,30 +1,34 @@
-import { createHash } from 'crypto'
+import { createHash, randomBytes } from 'crypto'
 
 import { CreateUserError } from '../../lib/errors'
 import logger from '../../lib/logger'
 
-const Adapter = (prismaConfig, options = {}) => {
+const Adapter = (config) => {
   const {
-    prisma, modelMapping = {
+    prisma,
+    modelMapping = {
       User: 'user',
       Account: 'account',
       Session: 'session',
       VerificationRequest: 'verificationRequest'
     }
-  } = prismaConfig
+  } = config
 
   const { User, Account, Session, VerificationRequest } = modelMapping
 
+  function getCompoundId (providerId, providerAccountId) {
+    return createHash('sha256').update(`${providerId}:${providerAccountId}`).digest('hex')
+  }
+
   async function getAdapter (appOptions) {
-    function debugMessage (debugCode, ...args) {
-      if (appOptions && appOptions.debug) {
-        logger.debug(`PRISMA_${debugCode}`, ...args)
-      }
+    function debug (debugCode, ...args) {
+      logger.debug(`PRISMA_${debugCode}`, ...args)
     }
 
     if (appOptions && (!appOptions.session || !appOptions.session.maxAge)) {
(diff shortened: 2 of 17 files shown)


Fumler (approved):
LGTM 👍

yareyaredesuyo:
I think default db should use sqlite, because it is easiest way and can use without configuration!

And in official prisma examples, almost all the schema.prisma is currently using sqlite as default datasource by default.

like this

```prisma
datasource db {
  provider = "sqlite"
  url      = "file:./dev.db"
}
```

- https://github.com/prisma/prisma-examples/blob/master/typescript/graphql-nextjs/prisma/schema.prisma

And I found your prisma scheme syntax looks like using obsolate style,  it changed drastically in 2.0.0-beta.1

Check This [Article](https://github.com/prisma/prisma/releases/tag/2.0.0-beta.1) ! 

> New syntax for defining relations
> 
> The Beta release introduces a new relation syntax which makes the @relation attribute required in each relation in the Prisma schema. Note that it often is enough to only declare the attribute only on the side of the relation (the side that stores the foreign key in the underlying database).
> 
> Additionally, for one-to-one and one-to-many relations, you need add a relation scalar field to the model which is used in the @relation attribute. This relation scalar field directly maps to the foreign key in the underlying database. Note that the foreign key is read-only in the Prisma Client API, to modify a relation you can keep using nested write queries as before.
> 
> Here's an overview for how relations need to be updated.
> One-to-one
> 
> During the Preview period, a 1-1-relation could be defined as follows:
> 
> model User {
>   id        Int       @id @default(autoincrement())
>   profile   Profile
> }
> 
> model Profile {
>   id      Int    @id @default(autoincrement())
>   user    User
> }
> 
> With the new Beta, you now must determine which side should store the foreign key. You can do so by adding the @relation attribute with its corresponding relation scalar field to the model:
> 
> model User {
>   id        Int       @id @default(autoincrement())
>   profile   Profile
> }
> 
> model Profile {
>   id      Int    @id @default(autoincrement())
> + user    User   @relation(fields: [userId], references: [id])
> + userId  Int    // relation scalar field (used in the `@relation` attribute above)
> }
>

iaincollins:
>I think default db should use sqlite, because it is easiest way and can use without configuration!
> And in official prisma examples, almost all the schema.prisma is currently using sqlite as default datasource by default.

That's good feedback and a valid reason to consider that in the example.

Prisma apparently ships with a version of SQLite too.

On the flip side, this means extra work, as we will still need other tutorials that can actually be used in production (i.e. for Postgres, MySQL) because SQLite is sufficiently different (in set up and behaviour) from a production database.

> And I found your prisma scheme syntax looks like using obsolate style, it changed drastically in 2.0.0-beta.1

The example is valid syntax for the current version (v2.3.0). The notes explain why NextAuth.js doesn't the relation syntax.

_tl;dr It's possible to create an adapter and use enforce relationships if you want that, but NextAuth.js doesn't currently do it for you out of the box because it's non-trivial to support well in a database agnostic way (it would likely involve some code branching for non-relational databases; but at some point it's worth considering splitting off adapter logic from the core to simplify things, but I don't think we want to there until we've resolved some other issues around automation of testing)._

Issue #472, opened 2020-07-24
Error setting up Local Environment following Contribution Guide: "Module not found: Can't resolve 'react'"


BenjaminWFox:
**Describe the bug**
Setting up `next-auth` local environment following the steps in the contribution guide fails without explicitly installing `react` in `next-auth` 

**To Reproduce**
`git clone https://github.com/BenjaminWFox/next-auth.git`
`git clone https://github.com/iaincollins/next-auth-example.git`

`cd next-auth-example`
`npm install`

`cd ../next-auth`
`npm i`
`npm run build`
`npm link ../next-auth-example/node_modules/react`

`cd ../next-auth-example`
`npm link ../next-auth`
`npm run dev`

```
> next-auth-example@3.0.0-beta.19 dev /Users/benfox/Documents/Projects/Lumedic/Repositories/next-auth-example
> next

ready - started server on http://localhost:3000
error - ../next-auth/dist/client/index.js
Module not found: Can't resolve 'react' in '/Users/benfox/Documents/Projects/Lumedic/Repositories/next-auth/dist/client'
```

**To Fix**
`cd ../next-auth`
`npm i -D react`
`npm link ../next-auth-example/node_modules/react`

`cd ../next-auth-example`
`npm run dev`

**Additional context**
The peer dependencies are called out at the various `npm` stages. I'm unsure, and didn't find via google, whether there is some other, better, way to solve this issue.

```
npm WARN next-auth@2.2.0 requires a peer of react@^16.13.1 but none is installed. You must install peer dependencies yourself.
npm WARN next-auth@2.2.0 requires a peer of react-dom@^16.13.1 but none is installed. You must install peer dependencies yourself.
```

If there is not a better way to resolve this issue (and/or if I'm not missing something obvious) and installing react (`npm i -D react`) is, in fact, the best way to resolve this situation I can submit a PR for the documentation in the Contributing Guide.

I would change [step 2 "Install packages and run the build command:"](https://github.com/iaincollins/next-auth/blob/main/CONTRIBUTING.md#setting-up-local-environment) to:
```
npm i
npm i -D react
npm run build
```

**Documentation feedback**
*Documentation refers to searching through [online documentation](https://next-auth.js.org), code comments and issue history. The example project refers to [next-auth-example](https://github.com/iaincollins/next-auth-example).*

* [ ] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [ ] Could not find relevant documentation
* [ ] Found the example project helpful
[...]

PR #475, opened 2020-07-25
docs (contributing) Update local environment setup steps to fix npm link problems per issue #472

BenjaminWFox:
In the Contributing documentation, the order of steps in the local environment setup section was in an order that broke the link between React in `next-auth` and the consuming application.

Reversing the order of Steps 3 & 4 fixed this issue and allowed me to link my project with next-auth for local development.

https://github.com/iaincollins/next-auth/issues/472


iaincollins:
Thanks Ben! This makes sense. I'll check this out on a clean folder!

PR #488, opened 2020-07-27
Re-sync entities on connection in TypeORM if changed in development mode

### LINKED ISSUE #467: RepositoryNotFoundError: v3 related to HMR
*Please refer to the [documentation](https://next-auth.js.org/getting-started/introduction), the [example project](https://github.com/iaincollins/next-auth-example) and existing issues before creating a new issue.*

**Your question**
I am looking an integrating the latest version (v3) into my nextjs application. I have created a custom model and configured the connection options as follows:

```
import Providers from 'next-auth/providers'
import Adapters from "next-auth/adapters"
import Entity from 'entity'

const databaseUrl = 'postgres://postgres:password@localhost:5432/bss?synchronize=true'

export const connectionOptions = {
    providers: [
        Providers.Auth0({
            domain: process.env.AUTH0_DOMAIN,
            clientId: process.env.AUTH0_CLIENT_ID,
            clientSecret: process.env.AUTH0_CLIENT_SECRET,
        })
    ],
    adapter: Adapters.TypeORM.Adapter(
        databaseUrl,
        {
            models: {
                User: Entity.User,
            },
        }
    ),
    database: databaseUrl,
    secret: process.env.SECRET,
}
```
This is imported and used in my [...nextauth].ts file. I am also using a ormconfig.json file with the following:

```
{
    "type": "postgres",
    "host": "localhost",
    "port": 5432,
    "username": "postgres",
    "password": "password",
    "database": "bss",
    "synchronize": true,
    "logging": true,
    "entities": [
       "src/entity/**/*.ts"
    ],
    "migrations": [
       "src/migration/**/*.ts"
    ],
    "subscribers": [
       "src/subscriber/**/*.ts"
    ],
    "cli": {
        "entitiesDir": "src/entity",
        "migrationsDir": "src/migration",
        "subscribersDir": "src/subscriber"
     }
 }
```

This works for as long as i dont make a change to the ts files. If i make a change when the db connection is utilised again the following error is thrown 

> RepositoryNotFoundError: No repository for "User" was found. Looks like this entity is not registered in current "default" connection?

After doing some googling it seem like other libraries have the same issue: https://github.com/nestjs/nest/issues/4283, https://github.com/vercel/next.js/discussions/12254#discussioncomment-7846, https://github.com/typeorm/typeorm/issues/3017. From what i understand its because im referencing the ts files. What is the appropriate way to set this up so that it works with HMR?

There is this solution but it looks like this might need to be implemented in this library itself? 
https://github.com/typeorm/typeorm/issues/6241#issuecomment-643690383

**What are you trying to do**
Define a custom User model and utilise the v3 version of nextAuth.js inside my next.js application.

**Documentation feedback**

deep-c:
Closes #467

### DIFF
--- a/src/adapters/typeorm/index.js
+++ b/src/adapters/typeorm/index.js
@@ -7,6 +7,7 @@ import adapterConfig from './lib/config'
 import adapterTransform from './lib/transform'
 import Models from './models'
 import logger from '../../lib/logger'
+import { updateConnectionEntities } from './lib/utils'
 
 const Adapter = (typeOrmConfig, options = {}) => {
   // Ensure typeOrmConfigObject is normalized to an object
@@ -66,6 +67,10 @@ const Adapter = (typeOrmConfig, options = {}) => {
     } else {
       // If the connection object already exists, ensure it's valid
       await _connect()
+    }
+
+    if (process.env.NODE_ENV !== 'production') {
+      await updateConnectionEntities(connection, config.entities)
     }
 
     // Get manager from connection object
--- a/src/adapters/typeorm/lib/utils.js
+++ b/src/adapters/typeorm/lib/utils.js
@@ -1,0 +1,18 @@
+const entitiesChanged = (prevEntities, newEntities) => {
+  if (prevEntities.length !== newEntities.length) return true
+  for (let i = 0; i < prevEntities.length; i++) {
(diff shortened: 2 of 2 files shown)


iaincollins:
Thanks for this! Going to do some integration testing to try it out in practice before release, but seems good to go!

PR #505, opened 2020-07-29
Fix typo in typeorm-custom-models.md

blms:
[Adapter](https://github.com/nextauthjs/next-auth/blob/main/src/adapters/typeorm/index.js) expects `customModels` property, but doc uses `models`.


iaincollins:
Thanks for spotting and fixing!

PR #511, opened 2020-07-30
Add Provider: Basecamp

marcmll:
Currently working on a project where I am making use of the Basecamp OAuth 2.0 Authentication. Managed to get Basecamp working as a custom provider and thought it might be nice to have it as part of the built-in support.

### Major Changes
• Added Basecamp as a provider
• Added documentation for Basecamp provider

### Minor Changes
• Bumped Version to 3.0.1 in package-lock

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "3.0.0",
+  "version": "3.0.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/providers/basecamp.js
+++ b/src/providers/basecamp.js
@@ -1,0 +1,20 @@
+export default (options) => {
+  return {
+    id: 'basecamp',
+    name: 'Basecamp',
+    type: 'oauth',
+    version: '2.0',
+    accessTokenUrl: 'https://launchpad.37signals.com/authorization/token?type=web_server',
+    authorizationUrl: 'https://launchpad.37signals.com/authorization/new?type=web_server',
+    profileUrl: 'https://launchpad.37signals.com/authorization.json',
+    profile: (profile) => {
+      return {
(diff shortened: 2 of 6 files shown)


LoriKarikari:
thanks @marcmll!

iaincollins:
Really appreciate the docs too, thank you!

balazsorban44:
@marcmll Sorry for coming back to this, but I have a question. Basecamp says:

> We don't recommend using the OAuth API to authenticate users in third party services. For example, to implement a button "Login with Basecamp" similar to others like "Login with Google" or "Login with GitHub". The reason is that we don't verify email addresses. As a result, an attacker could gain access to these services using email addresses they don't own.

https://github.com/basecamp/api/blob/master/sections/authentication.md#on-authenticating-users-via-oauth

What do you think about that?

marcmll:
> @marcmll Sorry for coming back to this, but I have a question. Basecamp says:
> 
> > We don't recommend using the OAuth API to authenticate users in third-party services. For example, to implement a button "Login with Basecamp" similar to others like "Login with Google" or "Login with GitHub". The reason is that we don't verify email addresses. As a result, an attacker could gain access to these services using email addresses they don't own.
> 
> https://github.com/basecamp/api/blob/master/sections/authentication.md#on-authenticating-users-via-oauth
> 
> What do you think about that?

@balazsorban44 this is definitely a concern that I didn't consider when adding the Basecamp login. At the time of implementation, I was building a company internal service that utilized authentication with Basecamp to match a basecamp user id to a Google account (using Google OAuth). I'm not sure how many users are making use of the Basecamp login. At the very least, the NextAuth documentation could include this note, forwarding the message to developers using NextAuth to add Basecamp OAuth. Wdyt?

balazsorban44:
I might be inclined to just remove the provider all together in our next version, to be honest.

PR #512, opened 2020-07-30
Documentation/server reload local environment

BenjaminWFox:
As noted in the documentation, Next.js will not hot reload the server when working with local packages via `npm-link`. 

While trying to find more info on this I found a webpack plugin in [a thread related to the issue](https://github.com/vercel/next.js/issues/5463#issuecomment-471308868) that facilitates clearing & reloading server modules, resulting in a much happier dev experience. Implementing it was pretty straight forward, and these instructions could be added to the documentation if desired.


iaincollins:
That's really cool, thank you!

PR #514, opened 2020-07-30
Add support for Microsoft SQL Server

iaincollins:
Testing out the new Microsoft SQL Server support added by @D10221 

* I'm happy this seems to work well and that it is ready for release.
* I've tested positive and negative flows with a range of providers, including email.
* The only changes to the original excellent Pull Request have been to the index (to allow multiple UNIQUE NULL values on email address, which is behaviour SQL Server handles slightly differently by design) and minor edits to documentation.

I propose we release this as v3.1 - along with the Basecamp provider PR in #511 and the support for entity hot reloading in #488, unless there are any objections or any other features or changes people would like to see make the cut?

Thanks again @D10221!


iaincollins:
@D10221 Thank you for the quick response, that's a really complete answer and I appreciate it! 

I'm not familiar with mssql culture / conventions so wanted to avoid us doing something unconventional by default.

PR #518, opened 2020-07-31
standardize the behaviour of the Docker images

D10221:
Hi,
Answering to your FIXME: 'standardize the behaviour of the Docker images'

The setup is based on a mix of the [Microsoft sample](https://github.com/microsoft/mssql-docker/blob/master/linux/mssql-server-linux/Dockerfile) sample/demo and the [Knex](https://github.com/knex/knex/blob/master/scripts/docker-compose.yml) implementation, which is closer to this project setup.

Let me know what you think.
Thanks
D.

NOTE: let me know if you prefer to be open,an issue  before or after, any pr ;)

### DIFF
--- a/test/docker/mssql.yml
+++ b/test/docker/mssql.yml
@@ -1,20 +1,15 @@
-# @FIXME The behaviour fo the MSSQL image differs slightly from the other images
-#        To make it easier to setupm run tests and maintain we probably want to
-#        standardize the behaviour of the Docker images, but not a priority yet.
-#        There might be a third party image that does this out of the box
-#          e.g. like the one from bitnami we use for mongodb
-#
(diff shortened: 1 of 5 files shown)


iaincollins:
This is great, thank you so much for picking up on that! 

(I tried to do this myself initially but realised how much work it was so took a step back and just left a comment.)

This will be really helpful for our plans for automated tests. :-)

If anyone else has time to review and it works for them am happy to merge!

Otherwise test this and get it merged as soon as I can.

iaincollins:
I forget where I was with this, but am going to merge it in!

It looks good to me and if we run into any issues we can just fix forward. :-)

Thanks again for doing this, it's super helpful!

PR #524, opened 2020-08-02
Update pages.md

hariria:
I believe the path should be /api/auth/signin


iaincollins:
Thanks for the PR!

These paths are actually correct in the documention as they are for custom pages, which can be used instead of the built in pages.

PR #529, opened 2020-08-02
Update pages.md

bowen31337:
as per Next.js doc https://nextjs.org/docs/api-reference/data-fetching/getInitialProps, it is a typo which breaks the expected functionality.


nyedidikeke:
Nice catch @bowen31337  👍🏿

iaincollins:
Already resolve - see #542 and #534

PR #534, opened 2020-08-03
Update pages.md

bowen31337:
correct all typos


iaincollins:
Thanks for the PR! Rolled in some other updates due to conflict with another PR that updated this page.

Issue #544, opened 2020-08-05
redis adapter


nikitapilgrim:
Hi, don't you have an example of connecting a redis as a database?

nyedidikeke:
Following the [documentation](https://next-auth.js.org/configuration/databases), you *should* be able to connect to redis as below;

```js
// pages/api/auth/[...nextauth].js
import NextAuth from 'next-auth'
import Providers from 'next-auth/providers'

const options = {
  // OAuth Service Providers and other details
  // Provide URL to redis to connect (here, from .env file)
  database: process.env.DATABASE_URL
}

export default (req, res) => NextAuth(req, res, options)
```

```
# .env
DATABASE_URL=redis://user:pass@host:port
```

You may want to check [`node-redis`](https://github.com/NodeRedis/node-redis) and [rediSQL](https://github.com/RedBeardLab/rediSQL) while experimenting and share details as redis may/is not currently be supported by default.

iaincollins:
MySQL, Postgres, Microsoft SQL Server and MongoDB are supported.

If you want to use a Redis database you would need to write your own adapter.

Redis would be a really unconventional choice in a context like this and not something support is planned for.

D10221:
Sorry 
I referenced this issue by mistake 
Miss typed it  :(

JakubKoralewski:
@iaincollins 

> Redis would be a really unconventional choice in a context like this and not something support is planned for.

Isn't storing sessions in Redis a good idea? https://stackoverflow.com/a/9153199/10854888
Or do you suggest just using JWT in this case?

designbyadrian:
@nikitapilgrim Did you get anywhere with this? Care to share your solution, pretty please?

LubomirGeorgiev:
@designbyadrian
> @nikitapilgrim Did you get anywhere with this? Care to share your solution, pretty please?

You can use that one https://github.com/quanhua92/next-auth-ioredis-adapter-example/blob/main/lib/IORedisAdapter.ts

Issue #540, opened 2020-08-04
Provide database configuration parameters to the custom adapter 


D10221:
**Summary of proposed feature**
Provide database configuration parameters to the custom adapter 

**Purpose of proposed feature**
Enable composition of the custom Adapter and connection configuration, decoupling

**Detail about proposed feature**
on `server/index.js` the _custom_ adapter is not initialized but left alone to provide it's own database configuration.  
While this is NOT a problem, it forces to implement the Adapter with the connection .
What I'm saying is ...
``` javascript
// line 53 of server/index.js (currently)
let adapter
if (userSuppliedOptions.adapter) {
    // If adapter is provided, use it (advanced usage, overrides database)
    adapter = userSuppliedOptions.adapter
} else if (userSuppliedOptions.database) {
    // If database URI or config object is provided, use it (simple usage)
    adapter = adapters.Default(userSuppliedOptions.database)
}
```

**Potential problems**
Backwards compatibility, currently the documentation, that explicitly suggests to use a built Adapter instead on "constructor" or 'callback' like pattern.  
``` javascript
// line 53 of server/index.js
let adapter
    if (userSuppliedOptions.adapter) {
      // If adapter is provided, use it       
      adapter =  //simplistic: but it might work
        typeof userSuppliedOptions.adapter.getAdapter === "function" 
          ? userSuppliedOptions.adapter 
          : userSuppliedOptions.adapter(userSuppliedOptions.database)
    } else if (userSuppliedOptions.database) {
      // If database URI or config object is provided, use it (simple usage)
      adapter = adapters.Default(userSuppliedOptions.database)
    }
```

**Describe any alternatives you've considered**
I didn't find a workaround that is not providing a built Adapter

**Additional context**
This came up while I was trying to build a simple cli tool to 'sync' the DB without starting the whole application.  
Or changing the url string and been   
And  swap adapters and database configuration  
And  use the inbuilt adapter is to don't re-type  the schemas 

Please see here the implementation
https://github.com/D10221/next-auth-cli   
... or https://github.com/D10221/next-auth-cli/tree/minimal  

NOTE: if you know a way to do this without any change please let me know :)

*Please indicate if you are willing and able to help implement the proposed feature.*
Happy to try the change my self 

Thanks

forgot to mention this would enable a bit more modular configuration, as an example.  
```javascript
/* file: my-custom-adapter.js */
import Adapters from "next-auth/adapters.js";
/**
 * Adds extra field to standard models
[...]

PR #546, opened 2020-08-05
#540 custom adapter initialization options

D10221:
Build custom adapter with the same mechanics as as the 'default' adapter
Provide 'backwards' compatibility with current implementation

Not sure about the testing strategy here, made a demo I'll try to attach it

sorry about the wrong reference , I meant #540  not #544 .... 
I renamed the branch  but git picked it up anyway :)

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -52,8 +52,16 @@ export default async (req, res, userSuppliedOptions) => {
     // Parse database / adapter
     let adapter
     if (userSuppliedOptions.adapter) {
-      // If adapter is provided, use it (advanced usage, overrides database)
-      adapter = userSuppliedOptions.adapter
+      // If adapter is provided, use it
+      adapter =
+        typeof userSuppliedOptions.adapter.getAdapter === "function" // 
+          ? userSuppliedOptions.adapter // ready to use 
+          : typeof userSuppliedOptions.adapter === "function"
+            ? userSuppliedOptions.adapter(userSuppliedOptions.database)
+            : undefined; // needs to be built
+      if(!adapter){
+        logger.warn('NEXTAUTH_ADAPTER', `Invalid Adapter: Expected "function" instead of "${typeof userSuppliedOptions.adapter}"`)
+      }
     } else if (userSuppliedOptions.database) {
       // If database URI or config object is provided, use it (simple usage)
       adapter = adapters.Default(userSuppliedOptions.database)


D10221:
Test-case/demo : https://github.com/D10221/nextauth_custom_adapter_initialization

D10221:
Small update to the demo, I realized I labeled them backwards, 
... no my day :(

iaincollins:
Hi there,

Following discussion in #540 I'm still not convinced of the need for this change or that it's something we want to do.

As it says in the [options documentation](https://next-auth.js.org/configuration/options#adapter), the 'database' option is an alternative to specifying an adapter and should only be used instead (not as well as) an adapter. This is very much by design to steer users in the right direction and avoid confusion.

Not all adapters can even use a connection string - the built in Prisma adapter, for example, only takes an object - but all adapters can take an object and may have 0 or more arguments. Even some people using the default adapter will be using an object and not the database prop, as it's the only way to pass some more advanced options through to TypeORM.

[As per the default adapter documentation](https://next-auth.js.org/schemas/adapters#typeorm-adapter), the default adapter is the TypeORM adapter, the following configuration options are exactly equivalent:

```js
database: {
  type: 'sqlite',
  database: ':memory:',
  synchronize: true
}
```

```
adapter: Adapters.Default({
  type: 'sqlite',
  database: ':memory:',
  synchronize: true
})
```

```js
adapter: Adapters.TypeORM.Adapter({
  type: 'sqlite',
  database: ':memory:',
  synchronize: true
})
```

If you can support all these forms, it's possible to support any adapter. There is nothing special about the default adapter, the database option is just a shortcut. An adapter should take all the arguments it needs that are specific to it as arguments when it is constructed. These arguments can accessed when `getAdapter()` is invoked (which is why it they are designed this way).

Anything that is possible with the default adapter is possible with a custom adapter.

If you want to share a configuration object, you can easily place it in a file and pass it to the Adapter when it is initialised.

D10221:
Thanks!

I just thought o would be nice to instantiate the current 'default' adapter with different parameters at different times.
Depending on how it is build, and when, 
it could open many opportunities, 
And with this small change, it was very clean IMMO.

But I  do understand, and You don't have to agree :) 
Its all good if you don't like it, 
Or think doesn't align with the current vision

Thanks again

iaincollins:
Thanks for being understanding!

I don't hate this and appreciate it's a small change, I'm just a bit reluctant to introduce complexity and another way of doing something (given it how it already feels like there are a lot of moving pieces in NextAuth.js) without exploring possibilities a bit more and really having a grasp on what unique behaviour it enables or if it's something that could be handled by taking different approach with the exiting functionality.

I'll try reading #540 again in a bit and looking at the example to see if I get it.

(I feel this would be a lot easier to discuss and understand in room a whiteboard! :-)

D10221:
Hi, 
I think we can close this.
Found a solution that doesn't require any change from your side.
Not that pretty, ... but it seems to work. :) 

Thank you for your time and attention !

_If you are curious, it can be found [here](https://github.com/D10221/next-auth-cli/blob/master/next-auth-cli/test/next-auth-external-adapter.js), it goes away with customizing the models/initializing/re-initializing/and being swapable, without retyping the original adapter, as I said not that pretty :)  _

PR #549, opened 2020-08-05
Update Google Provider example to be functional

NickParks:
Currently the Google Provider example will always fail due to checking for `email_verified` when the correct response from the server is `verified_email`

next-auth debug output for validation:

```
[next-auth][debug][profile_data] {
  id: 'XXXXXXX',
  email: 'nick@example',
  verified_email: true,
  name: 'Nick Parks',
  given_name: 'Nick',
  family_name: 'XXXX',
  picture: 'XXXX,
  locale: 'en',
  hd: 'example.com'
}
```

### DIFF
--- a/www/docs/providers/google.md
+++ b/www/docs/providers/google.md
@@ -51,7 +51,7 @@ const options = {
 :::
 
 :::tip
-Google also return an `email_verified` boolean property in the OAuth profile.
+Google also return an `verified_email` boolean property in the OAuth profile.
 
 You can use this property to restrict access to people with verified accounts at a particular domain.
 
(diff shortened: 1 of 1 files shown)


iaincollins:
Thanks for the PR.

Oh that is wild, Google return either `verified_email` or `email_verified` depending on if you request a JSON or JWT (which follows the Open ID spec and returns it as `email_verified`) from the `usertoken` endpoint.

It seems `verified_email` is correct for this flow though, as the provider currently defaults to requesting JSON.

That's handy to know about in case we ever switch it open to use Open ID as would be a confusing breaking change for folks relying on the profile object being reasonably consistent.

itwasmattgregg:
Does this mean if I'm using the Google provider and prisma I should @map my `emailVerified` to `verified_email` like: `emailVerified DateTime? @map(name: "verified_email")`? Is it still a DateTime or a Boolean?

Issue #538, opened 2020-08-03
LDAP Authentication


giraffesyo:
I'm trying to implement LDAP authentication in Next.js. Is there an example out there of this? I only saw it mentioned but couldn't find anything concrete in the docs.

Winwardo:
Hey there,

I was trying to do this a couple of days ago as well, I managed to get the following piece of code working (connecting to a local OpenLDAP server using [LDAPJS](http://ldapjs.org/))
However I'm no longer working on that particular project, so can't guarantee this is long-term the best way of doing this. Hopefully this will get you started though:

```typescript
const ldap = require("ldapjs");
import NextAuth from "next-auth";
import Providers from "next-auth/providers";

const options = {
  providers: [
    Providers.Credentials({
      name: "LDAP",
      credentials: {
        username: { label: "DN", type: "text", placeholder: "" },
        password: { label: "Password", type: "password" },
      },
      authorize: async (credentials) => {
        // You might want to pull this call out so we're not making a new LDAP client on every login attemp
        const client = ldap.createClient({
          url: process.env.LDAP_URI,
        });

        // Essentially promisify the LDAPJS client.bind function
        return new Promise((resolve, reject) => {
          client.bind(credentials.username, credentials.password, (error) => {
            if (error) {
              console.error("Failed");
              reject();
            } else {
              console.log("Logged in");
              resolve({
                username: credentials.username,
                password: credentials.password,
              });
            }
          });
        });
      },
    }),
  ],
  callbacks: {
    jwt: async (token, user, account, profile, isNewUser) => {
      const isSignIn = user ? true : false;
      if (isSignIn) {
        token.username = user.username;
        token.password = user.password;
      }
      return token;
    },
    session: async (session, user) => {
      return { ...session, user: { username: user.username } };
    },
  },
  secret: process.env.NEXT_AUTH_SECRET,
  jwt: {
    secret: process.env.NEXT_AUTH_SECRET,
    encryption: true, // Very important to encrypt the JWT, otherwise you're leaking username+password into the browser
  },
};

export default (req, res) => NextAuth(req, res, options);
```

The idea is that once we've authenticated with the LDAP server, we pass through both the username/DN and password to the JWT stored in the browser. This is then passed back to any API routes and retrieved as such:

```
// ./pages/api/doLDAPWork.js

    token = await jwt.getToken({
      req,
      secret: process.env.NEXT_AUTH_SECRET,
    });
    const {username, password} = token;
```

[...]

PR #566, opened 2020-08-10
Add: ldap auth tutorial example

ndom91:
Based upon #538 

I wrote up the LDAP Auth example into a tutorial page.

Thanks @Winwardo

### DIFF
--- a/www/docs/tutorials.md
+++ b/www/docs/tutorials.md
@@ -19,5 +19,9 @@ title: Tutorials and Explainers
 
 * [Testing with Cypress](tutorials/testing-with-cypress)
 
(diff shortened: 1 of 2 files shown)


iaincollins:
@ndom91 Thanks for this! 

I've added a credit card to Vercel - I thought it had expired, but actually it's just because I made a new team and it it seems billing details don't automatically copy across from a personal account when you do that so should work now!

Should we use `NEXTAUTH_SECRET` instead of `NEXT_AUTH_SECRET` for the env var in the example (just for consistency)?

iaincollins (approved):

ndom91:
Yeah good idea!

PR #567, opened 2020-08-10
Add: warnings page

ndom91:
In the logger.js you also have warning URLs being printed, but the page for it had not been created yet so people were getting 404s.

With this PR we have a basic warnings page, modeled after the error one, with all the warnings I could find in the codebase.

### DIFF
--- a/www/docs/warnings.md
+++ b/www/docs/warnings.md
@@ -1,0 +1,50 @@
+---
+id: warnings
+title: Warnings
+---
+
+This is a list of warning output from NextAuth.js.
+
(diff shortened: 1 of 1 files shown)


iaincollins:
Awesome! Totally forgot to get round to this and was nagging at the back of my brain…

iaincollins:
Thank you for adding `JWT_AUTO_GENERATED_SIGNING_KEY` docs!

I bet that is a WTF for a lot of folks.

It's not super sophisticated but felt it was worth doing for 3.1 as provides a better base for the future (i.e. something we can improve on, without needing to introduce breaking changes).

PR #575, opened 2020-08-13
Documentation typos

thoughtcodex:
Am I misreading this?  These changes should make the parameter to attribute mapping less ambiguous...


iaincollins:
Thank you! Yes this looks better!

We probably need to check the other examples too.

PR #584, opened 2020-08-17
feat(provider): Foursquare

joe-bell:
I've been super keen to contribute to this project for a long time, so I guess this is a PR and an issue? I would really appreciate any support to tie things off.

I've been playing around with the example repo and set up the following as a custom provider (along with configuring my Foursquare account to use the redirect_uri)
The login works successfully (I get an email from Foursquare), but on callback I get the following error when trying to access profile data:

```sh
[next-auth][error][callback_oauth_error] {
  statusCode: 400,
  data: '{"meta":{"code":400,"errorType":"invalid_auth","errorDetail":"Missing access credentials. See https:\\/\\/developer.foursquare.com\\/docs\\/api\\/configuration\\/authentication for details.","requestId":"5f3a574f2bbd3313800d2bbb"},"response":{}}'
} 
```

I have 2 theories on what this could be:
- foursquare isn't very friendly on localhost support
- oauth credentials aren't passing over to the profile step (but I'm not quite sure how to debug this)


LoriKarikari:
I've encountered this issue too with some providers. We should definitely find out why the credentials aren't passed.

iaincollins:
Thanks @joe-bell! I guess this provider might require some special handling (a few providers are like that).

Thanks for this! I'll take a look at what it needs once I've made some headway on the backlog. :)

PR #589, opened 2020-08-18
feat(provider): Add Bungie

RobertCraigie:
Bungie's implementation of the OAuth 2.0 protocol requires accessing an extra property (membership_id) that is returned from
the access token request.

I simply added a new provider property (prepareProfileRequest) and call it before making the profile request.
Please let me know if there is a better method for implementing this.

Also I was not sure how to properly handle errors, whether they be developer errors or internal errors and if you could point me in the right direction, that would be great, thanks.


iaincollins:
Thanks @RobertCraigie! This looks good, just working my way through the backlog!

jaaneh:
Any reason this can not be merged? Interested in using this for an upcoming project.

RobertCraigie:
@jaaneh

It should be merged as soon as possible, in the meantime you can use my fork.

ndom91:
Looks good to me, thanks for the extensive documentation as well! :tada:

PR #541, opened 2020-08-04
feat(provider): Add option to generate email verification token

firede:
see #537

PR #594, opened 2020-08-20
Fix redirect reponseAsJson

firede:
**Issue**

```tsx
const res = await fetch("/api/auth/signin/email", {
  method: "post",
  headers: { "Content-Type": "application/json" },
  body: JSON.stringify({ ...values, json: true /* NOTE */ }),
})
console.log(res)
```

**NOTE**
* `json: true` triggers a 302 redirect
* `json: "true"` reponse as JSON

**Expected**

* `json: true` reponse as JSON

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -180,7 +180,7 @@ export default async (req, res, userSuppliedOptions) => {
     // @TODO Refactor into a lib instead of passing as an option
     //       e.g. and call as redirect(req, res, url)
     const redirect = (redirectUrl) => {
-      const reponseAsJson = !!((req.body && req.body.json === 'true'))
+      const reponseAsJson = !!((req.body && String(req.body.json) === 'true'))
       if (reponseAsJson) {
         res.json({ url: redirectUrl })
       } else {


iaincollins:
The current behaviour is correct as the form is `application/x-www-form-urlencoded` rather than anticipating JSON because it supports clients that do not have JavaScript enabled.

We could introduce the option to support JSON as well, but there would need to be quite a bit of regression testing, because there are other parts of the flow that would break, as the expected submission is currently always `application/x-www-form-urlencoded`.

The easiest way to submit a sign in form, is with the sign in client method, which handles both form encoding and CSRF token submission.

```js
import { signIn } from 'next-auth/client'

export default ({ email }) => (
  <button onClick={() => signIn('email', { email })}>Sign in with Email</button>
)
```

firede:
@iaincollins Thanks for the explanation!

Related case: https://github.com/nextauthjs/next-auth/pull/541

I am try to implement it on the same page:

1. Send verification email
2. Paste verification code
3. Verify and login

The `signIn` API will redirect to the verification request page by default, which cannot satisfy the above interaction.
Do you have any suggestions?

PR #599, opened 2020-08-23
[providers]: Add FusionAuth provider

mprasanjith:
I'm working on a project that uses a FusionAuth instance as the OAuth2 server. Wrote a custom provider and thought it would be great to have built-in support.


LoriKarikari:
@mprasanjith thanks for the PR!

PR #601, opened 2020-08-24
Change check on being secure

ramiel:
When calling `jwt.getToken` let's consider the cookie to be secure if `NEXTAUTH_URL` is not set.
Documentation asks the user to set that variable in development and, in production, that variable cannot be set because the url is
dynamically created. That said, the check in getToken function should be changed as in this PR.

I'm opening this PR more as a discussion point. Generally speaking I feel this project should try to work without `NEXTAUTH_URL` at all and use instead `VERCEL_URL` or `req.headers.host` if the domain name is really needed

### DIFF
--- a/src/lib/jwt.js
+++ b/src/lib/jwt.js
@@ -95,8 +95,8 @@ const getToken = async (args) => {
   const {
     req,
     // Use secure prefix for cookie name, unless URL is NEXTAUTH_URL is http://
-    // or not set (e.g. development or test instance) case use unprefixed name
-    secureCookie = !(!process.env.NEXTAUTH_URL || process.env.NEXTAUTH_URL.startsWith('http://')),
+    // If NEXTAUTH_URL is not set, consider it to be production and so secure
+    secureCookie = !(process.env.NEXTAUTH_URL && !process.env.NEXTAUTH_URL.startsWith('http://')),
     cookieName = (secureCookie) ? '__Secure-next-auth.session-token' : 'next-auth.session-token',
     raw = false
   } = args


iaincollins:
Thanks for the PR!

So we don't want to assume the cookie is secure / HTTPS here as it should work without `NEXTAUTH_URL` being configured in local environments, and it should assume that is HTTP (rather than HTTPS).

We could add a check for `VERCEL_URL` being checked to this function, and assume the cookie is secure if it is, that would address the problem in this context for users hosting on Vercel.

We currently use `VERCEL_URL` where possible elsewhere (in both the API routes and in the Client) but with the caveat that is not a production ready URL as it is an instance name and probably not something most people would want to expose publicly (it's really intended for test instances), but if it is set, we know the URL is HTTPS (and that the cookie is secure)..

[Alternatively, we could also check the `Forwarded` and/or `X-Forwarded-Proto` headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) to determine this, so support it across providers (at least one of which which should also solve it for Vercel).

As of version 2 onwards, we don't use `req.headers.host` as while it's a great developer experience there are security issues with it - specifically relying on a value supplied in a header opens up a vector of using an API route to launch a DDoS attack on another host, we can't trust it as it's supplied by the client.

jan-wilhelm:
In the meantime, what do you think about setting an env variable in Vercel that exposes the value of `VERCEL_URL` as `NEXTAUTH_URL` to javascript? Are there any security drawbacks to this, as long as you keep the Vercel domains private?

PR #590, opened 2020-08-18
[providers]: Updating Discord provider domain

PermissionError:
Discord is migrating to discord.com, including their OAuth2 API routes. Support for the old domain, discordapp.com, will be dropped on 7 Nov 2020.

Note that the cdn.discordapp.com domain is unchanged. This is intentional, as the cdn domain will not be migrated due to technical restraints on Discord's side.

PR #605, opened 2020-08-26
Set Discord to Prompt = None

anishanne:
When you set the value of `prompt` to `none` in the get params, Discord will not prompt the user if they have auth'd with your app before. (It will just display the page and then the page disappears and they are in.)

Docs: https://discord.com/developers/docs/topics/oauth2#authorization-code-grant-authorization-url-example

### DIFF
--- a/src/providers/discord.js
+++ b/src/providers/discord.js
@@ -8,7 +8,7 @@ export default (options) => {
     params: { grant_type: 'authorization_code' },
     accessTokenUrl: 'https://discordapp.com/api/oauth2/token',
     authorizationUrl:
-      'https://discordapp.com/api/oauth2/authorize?response_type=code&prompt=consent',
+      'https://discordapp.com/api/oauth2/authorize?response_type=code&prompt=none',
     profileUrl: 'https://discordapp.com/api/users/@me',
     profile: (profile) => {
       return {
--- a/src/providers/discord.js
+++ b/src/providers/discord.js
@@ -8,8 +8,8 @@ export default (options) => {
     params: { grant_type: 'authorization_code' },
     accessTokenUrl: 'https://discord.com/api/oauth2/token',
     authorizationUrl:
-      'https://discordapp.com/api/oauth2/authorize?response_type=code&prompt=none',
-    profileUrl: 'https://discordapp.com/api/users/@me',
+      'https://discord.com/api/oauth2/authorize?response_type=code&prompt=none',
+    profileUrl: 'https://discord.com/api/users/@me',
     profile: (profile) => {
       return {
         id: profile.id,


LoriKarikari:
I think it's always better to provide a prompt so the user knows which data the application has access to. @iaincollins what are your thoughts about this one?

nyedidikeke:
@LoriKarikari I agree, but it will be great as a configurable option to the application owner.

LoriKarikari:
@nyedidikeke yes we should add an option to change this.

LoriKarikari:
@iaincollins I think we should leave the option as it is now but make it possible to change this value?

anishanne:
The way that the prompt works is the first time the user auths with the app, they have to give consent. Then every time after, they don't (provided the scopes are the same). It's basically cookie/session based login just if you have a Discord session it still works.

LoriKarikari:
In that case, it's fine to merge this PR! Thanks for the clarification @anishanne!

nyedidikeke (changes requested):
@anishanne @LoriKarikari I don't think it's a good idea to change the current behaviour unless programmatically optional and easily configurable but should this be considered, let's not override #590 as Discordapp.com is now [Discord.com](https://support.discord.com/hc/en-us/articles/360042987951-Discordapp-com-is-now-Discord-com).

anishanne:
@nyedidikeke I think it would be better to not request users for consent each time they decide to login, if the scopes are the same & its the same user, just log them right in.

nyedidikeke (approved):
The current no longer override #590 👍🏿

LoriKarikari:
I'm going to leave this one open and see what @iaincollins thinks about this.

iaincollins:
@anishanne @LoriKarikari @nyedidikeke Thanks everyone!

I think `prompt=none` makes sense as that is how the other providers work, users shouldn't see a prompt on subsequent sign in attempts (unless there is a technical reason for that specific to Discord?). 

Most providers with this way, though Google actually has the same option. By default they do not prompt after the initial sign in (unless you have multiple accounts, then sometimes you are asked to pick one). However with Google, you can request a prompt on sign in. Usually there are only some specific scenarios where a prompt is required and specific reasons for doing it.

e.g. In the case of Google it can be used to to issue a Refresh Token again - as, unlike other providers, they usually only provide one on initial sign in - their intended use case is mobile or desktop applications (that are persist sessions forever, but need to get it again if the user uninstalls and reinstalls the app).

We can always add an example to the documentation for this for folks who might want to prompt if there is use case for it. If folks want to configure this they can always override the `authorizationUrl` option on the provider, and we could show how to do that in the docs.

I agree it seems like `discord.com` is the right domain to use.

iaincollins:
Thanks folks! Will merge in now and will go out in the next release!

PR #612, opened 2020-08-27
Spotify provider: fix error occurred when profile image is not set

youpy:
> [next-auth][error][oauth_parse_profile_error] [ TypeError: Cannot read property 'url' of undefined


LoriKarikari:
thanks for the PR @youpy!

PR #630, opened 2020-09-02
fix(adapter): Fix ISO Datetime type error in updateSession

HaNdTriX:
## Issue

When running `updateSession` the adapter didn't convert the `expires` date into an ISO 8601-formatted string.

```log
...
Mutation (object)
  ↳ updateOneSession (field)
    ↳ data (argument)
      ↳ SessionUpdateInput (object)
        ↳ expires (field)
          ↳ Value types mismatch. Have: String("2020-10-02T08:24:27.840Z"), want: Object((Weak))` at `.Mutation.updateOneSession.data.SessionUpdateInput.expires`
...
```

[more info](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema/prisma-schema-reference#datetime)

## Solution 

Convert the date object via `expires.toISOString()` into an ISO 8601-formatted string.


ndom91:
Looks good to me! 

I'm not super experienced with graphql though, so before I merge this I'd like to get some feedback on how (if at all) this would affect users with preexisting databases, etc.?

PR #634, opened 2020-09-02
Run integration tests against build running in Docker container

iaincollins:
You can now run `npm test` to run end to end tests.

* Cypress runs against a Docker container, which exposes a self-contained test app of the build in the working directory, on localhost:3000 (we might change this port).
* You need Docker installed to run `npm test` but all the Docker images and Cypress will be downloaded automatically.
* Initial run can take 10 minutes or more as it needs to download Docker images and Cypress, but subsequent test re-runs should only take a few seconds (~30 seconds; but can take 1-2 minutes for a fresh build).
* GitHub Workflows has been refactored to split out build and linting from end-to-end integration tests.
* The tests are very basic (checks the module loads, pages are generated, buttons and text is displayed) but will expand on them in future. Environment variables are passed through to the Docker image so will work if configured in GitHub as secrets *or* if put in a `.env` file in the root directory (I have tested both work) so should be a good developer experience when developing locally.

I'll update the contributing docs in a bit.

### About the Dockerization

Adds commands to start/rebuild/stop a Docker image of a sample Next.js app that loads the latest build of NextAuth.js from the current directory.

* `npm run test:app:start` - Start local instance of a sample app running on port 3000
* `npm run test:app:rebuild` - Rebuild and restart sample app 
* `npm run test:app:stop` - Stop sample app on port 3000

Notes:
*  It loads the current build in the work directory into the Docker container, so you need to have run `npm i` and `npm run build` in the `next-auth` root directory at least once.
* If you are running `npm run watch` in the background, you can run `npm run test:app:rebuild` to rebuild the sample app at any time (but you must trigger the rebuild of the app in Docker manually each time).

* The Dockerfile uses a multi-stage build process to optimise build performance, but the nature of the process is intensive and may be slow, build times vary depending on computer speed and internet connection.
* Initial local build times may be slow due to images to support testing being downloaded (it may take 10 minutes or more) but subsequent builds on the same computer should be faster (1 minute or less). Building in CD/CI (GitHub Actions on a Pull Request) should only take around 5 minutes.
* To ensure the `package.json` and `package-lock.json` files are valid, dependencies listed are re-installed fresh on every build of the Docker image.
* A Docker compose file is used to allow us to extend the test app to run it against multiple databases, which are self-contained inside the container (so they will not conflict with any databases running on the host OS). Docker is configured so that environment variables can be passed through to re-configure the app without needing to rebuild it (so it's very fast to test different configuration options, databases, etc).
* Subsequent updates may look to improve performance and developer experience when working on the library locally (e.g. another Docker image that supports fast reloading without re-installing modules) but it's important checks like checking package.json is valid and running the build in isolation are performed in CD/CI so that we can have the best possible confidence the module will work as expected when published to NPM.


iaincollins:
cc @JeffersonBledsoe just to let you know I'm working on this :-)

iaincollins:
@JeffersonBledsoe @ndom91 @LoriKarikari  I've integrated the Cypress tests and it's working, running against the local build!

You can now just  run `npm test` to run end to end tests! Have updated the comment for the PR.

I had to skip some of the tests because will need work to refactor as looks like the changes for enhanced Cross Site Request Forgery in 3.0 means they need some refactoring, but that shouldn't be too hard. I'll start one provider at a time for a full end to end test, I probably just pick one for each flow (OAuth 1, OAuth 2, Email, dummy Credentials provider, etc).

As you can see in Package.json `npm test` it actually runs `npm run lint && npm run build && npm run test:app:rebuild && npm run test:e2e:run && npm run test:app:stop`.

On first run it can take a while (10 min or more download Docker images, dependancies and Cypress) but subsequent runs only take 30 seconds, even on this crusty i5 iMac from 2013.

This is really wonderful, I'm so happy! This will make making changes so much easier and safer and allow things to go much faster. Cypress is really easy to understand and work with once it's in place, but I wasn't familiar with it before and would have taken me much longer to get my head around it to get it to this point, so I'm very grateful for the work done which has made this possible.

@JeffersonBledsoe I hope merging the PR from draft wasn't too annoying and is worth the pain, now we have a way of running end to end tests that I hope is fairly easy and reasonably reliable. :-)

PR #669, opened 2020-09-12
Add missing comma in type object

JLHasson:
This PR adds a missing comma to the database type object to make it work when copy-pasted


LoriKarikari:
Thanks for the PR @JLHasson!

PR #678, opened 2020-09-15
Display error if no [...nextauth].js found

### LINKED ISSUE #647: Display warning or error if filename not set to [...nextauth].js

Display warning or error if the filename is not set to `[...nextauth].js`

See #636 for context.

Checking for the existence of `nextauth` parameter on the request would prevent people from running into errors when they have named the file something else. 

e.g. `[...nextAuth].js`, `[...id].js]` etc.

This is a problem a few folks have run into as the error that is currently displayed when this happens isn't very helpful.

We should check for `req.query.nextauth` in `src/server/index.js` and exit and return an HTTP 500 error (and log a message to the console) if it does not exist.

**Potential problems**

There are no problems doing this I can foresee.

None.

**Additional context**

None.

I'm very happy if anyone wants to submit a PR for this.

jaknas:
Let me know if you have any suggestions for changes.

Fixes #647


tpatel:
Just had this issue, this PR would have helped me save time!

I think it would be even better if you used the logger to link this issue to the documentation. This logger: https://github.com/nextauthjs/next-auth/blob/73d21e66ddb73da968d93cec32d4b46b152e217a/src/lib/logger.js#L2 with the error on https://github.com/nextauthjs/next-auth/blob/main/www/docs/errors.md.

kleyu:
> Just had this issue, this PR would have helped me save time!
> 
> I think it would be even better if you used the logger to link this issue to the documentation. This logger:
> 
> https://github.com/nextauthjs/next-auth/blob/73d21e66ddb73da968d93cec32d4b46b152e217a/src/lib/logger.js#L2
> 
> with the error on https://github.com/nextauthjs/next-auth/blob/main/www/docs/errors.md.

Thanks for taking a look! I've added new error code and logged it when the server can't find `[...nextauth]` in new commit.

Feel free to leave more suggestions.

**Side note**: while testing it, the client-side error handling for this particular case was not ideal. I had already some basic `login` and `session` setup and after removing the `[...nextauth]` from my project dir, browser error messages would look like this:

image

And when clicking client-side login:

image

(`providers` is `null` here): 
 https://github.com/nextauthjs/next-auth/blob/e6995d21cd1c7c0e701a305c91062ece83d48dfe/src/client/index.js#L231


The generic error message comes from `_fetchData` as it assumes that `response` type is `json` while it's actually `text` -- because we're returning using `res.end(error)`.


Not sure where to go from here. I presume that percentage of people who'll encounter this type of not-ideal-error is really small, it could happen when you move some files around or mistakenly remove `[...nextauth].js` while you have already setup some client side auth and miss the very prominent server-side error log.

tpatel:
Looks great to me!

I've had other errors and I believe they display the same error message. So now when I `provider is null` in the frontend I usually go check what's up with the backend.
So this should be good.

iaincollins:
@kleyu Awesome! Thank you! I'll get round to rolling this out in the next update as soon I can.

iaincollins:
> Not sure where to go from here. I presume that percentage of people who'll encounter this type of not-ideal-error is really small, it could happen when you move some files around or mistakenly remove [...nextauth].js while you have already setup some client side auth and miss the very prominent server-side error log.

Ugh yeah the client should really handle errors better!

Thanks for raising that issue in the comment - it came up in another discussion recently and could do with improving.

Issue #682, opened 2020-09-17
faunadb adapter 


s-kris:
I wanted to use faunadb with next-auth and saw that it's not yet supported. I saw some discussions around it.
So sharing this fauna-adapter here in case if someone wants to quickly integrate faunadb with next-auth.

https://gist.github.com/s-kris/fbb9e5d7ba5e9bb3a5f7bd11f3c42b96

LoriKarikari:
Looks good @s-kris! You can open a PR to add this adapter if you want.

s-kris:
Thank you @LoriKarikari 
Raised PR: #686

s-kris:
#708 was merged tentatively

PR #686, opened 2020-09-18
Add fauna adapter

s-kris:
@LoriKarikari 

Adding faunadb adapter #682


src/adapters/fauna/index.js
```diff
@@ -0,0 +1,420 @@
...
+        dateSessionIsDueToBeUpdated.setTime(
+          dateSessionIsDueToBeUpdated.getTime() + sessionUpdateAge
+        )
+
+        // Trigger update of session expiry date and write to database, only
+        // if the session was last updated more than {sessionUpdateAge} ago
+        if (new Date() > dateSessionIsDueToBeUpdated) {
+          const newExpiryDate = new Date()
+          newExpiryDate.setTime(newExpiryDate.getTime() + sessionMaxAge)
+          session.expires = newExpiryDate
```
raimille1, line 250: Shouldn't this be?
``session.expires = newExpiryDate.valueOf()``
If not fauna just stores an empty object (at least that's what I've been getting)

src/adapters/fauna/index.js
```diff
@@ -0,0 +1,420 @@
...
+      console.log('no default options for session')
+    }
+
+    const defaultSessionMaxAge = 30 * 24 * 60 * 60 * 1000
+    const sessionMaxAge =
+            appOptions && appOptions.session && appOptions.session.maxAge
+              ? appOptions.session.maxAge * 1000
+              : defaultSessionMaxAge
+    const sessionUpdateAge =
+            appOptions && appOptions.session && appOptions.session.updateAge
```
raimille1, line 39: Noticed this while testing with 0, was using the default.

``appOptions && appOptions.session && (appOptions.session.updateAge || appOptions.session.updateAge === 0)``

src/adapters/fauna/index.js
```diff
@@ -0,0 +1,420 @@
...
+              ? appOptions.session.updateAge * 1000
+              : 24 * 60 * 60 * 1000
+
+    async function createUser (profile) {
+      console.log('-----------createUser------------')
+      // console.log(profile);
+      return faunaWrapper(
```
raimille1, line 46: I'm not sure if this is a pattern with fauna (new user as well) but to use the ref's across the other methods I basically return:
```
return await faunaClient.query(q.Create(q.Collection('users'), { data: profile }))
          .then(res => { return { ...res.data, id: res.ref } })
```

This way further down on other methods like getUser, I can directly use the id as follows:
```
    async function getUser (id) {
      try {
        return await faunaClient
          .query(q.Get(q.Ref(id)))
          .then(res => { return { ...res.data, id: res.ref } })
          .catch(err => Promise.resolve(null))
      } catch (error) {
        return Promise.reject(new Error('GET_USER_ERROR'), error)
      }
    }
```

s-kris: It is a fauna pattern. It's optimal to use FQL way to execute multiple operations instead of using js patterns.

src/adapters/fauna/index.js
s-kris: The logic is kept in sync with other adapters: https://github.com/nextauthjs/next-auth/tree/main/src/adapters
Can you raise an issue for this?

src/adapters/fauna/index.js
s-kris: The logic is kept in sync with other adapters: https://github.com/nextauthjs/next-auth/tree/main/src/adapters
Let me check why fauna isn't saving this

src/adapters/fauna/index.js
raimille1: maybe it wasn't clear because of the line # of the comment, but I wasn't referring to the faunaWrapper. I was referring to the `q.Select('data'` part ... by returning the ref object as well, I was able to use the fauna IDs directly (Ref) rather than creating yet another set of `id:` fields for each object. 

Which I'm not sure if fauna encourages that or encourages devs to handle their own IDs.

src/adapters/fauna/index.js
raimille1: checked master, and both prisma and typeorm adapters have the right logic (fallback to 0). 

Am I missing something here?
```    
const sessionUpdateAge = (appOptions && appOptions.session && appOptions.session.updateAge)
      ? appOptions.session.updateAge * 1000
      : 0
```

src/adapters/fauna/index.js
s-kris: You are correct. We can directly use fauna ID but I just want to have id field with uuid so that I can simply extract the 'data' part of the document in case of migration to other db.

ndom91:
Hi guys, tentative Fauna support was merged from PR #708. We still need some docs re Fauna, and any other PRs improving the support or tests are always welcome as well!

s-kris:
@ndom91 thanks for the update, closing this.

PR #706, opened 2020-09-23
docs: More better slack documantation link

terrierscript:
I think it seems slack has more better documentations


iaincollins:
Thanks!

PR #714, opened 2020-09-26
Fix: rename refresh token to access token secret if provider oauth version is 1

euxn23:
## Issue

OAuth 1.0 prorocol and node-oauth's `getOAuthAccessToken` method returns `access_token` and `access_token_secret`,
but in next-auth app, we get `access_token` as `reflesh_token`, named suit to OAuth 2.0.

e.g. Twitter, 
```
[next-auth][debug][oauth_callback_response] {
  profile: {
    name: 'euxn',
    email: null,
    image: 'https://pbs.twimg.com/profile_images/1262491378291884032/TH28GyIm.jpg'
  },
  account: {
    provider: 'twitter',
    type: 'oauth',
    id: '123857414',
    refreshToken: 'XXXXXXXXXXXX',
    accessToken: '123857414-XXXXXXXXXXXXXXXXX',
    accessTokenExpires: null
  },
  OAuthProfile: {
    ...
  }
```

## Solution

rename returning object parameter name if provider used OAuth 1.0.
This is breaking change, but `reflesh_token` is incorrect as OAuth 1.0 response.

I'll appliciate if you consider to merge this breaking change for to be.

### DIFF
--- a/src/server/lib/oauth/callback.js
+++ b/src/server/lib/oauth/callback.js
@@ -107,7 +107,7 @@ export default async (req, provider, csrfToken, callback) => {
       oauth_token,
       null,
       oauth_verifier,
-      (error, accessToken, refreshToken, results) => {
+      (error, accessToken, accessTokenSecret, results) => {
         // @TODO Handle error
         if (error || results.error) {
           logger.error('OAUTH_V1_GET_ACCESS_TOKEN_ERROR', error, results)
@@ -116,9 +116,9 @@ export default async (req, provider, csrfToken, callback) => {
         client.get(
           provider.profileUrl,
           accessToken,
-          refreshToken,
+          accessTokenSecret,
           async (error, profileData) => {
-            const { profile, account, OAuthProfile } = await _getProfile(error, profileData, accessToken, refreshToken, provider)
+            const { profile, account, OAuthProfile } = await _getProfile(error, profileData, accessToken, accessTokenSecret, provider)
             callback(error, profile, account, OAuthProfile)
           }
         )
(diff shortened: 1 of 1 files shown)


iaincollins:
Thanks for raising this! I completely overlooked that these are named differently for OAuth 1.

I think it's valid to raise this although I'm inclined to think we don't change it, as it's a breaking change and having things normalized makes them easier to work with in some ways, but that we should add information to the documentation for Twitter to help folks avoid confusion. Happy to leave this open for discussion if folks feel strongly about it one way or the other.

There is some precedence here, as actually *deliberately* don't use the same names Twitter uses for ID and Secret as historically Twitter has not even been consistent with this in it's own documentation, and it seemed sensible just to use the same `clientId` and `clientSecret` values as other OAuth providers, as Twitter was the only outlier (even if it's where OAuth was created!).

Twitter is are currently the only OAuth 1 provider left, and they have a new API in development which uses OAuth 2 so I'm hopeful it being different to other OAuth providers will resolve itself over time.

euxn23:
Thanks for review this! I now understood `clientSecret` is not suitable for all OAuth providers.
I know Twitter is developing new API using OAuth2, but I think it is kindness to write note about this into document, so I'll create new pull-request about it.

euxn23:
Thanks for review this! I now understood `clientSecret` is not suitable for Twitter document.
But [node-oauth](https://github.com/ciaranj/node-oauth) uses `client_secret`, so I think using `clientSecret` is better.
(But as you say, this is historically problem, so I don't think this should be fixed just now. I'll appreciate if you consider this request and fixing in the future even a little. So I keep this pull-request open, but if you decided not to change, I'd be respect your dicision.)

I know Twitter is developing new API using OAuth2, but OAuth1 will be used for a while, so I also think it is kindness to write note about this into document. I'll create new pull-request about it.

euxn23:
I have additional opinion, custom providers using OAuth1 also requires to be careful to use OAuth response, and this should be also documented.

PR #730, opened 2020-09-30
send signInCallbackResponse to jwt callback

### LINKED ISSUE #728: Return new token from signIn callback
**UPDATE:**
I eagerly created both a [PR](https://github.com/nextauthjs/next-auth/pull/730) and updated the [type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48375) to mitigate this. :crossed_fingers: 

**Describe the bug**
I am trying to implement #583 in user land to create a Proof of Concept.

According to [the docs](https://next-auth.js.org/configuration/callbacks#sign-in-callback), it should be possible to return a modified JWT token in the `signIn` callback, which I was assuming I could get in the `jwt` callback.

**Steps to reproduce**
```js
async signIn(...args) {
  const account = args[1]
  const MAX_CONCURRENT_LOGIN = 1
  const numberOfSessions = await sessionsDB.getNumberOfSessions(account.id)
  if (numberOfSessions - 1 < MAX_CONCURRENT_LOGIN) {
    const sessionId = await sessionsDB.addSession(account.id)
    return {
      ...args[0],
      sessionId,
    }
  }
  return false
},
```

**Expected behavior**
I would like to pass a sessionId to the `jwt` callback.
**Screenshots or error logs**
If applicable add screenshots or error logs to help explain the problem.

**Additional context**
On line 71, the signIn callback response is saved, but the value is never used at line 92.
https://github.com/nextauthjs/next-auth/blob/8115a7c66cdb1d86b0d2a0d76b9aa33c2bfaa33b/src/server/routes/callback.js#L70-L92

Wouldn't it be enough to resue `signInCallbackResponse` at line 87 to create a `defaultJwtPayload`

**Feedback**

* [x] Found documentation but was incomplete

balazsorban44:
According to [the docs](https://next-auth.js.org/configuration/callbacks#sign-in-callback), it should be possible to return a modified JWT token in the `signIn` callback, which (if it is an object) should probably be sent further to the `jwt` callback.

Fixes #728

### DIFF
--- a/src/server/routes/callback.js
+++ b/src/server/routes/callback.js
@@ -67,8 +67,9 @@ export default async (req, res, options, done) => {
             }
           }
 
+          let signInCallbackResponse
           try {
-            const signInCallbackResponse = await callbacks.signIn(userOrProfile, account, OAuthProfile)
+            signInCallbackResponse = await callbacks.signIn(userOrProfile, account, OAuthProfile)
             if (signInCallbackResponse === false) {
               return redirect(`${baseUrl}${basePath}/error?error=AccessDenied`)
             }
@@ -87,7 +88,8 @@ export default async (req, res, options, done) => {
             const defaultJwtPayload = {
               name: user.name,
               email: user.email,
-              picture: user.image
+              picture: user.image,
+              ...(typeof signInCallbackResponse === "object" ? signInCallbackResponse : {})
             }
             const jwtPayload = await callbacks.jwt(defaultJwtPayload, user, account, OAuthProfile, isNewUser)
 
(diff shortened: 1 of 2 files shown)


iaincollins:
This is a typo in that comment. 

The return type was `boolean`, but now can be an Error object or a `string` (in which case a URL to redirect to is assumed) but cannot be a JWT. The intended use for this function is only to determine if a user should be allowed to sign in or not (and optionally to control what specific error message they see if they cannot; e.g. account suspended).

The only intended way to customise the the JWT response is to use the `jwt` callback, as in the example.

balazsorban44:
Thank you! I will close this PR then.

PR #734, opened 2020-10-01
fix: ensure Images are produced for discord

daggy1234:
0 Current Discord does not provide images if the avatar is the default one. This ensures a default avatar is generated.
Also ensures based on profile premium gifs are generated.


nyedidikeke (changes requested):
Great work @Daggy1234 
Please consider following;
- Removal of redundant console.log() instances.
- `const = format` may be a single line.
- Replacing discordapp.com with discord.com based on #590.

Is `defaultAvatarNum` meant to represent default _Avatar Number_, _Avatar Name_ or something else?

Daggy1234:
okay so I have a few comments to that 

- `console.log` will be removed 

- Single line is good

However the cdn is still `discordapp.com` and not `discord.com` therefore it does not make sense to change that

Also defaultAvatarNum is the Number! Not avatar Name.

nyedidikeke (changes requested):
Thanks @Daggy1234 

> Also defaultAvatarNum is the Number! Not avatar Name.

`defaultAvatarNumber` is more readable and preferred.

Daggy1234:
Will Change 👍🏽

Daggy1234:
lmk if there is any update!

nyedidikeke (approved):

PR #746, opened 2020-10-06
Update branding (icons, homepage)

iaincollins:
* Refreshed branding (icons, colours)
* Improved existing theme, especially in dark mode
* Minor tweaks to copy on homepage


iaincollins:
No changes to code, just the website - the integration test is marked as failing as need to improve logic to handle the test runner on different operating systems (e.g. Mac vs Linux) as Chromium binary is different on each (and across architecture, e.g. x86 vs ARM; which is already supported). I can resolve that for now by only running them on Linux.

Could run tests themselves inside Docker; though that might increase test execution time, so maybe better to just to improve the logic in the test runner.

PR #750, opened 2020-10-06
Correct multiple typos in the docs

kripod:
"relev**e**nt" → "relev**a**nt" and other fixes.


iaincollins:
Thanks for these fixes!

I wonder if it's time to implement something like [markdown-spellcheck](https://www.npmjs.com/package/markdown-spellcheck)…

PR #763, opened 2020-10-09
fix: update Okta routes

ohheyalanray:
the current routing for the Okta provider does not follow the standard set by Okta, and as such doesn't allow for custom subdomains. this update amends the routes to allow for customer subdomains, and also aligns next-auth with Okta's documentation (or, at least, the Okta-React docs, [here](https://github.com/okta/okta-oidc-js/tree/master/packages/okta-react))


balazsorban44:
I checked https://github.com/okta/okta-auth-js#required-options, and looks like you are right. Thanks!

PR #767, opened 2020-10-12
Expose "reloadSession" function in useSession()

brunocrosier:
When building with `next-auth` I found myself wanting more from the `useSession()` hook, which currently returns `data` and `loading`.

I wanted the `useSession()` hook to expose a function that would allow me to declaratively reload the session when I **know** the session's data has changed - without needing to refresh the whole page.

This is the end result: https://streamable.com/w4tg6q

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -27,7 +27,7 @@ const __NEXTAUTH = {
   keepAlive: 0, // 0 == disabled (don't send); 60 == send every 60 seconds
   clientMaxAge: 0, // 0 == disabled (only use cache); 60 == sync if last checked > 60 seconds ago
   // Properties starting with _ are used for tracking internal app state
-  _clientLastSync: 0, // used for timestamp since last sycned (in seconds)
+  _clientLastSync: 0, // used for timestamp since last synced (in seconds)
   _clientSyncTimer: null, // stores timer for poll interval
   _eventListenersAdded: false, // tracks if event listeners have been added,
   _clientSession: undefined, // stores last session response from hook,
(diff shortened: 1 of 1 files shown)


iaincollins:
Hi there,

Appreciate the typo fixes!

If you want to update the session client side, you should be able to do this already, by calling`getSession()`, which should update both the current window and all other open windows. That is probably a better pragma than exposing an additional method.

If that doesn't work for any reason I'd be happy to investigate that.

brunocrosier:
Thanks @iaincollins 

I had tried `getSession()` on the client side and that didn't work, hence why I thought this was necessary.

I'll send over a demo repo soon but basically it's doing this:
1) get session on frontend
`const [session, loading] = useSession()`
2) populating `role` on the `session.user` object via next-auth session callback function
3) button which has an onClick function that fetches `/api/switchRole` . On this route, I'm using `getSession()` to find the current session and flip the value of `session.user.role` between `ADMIN` and `USER`

All of the above works perfectly, but the value of `session.user.role` doesn't update on the frontend until I either refresh the page or switch tab. Exposing a `reloadSession()` function and calling it after the API call has resolved, updates the frontend as you see in the video

Any thoughts based on the above? Possibly the reason `getSession()` isn't working on the client side is because it's being called on server side first, in conjunction with a session callback?

brunocrosier:
@iaincollins I've invited you to a very basic private repo, which demonstrates the issue. Please check the "reloadSession" branch. It's private because I've added the `.env` file to it.

You can run `npx prisma studio` to open up Prisma Studio, and the frontend code / issue with `getSession()` is on `/pages/protected` 

Any questions, let me know!

brunocrosier:
Just in case you haven't received the GitHub invitation email to the repo (just happened with a different collaborator), [here is the link](https://github.com/brunocrosier/rmenu)

Issue #695, opened 2020-09-19
Auth0 callback url


waptik:
**Your question**
What should be the callbackUrl to set?

**What are you trying to do**
I'm trying to use nextauth with auth0 as a provider. I followed the [guide](https://next-auth.js.org/providers/auth0) as suggested and set callback url as `http://localhost:3000/api/auth/callback` in my application settings at auth0 because i'm testing it locally.
But when i click on the login button and select auth0, I'm being redirected to their error page saying `Callback URL mismatch`.
The docs doesn't make mention of how to setcallbacks.
A [snapshot](https://web.archive.org/web/20200910061218/https://next-auth.js.org/providers/auth0) of it.
Any help with that?

Edit: I finally fixed my issues after changing callbackUrl  several times. So the correct callbackUrl is `http://localhost:3000/api/auth/callback/auth0`. I just needed to append `/auth0` at the end.

* [x] Found the documentation helpful
* [x] Found documentation but was incomplete
* [ ] Could not find relevant documentation
* [ ] Found the example project helpful
* [ ] Did not find the example project helpful

iaincollins:
> Edit: I finally fixed my issues after changing callbackUrl several times. So the correct callbackUrl is http://localhost:3000/api/auth/callback/auth0. I just needed to append /auth0 at the end.

Thanks for coming back and updating with the resolution, always apprecaited!

spsaucier-bakkt:
@waptik How were you able to set localhost as your callback URL? Auth0 appears to block that domain in the "Application Login URI" field.

waptik:
> @waptik How were you able to set localhost as your callback URL? Auth0 appears to block that domain in the "Application Login URI" field.
> 
> **EDIT** : I was using the wrong field -- it should instead be `Allowed Callback URLs`

So it's working now?

spsaucier-bakkt:
Yes, it is, thanks.

waptik:
No problem. I'm glad it works

ChrisCates:
@iaincollins, could you add this to the documentation on the npm module `README.md`.

Would be handy to newcomers.
Thank you.

scarstens:
Just echoing that this would be useful on the Auth0 provider documentation as its not noted anywhere right now.

PR #768, opened 2020-10-12
Auth0: added callbackUrl documentation

albertpak:
Based on callbackUrl [comment](https://github.com/nextauthjs/next-auth/issues/695) and my searching around, thought it'd be helpful to add it to the actual documentation

### DIFF
--- a/www/docs/providers/auth0.md
+++ b/www/docs/providers/auth0.md
@@ -15,6 +15,10 @@ https://manage.auth0.com/dashboard
 Configure your application in Auth0 as a 'Regular Web Application' (not a 'Single Page App').
 :::
 
+:::tip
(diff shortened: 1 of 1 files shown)


iaincollins:
Hi there!

Thanks for the pull request!

This is a good nudge, I think we actually want to go through all the providers and add that information in to every provider so it's better surfaced as we get questions about it from time to time.

The list of services available is always exposed at [/api/auth/providers](https://next-auth-example.now.sh/api/auth/providers) but I think people don't always see that as the documentation page for providers is a little long.

I think probably we shouldn't merge this in, but we should update each provider page with this info in the same place, so it's clear for everyone.

balazsorban44:
As per @iaincollins's suggestion, this individual PR should not be merged, but your change should be added to each providers' page documentation

Issue #361, opened 2020-06-29
How can I register users with email and password?


dhuber666:
Hiho,

I read the whole documentation and also the part with the "Credentials" Provider. It seems that it's the thing I need but it doesn't get persisted in the database according to the docs. 
There is also an email provider, but I don't want it that way (email gets sent to the user and then he can register). I want a pretty normal and boring email + password signup, is this possible?

Is there really no way to add a normal email / password provider or am I missing something? I think that's still the most common way to signup to web apps todate. 

BTW: I love this library everything works just out of the box. Now I only need email / password besides facebook and google signup

Appreciate your help 👍

iaincollins:
You can do this [with the credentials provider](https://next-auth.js.org/providers/credentials).

It is not a path we go out of our way to support as it's less secure than passwordless mechanisms (and often less convenient for users) but if you have an existing database of usernames and passwords it's possible.

PR #771, opened 2020-10-14
Add password into 'user' of the database and change the default adapter

li3939108:
This change can enable signup with email and password. 
The signin with email and password can also be easier 
No need to use another database to store the password.
The updateUser() can also update password if provided current password matched

The password is hashed with bcrypt using a randomly generated salt.

This issue is related https://github.com/nextauthjs/next-auth/issues/361


darshkpatel:
On a side note do you have an implementation example for this?

li3939108:
I just expose the Adapter in/api/auth/[...nextauth].js (and
 then you can directly pass in the adapter not the DB url ) and call the adapter.createUser() and adapter.updateUser() in the provider.authorize function

gamedevsam:
The prisma schema should also be updated so this functionality can be used with Prisma.

gamedevsam:
I think providing a fully featured credential provider with forget / reset password functionality is a valuable addition to NextAuth. We can still encourage users to use email auth, but this might be a great option during development, or for users who prefer to go this route.

darshkpatel:
> I think providing a fully featured credential provider with forget / reset password functionality is a valuable addition to NextAuth. We can still encourage users to use email auth, but this might be a great option during development, or for users who prefer to go this route.

Completely agree on that one, I have to manually implement auth using ironsession if I have a requirement of email password based auth. Next-Auth providing that feature means it'd the one-stop authentication solution.

balazsorban44:
So I am going to close this, as we don't pursue any addition of an email+password combination as it is harder to maintain, and there are alternatives that are much more secure than this. @iaincollins has explained it very well, but I cannot seem to find his comment at the moment. I will try to remember posting it here when I see it again.

jsgv:
> So I am going to close this, as we don't pursue any addition of an email+password combination as it is harder to maintain, and there are alternatives that are much more secure than this. @iaincollins has explained it very well, but I cannot seem to find his comment at the moment. I will try to remember posting it here when I see it again.

Could you share some of the secure alternatives you mentioned?

Issue #718, opened 2020-09-28
Allow diacritics/accents characters for User schema in SQL Server DB - TypeORM


jiriludvik:
**Summary of proposed feature**
Hello, recently I discovered a problem when working with database sessions stored in SQL Server (TypeOrm adapter). My setup uses Azure Active Directory (oauth2.0) to sign users in. The thing is that as I have there user with name "Jiří" - this creates new record in DB table "users" with name = "JiYí". 

Having checked the [docs for MS SQL schema](https://next-auth.js.org/schemas/mssql) seems the problem is that column "name" is of type varchar. Another problem could be in [schema definition](https://github.com/nextauthjs/next-auth/blob/main/src/adapters/typeorm/models/user.js) which states the same.

**Purpose of proposed feature**
Would it be possible to update required definitions to use nvarchar instead of varchar to handle accented characters?

**Potential problems**
Not sure about potential problems as I'm quite new in Javascript and Next.JS field ;)

iaincollins:
Oh no! Sorry you ran into this issue and thank you for raising it.

Yes that change seems sensible, if it's mangling names that's a bug.

ndom91:
Re: diacritics and other special latin characters, usually you want to set the db charset to utf-8. I did a bit of digging on what to do there with typeorm and came across this: https://github.com/typeorm/typeorm/issues/390

balazsorban44:
@ndom91 is there anything to do here in the adapter code?

ndom91:
I believe you can define the `utf8` character set when defining the db. 

Might make sense to append to our seed file. 

EDIT: after some research, it seems this is the default character set with mysql8 👍 

So nothing to do as far as I see, should be supported out of the box

PR #777, opened 2020-10-14
update(db): Change image to text from varchar

Xodarap:
I got an error earlier today because someone with a Google account had an image URL longer than 255 characters. My understanding is that there isn't much of a reason to use varchar over text anyway, so I'm changing this to just use text.

I just changed the one place where I got an error – not sure if it's worth changing the others. Also I'm less familiar with the other databases but it's possible they should be updated as well.


iaincollins:
Thanks for this! Let me check if there is (a) a good reason for this or (b) if the other types are wrong too.

I suspect name and maybe email should be changed too, and that should change across schemas for different databases, but want to take this as a nudge to review all types again (and to add tests and comment/if why is varchar rather than text).

I have an odd feeling there were some things that are like that for a weird reason, but I could be wrong as it's just a mistake; but if can't get to the bottom of it would certainly favour changing them and then documenting if we have to revert.

Seems likely to relate to issue seen in #718

ndom91:
Theres another PR addressing this same issue (#845). I've done a bit of research and there doesn't seem to be much difference between `text` vs `varchar`. Other than text you don't need to specify a length, you can automatically add content up to ~64k characters. And it cannot be part of an index, however I doubt we'll ever want to index on the user image URL..

`text` is supported in postgres, mysql/mariadb, and mssql as well, unlike `nvarchar` (which was suggested in the other related issue (#718)) which doesn't seem to be supported in postgres (source: https://orkhan.gitbook.io/typeorm/docs/entities)

https://stackoverflow.com/questions/25300821/difference-between-varchar-and-text-in-mysql

My vote is for changing it to text then.

danielstocks:
I just ran into the same issue on my first attempt to use next-auth:

`[next-auth][error][create_user_error] QueryFailedError: value too long for type character varying(255)`

I see no particular reason for using varchar(255) over text?

Postgres WIKI:
https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_varchar.28n.29_by_default

Discussion on HN:
https://news.ycombinator.com/item?id=19817730

ndom91:
@danielstocks thanks for the info. We're currently in the process of releasing a canary release of next-auth with a bunch of merged PRs, bug fixes, etc.

As soon as that is tested and out, we'll go about merging others - I think this one is a no brainer though based on our research :+1:

balazsorban44:
@ndom91 is this OK to merge now?

Issue #625, opened 2020-08-31
Account linking does not work correctly when using JWT


RobertCraigie:
**Describe the bug**
Signing in with email and then signing in with an OAuth provider creates two separate users when using JWT sessions.

It should be noted that I only tested this issue with an OAuth provider that does not grant access to the user's email address

**Steps to reproduce**
https://github.com/RobertCraigie/next-auth-multiple-user-bug

* Sign in with email
* Sign in with Bungie

**Expected behavior**
One user is  created and the OAuth account is linked to said user.

**Screenshots or error logs**
https://imgur.com/a/sI2pNzp

**Additional context**
I have traced the root cause of this issue to https://github.com/nextauthjs/next-auth/blob/73d21e66ddb73da968d93cec32d4b46b152e217a/src/server/lib/callback-handler.js#L55

This block of code expects the decoded JWT to have a user property when it does not.

A simple fix is to add the user property in the JWT callback

```js
...
  callbacks: {
    jwt: async (token, user, account, profile, isNewUser) => {
      const isSignIn = user ? true : false;

      if (isSignIn) {
        token.user = { id: user.id }
      }

      return Promise.resolve(token);
    }
  },
...
```

**Feedback**

* [x] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [ ] Could not find relevant documentation
* [x] Found the example project helpful
* [ ] Did not find the example project helpful

iaincollins:
Thanks for the bug reprot @RobertCraigie I think this makes sense!

In v3 release the JWT contents got refactored to use [commonly supported JWT claims](https://tools.ietf.org/html/rfc7519#section-4.1.2) instead of a `user` object (e.g. it used to have `.user` object with `.user.name`, `.user.email`, etc but now the property names are just `.name`, `.email`).

It looks like the bug was introduced then.

That's for tracing the bug down! Line 55 and 56 in `src/server/lib/callback-handler.js` do seem to be the problem!

I'm adding some extra detail in below in case I forget - or if anyone else is able to pick this up!

### Context

As part of the changes v3 we also removed exporting the User ID from the token by default.

This was done as we switched from encrypted by default (using a signed token that was then encrypted with AES) to only signed by default (with optional JWE compatible encryption able to be enabled by setting `jwt.encryption: true`) and I wanted to avoid exposing a User ID to the client if we didn't need to to enhance security and to reduce the default JWT payload size.

[...]

PR #784, opened 2020-10-18
Store user ID in sub claim of default JWT

lukel97:
This allows us to check if the user is signed in when using JWTs
Part of #625, not sure if this is enough to close that entire issue though

### DIFF
--- a/src/server/lib/callback-handler.js
+++ b/src/server/lib/callback-handler.js
@@ -52,8 +52,8 @@ export default async (sessionToken, profile, providerAccount, options) => {
       if (useJwtSession) {
         try {
           session = await jwt.decode({ ...jwt, token: sessionToken })
-          if (session && session.user) {
-            user = await getUser(session.user.id)
+          if (session && session.sub) {
+            user = await getUser(session.sub)
             isSignedIn = !!user
           }
         } catch (e) {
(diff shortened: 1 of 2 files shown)


iaincollins:
Thanks @bubba!

I think this is a good direction and want to take the time to test it out with different options (and maybe write some tests for it).

sughodke:
> Thanks @bubba!
> 
> I think this is a good direction and want to take the time to test it out with different options (and maybe write some tests for it).

@iaincollins would you be averse to creating a "bleeding-edge" branch while these tests are being written? This way useful PRs like this one get merged and tested out without affecting "stable."

bubba:
@sughodke I've uploaded a fork of this package that contains this change here on GitHub packages: https://github.com/nuvest/next-auth/packages/510172 
You'll need to add the GitHub package registry to npm but other than that it should be public

balazsorban44:
@iaincollins as we now have a "bleeding-edge" branch (canary), could this be merged? I'm not entirely sure if the changes are breaking or not.

iaincollins:
> @iaincollins as we now have a "bleeding-edge" branch (canary), could this be merged? I'm not entirely sure if the changes are breaking or not.

I think this is good to go :D

PR #790, opened 2020-10-20
Include callbackUrl in newUser page

ramiel:
It apperas that the original callback url is not accessible from the newUser page. This includes it so that the journey can be completed. This may be wrong and I'm missing something, but it looks there's no way to access original callback url right now


iaincollins:
This is a good spot, thank you! It should still be readable via a cookie but this is a much better approach.

Will make a couple of small suggestions so it works in the unusual case of URLs that already have query string parameters in them would be great if you can review them to see if they have any typos.

src/server/routes/callback.js
```diff
@@ -110,7 +110,7 @@ export default async (req, res, options, done) => {
           // e.g. option to send users to a new account landing page on initial login
           // Note that the callback URL is preserved, so the journey can still be resumed
           if (isNewUser && pages.newUser) {
-            return redirect(pages.newUser)
+            return redirect(`${pages.newUser}?callbackUrl=${encodeURIComponent(callbackUrl)}`)
```
iaincollins, line 113: Suggestion so we check if the URL already has a ? in it (and adds the option with & instead).

```suggestion
            return redirect(`${pages.newUser}${pages.newUser.includes('?') ? '&' : '?'}callbackUrl=${encodeURIComponent(callbackUrl)}`)
```

src/server/routes/callback.js
```diff
@@ -200,7 +200,7 @@ export default async (req, res, options, done) => {
       // e.g. option to send users to a new account landing page on initial login
       // Note that the callback URL is preserved, so the journey can still be resumed
       if (isNewUser && pages.newUser) {
-        return redirect(pages.newUser)
+        return redirect(`${pages.newUser}?callbackUrl=${encodeURIComponent(callbackUrl)}`)
```
iaincollins, line 203:
```suggestion
        return redirect(`${pages.newUser}${pages.newUser.includes('?') ? '&' : '?'}callbackUrl=${encodeURIComponent(callbackUrl)}`)
```

ramiel:
Suggestion is welcome and great. Probably there was a more elegant way of doing it with URL, but it's fine like this :+1:

ndom91:
Thanks for updating it re: suggestions, looks good to me!

PR #802, opened 2020-10-27
WIP: Update Docusaurus + Site dependencies

ndom91:


ndom91:
Other than that test, this all seems to work. I will double check the test later, but otherwise should be good to merge.

ndom91:
Okay very strange, the e2e test passes completely with twitter, but has trouble with the other providers:

image

Im gonna try to figure out whats going on there tonight

PR #817, opened 2020-11-01
feat(provider): Add instagram

PolMrt:
Add Instagram as a provider.

Note that I've add `node-fetch` npm package because the Instagram api only returns some informations such as username, id, account_type. I've add a request to Instagram public api (https://www.instagram.com/instagram/?__a=1) which, strangely, returns more informations such as full name and profile pictures (this works for public and private accounts).


ndom91:
@PolMrt next.js ships with an isomorphic version of fetch since 9.4+. 

Take a look at this example for usage:

https://github.com/nextauthjs/next-auth/blob/canary/src/client/index.js

It's wrapped up in a custom function `_fetchData` located towards the bottom of the file. Can you please use that instead of adding a new dependency? Otherwise it looks good, thanks! :tada:

PolMrt:
Thanks I'll update that. But I tried to use this "_public API_" for another project and it seems to not be usable as an API, sometimes it will redirects you to a login page if you haven't logged to Insta with your ip in a while... So I don't think this is the best approach.

Would it be okay if we only stuck with the username ? Or should we still use this unstable API and return no email, picture, name (as [the Instagram API don't return that kind of information for non-business account](https://developers.facebook.com/docs/instagram-basic-display-api/reference/user#fields)) if there's a redirection ?

package.json
```diff
@@ -46,6 +46,7 @@
     "jose": "^1.27.2",
     "jsonwebtoken": "^8.5.1",
     "jwt-decode": "^2.2.0",
+    "node-fetch": "^2.6.1",
```
balazsorban44, line 49:
```suggestion
```

src/providers/instagram.js
```diff
@@ -0,0 +1,25 @@
+import fetch from 'node-fetch'
```
balazsorban44, line 1:
```suggestion
import logger from "../../lib/logger"
```

src/providers/instagram.js
```diff
@@ -0,0 +1,25 @@
...
+    profileUrl: 'https://graph.instagram.com/me?fields=id,username,account_type,name',
+    profile: async (profile) => {
+      const instagramData = await fetch('https://www.instagram.com/' + profile.username + '/?__a=1').then(res => res.json())
+      return {
+        id: profile.id,
+        name: instagramData.graphql.user.full_name,
+        email: null,
+        image: instagramData.graphql.user.profile_pic_url_hd
+      }
+    },
```
balazsorban44, line 22:
```suggestion
    async profile (profile) {
      try {
        const response = await fetch('https://www.instagram.com/' + profile.username + '/?__a=1')
        const data = await response.json()
        return {
          id: profile.id,
          name: data.graphql.user.full_name,
          email: null,
          image: data.graphql.user.profile_pic_url_hd
        }
      } catch(error) {
        logger.error("PROFILE_FECTH_ERROR", error)
      }
    },
```

www/docs/providers/instagram.md
```diff
@@ -0,0 +1,35 @@
...
+}
+...
+```
+
+:::warning
+Email address is not returned by the Instagram API.
+:::
+
+:::tip
+Instagram display app required callback url to be configured in your Facebook app and Facebook required you to use **https** even for localhost ! In order to that, you either need to [add a ssl to your localhost](https://www.freecodecamp.org/news/how-to-get-https-working-on-your-local-development-environment-in-5-minutes-7af615770eec/) or use a proxy such as [ngrock](https://ngrok.com/docs).
```
balazsorban44, line 33:
```suggestion
Instagram display app required callback URL to be configured in your Facebook app and Facebook required you to use **https** even for localhost! In order to do that, you either need to [add an SSL to your localhost](https://www.freecodecamp.org/news/how-to-get-https-working-on-your-local-development-environment-in-5-minutes-7af615770eec/) or use a proxy such as [ngrock](https://ngrok.com/docs).
```

balazsorban44 (changes requested):
I added some suggestions to eliminate `node-fetch`. Could you please check out if this still works for you locally?

PolMrt:
Thanks for your help @balazsorban44. Everything seem to be fine, gonna try this tomorrow morning !

src/providers/instagram.js
```diff
@@ -0,0 +1,30 @@
...
+        const response = await fetch(`https://www.instagram.com/${profile.username}/?__a=1`)
+        const data = await response.json()
+        return {
+          id: profile.id,
+          name: data.graphql.user.full_name,
+          email: null,
+          image: data.graphql.user.profile_pic_url_hd
+        }
+      } catch(error) {
+        logger.error("PROFILE_FETCH_ERROR", error)
```
balazsorban44, line 25:
```suggestion
        logger.error("PROFILE_FETCH_ERROR", error)
        return {
          id: profile.id,
          name: profile.username
        }
```

Maybe return the profile `id` and `username` in case of an error also?

balazsorban44 (changes requested):

balazsorban44:
@PolMrt I added another suggestion. Also, could you make sure that there are no linting errors, and could you report back if you have tested these changes locally successfully? Thank you!

PolMrt:
I don't think we're in the good path right now. Let me explain :

My initial idea (using Instagram *public* api) was working because I tried it on localhost. Next's server side and client side were using the same IP. So it was always working because of it. But when it will be used in production, this would never work, not even once since the IP of the server has never been used in Instagram. 

I think we should just return user id and username, nothing else. Implementing a solution that might works some time but not most of the time will be very confusing for developers. 

We should say in the doc that only the username is returned and maybe add a second Instagram provider for business account using the API reserved to that kind of account (which returns a lot more of informations).

Please tell me if that sounds good to you and I'll update the PR.

balazsorban44:
Sorry for the delay. So I found this documentation: https://developers.facebook.com/docs/instagram-basic-display-api/reference/user

if I understand it correctly, you simply don't have the option to get the user's email?! 😕 Very strange from Instagram.

I'm not sure how to proceed, cc @iaincollins

Ericrobert22:
> Sorry for the delay. So I found this documentation: https://developers.facebook.com/docs/instagram-basic-display-api/reference/user
> 
> if I understand it correctly, you simply don't have the option to get the user's email?! 😕 Very strange from Instagram.
> 
> I'm not sure how to proceed, cc @iaincollins

I had similar issues like this on till i mete acehack1 on Instagram he helped me got my successfully

PR #819, opened 2020-11-01
feat: allow react 17 as a peer dependency

### LINKED ISSUE #814: Widen peer dependency range to include React 17

[React 17](https://reactjs.org/blog/2020/10/20/react-v17.html) was released recently with minimal breaking changes. `next-auth` currently supports only `^16.13.1`:

https://github.com/nextauthjs/next-auth/blob/5126f4e3428f2b4b8ebdd38b10066edfdd4ea2fa/package.json#L57-L60

Could this be widened to `^16.13.1 || ^17.0.0`?

Allows apps using React 17 to also use `next-auth` without peer dependency warnings.

n/a

**Potential problems**
There are [subtle differences](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#changes-to-event-delegation) between React 16 and React 17, so it's worth reviewing to see if any of them might affect the library's functionality.

You could alternatively release a new major version of `next-auth` that peer-depends only on React `^17.0.0`, but since the changes are minimal, you may not want to deprecate support for React 16 going forward.

**Additional context**
n/a

Yes.

lukel97:
Closes #814

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,8 +55,8 @@
     "typeorm": "^0.2.24"
   },
   "peerDependencies": {
-    "react": "^16.13.1",
-    "react-dom": "^16.13.1"
+    "react": "^16.13.1 || ^17",
+    "react-dom": "^16.13.1 || ^17"
   },
   "peerOptionalDependencies": {
     "mongodb": "^3.5.9",


package.json
```diff
@@ -55,8 +55,8 @@
     "typeorm": "^0.2.24"
   },
   "peerDependencies": {
-    "react": "^16.13.1",
-    "react-dom": "^16.13.1"
+    "react": "^16.13.1 || 17",
+    "react-dom": "^16.13.1 || 17"
```
Zertz, line 59: May I suggest using this instead?

```suggestion
    "react": ">=16.13.1",
    "react-dom": ">=16.13.1"
```

iaincollins: Thanks both! Am thinking we tweak that further to `"^16.13.1 || ^17"`?

Will try and get this released over the weekend.

Zertz: It's hard to tell without knowing how breaking an eventual React 18 would be but in the past upgrading has been very smooth as long as deprecations are fixed in a timely manner.

Also, npm seems to be somewhat nudging toward having loose peer dependencies to avoid, for example, installing two copies of React when a user tries out a v18 pre-release.

That's why I personally prefer the `>=` approach but I don't have a very strong opinion either way!

bubba: I don't have much of an opinion either, but tweaked it to `^16.13.1 || ^17` anyway

balazsorban44: Thanks!

PR #823, opened 2020-11-02
feat: forward auth params from signin to provider

### LINKED ISSUE #737: Send ui_locales to external provider trough login
Our external provider (Identity server 4) supports multiple languages. Therefore we would like to send the parameter `ui_locales ` to the provider.

Think the nices way is when we can add extra query params to the login function. Like this:

`signIn('provider', params: { ui_locales : "en-US" });`

balazsorban44:
According to the spec, additional parameters can be sent to the authorization url, please see the spec:

https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

I would like to send the `prompt: "login"` param, so my users will be forced to log in on the Provider, whenever they click on a button with the handler
```js
signIn("identity-server4", {}, {
  prompt: "login"
})
```

NOTE: These params should not be sent in the body as `args` here:
https://github.com/nextauthjs/next-auth/blob/5126f4e3428f2b4b8ebdd38b10066edfdd4ea2fa/src/client/index.js#L238-L251

This is why I decided to add a third argument to the `signIn` function instead. This also keeps this change **backward compatible**.

**Tested this locally, and it works as intended**

This fixes #737, as far as I can tell.

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -235,7 +235,7 @@ const signIn = async (provider, args = {}, authParams = {}) => {
     let signInUrl = (providers[provider].type === 'credentials')
       ? `${baseUrl}/callback/${provider}`
       : `${baseUrl}/signin/${provider}`
-    
+
     if (authParams) {
       signInUrl += `?${new URLSearchParams(authParams).toString()}`
     }
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -222,7 +222,7 @@ const _useSessionHook = (session) => {
 }
 
 // Client side method
-const signIn = async (provider, args = {}) => {
+const signIn = async (provider, args = {}, authParams = {}) => {
   const baseUrl = _apiBaseUrl()
   const callbackUrl = (args && args.callbackUrl) ? args.callbackUrl : window.location
   const providers = await getProviders()
@@ -232,9 +232,14 @@ const signIn = async (provider, args = {}) => {
     // If Provider not recognized, redirect to sign in page
     window.location = `${baseUrl}/signin?callbackUrl=${encodeURIComponent(callbackUrl)}`
   } else {
-    const signInUrl = (providers[provider].type === 'credentials')
+    let signInUrl = (providers[provider].type === 'credentials')
       ? `${baseUrl}/callback/${provider}`
       : `${baseUrl}/signin/${provider}`
+    
+    if (authParams) {
+      signInUrl += `?${new URLSearchParams(authParams).toString()}`
+    }
+
     // If is any other provider type, POST to provider URL with CSRF Token,
     // callback URL and any other parameters supplied.
     const fetchOptions = {
--- a/src/server/lib/signin/oauth.js
+++ b/src/server/lib/signin/oauth.js
@@ -2,7 +2,7 @@ import oAuthClient from '../oauth/client'
 import { createHash } from 'crypto'
 import logger from '../../../lib/logger'
 
-export default (provider, csrfToken, callback) => {
+export default (provider, csrfToken, callback, authParams) => {
   const { callbackUrl } = provider
   const client = oAuthClient(provider)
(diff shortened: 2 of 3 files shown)


makezi:
This would greatly benefit the situation in which I would like to append form data (along with `register=true`) to the URL for my custom Auth0 lock screen to default to the signup tab and process said form data.

jmfury:
> This would greatly benefit the situation in which I would like to append form data (along with `register=true`) to the URL for my custom Auth0 lock screen to default to the signup tab and process said form data.

@makezi  or @balazsorban44  Did either of you happen to get this option working with Auth0? 

I'm passing in `mode: 'signUp'` to no avail 😢 

Adding this as a param in the `authorizationUrl` within the `next-auth` instance options works but for some reason I'd like to preserve a more dynamic option.. i.e. Support both `login` or `signUp`

balazsorban44:
Hi there! I removed this option after #1022 in canary, but if there is still a need for it, maybe I could add it back. Maybe open a feautr request issue on it?

jmfury:
> Hi there! I removed this option after #1022 in canary, but if there is still a need for it, maybe I could add it back. Maybe open a feautr request issue on it?

@balazsorban44  I'll plan to PR something in addition since that'll help me describe the interface & useful options for Auth0 specifically 👍

makezi:
@balazsorban44 I know that having `authParams` is still necessary for Auth0's case for customisation/passing params to [Auth0 Lock](https://auth0.com/docs/libraries/lock/lock-configuration). This is blocking for our situation hence have to remain on < version 3.2.0-canary.15

@jmfury My case was a bit more complicated hence had to [customise the lock sign in page](https://auth0.com/docs/universal-login/universal-login-page-customization#lock) to read some other user registration data from the app client. Either way, with `authParams` you could still pass params to lock, mainly [initialScreen](https://auth0.com/docs/universal-login/universal-login-page-customization#lock) to open lock on sign up. We made it easier to just provide an auth interface across our app, an example:

```javascript
// utils/auth.js
import { signIn } from 'next-auth/client';

function handleSignIn(event, params) {
  event.preventDefault();
  signIn('auth0', { callbackUrl: '/' }, params);
}

function handleRegister(event, params) {
  event.preventDefault();
  // This is what we did but in your case you would do { initialScreen: 'signUp', ...params }
  signIn('auth0', { callbackUrl: '/' }, { register: true, ...params} );
}

...

<a href="/register" onClick={handleRegister}>Register</button>
```

Hope that helps! I'll upvote that PR 👍!

jmfury:
@makezi  This was hugely helpful.. 

After digging thru the client code I realized I was [wrong about the above](https://github.com/nextauthjs/next-auth/pull/823#issuecomment-759756932) (re: server route for `signin`).

The following is a customized version of the original `signIn()` helper for my use case, but perhaps it is useful for you as well?

```js
async function signUp(provider, args) {
  const csrfToken = await getCsrfToken()
  const signInUrl = `${process.env.NEXT_PUBLIC_NEXTAUTH_URL}/api/auth/signin/${provider}`
  const fetchOptions = {
    method: 'POST',
    headers: {
      'Content-Type': 'application/x-www-form-urlencoded'
    },
    body: encodeForm({ // Taken from the `next-auth` helper
      csrfToken,
      callbackUrl: args?.callbackUrl || process.env.NEXT_PUBLIC_NEXTAUTH_URL,
      json: true
    })
  }
  const res = await fetch(signInUrl, fetchOptions)
  const data = await res.json()
  return (window.location = data?.url ? data.url + `&mode=signUp`  : '/my-signin-error-page') // custom params or error page
}

```

Given the above, another option to consider is some sort of `getSignInUrl()` helper. This could simply return the URL to the client and it would be the consumer's responsibility to add custom params and navigate to the URL itself

balazsorban44:
So this is available in `next-auth@canary.26` again! 🎉

PR #809, opened 2020-10-29
feat(provider): Add Azure Active Directory B2C

vladimir-e:
Adds a provider for authenticating with Azure OAuth 2.0 API.

Update: 
[How to Configure Azure AD B2C Authentication with Next.js](https://benjaminwfox.com/blog/tech/how-to-configure-azure-b2c-with-nextjs)

PR #827, opened 2020-11-03
Azure AD Provider

bjrb20:
Added a provider for Azure AD. When this is merged I can also update the types accordingly.

### DIFF
--- a/src/providers/azure-ad.js
+++ b/src/providers/azure-ad.js
@@ -1,0 +1,25 @@
+export default (options) => {
+  return {
+    id: 'azure-ad',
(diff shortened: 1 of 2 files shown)


PolMrt:
Seems to be a duplicate of #809 ?

balazsorban44:
Hi @BenRedmondBenham, and thanks for this PR!

I think I am going to close this in favor of #809, as that one came before yours, and also adds some documentation.

If you feel you could contribute to the other PR with some comments, please do! I see for example you added `image` to the `profile` callback, which I cannot see on the other PR.

PR #831, opened 2020-11-04
Allow customization of verification requests

fromi:
Transfer options to "sendVerificationRequest" function to be able to create custom email depending on the user request (for instance, send the email in their language)

### DIFF
--- a/src/adapters/typeorm/index.js
+++ b/src/adapters/typeorm/index.js
@@ -288,7 +288,7 @@ const Adapter = (typeOrmConfig, options = {}) => {
       }
     }
 
-    async function createVerificationRequest (identifier, url, token, secret, provider) {
+    async function createVerificationRequest (identifier, url, token, secret, provider, options) {
       debug('CREATE_VERIFICATION_REQUEST', identifier)
       try {
         const { baseUrl } = appOptions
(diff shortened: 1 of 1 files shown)


balazsorban44:
Hi there! Thanks for this PR!

I was wondering, what information are you expecting to use from `options` here? You are forwarding the following object:
https://github.com/nextauthjs/next-auth/blob/71c78e8e24b0ac4d0de17a790b7ae8a753d79436/src/server/index.js#L91-L135

You mention you could use it to send an e-mail in a different language. I don't think we have any information about that from the user in `options`. That said, there is an open issue https://github.com/nextauthjs/next-auth/issues/1154 where this could fit. We could register the user's preferred language by utilizing cookies and headers (and/or https://nextjs.org/docs/advanced-features/i18n-routing). Would this cover your use case?

This includes properties and information we probably don't wish to expose to user-facing APIs, so options should be at least stripped.

In any case, I am going to close this now. If you are still interested in this to be added, could you please open a new PR against `main`? We have changed the release flow, and simply changing the base creates a lot of merge conflicts and unfortunately screws up our automatic release process :slightly_frowning_face:

I would like to hear more about your use case!

fromi:
I honestly can't remember after 4 month what I had tracked. Maybe I was wrong.
Use case is simple enough: send the verification emails (magic link) in the correct language. #1154 is fine. For now I worked it around using a global variable… ugly and not guaranteed to work all the time!

balazsorban44:
Thanks! I'll try keeping this in mind when/if we get to Multi Language Support! 😉 And if I don't forget, I can ping you about this being built-in

Issue #836, opened 2020-11-06
Federated logout (OpenID Connect)


balazsorban44:
**Summary of proposed feature**

Acces `id_token` in `jwt` callback, when `idToken: true` in a provider's option.

**Purpose of proposed feature**

Make it possible to start a logout process from a `next` app using `next-auth` that will log out from the Identity Provider entirely, if it is OIDC compliant.

**Detail about proposed feature**

OpenID Connect compliant IdPs (like IdentiyServer4, which is also supported by `next-auth`) have a federated logout.
To initiate such a logout process, when `signOut()` is called, `next-auth` should redirect to `https://idp.com/endsession`, and forward at least the two following query params:

1. `id_token_hint`: The original `id_token` received from the IdP at login.
2. `post_logout_redirect_uri`: A preconfigured URL in the IdP client, that the IdP should redirect to after a successful logout. This could default to `NEXTAUTH_URL`, but could be overridden in a provider setting. (Note that the same url must be registered in the IdP client)

Additional optional query params exist, see the [relevant spec section](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout)

Obtaining the `id_token` at the initial `jwt()` callback call would make it possible to preserve it in a session database, or in the jwt token itself.

**Potential problems**

- According to the [JWT callback docs](https://next-auth.js.org/configuration/callbacks#jwt-callback) `jwt` callback has already many parameters, and adding a new `idToken` would just make it worse. In my opinion should/could be provided as a single object to basically be able to use named params, but that would be a breaking change.

- Setting the `id_token` in the JWT cookie is kind of an overhead, as the cookie most possibly contains some kind of a mapping of the id_token (name, email, etc.) already. My solution for this would be to not store it in the cookie, but create an in-memory database (just a JS class with key value pairs, where keys are user ids, and each user contains a list of sessions, containing sessions' `id_token`s. Hint: The id_token usually contains a `sid` or Session ID, and timestamps if needed). Adding such an in-memory database should not be required from most of the users though. (Although this could very well be an optional part of `next-auth`, that could also complement #583 nicely, which I am already working on at work, and some day be able to share it as a PR to `next-auth`) _*I am open for suggestions here!*_

- There is the 4096 bytes cookie size limit in browser, and because of that every extra bytes count, if we cannot somehow divide the data into multiple cookies somehow. (I am no expert on cookies though, so I do not know hard it would be to split/merge cookies)

**Describe any alternatives you've considered**
[...]

PR #837, opened 2020-11-07
feat: Forward idToken to callbacks

### LINKED ISSUE #378: Not getting the ID token, only the access token from Okta
Should I be able to get the ID token when using Okta? 

I have a working project using Okta. I am requesting the following scopes in my [...nextauth].js file...
scope: "openid profile email offline_access"

The [Okta docs](https://developer.okta.com/docs/reference/api/oidc/#response-properties-2) indicate the id_token is included in the response from /v1/token if the "openid" scope was requested.

console output of "message" in session Event...

```
{
  session: {
    user: {
      name: 'Bruce',
      email: 'bruce@redacted',
      image: null
    },
    expires: '2020-07-30T21:12:10.158Z'
  },
  jwt: {
    user: { name: 'Bruce', email: 'bruce@redacted' },
    account: {
      provider: 'okta',
      type: 'oauth',
      id: '00ubk926sxN1qb8ZW4x6',
      refreshToken: 'IjQMrKDYYgtuIIxssqrnT_ZuHEYHNZXY_uyZIOjOp7U',
      accessToken: 'redacted',
      accessTokenExpires: null
    }
  }
}
```
The access token decodes to

```
{
  "ver": 1,
  "jti": "AT.l1a5pxB-uVn5FpbUGG_Mo0OUF7ySZQNu361JKsnu5qE.m/3Vy0DhG5LJKBRVs/SLm18+v8jnsNOC1iTxwTJLCUo=",
  "iss": "https://dev-726851.okta.com",
  "aud": "https://dev-726851.okta.com",
  "sub": "bruce@redacted",
  "iat": 1593551529,
  "exp": 1593555129,
  "cid": "0oac1u4f77TZisppR4x6",
  "uid": "00ubk926sxN1qb8ZW4x6",
  "scp": [
    "openid",
    "profile",
    "email",
    "offline_access"
  ]
}

```

### LINKED ISSUE #393: How to get Google provider ID token and not access token
When using the Google provider I want to be able to authenticate with a backend server and for this I need to send the `ID token` to that API and get back the access token. Further info about this flow can be found [here](https://developers.google.com/identity/sign-in/web/backend-auth).

This extra request I plan to do in the JWT callback, but in there I cannot access `ID token`. Only `accessToken` is given back in `token.account`.

This is what I get in the JWT callback:
token (first argument):
```
{
  user: {
    name: <name>,
    email: <email>,
    image: <image>
  },
  account: {
    provider: 'google',
    type: 'oauth',
    id: <id>,
    refreshToken: undefined,
    accessToken: <accessToken>,
    accessTokenExpires: null
  },
  isNewUser: undefined
}
```

profile (second argument):
```
{
  id: <id>,
  email: <email>,
  verified_email: true,
  name: <name>,
  given_name: <given_name>,
  family_name: <family_name>,
  picture: <picture>,
  locale: 'en'
}
```

* [X] Did not find the example project helpful

balazsorban44:
This is a bare minimum change required so an OpenID Connect compliant Identity Provider, (like IdentityServer4) can initiate a logout process from the IdP from within an app using `next-auth`, and return to the application afterwards.

According to the [spec](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout), an `id_token` must be present as either a query param (in case of `GET`) or as a body in case of `POST` to `https://IdP.com/connect/endsession` to be able to redirect back to the application (or anywhere else explicitely allowed so in the IdP client by `post_logout_redirect_uri`) after a successful logout.

This change forwards the `id_token` as-is to `callbacks.signIn` and `callbacks.jwt` initially, and leaves it up to the user to store that as they want to (probably either in a session database or in the JWT token).

There may be other situations where access to the `id_token` in that manner might be useful.

Related issues: #836, closes #393, closes #378

**UPDATE**:

I decided that in case an `id_token` is present, it should always be sent to the callbacks, regardless of `idToken: true` is set, or not.

The reasoning is that I thought that by setting `idToken: true`, I could skip an extra request to the `/userinfo` IdP endpoint. The problem is, that if I use `id_token`, all the claims must be sent from the IdP, making the `id_token` too big to send it again while using the `/connect/endsession` endpoint, destroying the purpose of this PR.

So here is a rough breakdown of how this PR solves my problem.

1. Create an `/api` endpoint in Next.js, call it `/api/auth/federated-logout`
```ts
import sessionsDB from "lib/session-db" // an in-memory session database
import jwt from "next-auth/jwt"
import log from "utils/server-logger"

export default async function federatedLogout (req, res){
  const token = await jwt.getToken({req, secret: process.env.SECRET, encryption: true})

  if (token) {
    const { userId, sessionId } = token
    const userSession = await sessionsDB.getUserSession(userId, sessionId)
    const endsessionURL = `https://identity-provider.com/connect/endsession`
    const endsessionParams = new URLSearchParams({
      id_token_hint: userSession?.rawIdToken, // since it is an in-memory DB, a server restart would make the redirect not possible. 😢 If needed though, the DB can be moved out of memory. I tried to keep it simple.
      post_logout_redirect_uri: `http://localhost:3000?forceLogout=true`,
    })
    res.redirect(`${endsessionURL}?${endsessionParams}`)
    return
  }
  res.redirect("http://localhost:3000?forceLogout=true")
}
```

2. With the provided `id_token_hint` to `/connect/endsession` will be able log the user out without consent, and redirect back to `http://localhost:3000?forceLogout=true`

3. In `_app` add the following:
```js
  import { signOut } from "next-auth/client"

  //...
  const router = useRouter()
  React.useEffect(() => {
    if (router.query.forceLogout) {
      signOut({ callbackUrl: "http://localhost:3000" }) // get rid of the `forceLogout=true` param
    }
  }, [router.query.forceLogout])
```

4. Step 3. makes sure that after the user is logged out of the IdP, the app also drops the now invalid session.
5. When a user want to initiate a federated logout, we just add a `<a href="/api/auth/federated-logout">Log out everywhere</a>` link somewhere in the app.


balazsorban44:
@iaincollins Not sure if I should update the docs here as well:
1. https://github.com/nextauthjs/next-auth/blame/canary/www/docs/providers/google.md#L62
2. https://github.com/nextauthjs/next-auth/blame/canary/www/docs/tutorials/ldap-auth.md#L50

Can `id_token` be relevant in any of these places?

www/docs/configuration/callbacks.md
```diff
@@ -25,7 +25,7 @@ You can specify a handler for any of the callbacks below.
     session: async (session, user) => {
       return Promise.resolve(session)
     },
-    jwt: async (token, user, account, profile, isNewUser) => {
+    jwt: async (token, user, account, profile, isNewUser, idToken) => {
```
balazsorban44, line 28: @iaincollins This (and probably any other public-facing function signature using more than 2 params) should at some point (maybe v4) become an object to leverage named parameters It is much easier for the user to pick only those params that they are going to use.

ryangittings:
Can we get this merged? Happy to help if I can but desperate to use this on a project. 

Cheers! :)

balazsorban44:
@ryangittings hope to get this merged soon. As it's a user facing change, I would like to review it with someone else from the core team before merge. (will have a chat with @iaincollins tomorrow, so hopefully then 🤞)

balazsorban44:
Continued at #1024

PR #839, opened 2020-11-08
docs: Update default ports for supported databases

manishrc:
Updated default port number for Postgres & MongoDB:
https://next-auth.js.org/configuration/databases


ndom91:
Looks good to me! Thanks for catching that oversight!

PR #867, opened 2020-11-20
feat: simplify NextAuth instantiation

### LINKED ISSUE #868: Simplify next-auth initialization
Simplifying the initialization of `next-auth`

The first thing that got me with `next-auth` was how simple it is to initialize in a single file. Alternatives I have tried always asked me to create multiple Next.js `/api` routes in multiple files. Building on the simplicity of `next-auth` it already brings, I propose a change that could further simplify it.

```diff
import NextAuth from 'next-auth'

const options = {
  //...
}

- export default (req, res) => NextAuth(req, res, options)
+ export default NextAuth(options)
```

Since req and res are not expected to be used in the default exported function, it can be "hidden" from the user if they wish and have no use for it.

**Potential problems**
It can be implemented in a backwards compatible way, so none.

None

**Additional context**
The implementation is available at #867

balazsorban44:
```diff
import NextAuth from 'next-auth'

const options = {
  //...
}

- export default (req, res) => NextAuth(req, res, options)
+ export default NextAuth(options)
```

We could simplify the `NextAuth` instance even further, by checking the number of arguments. If there is only one argument, the default function in `server/index.js` could return a `NextApiHandler`. Otherwise preserve the current behaviour, for backwards compatibility.

Fixes #868

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ const options = {
   database: process.env.DATABASE_URL
 }
 
-export default (req, res) => NextAuth(req, res, options)
+export default NextAuth(options)
 ```
 
 ### Add React Component
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Alternatively you can raise a PR directly with your fixes on [**DefinitelyTyped*
 import NextAuth from 'next-auth'
 import Providers from 'next-auth/providers'
 
-const options = {
+export default NextAuth({
   providers: [
     // OAuth authentication providers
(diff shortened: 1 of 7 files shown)


kripod:
Great change!

Typings should be updated accordingly [in the DefinitelyTyped project](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/next-auth), so TypeScript users may also benefit from this change.

PR #870, opened 2020-11-20
feat: user configurable logging

### LINKED ISSUE #869: Configurable logging
A clear and concise description of the feature being proposed.

Make it possible for the user to decide how/where to log messages by `next-auth`, on all level.

A clear and concise description description of why this feature is necessary and what problems it solves.

`next-auth` already gives us the possibility to log some errors when they arise, through the 
error [event handler](https://next-auth.js.org/configuration/events) (although I am not sure when and where this is called, @iaincollins. I would even argue, that this proposal would make `events.error` unnecessary, but that would make the feature a breaking change. I need your opinion on this). I think we can get a bit further, if we make it possible for the user to decide where to send all the logs put out by `next-auth`. You see, when the Next.js app gets deployed somewhere, logging to the console will probably be lost or at least hard to reach. I propose a way for the user to get hold of those logs, and let them send it to somewhere else, like a third-party logging service.

Here is how I can see the user facing API, following the signatures of built-in `console` logger functions at https://github.com/nextauthjs/next-auth/blob/main/src/lib/logger.js:

```ts
import myLogger from "my-logger"

const options = {
  logger: {
    error(params) {
      myLogger.error(params, "{code}: {text} ({link})")
    },
    warn(params) {
      myLogger.warn(params, "{code}: {text} ({link})")
    },
    debug(params) {
      myLogger.debug(params, "{code}: {text}")
    },
  }	as Logger
}

interface LogParams {
  code: string
  text: string
  link?: string
}

// This is the form of the logger, not required by user, could be added to `@types/next-auth`
interface Logger {
  error?: (params: LogParams) => void
  warn?: (params: LogParams) => void
  debug?: (params: LogParams) => void
}
```

Here, when `myLogger#{level}` is called, it will send that message to a third party logging server. 

Its implementation is less relevant than the signature of the `logger` methods, But here I provided the format which our logger takes. The first argument is an object, the second is the actual message, where I can reference values from the object. The message is shown first in our logger service, and on click it will also expand the object. In case `myLogger` has the same signature as `logger` expects, the user could simply do :
```ts
//...
import myLogger from "my-logger"

const options = {
  logger: myLogger
}

```

The idea is, that the user could provide as many of these methods as they want, so if they only define `error` inside `logger`, then `warn` and `debug` will keep logging to `console`. Or we could require the user to provide all 3 levels all the time, for consistency.

**Potential problems**

This can be implemented in a backward-compatible way, so no problems arise.

When having a problem with `next-auth`, I can only debug it locally, cause accessing the logs in our deployment environment is not always easy (For example a Docker container in Azure App Service)

**Additional context**

Implemented in #870 

Would love to here your thoughts!

### LINKED ISSUE #907: How to log errors to file instead of console?
Is it possible to avoid logging errors to console and instead log them to file? 

I am using the default TypeOrm Postgres adapter. In the console, there is an error ([next-auth][error][create_user_error] QueryFailedError). Checking console for errors will be a problem when on production. I tried to use the config option "events" -> "error" to call my custom logger but this event is not fired for this particular error.

balazsorban44:
It adds a new option called `logger` to the `options` object taken by `NextAuth`.

closes #869
closes #907

### DIFF
--- a/src/lib/logger.js
+++ b/src/lib/logger.js
@@ -1,30 +1,60 @@
+const userLogger = {
+  error: undefined,
+  warn: undefined,
+  debug: undefined
+}
+
 const logger = {
-  error: (errorCode, ...text) => {
-    if (!console) { return }
+  error (code, ...text) {
     if (text && text.length <= 1) { text = text[0] || '' }
-    console.error(
-      `[next-auth][error][${errorCode.toLowerCase()}]`,
-      text,
-      `\nhttps://next-auth.js.org/errors#${errorCode.toLowerCase()}`
-    )
+    code = code.toLowerCase()
+    const link = `https://next-auth.js.org/errors#${code}`
+    if (userLogger.error) {
+      userLogger.error(code, text, link)
+    } else {
+      console.error(
+        `[next-auth][error][${code}]`,
+        text,
+        `\n${link}`
+      )
+    }
   },
-  warn: (warnCode, ...text) => {
-    if (!console) { return }
+  warn (code, ...text) {
     if (text && text.length <= 1) { text = text[0] || '' }
-    console.warn(
-      `[next-auth][warn][${warnCode.toLowerCase()}]`,
-      text,
-      `\nhttps://next-auth.js.org/warnings#${warnCode.toLowerCase()}`
-    )
+    code = code.toLowerCase()
+    const link = `https://next-auth.js.org/warnings#${code}`
+    if (userLogger.warn) {
+      userLogger.warn(code, text, link)
+    } else {
+      console.warn(
+        `[next-auth][warn][${code}]`,
+        text,
+        `\n${link}}`
+      )
+    }
   },
-  debug: (debugCode, ...text) => {
-    if (!console) { return }
+  debug (code, ...text) {
     if (text && text.length <= 1) { text = text[0] || '' }
+    code = code.toLowerCase()
+    if (userLogger.debug) {
+      userLogger.debug(code, text)
+    }
     if (process && process.env && process.env._NEXTAUTH_DEBUG) {
       console.log(
-        `[next-auth][debug][${debugCode.toLowerCase()}]`,
+        `[next-auth][debug][${code}]`,
         text
       )
+    }
+  },
+  setLogger (logger = {}) {
+    if (typeof logger.error === 'function') {
+      userLogger.error = logger.error
+    }
+    if (typeof logger.warn === 'function') {
+      userLogger.warn = logger.warn
+    }
+    if (typeof logger.debug === 'function') {
+      userLogger.debug = logger.debug
     }
   }
 }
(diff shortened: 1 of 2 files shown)


src/lib/logger.js
```diff
@@ -1,31 +1,61 @@
...
   },
-  debug: (debugCode, ...text) => {
-    if (!console) { return }
+  debug (code, ...text) {
     if (text && text.length <= 1) { text = text[0] || '' }
+    code = code.toLowerCase()
+    if (userLogger.debug) {
+      userLogger.debug(code, text)
+    }
     if (process && process.env && process.env._NEXTAUTH_DEBUG) {
```
balazsorban44, line 42: I am not checking for `process.env._NEXTAUTH_DEBUG` here, because if the user implemented the `logger.debug` method, they probably expect it to be called. In this case, they can decide how debugging logs should be handled. (I for example have my own `process.env.LOG_LEVEL` env variable, which decides which logs should be sent to my third-pary and which one shouldn't. It also works in a way that when `process.env.NODE_ENV` is not `"production"`, instead of sending the logs, it will fall back to `console`)

src/lib/logger.js
```diff
@@ -1,31 +1,61 @@
...
+  setLogger (logger = {}) {
+    if (typeof logger.error === 'function') {
+      userLogger.error = logger.error
+    }
+    if (typeof logger.warn === 'function') {
+      userLogger.warn = logger.warn
+    }
+    if (typeof logger.debug === 'function') {
+      userLogger.debug = logger.debug
+    }
```
balazsorban44, line 58: We could make so that all three methods are required and only accepted together, and provide a nice `TypeError` to the user if they forgot one or did not define it correctly. I don't have strong opinions about this.

src/server/index.js
```diff
@@ -22,6 +22,9 @@ if (!process.env.NEXTAUTH_URL) {
 }
 
 export default async (req, res, userSuppliedOptions) => {
+  if (userSuppliedOptions.logger) {
+    logger.setLogger(userSuppliedOptions.logger)
+  }
```
balazsorban44, line 27: Providing this in `userSuppliedOptions` would make this totally backwards compatible and opt-in.

src/lib/logger.js
```diff
@@ -1,31 +1,61 @@
+const userLogger = {
+  error: undefined,
+  warn: undefined,
+  debug: undefined
+}
+
 const logger = {
-  error: (errorCode, ...text) => {
-    if (!console) { return }
```
balazsorban44, line 3: I could not understand when console would not be defined, so I took the liberty and removed these checks, also further down in `warn` and `debug`. Please do tell if this has a real importance, and I can add them back in a way that it will not collide with the PR's implementation.

balazsorban44:
Need to handle `log.debug` on client side

PR #875, opened 2020-11-23
Sync TypeORM provider support with Prisma provider

damieng:
Specifically:

- Adds usual try/catch error semantics to `updateUser`
- Implements `deleteUser`
- Implements `unlinkAccount`
- Adds deleting of expired sessions to `createSession`

I need to test this a bit first. Additionally I think all providers `deleteUser` should also delete the users associated `Account`, `Sessions` and `VerificationRequest` entries for that user but will leave that for another PR.

### DIFF
--- a/src/adapters/typeorm/index.js
+++ b/src/adapters/typeorm/index.js
@@ -176,13 +176,23 @@ const Adapter = (typeOrmConfig, options = {}) => {
 
     async function updateUser (user) {
       debug('UPDATE_USER', user)
-      return manager.save(User, user)
+      try {
+        return manager.save(User, user)
+      }
+      catch (error) {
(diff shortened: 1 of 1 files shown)


ndom91:
Thanks for finishing up some of those functions and bringing the typeorm provider up to date!

You mentioned still wanting to test a bit, do you still consider this PR a WIP then? 

Just give us a thumbs up when you feel its ready to merge from your side :+1:

damieng:
I haven't been able to test them and I don't think I will be able to for a little while as my project won't be reaching the point where we'd be calling them for several more weeks. Somebody can either take this over or I can close it down. (I don't see how these changes can fail but it's those famous last words)

PR #895, opened 2020-12-01
Update Slack provider to use V2 OAuth endpoints

### LINKED ISSUE #798: Support Slack V2 endpoints
The [current implementation](https://github.com/nextauthjs/next-auth/blob/main/src/providers/slack.js) of Slack is for [legacy Slack apps](https://api.slack.com/methods/oauth.access)

New Slack Apps must use the newer `oauth.v2.access` endpoint. I'd like `next-auth` to support that

Slack now has new OAuth endpoints for new Slack Apps
```
AuthURL:  "https://slack.com/oauth/v2/authorize",
TokenURL: "https://slack.com/api/oauth.v2.access",
```

cathykc:
# In this PR
1. Update Slack provider to use updated Slack OAuth endpoints instead of legacy
Fixes #798
- https://api.slack.com/docs/sign-in-with-slack
- The new endpoint returns a response with the token nested inside a `authed_user` object so I added an option (`accessTokenGetter`) in the Slack provider to get the correct entry from the response json. Would imagine other OAuth providers in the future not necessarily returning an `access_token` value in the top level of a response object.
```
{
  ok: true,
  authed_user: { id: 'USER_ID', access_token: 'xoxp-XXX' },
}
```

2. Allow for additional params in authorization URL
- These params can be set in the provider options dict with `additionalAuthorizeParams` key (happy to update documentation if someone points me to where that would be)
- This is useful as a couple OAuth providers (like Slack) have more than just `redirect_uri`, `scope`, and `state` options in their authorization URLs (e.g. `user_scope`). 

Let me know if there's any more information I can add here! Very grateful for this open source project :)

------------

Build failed on lint step so I ran `standard --fix` which edited `src/server/pages/error.js`. No functional change there.


tarzain:
Wow, I've been waiting for this edit for a while - thanks for doing it! 🙌

ndom91:
@cathykc thanks for this update! Do you mind updating the Slack provider docs to mention your new getter and how to use it? Otherwise looks good to me!

cathykc:
@ndom91 Done! Let me know if there's anything else

PR #881, opened 2020-11-25
refactor: update Prisma calls to support 2.12+

kripod:
[Prisma 2.12](https://github.com/prisma/prisma/releases/tag/2.12.0) deprecates the `findOne` method in favor of `findUnique`. I've updated the corresponding adapter to reflect these changes.

Please consider releasing a new minor version of NextAuth.js soon, as the current one has severe breakages due to #559 (fixed by #626 but not yet released).

PR #898, opened 2020-12-02
Fix findOne deprecation warning in Prisma adapter

royletron:
Hey authers,

So Prisma has deprecated `findOne` in favour of `findUnique`, and this PR simply changes it in NextAuth so we don't get the annoying warning messages that this is to be deprecated: https://www.prisma.io/docs/concepts/components/prisma-client/crud#findunique


royletron:
Just realised this is a duplication of #881

PR #903, opened 2020-12-03
Enable login hinting for OAuth services like Google

jdkula:
tl;dr: Allows additional options passed from `signIn` to be used in order to build additional query params that are eventually passed to the OAuth authorization URL.

Some OAuth providers, like Google, allow you to customize the login flow on their side of things by adding additional query parameters. In particular, for Google, if you know the email of the user trying to sign in, you can give them a `login_hint` parameter that pre-fills that email address when logging in, or pre-selects that account if the user is already signed into multiple.

Pull request includes a modification to the google provider that allows the above if `email` is passed in as extra data to the `signIn` function. Also updated the docs with information about this usage. (While doing so, I also discovered that the docs claim that the `params` provider option affects the authorization URL, when it actually affects the access token URL – that was fixed to avoid confusion.)

I'm using a sign-in flow on my site where users are first asked to enter their email before being redirected to their sign-in method of choice – this just helps with the user experience on their end!

If this gets merged I'm happy to update the typings over at DefinitelyTyped.

Thank you all!!

### DIFF
--- a/src/providers/google.js
+++ b/src/providers/google.js
@@ -18,6 +18,11 @@ export default (options) => {
         image: profile.picture
       }
     },
+    authorizationParams: (body) => {
+      return {
+        login_hint: body.email || undefined
+      }
+    },
     ...options
   }
 }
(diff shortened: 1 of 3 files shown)


balazsorban44:
Could this be a duplicate of #823?

jdkula:
I think it addresses the same thing in a slightly different way (this keeps processing as part of the provider vs for all `signIn` calls)

Closing in favor of #823, you've got more progress there! Thanks!

Issue #897, opened 2020-12-01
signIn callback returning before promise resolves


wwsalmon:
**Your question**
I use the `signIn` callback to access MongoDB before letting the user through. My code looks like this:

```
signIn: async (user, account, profile) => {
    try {
        mongoose.connect(process.env.MONGODB_URL, {
            useNewUrlParser: true,
            useUnifiedTopology: true,
            useFindAndModify: false,
        });

        userModel.findOne({ email: user.email }, (err, foundItem) => {
            if (err) return Promise.reject(new Error(err));

            // if user object already exists, return
            if (foundItem) return Promise.resolve(true);

            const urlName = user.name.split(" ").join("-") + "-" + short.generate();

            // otherwise, create new user object
            userModel.create({
                email: user.email,
                name: user.name,
                image: user.image,
                urlName: urlName,
                private: false,
            }, (err, newUser) => {
                if (err) return Promise.reject(new Error(err));
                return Promise.resolve(newUser);
            });
        });
    } catch (e) {
        console.log(e);
        return Promise.reject(new Error(e));
    }
}
```

However, the callback returns true and redirects to the homepage before the promise is even returned. As a minimal reproduction:

```
signIn: async (user, account, profile) => {
     setTimeOut(() => {
		return Promise.resolve(false);
	}, 2000);
}
```

does not stop the user, but rather lets them through. Wrapping everything in a promise doesn't work either:

```
signIn: async (user, account, profile) => {
	return Promise((res, rej) => {
		setTimeOut(() => {
			res(false);
		}
	});
}
```

This request just returns a pending promise and never resolves.

How do I make the callback return, and the page redirect, only after the function has fun and the promise has resolved?

Here's my whole project repo for further context: https://github.com/wwsalmon/updately

**Feedback**
*Documentation refers to searching through [online documentation](https://next-auth.js.org), code comments and issue history. The example project refers to [next-auth-example](https://github.com/iaincollins/next-auth-example).*

* [ ] Found the documentation helpful
* [x] Found documentation but was incomplete
* [ ] Could not find relevant documentation
* [ ] Found the example project helpful
* [x] Did not find the example project helpful

balazsorban44:
Hi there! Since you are using Promises inside `userModel.findOne`, I assume it is also an async method. I think you should actually return `userModel.findOne` in your try block, or if I may suggest, you could try using async/await for better readability like so:

```js
signIn: async (user, account, profile) => {
[...]

PR #915, opened 2020-12-05
docs: Remove unnecessary promises

### LINKED ISSUE #731: Simplify code/examples
I have seen this many places, like [this example code](https://next-auth.js.org/configuration/callbacks#sign-in-callback):

```js
signIn: async (user, account, profile) => {
  const isAllowedToSignIn = true
  if (isAllowedToSignIn) {
    return Promise.resolve(true)
  } else {
    // Return false to display a default error message
    return Promise.resolve(false)
    // You can also Reject this callback with an Error or with a URL:
    // return Promise.reject(new Error('error message')) // Redirect to error page
    // return Promise.reject('/path/to/redirect')        // Redirect to a URL
  }
}
```

When defining an async function, the returned value is automatically resolved, meaning one could refactor the above code. (Notice that throwing an Error would probably also feel more natural):

```js
async signIn(user, account, profile) {
  const isAllowedToSignIn = true
  if (isAllowedToSignIn) {
    return true
  } else {
    // Return false to display a default error message
    return false
    // You can also Reject this callback with an Error or with a URL:
    // throw new Error('error message')) // Redirect to error page
    // return Promise.reject('/path/to/redirect')        // Redirect to a URL
  }
}
```

This may be only a personal preference, but I think it would also be easier to parse the code for newcomers. With the rise of "async/await", maybe there is lesser need of using `Promise`s all-together. Any special reason I am unaware for using `Promise.resolve()` everywhere?

balazsorban44:
There is an open discussion about this already in #731, but after looking at some of the issues and trying to help people, I came across MANY code examples from users using `Promise.resolve()` and `Promise.reject()` where it is not needed.

It is not strange, after all, we currently encourage this through the docs.

The following search yields 72 issues using `Promise.resolve`:
https://github.com/nextauthjs/next-auth/issues?q=is%3Aissue+is%3Aopen+Promise.resolve

I argue that using `Promise` and `async/await` together has a negative effect since new developers might not know the connection between the two and would think that they have to rely on it for things to work. The above issue query proves just that. Furthermore, if someone starts looking up how Promises are handled, they might go down a path where they come across a lot of code using `.then` that will result in callback hell.

Even though the intention was to "self-document", I am certain it had the opposite effect what the original author intended. (#897 got simply resolved by removing `Promise`)

Besides, I don't know of any other library using this "technique" to teach `async/await`.

closes #731


balazsorban44:
@iaincollins please do reconsider your decision on this, to prevent further issues being raised deriving from this easily misunderstandable behavior in our examples.

iaincollins (approved):
Thanks for talking though this!

foaadnami:
@balazsorban44 is on fire this new years eve!

Happy new year everyone

balazsorban44:
Thanks! 🎇🥂

Issue #825, opened 2020-11-03
502: BAD_GATEWAY from Vercel


dasveloper:
**Your question**
I'm getting sent to a Vercel error page showing "502: BAD_GATEWAY" on the callback from Google auth and Auth0. Both are working fine locally. I've checked the env variables multiple times to ensure they're correct. I tired creating new Google OAuth credentials and Auth0 credentials, a new Vercel app, a new Git repository, and a new Mongo Atlas DB, all of which still work locally but not on Vercel. 

```The error logs are correctly reading out all of my Google auth info followed by the error:
[next-auth][debug][typeorm_get_user_by_provider_account_id] [ 'google', '115431037803995017525' ]
[next-auth][debug][typeorm_get_user_by_provider_account_id] [ 'google', '115431037803995017525' ]
2020-11-03T01:15:57.090Z	c966b832-993e-4ed3-afe8-c18c37b39d37	ERROR	Unhandled Promise Rejection 	{"errorType":"Runtime.UnhandledPromiseRejection","errorMessage":"","promise":{},"stack":["Runtime.UnhandledPromiseRejection","    at process.<anonymous> (/var/runtime/index.js:35:15)","    at process.emit (events.js:327:22)","    at processPromiseRejections (internal/process/promises.js:209:33)","    at processTicksAndRejections (internal/process/task_queues.js:98:32)"]}
Unknown application error occurred
```
I've run out of ways to troubleshoot this error, it seems as if the callback URL `https://{domain}.com/api/auth/callback/google` just doesn't exist.

**What are you trying to do**
I'm trying to just get authentication to work, similar to how it is on my local copy.
**Feedback**
*Documentation refers to searching through [online documentation](https://next-auth.js.org), code comments and issue history. The example project refers to [next-auth-example](https://github.com/iaincollins/next-auth-example).*

* [ ] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [x] Could not find relevant documentation
* [ ] Found the example project helpful
* [ ] Did not find the example project helpful

iaincollins:
> I've run out of ways to troubleshoot this error, it seems as if the callback URL https://{domain}.com/api/auth/callback/google just doesn't exist.

Like https://next-auth-example.now.sh/api/auth/callback/google I'd expected to return https://next-auth-example.now.sh/api/auth/error?error=oAuthCallback if visited in the browser as the request will be incomplete.

This should return all configured providers and their callback URLs:
[...]

PR #924, opened 2020-12-07
Updated docs to help prevent signIn error bug

abcd-ca:
Documents how to avoid a signIn callback [bug where Vercel displays a 502](https://github.com/nextauthjs/next-auth/issues/825) error and doesn't show your intended error page. Vercel shows a 502 when you have an uncaught exception.

### DIFF
--- a/www/docs/configuration/callbacks.md
+++ b/www/docs/configuration/callbacks.md
@@ -70,6 +70,29 @@ callbacks: {
 
 * When using the **Credentials Provider** the `user` object is the response returned from the `authorization` callback and the `profile` object is the raw body of the `HTTP POST` submission.
 
+:::warning
+Setting your callback as `async` and using try/catch and doing `return Promise.reject(error)` from within the `catch` block will raise a `UnhandledPromiseRejectionWarning`.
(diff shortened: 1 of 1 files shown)


balazsorban44:
I have some thoughts about using async/await vs Promises. I'll get back to you with feedback on this.

abcd-ca:
I should add that you do not need to deploy to Vercel to test this issue. All you need to do is use localhost and look for the `UnhandledPromiseRejectionWarning` message. As long as you don't get that from your `signIn` callback you wont get the 502 on Vercel.

balazsorban44:
I'm on my phone, but here is  what I think.

When using async/await and try/catch, you really don't need to return Promise.reject anymore, as the catch block is your reject case. If you return a Promise.reject, you will create a new Promise that is indeed not caught anywhere. (you would need a new try..catch inside your catch)

Putting async/await together with Promise complicates things in my opinion. Promise were kind of a jump start for async/await but in modern js, I would argue that you rarely have a usecase for Promise. Maybe if you convert synchronous calls to async ones.

PR #925, opened 2020-12-07
docs: added mssql indexes in docs, fixed typo

jrvidotti:


www/docs/schemas/mssql.md
```diff
@@ -58,10 +58,31 @@ CREATE TABLE verification_requests
...
+
+CREATE UNIQUE INDEX email
+  ON users(email);
+
+CREATE UNIQUE INDEX token
+  ON verification_requests(token);
+```
 
-When using NextAuth.js with SQL Server fir the first time, run NextAuth.js once against your database with `?syncronize=true` on the connection string and export the schema that is created.
+When using NextAuth.js with SQL Server fir the first time, run NextAuth.js once against your database with `?synchronize=true` on the connection string and export the schema that is created.
```
balazsorban44, line 87:
```suggestion
When using NextAuth.js with SQL Server for the first time, run NextAuth.js once against your database with `?synchronize=true` on the connection string and export the schema that is created.
```

While at it... 😅 Thanks

balazsorban44 (changes requested):
Caught another typo! Could you fix? :slightly_smiling_face:

balazsorban44 (approved):

Issue #311, opened 2020-06-23
Support multiple SAML 2 Identity Providers (IDP)


nick-myers-dt:
**Summary of proposed feature**
Allow users of a SAML 2.0 Identity Provider (IDP) to sign in using next-auth

**Purpose of proposed feature**
In the education sector, many organisations use a SAML 2 IDP to authenticate their users (e.g. students) and seek to use third party services which allow their users to sign in to the third party service using their own IDP.    This means that businesses that sell into the education sector must provide sign in capabilities using SAML.

**Detail about proposed feature**
next-auth would allow configuration of email domains which would redirect users who enter an email address in that domain to the relevant IDP.  This is referred to as Identity Provider Discovery.

next-auth would allow configuration of metadata to be exchanged during authentication process.

Essentially, next-auth would allow itself to be set-up as a SAML 2 Service Provider (SP)

**Potential problems**
SAML 2 is pretty well established as a standard, I don't perceive any possible problems.  

**Describe any alternatives you've considered**
A clear and concise description of any alternative options you've considered.

**Additional context**
[Understanding SAML on Okta.com](https://developer.okta.com/docs/concepts/saml/)
[Identity Provider Discovery on Shibboleth.net](https://www.shibboleth.net/index/intermediate/)
[SAML Service Provider on Auth0.com](https://auth0.com/learn/saml-service-provider/)

*Please indicate if you are willing and able to help implement the proposed feature.*
I don't think I have the technical expertise to do so!

iaincollins:
Thanks for the feature request and for the relevant links.

Oh SAML gives me horrible flashbacks to supporting SSO with it years ago. 🙀

Given my experience with it and I as have no need to work on this, it's not something I'd work on for free, but it's a reasonable feature request and I'd happy accept adding this as a pull request.

Maybe someone else who wants it will see this and pick it up.

If anyone would like integrate with SAML service in the meantime, you should be able to do it using a combination of a custom sign in page / Credentials Provider / `signin()` callback, without having to actually modify `next-auth` to add native support for SAML.

You could possibly even create a Provider based on the Credentials Provider that was specific to supporting sign in with SAML. 

You could also use integration with Okta and/or Auth0 to handle this - while they are also in many respects alternatives to NextAuth.js, NextAuth.js also works great with both of these services.

nick-myers-dt:
[...]

PR #927, opened 2020-12-07
Add Osso as open source SAML provider

sbauch:
Per https://github.com/nextauthjs/next-auth/issues/311 this adds a Provider [Osso](ossoapp.com) - we're like an open source Auth0, but exclusively for SAML 2.0. Osso provides a UI and persistence for configuring SAML identity providers, and wraps a SAML auth process inside in Oauth 2.0 auth code grant flow, which allows us to easily hook in to projects like this that implement Oauth.

Still TODO is I need to make a decision about how to surface an Identity Provider for the user who wishes to sign in.

Auth0 does this on their hosted sign in page - I think we will probably implement similar.

In previous contexts, we ask the developer integrating Osso to ascertain either the user's email or corporate domain, and use that to look up their IDP(s) and redirect them to the right place. I don't see a way of achieving that here - I need an email input, and then for that value to be passed as a query param to the authorize endpoint in our OAuth flow.

So that's hardcoded for now to one of my emails.

I figured before we went and made those changes I'd at least open this draft PR to see if this would be welcome, and to what extent we might be able to solve the issue of getting the user's email prior to kicking off the oauth flow, so this is draft for now.

Edit: sorry, not really sure whats up with git history here?

### DIFF
--- a/src/providers/index.js
+++ b/src/providers/index.js
@@ -17,6 +17,7 @@ import IdentityServer4 from './identity-server4'
 import LinkedIn from './linkedin'
 import Mixer from './mixer'
 import Okta from './okta'
+import Osso from './osso'
 import Slack from './slack'
 import Spotify from './spotify'
 import Twitch from './twitch'
@@ -43,6 +44,7 @@ export default {
   LinkedIn,
   Mixer,
   Okta,
+  Osso,
   Slack,
   Spotify,
   Twitter,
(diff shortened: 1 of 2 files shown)


sbauch:
Ok sorry, I had used the canary branch as the base. gonna close and reopen a clean one 🙈

balazsorban44:
Hi @sbauch The canary should be your base.

PR #928, opened 2020-12-07
Add Osso as open source SAML provider 

sbauch:
Per #311 this adds a Provider Osso - we're like an open source Auth0, but exclusively for SAML 2.0.

Osso provides a UI, persistence and documentation for configuring SAML identity providers, and wraps a SAML auth process inside in Oauth 2.0 auth code grant flow, which allows us to easily hook in to projects like this that implement Oauth. You can check out a demo Osso instance at https://demo.ossoapp.com and generally learn more about osso at https://ossoapp.com

In order to show how this all works, I've set up the following:

• a staging instance of Osso

this includes some work to add a "[hosted login page](https://staging.ossoapp.io/oauth/authorize?response_type=code&redirect_uri=https%3A%2F%2Fosso-next-auth-dw7gmmt27.vercel.app%2Fapi%2Fauth%2Fcallback%2Fosso&scope=&state=896484f73a7f60d84a3bb69aa288c9a28166eb92e07b78375ec811134e2e1e1b&client_id=960ba9f7e7e9b417b183f1e6960d6aca)" like Auth0 has

• a next app - https://osso-next-auth-mcwe2rmtw.vercel.app/

this is based on this project's example, but includes a module transpilation plugin to build a version of this package with support for Osso as a provider. it uses the review Osso app above via ENV vars for `OSSO_DOMAIN`, `OSSO_ID` and `OSSO_SECRET`. repo: https://github.com/enterprise-oss/osso-next-auth-example

• a fake Identity Provider

It takes a bit of work to configure an IDP for SAML login, so we've developed a little ruby app that acts as a fake IDP. it takes a SAMLRequest, allows a user to "log in" using any values, and crafts a SAMLResponse, sending the user back to osso and eventually the next app on vercel. repo: https://github.com/enterprise-oss/sinatra-ruby-idp

So you can test this end to end: 

1. visit https://osso-next-auth-dw7gmmt27.vercel.app/ and sign in with SAML SSO
2. enter anything@example.com on the next screen (this is our "hosted login page" similar to the Auth0 example)
3. enter any values for email and password on our fake idp
4. you'll be redirected back to the vercel app, and the value you entered for email on step 3 should be your "signed in as"

I'll leave some code specific comments inline. 

Thanks gang, and apologies again for my nonsense with the canary branch etc. Would be really excited to get support for Osso into such a great project!


balazsorban44:
The intention IS to use canary as your base. This we can merge things faster and get them out to npm with a `canary` tag, while testing and docs will be polished before it is pushed to `main` and released as `latest` on npm.

sbauch:
thanks... wasn't clear in the contribution guidelines that was the intention

balazsorban44:
Sorry about that! We are currently transitioning to a new structure we started a few days ago. Hopefully, this will allow a faster release cycle for those who don't mind some bugs here and there. Once the details are set, we will probably update the CONTRIBUTING document.

sbauch:
no problem! sorry bout it all, hate to create more work for maintainers. its a clever approach now that i understand a bit!

anyhow, it feels like too big a change to find a way to add another query param to my authorize url based on user input? so we'll get cracking on more of a hosted sign in flow so I can at least demo this in action. It's a bit tough because in order to do a full SAML login you'll need an Osso instance (we have one at demo.ossoapp.com) and to have configured a SAML provider (Okta has a good developer account offering) in it, and assigned users, etc.

src/providers/osso.js
```diff
@@ -0,0 +1,20 @@
+export default (options) => {
+  return {
+    id: 'osso',
+    name: 'SAML SSO',
```
sbauch, line 4: can i read this from options? seems like the best approach. we want to be mroe white labelled, so "SAML SSO" feels most appropriate, but i'd want the implementer to be able to decide on this label.

does this work?

```suggestion
    name: options.label || 'SAML SSO',
```

src/providers/osso.js
```diff
@@ -0,0 +1,20 @@
...
+    version: '2.0',
+    params: { grant_type: 'authorization_code' },
+    accessTokenUrl: `https://${options.domain}/oauth/token`,
+    authorizationUrl: `https://${options.domain}/oauth/authorize?response_type=code`,
+    profileUrl: `https://${options.domain}/oauth/me`,
+    profile: (profile) => {
+      return {
+        id: profile.id,
+        name: profile.name,
+        email: profile.email
```
sbauch, line 15: we currently offer a minimal profile, but could envision adding more attributes as needed. can we `...spread` the profile object? are any of these required?

sbauch:
Our vercel teams trial expired, so the demo app for this is now at https://nextjs-demo.ossoapp.com/

lmk what i can do to push this along! more than happy to write docs etc, not sure what the expectation is for a canary merge

🙏

src/providers/osso.js
iaincollins: Great question! The docs on this could use improving.

To work with a database, a `id` is required from a profile, it is stored as `accountId`, but name and email are optional as neither are guaranteed from a provider.

From memory:

* If using a database then the `name` and `email` will be used when creating a new user, but if the account exists and is associated with a user, then it will use the `id` to look up the account, then from there look up the `user` associated with that account and return them for the user info when `getSession()` is called.
*  If using JSON Web Tokens without a database, then it doesn't matter what these values are, but if blank they will be blank in the JSON Web Token / Session.
* The full profile returned from a provider is exposed in various callbacks. It is used here for very limited purposes. Returning extra fields here has no impact.

src/providers/osso.js
iaincollins: I think best just to do `name: 'SAML SSO'` as matches convention for the other providers.

If `name` is specified in options then it will automatically override the value. 👍

iaincollins:
@sbauch I just wanted to say have checked out the demo site, thanks for setting that up. Osso looks really cool!

Having a way to integrate with SAML auth directly would be a lot of work for something like NextAuth.js, so having an solution like this is wonderful, especially an open source product.

It would be lovely to chat sometime!

sbauch:
thanks @iaincollins! glad to hear you think this could be a good addition - it's admittedly a little weird to add support for a provider like this where the library users have to deploy or pay for an Osso instance, but since you already support a provider like Auth0 it seems reasonable to us.

I'm planning on responding to feedback here and writing some docs this week, but would be happy to connect. to avoid putting my email on github, maybe you could drop a note in our chat widget at ossoapp.com and we can take it offline from there?

src/providers/osso.js
sbauch: neat!

src/providers/osso.js
```diff
@@ -0,0 +1,20 @@
...
+    type: 'oauth',
+    version: '2.0',
+    params: { grant_type: 'authorization_code' },
+    accessTokenUrl: `https://${options.domain}/oauth/token`,
+    authorizationUrl: `https://${options.domain}/oauth/authorize?response_type=code`,
+    profileUrl: `https://${options.domain}/oauth/me`,
+    profile: (profile) => {
+      return {
+        id: profile.id,
+        name: profile.name || profile.email,
```
sbauch, line 14: SAML spec is a little odd wrt NAMEID format things. Osso's docs don't instruct IDP admins to include a "name" attribute, but we do sometimes get them back.

I'm thinking here that if the database or user model depends on there being a name, we should at least fall back to email, which is guaranteed to be available.

sbauch:
Added some docs and got up to date here 📖 👩‍⚕️ 👨‍⚕️

sbauch:
added docs for this to our docs site (borrowing a bit from your docs) - https://ossoapp.com/docs/consume/nextjs

now that the big release and security patch is out, chances we could take another look here?

balazsorban44:
Hi there, and thanks for your patience on this! I think we are ready to approve this, only one thing left... I am going to close this PR, because we changed our release strategy, and this PR should be opened against `main` instead. I could ask you to simply rebase, but we rely strictly on the commits to generate release notes, so after my findings, the easiest thing is just to re-open a PR against another branch. I am the one who asked you to change to `canary` in the first place, so terribly sorry for that! 🙈 If you don't wish to set up a new PR, I can happily copy your changes and make you a co-author in a new PR as well!

PR #941, opened 2020-12-09
feat: add PKCE support

### LINKED ISSUE #502: invalid_grant when using IdentityServer4
First of all thank you for this awesome project! I've using months ago in my project with the previous version. It is great to have sql database option further than mongoDB. 
I having trouble with this new version to get access_token, first part of the flow goes fine and can login but I'm getting  
**[next-auth][error][callback_oauth_error] { statusCode: 400, data: '{"error":"invalid_grant"}' }** after redirect back.

I just cloned the example project and add an IdentityServer4 provider and connect to the same idsvr4 that is now working with previous version using a passport-oauth2 strategy. 

Apparently _code_ is lose at some point according to idsvr4
`fail: IdentityServer4.Validation.TokenRequestValidator[0]
      Authorization code is missing, {
        "ClientId": "clientid",
        "ClientName": "clientname",
        "GrantType": "authorization_code",
        "Raw": {
          "grant_type": "authorization_code",
          "code": "",
          "client_id": "clientid",
          "client_secret": "***REDACTED***",
          "redirect_uri": "http://localhost:3000/api/auth/callback/identity-server4"
        }
      }`

I tried using a custom OAuth provider but get same error.

### LINKED ISSUE #685: Is "Authorization Code + PKCE" Supported?
**Your question**
Does next-auth support "Authorization Code + PKCE"?

**What are you trying to do**
Trying to add a new provider using the "Authorization Code + PKCE" flow (Akamai, using a Public Client).  I was able to generate a code_challenge and send that into the auth endpoint.  I couldn't figure out how to send in the code_verifier though.  I tried in the provider.params as it looks like that is what gets added to the postData, but its throwing an error.

* [X] Could not find relevant documentation

balazsorban44:
**What**:

Adding support for [Proof Key for Code Exchange by OAuth Public Clients](https://tools.ietf.org/html/rfc7636)

**Why**:

Many popular Identity Providers require us to support this flow: https://auth0.com/docs/flows/authorization-code-flow-with-proof-key-for-code-exchange-pkce

**How**:

A new option to providers:

```js
  protection: "pkce"
```

If this value is found, `next-auth` will create a `code_challenge` - `code_verifier` pair, which then will respectively be passed to the OAuth2 `/authorize` and `/token` endpoints. The `code_verifier` is saved in a cookie before redirecting to `/authorize`, and then read and removed before calling the `/token` endpoint.

**Checklist**:

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged
      

NOTE: This is highly experimental, and not sure yet if creating a single challenge-verifier pair for the entire app is good/secure enough. I just wanted to have this some place, so I can start experimenting.

Closes #685, closes #502 (potentially others as well, please link to this issue, if you stumble upon other open, related issues.)

Useful sources:
YouTube: https://www.youtube.com/watch?v=Gtbm5Fut-j8
Spec: https://tools.ietf.org/html/rfc7636
Auth0 article: https://auth0.com/docs/flows/authorization-code-flow-with-proof-key-for-code-exchange-pkce

Maybe needed, if we need to create the code in the browser(?):
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle


src/server/index.js
```diff
@@ -217,6 +218,8 @@ async function NextAuth (req, res, userSuppliedOptions) {
       redirect
     }
 
+    const pkce = options.providers[provider]?.pkce ? pkceChallenge(64) : undefined
```
balazsorban44, line 221: length of `64` was arbitrarily chosen here. The [requirement by the spec](https://tools.ietf.org/html/rfc7636#section-4.1) is 43-128 characters

src/server/lib/signin/oauth.js
```diff
@@ -24,14 +24,19 @@ export default (provider, csrfToken, callback, authParams) => {
...
-      const baseUrl = `${parseUrl.origin}${parseUrl.pathname}?`
-      url = url.replace(baseUrl, provider.authorizationUrl + '&')
+      const providerParams = new URLSearchParams(provider.authorizationUrl.split('?')[1])
+      const newParams = { ...url.searchParams, ...providerParams }
+      url = new URL(newParams.toString(), url)
+    }
+
+    if (provider.pkce) {
+      url.searchParams.append('code_challenge', codeChallenge)
+      url.searchParams.append('code_challenge_method', 'S256')
```
balazsorban44, line 34: See the [spec](https://tools.ietf.org/html/rfc7636#section-4.3)

src/server/lib/oauth/callback.js
```diff
@@ -187,13 +188,17 @@ async function _getProfile (error, profileData, accessToken, refreshToken, provi
...
+async function _getOAuthAccessToken (code, provider, callback, codeVerifier) {
   const url = provider.accessTokenUrl
   const setGetAccessTokenAuthHeader = (provider.setGetAccessTokenAuthHeader !== null) ? provider.setGetAccessTokenAuthHeader : true
-  const params = { ...provider.params } || {}
+  const params = new URLSearchParams({ ...provider.params } || {})
   const headers = { ...provider.headers } || {}
   const codeParam = (params.grant_type === 'refresh_token') ? 'refresh_token' : 'code'
 
+  if (provider.pkce) {
+    params.code_verifier = codeVerifier
```
balazsorban44, line 199: See the [spec](https://tools.ietf.org/html/rfc7636#section-4.5)

package.json
```diff
@@ -50,11 +50,12 @@
     "jwt-decode": "^2.2.0",
     "nodemailer": "^6.4.16",
     "oauth": "^0.9.15",
+    "pkce-challenge": "^2.1.0",
```
balazsorban44, line 53: This can probably be removed and we can create our own, added to be able to get started a bit faster

package.json
```diff
@@ -50,11 +50,12 @@
...
     "preact": "^10.4.1",
     "preact-render-to-string": "^5.1.7",
     "querystring": "^0.2.0",
     "require_optional": "^1.0.1",
-    "typeorm": "^0.2.24"
+    "typeorm": "^0.2.29"
```
balazsorban44, line 58: Fixes an issue with one of its dependencies, Highlight.js version 9 (which is not supported anymore). Constantly got an annoying warning in the terminal.

src/server/index.js
mikecabana: In terms of saving the challenge and verifier, it could be saved in local storage (maybe session) to be cleaned up after a successful log in.

balazsorban44: I am not sure we would like to store anything sensitive this way, cause it is perceptible with potentially malicious code on the client. If I understand it correctly, `sessionStorage` would be a better option in this cause, although I am still not sure it is good enough. Going to discuss it with @iaincollins hopefully this weekend, I hope he has some insights.

mikecabana: No problem! I inspired myself from [oidc-client-js](https://github.com/IdentityModel/oidc-client-js) which I'm currently using in a production spa. They use `localStorage` for the code_verifier and `sessionStorage` for the tokens. 

Via the spec every new authorization flow should use a new `code_verifier` so this could be why they've opted to go this route. I'm not really sure either what kind of security concerns this method raises.

In any case I've set up a simple sample to generate the `code_verifier` and `code_chalenge` in the browser. I don't think it's perfect but it can help to get you started. https://gist.github.com/mikecabana/9cae8b447afd2a36da5d38b94bfc2565

balazsorban44: In theory we wouldn't actually need PKCE to be secure I think, since our secrets are all server-side and never accessed by the client. So this is why I went with my implementation. Although there might be something important I leave out. We will have a talk about this soon.

balazsorban44:
https://stackoverflow.com/a/56873098/5364135 it might just be OK to consider our server app as one single client, and we don't need to generate separate challenges for each user

balazsorban44:
Added some changes. The code_verifier is now saved in a cookie, to be able to create a verifier/challenge pair per user.

Also, the previous method wouldn't have worked anyway, because each invocation would have created a new verifier/challenge pair, and the signin flow would have failed.

This still needs some re-thinking, just sharing my implementation in hope that someone may come with a good solution.

The way I see it, we have three alternatives:


1. generate a deterministic code_challenge/code_verifier at the start from the `secret` provided by the user, and save it in `req.options`
2. generate a code_challenge/code_verifier from the `secret` provided by the user at signin and save it in a cookie (encrypted)
3. generate a code_challenge/code_verifier client-side, store the code_verifier in a cookie, and provide it to `/api/auth/callback` when needed

thulstrup:
@balazsorban44 The beta version of nextjs-auth0 supports PKCE. Maybe you can use that to validate your implementation?
https://github.com/auth0/nextjs-auth0/tree/beta

balazsorban44:
Thanks! I'll definitely check!

I also checked the docs of `openid-client` which we might turn to later on (#)

and there they say this:

> store the code_verifier in your framework's session mechanism, if it is a cookie based solution // it should be httpOnly (not readable by javascript) and encrypted.

balazsorban44:
@thulstrup thanks again, your info was very useful! I went with option 2 (`generate a code_challenge/code_verifier from the secret provided by the user at signin and save it in a cookie (encrypted)`), as someone at `nextjs-auth0` confirmed they also saved the `code_verifier` in an encrypted cookie.

What I am starting to settle on, that we should introduce a new Provider option called `protection` (maybe a better name?), that can take either `pkce` or `state` as a value. Current providers that support PKCE can use the features added in this PR, while providers only supporting `state` can use the already present features. If a Provider does not support either `pkce` or `state`, then no extra checks will be added in the login flow. (As far as I know, currently only the Apple provider is opted-out of  the`state` check)

package.json
balazsorban44: `openid-client` has this built-in, we can utilize it when/if we finish #1105

src/server/index.js
iaincollins: @balazsorban44 FWIW for the record this is exactly my current understanding as well.

thulstrup:
@balazsorban44 That is awesome! I will test later today.

thulstrup:
@balazsorban44 I had to fix one tiny error to get it working:
`import logger from 'src/lib/logger'` should be `import logger from '../../lib/logger'`

After that change I was able to log in using PKCE 🎉

balazsorban44:
I see! Will fix immediately.

kierancrown:
I might be a little dumb here. But where do I add this within the config? I can't see anything in the type files.

aakash14goplani:
Can I use code-verifier from cookie as is or it needs some decryption?

Issue #844, opened 2020-11-11
SSR + useSession causes an extra network call and render cycle


joeltg:
**Describe the bug**

If you use `getSession` to return a `session` prop from `getServerSideProps` and then use `useSession` in the component, the `useSession` hook will cause the page to re-render twice (three times total) even though it returns the valid session object all three times.

In addition, the hook will still cause the client to make a network request to the server after its first render, even though it already has the session object.

**Steps to reproduce**

You can observe this behavior by adding a console log to the render method of /server in [next-auth-example](https://github.com/nextauthjs/next-auth-example/) and watching network requests that it makes.

**Expected behavior**

**As far as I can tell**, the expected behavior would go like this for a server-side rendered page:

1. the incoming request causes nextjs to to call `getServerSideProps`, which calls `getSession` and returns the session in `pageProps.session` which is passed into the context provider in _app.tsx
2. nextjs renders the page on the server, where `useSession` returns `[session, false]` since it is able to retrieve the session object from the Provider context
3. the client hydrates and renders for the first time, and `useSession` again returns `[session, false]` since it is able to retrieve the session object from the Provider context
4. no additional renders or network requests are triggered

I could definitely be mistaken about what's expected - but my understanding from the [Provider documentation](https://next-auth.js.org/getting-started/client#provider) is that there should _not_ be an extra network request. 

**Screenshots or error logs**

What instead happens is that the client renders three times:

1. on the first render, useSession returns `[session, true]` (where `session` is the full valid session object)
2. the client dispatches a network request to `/session`
3. the client re-renders, useSession returns `[session, true]` (the same as the first render)
4. the network request resolves
5. the client re-renders again, useSession returns `[session, false]`

<img width="878" alt="E16F1738-F7C5-4250-8BEA-31C85B6EFB70-658-000107070BB4C866" src="https://user-images.githubusercontent.com/8454646/98854863-20f33180-2429-11eb-8a58-c66ab4e116f8.png">

<img width="1156" alt="AAD5D18A-7144-4A4F-B89C-DD71D87F0208-658-00010708B7239CDB" src="https://user-images.githubusercontent.com/8454646/98854871-251f4f00-2429-11eb-9d9a-3770605772ce.png">

As another example, if you block network requests to `/session`, then the page first renders a logged-in view (once), and then a logged-out view (twice):

[...]

PR #943, opened 2020-12-10
refactor: improve client code

balazsorban44:
**What**:
Improving the code exported from `next-auth/client`

**Why**:

I found some interesting stuff when looking at an issue. https://github.com/nextauthjs/next-auth/issues/844#issuecomment-742126203

While I was fixing that, I thought I could clean up a bit in the React section.

**How**:

Removed some of the code to make it leaner, moved parts near each other where it made sense. (All React related code is near each other, waiting to be abstracted away to its own file. 😉)

I also added a new export, which is just a re-export of `Provider`. I feel `SessionProvider` is a more descriptive name, which would help not to confuse with the Auth Providers. I think `NextAuthProvider` would also be acceptable.

I could not understand why `useSession()` would take `session`, so I removed it. I don't think it would be breaking, looking at the code (and the fact that `useEffect()` ran way too often without a dependency array), it had no effect at all.

**Checklist**:

- [ ] Documentation (will add docs to `SessionProvider` if someone thinks it's a good idea to add)
- [ ] Tests
- [ ] Ready to be merged


src/client/index.js
```diff
@@ -11,7 +11,7 @@
 // We use HTTP POST requests with CSRF Tokens to protect against CSRF attacks.
 
 /* global fetch:false */
-import { useState, useEffect, useContext, createContext, createElement } from 'react'
```
balazsorban44, line 14: With this, we can prevent future line changes in commits. (Idea: https://epicreact.dev/importing-react-through-the-ages/)

src/client/index.js
```diff
@@ -326,14 +236,105 @@ const _sendMessage = (message) => {
...
+      {children}
+    </SessionContext.Provider>
+  )
+}
+
 export default {
+  // React
+  useSession,
+  Provider: SessionProvider,
+  SessionProvider,
```
balazsorban44, line 331: Grouped React related exports

src/client/index.js
```diff
@@ -326,14 +236,105 @@ const _sendMessage = (message) => {
...
+const SessionContext = React.createContext()
+
+/** React hook to get the session */
+function useSession () {
+  return React.useContext(SessionContext)
+}
+
+/** Provider to wrap the app in to make session data available globally */
+function SessionProvider ({ children, session, options }) {
+  setOptions(options)
```
balazsorban44, line 251: Could/Should this be:
```suggestion
  React.useEffect(() => {setOptions(options)}, [])
```

Notice the empty dependency array. Do we expect options to change here?

src/client/index.js
```diff
@@ -326,14 +236,105 @@ const _sendMessage = (message) => {
...
+    __NEXTAUTH._getSession = _getSession
+
+    _getSession()
+  }, [hasSession])
+
+  return React.createElement(
+    SessionContext.Provider,
+    { value: [data, loading] },
+    children
+  )
```
balazsorban44, line 324: Looks like `React.createElement` is safe, even with future versions:
> Note
The functions inside react/jsx-runtime and react/jsx-dev-runtime must only be used by the compiler transform. If you need to manually create elements in your code, you should keep using React.createElement. It will continue to work and is not going away.

https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

src/client/index.js
```diff
@@ -326,14 +236,105 @@ const _sendMessage = (message) => {
...
+  React.useEffect(() => {
+    if (hasSession) { // Don't fetch session, if it is already in state
+      return
+    }
+    const _getSession = async ({ event = null } = {}) => {
+      try {
+        const triggredByEvent = (event !== null)
+        const triggeredByStorageEvent = !!((event && event === 'storage'))
+
+        const clientMaxAge = __NEXTAUTH.clientMaxAge
```
amrnt, line 265: missing double underscore? `__NEXTAUTH__`

PR #513, opened 2020-07-30
add /api/auth/tokens endpoint to return oauth tokens

tomvoss:
A method to retrieve OAuth tokens from the database was needed in order to use them with third-party APIs as Bearer tokens.

As @iaincollins requested in the [original pull request](https://github.com/nextauthjs/next-auth/pull/425#issuecomment-665584387), I have opened a new pull request against `main` on the official organization account for next-auth.

PR #951, opened 2020-12-11
feat: /tokens endpoint

balazsorban44:
**What**:
The goal of this PR is to create a new `/tokens` endpoint, that the users can call, and if they have a valid session, the `access_token` will be returned. In the case of OpenID Connect (OIDC) compliant providers, we can also try to keep that token fresh, by doing using a [refresh token](https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them)

We can also add a new hook `useAccessToken`, to provide users a nice API to retrieve an access token, without needing to configure the `jwt` and `session` callbacks

**Why**:
One of the most popular usages of the `jwt` and `session` callbacks are to put an access_token in the `session` cookie, so the users can do client-side calls to protected APIs.

**How**:

~This will be a new `/tokens` endpoint. The `access_token` will be saved into its own cookie by default (along with the `refresh_token` and expiry date), to mitigate the 4096bytes limit common browser use. This way, the session cookie will free up enough space for the users to save even more stuff there about the user, if they wish.~

As of v4, this is irrelevant as we implement cookie chunking (if the session is larger than 4096 bytes, we split it into multiple cookies and merge them when reading.)

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

> NOTE ⚠ This is highly experimental add this moment, the tokens are not even encrypted in the cookies, which could be a major security issue. DO NOT USE THIS IN PRODUCTION.

This is a continuation of #513 

In addition to the above mentioned `access_token` and `refresh_token` cookie, it will also save the `id_token` in a separate cookie, that can be used when logging out of OIDC providers. This will make it possible to do federated logout


src/server/routes/callback.js
```diff
@@ -286,3 +300,56 @@ export default async (req, res, options, done) => {
...
+      accessTokenExpires = expires
+    }
+    // Save the access_token and refresh_token together
+    const accessToken = {
+      accessToken: account.accessToken,
+      refreshToken: account.refreshToken
+    }
+
+    // TODO: Encrypt with secret
+    const encryptedAccessToken = JSON.stringify(accessToken)
```
balazsorban44, line 335: Needs encryption

src/server/routes/callback.js
```diff
@@ -286,3 +300,56 @@ export default async (req, res, options, done) => {
...
+      res,
+      cookies.accessToken.name,
+      encryptedAccessToken,
+      { expires: accessTokenExpires.toISOString(), ...cookies.accessToken.options }
+    )
+  }
+
+  if (account.idToken) {
+    // TODO: Encrypt with secret
+    const encryptedIdToken = account.idToken
```
balazsorban44, line 347: Needs encryption

src/server/routes/tokens.js
```diff
@@ -0,0 +1,163 @@
...
+    if (useJwtSession) {
+      const provider = options.providers[providerName]
+      if (provider?.type !== 'oauth') {
+        res.json(response)
+        logger.error('INVALID_TOKEN_PROVIDER', 'Invalid ')
+        return done()
+      }
+      // TODO: decrypt with options.secret
+      /** @type {import('../cookies').AccessToken} */
+      let accessToken = req[cookies.accessToken.name]
```
balazsorban44, line 46: Needs decryption

src/server/routes/tokens.js
```diff
@@ -0,0 +1,163 @@
...
+    },
+    method: 'POST'
+  })
+
+  const refreshedToken = await response.json()
+
+  if (!response.ok) {
+    throw new Error({ refreshToken: refreshedToken })
+  }
+
```
balazsorban44, line 155: Should also update the cookie with these new tokens

iaincollins:
FYI there is already an open PR regarding this #513

I don't recall the status, IIRC the direction looked good but might have had some outstanding issues.

balazsorban44:
Yes, I know, the original author said he would probably not be able to work in it for a while, and if someone wants to jump in and help can do that. I copied the code from there to start.

MatteoGauthier:
Hope it will be merged, very useful features, thanks a lot !

balazsorban44:
@MatteoGauthier it probably will be merged in some form at some point, there are some rough edges still we have to work out. 🙂 you can try forking it and make it work if you have the capacity/time

balazsorban44:
Going to come back to this at a later point. A lot of bigger changes are being made currently, that need to be done first.

frrr1:
Hi @balazsorban44! 

I appreciate your work on this topic very much, your snippets are of great value. We are currently facing issues when logging out (using oauth/oidc with keycloak 12.0.2, next 10.2.0 and next-auth 3.23.1). Considering your comment "In addition to the above mentioned access_token and refresh_token cookie, it will also save the id_token in a separate cookie, that can be used when logging out of OIDC providers. This will make it possible to do federated logout", could you advise any workaround until the new features will be merged?

PR #959, opened 2020-12-15
refactor: code base improvements

balazsorban44:
**What**:

General, non user-facing code improvements in the code base.

1. moved some properties to `req` and `res`, so we need to only pass those two further down. This makes it way easier to follow what needs to be sent where.
Most notably:
- `options` is now `req.options`
- redirect is `res.redirect` and
- `done()` is `res.end()`

2. using the native `URL` constructor instead of concatenating strings and doing regex. Makes `baseUrl` globally available as a function, so we don't have to pass it around manually.

3. refactor some internal callbacks to Promises. Much easier to read and work with
4. remove `jwt-decode` dependency in favour of `jsonwebtoken` which is already added

**Why**:

Almost all functions were defined as anonymous arrow functions, which made it hard to follow the flow when those were imported in other modules. Even then, many modules have the same name, but fortunately commented at the top.

Naming the functions before export, and converting the `//` comments to multi line `/** */` right above the exported functions play nicely with IntelliSense, and adds some extra help when using those functions.

Example:
image

**How**:

None of these changes the user-facing API, it is only internal for increased readibility and DX while working on `next-auth`

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged


src/server/index.js
```diff
@@ -30,18 +31,19 @@ async function NextAuth (req, res, userSuppliedOptions) {
     // This is passed to all methods that handle responses, and must be called
     // when they are complete so that the serverless function knows when it is
     // safe to return and that no more data will be sent.
-    const done = resolve
+    // REVIEW: Why not just call res.end() as is, and remove the Promise wrapper?
+    res.end = () => {
+      resolve()
+      res.end()
+    }
```
balazsorban44, line 38: Don't really understand the need for Promise and `resolve()` here. `res.end()` should already signal that the response is complete.

See: https://nodejs.org/api/http.html#http_response_end_data_encoding_callback

src/lib/baseUrl.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+  const defaultUrl = 'http://localhost:3000/api/auth'
+  _url = _url || defaultUrl
+  try {
+    const parsedUrl = new URL(_url)
+    if (parsedUrl.pathname === '/') {
+      parsedUrl.pathname = '/api/auth'
+    }
+    parsedUrl.pathname = parsedUrl.pathname.replace(/\/$/, '')
+    parsedUrl.href = parsedUrl.href.replace(/\/$/, '')
```
balazsorban44, line 21: removes trailing slashes

src/lib/baseUrl.js
```diff
@@ -0,0 +1,30 @@
...
+function baseUrl (url) {
+  let _url = url || process.env.NEXTAUTH_URL || process.env.VERCEL_URL
+  if (typeof _url !== 'string' && !(_url instanceof URL)) {
+    throw new Error('baseUrl must be either a valid URL object or a valid string URL')
+  }
+  const defaultUrl = 'http://localhost:3000/api/auth'
+  _url = _url || defaultUrl
+  try {
+    const parsedUrl = new URL(_url)
+    if (parsedUrl.pathname === '/') {
```
balazsorban44, line 17: set a default pathname, if it wasn't provided

PR #708, opened 2020-09-23
add(db): Add support for Fauna DB

Joev-:
Adds support for [FaunaDB](https://fauna.com/) as a database adapter.

I have attempted to mimic the existing adapter implementations, whilst still keeping to the Fauna style where possible. Fauna refs are used to reference related data, although I have opted for storing just the 'id' rather than a ref object. All timestamps are stored as Time objects in UTC (not sure if this will cause any problems down the line?).

I am not very experienced with Fauna and FQL, so if anyone does have suggestions, I am all ears :-) 

There is some initial setup required by the user to create collections and indexes, although I was not sure if this could be fit into the `syncrhonize` option somehow. 

```
CreateCollection({name: 'account'});
CreateCollection({name: 'session'});
CreateCollection({name: 'user'});
CreateCollection({name: 'verification_reset'});

CreateIndex({
    name: 'account_by_provider_account_id',
    source: Collection('account'),
    unie: true,
    terms: [
        { field: ['data', 'providerId'] },
        { field: ['data', 'providerAccountId'] }
    ]
});

CreateIndex({
    name: 'session_by_token',
    source: Collection('session'),
    unie: true,
    terms: [
        { field: ['data', 'sessionToken'] }
    ]
});

CreateIndex({
    name: 'user_by_email',
    source: Collection('user'),
    unie: true,
    terms: [
        { field: ['data', 'email'] }
    ]
});

CreateIndex({
    name: 'verification_reset_by_token',
    source: Collection('verification_reset'),
    unie: true,
    terms: [
        { field: ['data', 'token'] }
    ]
});

```

PR #968, opened 2020-12-18
Add collections & indexes overrides for Fauna DB

n44ps:
**What**: Make the Fauna DB Adapter open to already existing databases

**Why**: What if we want to use Next-auth & Fauna DB using a running database ? The [previous pull requests](https://github.com/nextauthjs/next-auth/pull/708) was hardcoding collections & indexes. That was making it impossible, unless renaming those items in Fauna DB which could lead to code refactoring.

**How**: Just like Prisma, we use the config parameter of the adapter to override those options.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged


waptik:
This looks good. With this version of the adapter, using/changing collections & indexes can be smooth and easy.
It would be great do have its documentation on the website but at this moment its not needed, not until a canary version of the package has landed.

n44ps:
Any chance that this PR goes to the next release ? 🙏 

Have a nice day everyone ! Let's go to 2021 finally :)

PR #1002, opened 2020-12-27
Allow linking oauth with existing unlinked user

jgeurts:
**What**:

Provide an option to allow OAuth to link with existing user found by email address. By default, this is disabled to preserve existing behavior.

**Why**:

In some instances, accepting the risk of duplicate user accounts is a better user experience than having the user login with email and then link their account to their oauth provider.

**How**:

Added a boolean configuration option called `linkOAuthWithExistingUser`

**Checklist**:

- [x] Documentation
- [x] Tests N/A
- [x] Ready to be merged

### DIFF
--- a/src/server/lib/callback-handler.js
+++ b/src/server/lib/callback-handler.js
@@ -17,7 +17,7 @@ export default async (sessionToken, profile, providerAccount, options) => {
     if (!profile) { throw new Error('Missing profile') }
     if (!providerAccount || !providerAccount.id || !providerAccount.type) { throw new Error('Missing or invalid provider account') }
 
-    const { adapter, jwt, events, linkOAuthWithExistingAccount } = options
+    const { adapter, jwt, events, linkOAuthWithExistingUser } = options
 
     const useJwtSession = options.session.jwt
 
@@ -179,7 +179,7 @@ export default async (sessionToken, profile, providerAccount, options) => {
           // We end up here when we don't have an account with the same [provider].id *BUT*
           // we do already have an account with the same email address as the one in the
           // oAuth profile the user has just tried to sign in with.
-          if (linkOAuthWithExistingAccount) {
+          if (linkOAuthWithExistingUser) {
             await linkAccount(
               userByEmail.id,
               providerAccount.provider,
--- a/src/server/lib/callback-handler.js
+++ b/src/server/lib/callback-handler.js
@@ -17,7 +17,7 @@ export default async (sessionToken, profile, providerAccount, options) => {
     if (!profile) { throw new Error('Missing profile') }
     if (!providerAccount || !providerAccount.id || !providerAccount.type) { throw new Error('Missing or invalid provider account') }
 
-    const { adapter, jwt, events } = options
+    const { adapter, jwt, events, linkOAuthWithExistingAccount } = options
 
     const useJwtSession = options.session.jwt
 
@@ -179,7 +179,25 @@ export default async (sessionToken, profile, providerAccount, options) => {
           // We end up here when we don't have an account with the same [provider].id *BUT*
           // we do already have an account with the same email address as the one in the
           // oAuth profile the user has just tried to sign in with.
-          //
+          if (linkOAuthWithExistingAccount) {
+            await linkAccount(
+              userByEmail.id,
+              providerAccount.provider,
+              providerAccount.type,
+              providerAccount.id,
+              providerAccount.refreshToken,
+              providerAccount.accessToken,
+              providerAccount.accessTokenExpires
+            )
+            await dispatchEvent(events.linkAccount, { user: userByEmail, providerAccount })
+            session = useJwtSession ? {} : await createSession(userByEmail)
+            return {
+              session,
+              user: userByEmail,
+              isNewUser
+            }
+          }
+
           // We don't want to have two accounts with the same email address, and we don't
           // want to link them in case it's not safe to do so, so instead we prompt the user
           // to sign in via email to verify their identity and then link the accounts.


iaincollins:
Thanks for creating this PR!

I'd like us to pick this up, but would ask we don't merge it in. Do correct me if I've misunderstood the intent.

We don't support fully automatic linking (even as an option) because it introduces a security problem - not all OAuth providers verify email addresses (and there is nothing in an RFC to say they should) meaning that in some cases it's possible to hijack an existing account by creating an account with an OAuth provider and changing the email address on it to match that of an account you want to access.

Quite a few sites actually do something like this with their home grown OAuth integration but it's insecure - often teams don't realise this right away, and handling it better is hard. It's thankfully becoming less common to see though. Commercial authentication solutions, like. Auth0 explicitly don't do it (and don't give an option).

I would like us to provide a better user experience for this that provides a much better experience to link accounts that the current experience - and to generally improve the email account sign in as well - but I think that is something we should tackle by improving that flow rather than an option like this (which is safe in some circumstances, but is problematic as an option in a general purpose library because it can be dangerous).

I appreciate the current user experience for this flow isn't ideal and that's likely what's driving this. It can, and should be, much simpler for end users and have better messaging (eg "It looks like you already have an account, we've sent you an email, please follow the instructions to link your accounts" and in the email is a link which signs you in *and* links the accounts).

In the short term, using custom error pages to handle this specific error seems like a good strategy.

jgeurts:
If it comes down to each provider, could there be a flag on a provider like `emailVerified` or something like that so that if we're only using providers that verify emails, accounts could be linked? Or is there still a security risk? For example, my use case only uses the google provider. In that case, I imagine the email address is verified to be correct.

kadirselcuk (approved):

iaincollins:
I'm going to close this as unfortunately we can't expose this as a provider option as it is likely to be abused.

However, I appreciate raising it and it's still very helpful to have.

Very much agree we should follow up this issue and try improve the user experience, so that it's easier to onboard (e.g. sends an email, or prompts user to sign in with previous provider and then remembers the action they were trying to perform and links the accounts once they are signed in). This might require some thought to get right, but I'm sure we can do something incremental (like trigger sending an email, if an email provider is configured) that would improve the UX.

iaincollins:
PS: If enough people clamour for this I'd consider adding it with something like a silly variable name so it's clear the risks (which I agree are fine in some contexts, such as Google when used in conjunction with the email verified flag).

Feel free to comment on this PR if you find this and it applies to you.

LeunensMichiel:
We have an existing database with users, and are building a new website in Next. Administrators can sign in with an azure b2c account, so it would be nice to check if an email already exists in the database, so the account can be linked to that.

adrianfdez469:
> I appreciate the current user experience for this flow isn't ideal and that's likely what's driving this. It can, and should be, much simpler for end users and have better messaging (eg "It looks like you already have an account, we've sent you an email, please follow the instructions to link your accounts" and in the email is a link which signs you in _and_ links the accounts).

It will be a great if NextAuth could do something like that.

meotimdihia:
I think almost everyone just uses Google or Facebook provider.

djvid:
Would be nice to be able to link user accounts for an app I'm building as well since we only have Google and email provider. But super happy with how easy it is to implement authentication with this library, thank you.

slipo:
> PS: If enough people clamour for this I'd consider adding it with something like a silly variable name so it's clear the risks (which I agree are fine in some contexts, such as Google when used in conjunction with the email verified flag).
> 
> Feel free to comment on this PR if you find this and it applies to you.

Yes! If you are only using oauth providers which are safe to link, this feature would simplify things greatly. 👍

maximousblk:
We have a similar use case too where we want to support use:pass and GitHub but we don't have a way to link between the two.

@iaincollins  If not automatically linking on sign in, would it make sense to have a linking flow where the user could click a button to link another provider **after** they are logged in (say on a user settings page) so you know it is the same user and not an outsider trying to hijack the account?

apperside:
With next-auth version 3.27.0 I am able to force the autolinking by applying[ this patch](https://gist.github.com/apperside/f0103d281e80867dce8050e2eb43513b) to the file node_modules/next-auth/dist/server/lib/callback-handler.js 

Beside the comments, It just changes  at line 151 this

```
if (userByEmail) {
         throw new _errors.AccountNotLinkedError();
       }
```

to this 

```
if (userByEmail) {
         await linkAccount(userByEmail.id, providerAccount.provider, providerAccount.type, providerAccount.id, providerAccount.refreshToken, providerAccount.accessToken, providerAccount.accessTokenExpires);
       }
```

To apply the patch:
- save the gist in the root of your project and name it callback-handler.js.patch 
- run this command from the command line (from the project root)

`patch --forward node_modules/next-auth/dist/server/lib/callback-handler.js < callback-handler.js.patch`

check that the file node_modules/next-auth/dist/server/lib/callback-handler.js has actually been changed.

You can eventually set the command as a npm post-install script

balazsorban44:
@apperside without saying for sure, I'm not sure your suggestion doesn't create some security issues. I'm almost convinced that there was an initial reason by @iaincollins that auto linking did not happen at that point. I might be wrong, but I just wanted to clarify that patches like that means you are on your own.

apperside:
@balazsorban44  I updated the post, I think it should be up to the developer to choose if scarify security in favor of user experience

balazsorban44:
I can agree to a certain degree. As a library, we would like to provide a secure-by-default solution though, with little space for leaving this path. Auth is easy to screw up, and small changes can quickly introduce unforseen consequences. As I haven't looked at this PR in details, I cannot speak for your suggestion specifically, but as an example, that's one of the reasons we don't provide a simple username+password built-in solution either.

goforbg:
Is there any way we can control this as a business logic in the options?

nmpribeiro:
@iaincollins at least it would be nice to have some hints on how to implement this ourselves. I agree it shouldn't be supported by default, but we could decide on a case by case, for our own projects, if we could link. Or at least have some ability to have a profile page where the users could link their accounts - I am looking that right now.

mattsears18:
I just started a new app using the Google, Facebook, and GitHub providers. What's the current best practice to auto-link accounts from these providers? Should I review this PR or is there a better approach somewhere else?

PR #1015, opened 2020-12-31
feat: add profile callback to options

### LINKED ISSUE #597: How to pass more attributes to profile object than default values?
I have written custom adapter.
In that, we have createUser method which accepts profile.
This profile object has only 3 default values (name, image, email).

async function createUser(profile) {
            console.log('createUser')
            console.log(profile)
            const user = await userObj.createUser(profile)
            //console.log(user)
            return user            
        }

I want to get more values in profile object in createUser method in custom adapter.
In scope as you see i have added address and birthday, i dont get them in profile object in createUser method.

Providers.Google({
      scope : 'https://www.googleapis.com/auth/user.addresses.read https://www.googleapis.com/auth/userinfo.profile https://www.googleapis.com/auth/userinfo.email https://www.googleapis.com/auth/user.birthday.read',
      clientId: <clientid>,
      clientSecret: <client secret>
    }),

So my question is how to add more values from scope in profile object?

cyrus-za:
**What**:

Add ability to overwrite the profile option, which allows adding additional attributes to the profile.

NOTE: A custom adapter will need to be used to take advantage of this. 

**Why**:

Fixes #597

**How**:

Specify `profile` function in the provider options

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

PS. This will obviously need to be done for all providers. I am simply showing here with a single provider what can be done.

### DIFF
--- a/src/providers/auth0.js
+++ b/src/providers/auth0.js
@@ -9,7 +9,7 @@ export default (options) => {
     accessTokenUrl: `https://${options.domain}/oauth/token`,
     authorizationUrl: `https://${options.domain}/authorize?response_type=code`,
     profileUrl: `https://${options.domain}/userinfo`,
-    profile: (profile) => {
+    profile: options.profile ?? (profile) => {
       return {
         id: profile.sub,
         name: profile.nickname,


balazsorban44:
Hi there, I am not sure what this supposed to achieve. Could you give an example?

To return more data from the profile, you can already access all the `profileData`:
https://github.com/nextauthjs/next-auth/blob/91e26ca47585440db92e9e39d5a85a6b13e98a67/src/server/lib/oauth/callback.js#L164

You can override the returned data like so:

```js
providers: [
  Providers.Auth0({
    clientId: "",
    clientSecret: "",
    async profile(profileData) {
      // You can access profileData here and for example update your database with it, if necessary
      return profileData // You return as much of this as you want
    }
  })
]
```

balazsorban44 (changes requested):
You need to give a better description of the goal for this PR.

treykaws:
> Hi there, I am not sure what this supposed to achieve. Could you give an example?
> 
> To return more data from the profile, you can already access all the `profileData`:
> https://github.com/nextauthjs/next-auth/blob/91e26ca47585440db92e9e39d5a85a6b13e98a67/src/server/lib/oauth/callback.js#L164
> 
> You can override the returned data like so:
> 
> ```js
> providers: [
>   Providers.Auth0({
>     clientId: "",
>     clientSecret: "",
>     async profile(profileData) {
>       // You can access profileData here and for example update your database with it, if necessary
>       await writeProfileDataToDBForUser(profileData.id, profileData)
>       return profileData // You return as much of this as you want
>     }
>   })
> ]
> ```

Is this in the current release? It's not working for me at the moment - profile doesn't seem to exist on the auth0 provider type.

balazsorban44:
Yes, it is available in both the stable and canary verdson. When you define a Provider, usually you only need to override just a few properties like client id and client secret, maybe a few others, but you have the possibility to override any parameters, including the profile callback method! This method receives the full profileData object that contains all the data we receive from a provider, whuch you can than mutate how you want, or even fetch extra data inside the profile method to extend the user profile.

cyrus-za:
@balazsorban44 that does not work for me with auth0 provider. typescript does not even allow it

balazsorban44:
As you can see here: https://www.github.com/nextauthjs/next-auth/tree/canary/src%2Fproviders%2Fauth0.js

options will override anything defined by default, so you can override the profile callback as you wish. TypesScript is a community effort right now, so if you have TS problems, please check out the DefinitelyTyped repo.

https://www.npmjs.com/package/@types/next-auth


You can always override types yourself though. Not ideal, but we welcome the community to help out migrating to TS. 🙂

treykaws:
> As you can see here: https://www.github.com/nextauthjs/next-auth/tree/canary/src%2Fproviders%2Fauth0.js
> 
> options will override anything defined by default, so you can override the profile callback as you wish. TypesScript is a community effort right now, so if you have TS problems, please check out the DefinitelyTyped repo.
> 
> https://www.npmjs.com/package/@types/next-auth
> 
> You can always override types yourself though. Not ideal, but we welcome the community to help out migrating to TS.

I was able to effectively do exactly what you described with the profile data callback. It adds a great deal of flexibility. If I am able to get some time I'll try to contribute to the TS migration efforts.

treykaws:
> As you can see here: https://www.github.com/nextauthjs/next-auth/tree/canary/src%2Fproviders%2Fauth0.js
> 
> options will override anything defined by default, so you can override the profile callback as you wish. TypesScript is a community effort right now, so if you have TS problems, please check out the DefinitelyTyped repo.
> 
> https://www.npmjs.com/package/@types/next-auth
> 
> You can always override types yourself though. Not ideal, but we welcome the community to help out migrating to TS.

So the default user object is only name, email, image. If I wanted to extend this to add an id property or any additional properties. what is the recommended way of accomplishing this?

Issue #1022, opened 2021-01-01
Reduce user facing provider APIs


balazsorban44:
Currently, there are some "magical" 🧙 fields only utilized by specific providers. We could handle these edge cases internally instead. All providers have an `id`, which we can use to switch on different behaviors. It would also make it easier to keep a TypeScript provider interface simple, without the need for [Discriminated Union](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) or similar

Some of these fields and which provider(s) use them (introduced in which PR):
* **additionalAuthorizeParams**: slack (#895)
* **accessTokenGetter**: slack (#895)
* **setGetAccessTokenAuthHeader**: okta, identity-server4 (https://github.com/nextauthjs/next-auth/pull/213)
* ~**setGetAccessTokenProfileUrl**: mailru (https://github.com/nextauthjs/next-auth/pull/522 not yet implemented)~
* **clientSecretCallback**: apple (https://github.com/nextauthjs/next-auth/issues/176)
* **state**: apple (https://github.com/nextauthjs/next-auth/issues/176)
* **prepareProfileRequest**: bungie (https://github.com/nextauthjs/next-auth/pull/589)

balazsorban44:
@cathykc As the original author of #895, I would like to ask you, how strongly do you feel about `additionalAuthorizeParams` being a provider property, instead of using the (new) third argument of the `signIn` client function?

https://github.com/nextauthjs/next-auth/blob/f2ad69358fb6e650035c85d320440468d0286377/src/client/index.js#L226-L242

Example:
```jsx
<button
  onClick={() => signIn("slack", {}, {user_scope: 'identity.basic,identity.email,identity.avatar'})}
>
  Login
</button>
```

Right now, there are 2 ways of doing it, and we could eventually keep both, I was just curious about your opinion.
Setting authorize params may actually be useful for several providers, so I may even be OK with removing the `authParams` from the client `signIn` function.

balazsorban44:
@cathykc when #1023 will be merged, the name will change from

`additionalAuthorizeParams` to `authorizationParams`. This plays better with the already present `authorizationUrl`.

I changed the slack provider respectively, so no changes will be needed.

PR #1060, opened 2021-01-08
feat(provider): add vk.com provider

boreyko1:
Hi! 👋

**What**:

This PR adds [vk.com](https://en.wikipedia.org/wiki/VK_(service)) provider.
VK's api calls require explicit version so it is added to urls. Also, it is possible to change it via providers' options config.

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [x] Ready to be merged

Looks like everything works fine. Let me know if I missed anything!

Thanks!

### DIFF
--- a/src/providers/vk.js
+++ b/src/providers/vk.js
@@ -1,0 +1,32 @@
+export default (options) => {
+  const apiVersion = options.apiVersion || '5.126' // https://vk.com/dev/versions
+
+  return {
+    id: 'vk',
+    name: 'vk.com',
+    type: 'oauth',
+    version: '2.0',
+    scope: 'email',
+    params: {
+      grant_type: 'authorization_code'
+    },
+    accessTokenUrl: `https://oauth.vk.com/access_token?v=${apiVersion}`,
+    requestTokenUrl: `https://oauth.vk.com/access_token?v=${apiVersion}`,
+    authorizationUrl:
+      `https://oauth.vk.com/authorize?response_type=code&v=${apiVersion}`,
+    profileUrl: `https://api.vk.com/method/users.get?fields=photo_100&v=${apiVersion}`,
+    profile: (result) => {
+      const profile = result.response?.[0]
+
+      if (!profile) return {}
+
+      return {
+        id: profile.id,
+        name: [profile.first_name, profile.last_name].filter(Boolean).join(' '),
+        email: profile.email,
+        image: profile.photo_100
+      }
+    },
+    ...options
+  }
+}
--- a/src/providers/vk.js
+++ b/src/providers/vk.js
@@ -1,5 +1,5 @@
 export default (options) => {
-  const apiVersion = options.apiVersion || '5.126' // https://vk.com/dev/versions
+  const apiVersion = '5.126' // https://vk.com/dev/versions
 
   return {
     id: 'vk',
@@ -16,9 +16,7 @@ export default (options) => {
       `https://oauth.vk.com/authorize?response_type=code&v=${apiVersion}`,
     profileUrl: `https://api.vk.com/method/users.get?fields=photo_100&v=${apiVersion}`,
     profile: (result) => {
-      const profile = result.response?.[0]
-
-      if (!profile) return {}
+      const profile = result.response?.[0] ?? {}
 
       return {
         id: profile.id,
--- a/www/docs/providers/vk.md
+++ b/www/docs/providers/vk.md
@@ -23,3 +23,27 @@ providers: [
   })
 ]
 ...
+```
+
+:::note
+By default the provider uses `5.126` version of the API. See https://vk.com/dev/versions for more info.
+:::
+
+If you want to use a different version, you can pass it to provider's options object:
+
+```js
+// pages/api/auth/[...nextauth].js
+
+const apiVersion = "5.126"
+...
+providers: [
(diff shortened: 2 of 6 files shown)


src/providers/vk.js
```diff
@@ -0,0 +1,32 @@
+export default (options) => {
+  const apiVersion = options.apiVersion || '5.126' // https://vk.com/dev/versions
```
balazsorban44, line 2: In an ongoing effort (#1022), we try to reduce the API surface to a common minimum, so we have a consistent API across all providers. I understand that this would make it easier to define the different URLs a tiny bit simpler, but I kindly ask you to remove this, and let this be handled in user land.

Example:
```js
// /api/auth/[...nextauth].js
const apiVersion = "5.126"
...
providers: [
  Providers.VK({
    accessTokenUrl: `https://oauth.vk.com/access_token?v=${apiVersion}`,
    requestTokenUrl: `https://oauth.vk.com/access_token?v=${apiVersion}`,
    authorizationUrl:
      `https://oauth.vk.com/authorize?response_type=code&v=${apiVersion}`,
    profileUrl: `https://api.vk.com/method/users.get?fields=photo_100&v=${apiVersion}`,
  })
]
...
``` 

This will make it much easier to create consistent TypeScript types later on, as well as keeps the documentation simple, and the users will be less confused about the available provider options.

You could instead add an explanation to `www/docs/providers/vk.md` to the VK provider users, explaining how they can override the default `5.126` version (probably also worth mentioning to them, which version is used by default.)

balazsorban44 (changes requested):
I had a little change request and some small comments, but all in all looks good!

(Thanks for reordering providers alphabetically)

src/providers/vk.js
```diff
@@ -0,0 +1,32 @@
...
+    version: '2.0',
+    scope: 'email',
+    params: {
+      grant_type: 'authorization_code'
+    },
+    accessTokenUrl: `https://oauth.vk.com/access_token?v=${apiVersion}`,
+    requestTokenUrl: `https://oauth.vk.com/access_token?v=${apiVersion}`,
+    authorizationUrl:
+      `https://oauth.vk.com/authorize?response_type=code&v=${apiVersion}`,
+    profileUrl: `https://api.vk.com/method/users.get?fields=photo_100&v=${apiVersion}`,
```
balazsorban44, line 17: please hard code a default `apiVersion` here instead

src/providers/vk.js
```diff
@@ -0,0 +1,32 @@
...
+    profile: (result) => {
+      const profile = result.response?.[0]
+
+      if (!profile) return {}
```
balazsorban44, line 21: Why would it not return anything here?

www/docs/providers/vk.md
```diff
@@ -0,0 +1,25 @@
...
+---
+
+## Documentation
+
+https://vk.com/dev/first_guide
+
+## Configuration
+
+https://vk.com/apps?act=manage
+
```
balazsorban44, line 13: Do explain here the possibility to use different API versions, and that `5.126` is the default. Maybe add a link to the corresponding vk.com page as well.

src/providers/vk.js
boreyko1: Good point! Left `apiVersion` hardcoded and added an example with version override to docs.

boreyko1:
Made required changes and added info to docs page. Please, take a look one more time. Thanks!

src/providers/vk.js
```diff
@@ -0,0 +1,30 @@
...
+    params: {
+      grant_type: 'authorization_code'
+    },
+    accessTokenUrl: `https://oauth.vk.com/access_token?v=${apiVersion}`,
+    requestTokenUrl: `https://oauth.vk.com/access_token?v=${apiVersion}`,
+    authorizationUrl:
+      `https://oauth.vk.com/authorize?response_type=code&v=${apiVersion}`,
+    profileUrl: `https://api.vk.com/method/users.get?fields=photo_100&v=${apiVersion}`,
+    profile: (result) => {
+      const profile = result.response?.[0]
```
balazsorban44, line 19: I like the use of optional chaining here, but you should probably add a fallback value with `?? {}`, otherwise property access on `profile` will fail if profile is undefined.

Could you tell me in which case the response won't include any profile information? It seems strange you need to check for existence at all.

balazsorban44 (changes requested):
Thanks! Just a minor change request and a small question left. In which case would the provider not return any profile information?

src/providers/vk.js
boreyko1: You're right. Added fallback, previously there was just an early return with `if (!profile) return {}`.

Response will contain `error` object with error code and message instead of users info, for example, if one omits `v=${apiVersion}` param from url.

boreyko1:
> Thanks! Just a minor change request and a small question left. In which case would the provider not return any profile information?

Provider will return error object when url params or urls themselves are wrong or there is something with client id/secret keys.

balazsorban44:
I wonder if we already detect those kind of errors by redirecting to an error page if the oauth callback fails, and logging the reason. 

I'm going to merge this for now though, thanks. If it comes up as an issue later on, I'll reference this PR.

Thanks for contributing!

balazsorban44 (approved):

Issue #615, opened 2020-08-27
Invalid signature for the JWT token when verifying in jwt.io and via jsonwebtoken


roderik:
**Your question**

I have a NextJS app with NextAuth configured with Auth0. I can log in, it all works great.
I can expose the JWT token that is being created and modified, so I'm good there.

My options contain `jwt: {secret: process.env.SECRET, }` and by some creative console logging, I have got the derived key out. I also tried to use the `signingKey: process.env.JWT_SIGNING_PRIVATE_KEY` variant and bypassed the warning.

Now I wan to use the same JWT token to authenticate to a NestJS API, which uses jsonwebtoken and not jose for its JWT verification and I can either give it a secret or a PEM key. 

And for the love of me, I cannot get the JWT token to be verified

- jwt.io shows invalid signature, this already feels weird, the JWT should at least be valid right?
- using the `secret` from the options gives me `JsonWebTokenError: invalid signature`
- using the jose asKey from the `signingKey`, either as-is or getting the `k` field out of it, does not work with the same error
- i tried ec/rsa keys so i could use the `toPem()` from jose, same error.

The only option I see is getting the decoded jwt on Next, and re-coding my own JWT with jsonwebtoken and a shared plain text secret, but that seems like a step extra and feels like giving up.

roderik:
And at that exact moment i realise `k` is base64 encoded...

```
    super({
      jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
      ignoreExpiration: false,
      secretOrKey: Buffer.from(jose.JWK.asKey(JSON.parse(configService.get<string>('SIGNING_KEY')!)).k!, 'base64'),
    });
```

PR #1066, opened 2021-01-10
feat: add jsonwebtoken support to jwt options

Thexumaker:
**What**:
 `jwt.getToken()` doesn't seem to work when using custom encode/decode functions. And considering how much more popular [jsonwebtoken](https://www.npmjs.com/package/jsonwebtoken) is compared to jose I thought it'd be nice if there was an easier way to use jsonwebtokens. 
I add an extra config option in [...nextauth]/jwt so that when` jsonwebtoken = true` next-auth will use the library jsonwebtoken instead of jose for encoding and decoding. 

**Why**:

So I personally have preferred using jsonwebtokens over jose for JWT related code, and I noticed a small bug(or maybe it was inteded) where if we set custom encode/decode functions jwt.getToken would no longer work. I think the issue was that in getToken, it calls the default decode() instead of the new custom decode to be called which causes an invalid signature to occur when checking if the token is valid on jwt.io.

I also found that it's not well explained that by default that jose is used, issue  #615 helped me figure out how to use jsonwebtoken with next-auth and I just thought it'd be nice to use the more popular jsonwebtoken library without having to set custom encode/decode functions. 

**How**:

So in the config file I'm introducing a new variable `jsonwebtoken` in lib/jwt.js  which by default is set to false using `DEFAULT_JSONWEBTOKEN`. And fortunately the only real changes is that an extra if statement is added to both the decode and encode functions:

    if (jsonwebtoken) {

    // signing key using jsonwebtoken instead of jose

    const signedToken = jwt.sign(token, secret, { algorithm: DEFAULT_SIGNATURE_ALGORITHM })

    // no support for encrypting tokens yet

    // return signed token

    return signedToken
  
    } 

    if (jsonwebtoken) {
    return jwt.verify(token, secret)
    } 

Everything else should have been left the same. In order to use these changes only one small change is needed in [...nextauth].js

    
    secret: process.env.JWT_SECRET,

    jsonwebtoken:true,
    
 

Additionally when calling getToken, it might be required to send one extra option however I don't remember testing for it.

` const token = await jwt.getToken({ req, secret,jsonwebtoken:true})`

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      
   
   I didn't write any tests so no even though the changes are somewhat minor , however it did seem to be working fine in my local development. 

I don't know how to write tests(like I've written both bdd/tdd tests before but I have no idea where to write the tests in the project structure). 
Also sorry if I messed up anything from the pull request to even this explanation. It's my first time contributing to an open source project and I'm not 100% sure if I'm doing things right.

### DIFF
--- a/src/lib/jwt.js
+++ b/src/lib/jwt.js
@@ -1,6 +1,7 @@
 import jose from 'jose'
 import hkdf from 'futoin-hkdf'
 import logger from './logger'
+import jwt from 'jsonwebtoken'
 
 // Set default algorithm to use for auto-generated signing key
 const DEFAULT_SIGNATURE_ALGORITHM = 'HS512'
@@ -10,6 +11,7 @@ const DEFAULT_ENCRYPTION_ALGORITHM = 'A256GCM'
 
 // Use encryption or not by default
 const DEFAULT_ENCRYPTION_ENABLED = false
+const DEFAULT_JSONWEBTOKEN = false
 
 const DEFAULT_MAX_AGE = 30 * 24 * 60 * 60 // 30 days
 
@@ -17,6 +19,7 @@ const encode = async ({
   token = {},
   maxAge = DEFAULT_MAX_AGE,
   secret,
+  jsonwebtoken = DEFAULT_JSONWEBTOKEN,
   signingKey,
   signingOptions = {
     expiresIn: `${maxAge}s`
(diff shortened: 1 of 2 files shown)


balazsorban44:
So someone else from the team has to chip in on  this, but the way I see it, jsonwebtoken hasn't been updated in 2 years, while jose looks actively maintained. I guess I see that we use an old 1.x version, while the newest one is 3.x as of this writing. Maybe we should update jose  first, and see if it brings anything new to the table, before introducing new flags?

Thexumaker:
Definitely wouldn’t mind an update to Jose but jsonwebtoken has been updated pretty recently and their github repo is still pretty active. Not to mention it’s got 5 million downloads per week vs only 200k for jose(https://www.npmtrends.com/jsonwebtoken-vs-node-jose-vs-simple-oauth2 ). But for sure if it seems kinda extra it’s all good.

Thexumaker:
Also wanted to add that this might be a separate issue but when I'm using custom encode/decode functions, and I put in the token I get from .getToken at https://jwt.io/ the signature can't be verified. Like the payload and header are fine but the signature isn't getting verified which I feel like has something to do with the decode function not working properly. With my fix it seems to verify the signature perfectly fine.

balazsorban44:
I mean the npm package hasn't been updated in 2 years for `jsonwebtoken`. It's not necessarily a bad thing, more of an observation, maybe the creators think it's feature-complete.

iaincollins:
@Thexumaker 

> jwt.getToken() doesn't seem to work when using custom encode/decode functions

It should do. Are you passing the custom encode/decode functions into `getToken()`?

Note: Because Serverless functions must be assumed to be run in isolated process you must pass the same encode/decode routines into `getToken()` as are defined in your `[...nextauth].js` (other endpoints can't access the logic in that route).

The upshot is that if using custom encode/decode functions, you probably to define the encode/decode methods centrally and import them were needed, and/or create a wrapper for `getToken()` to simplify usage in different routes in your app.

I appreciate this is a bit of a gotcha. In the longer term I want to look at ways of making this easier but there are some constraints on different environments I'm mindful of, especially if we want to try and keep configuration simple and as centralised as possible.

iaincollins:
Related question about how to use options with `getToken()` that folks might find helpful: https://github.com/nextauthjs/next-auth/discussions/1496

The project actually used jsonwebtoken at one point (it is a bit simpler, which made it more accessible) but we have migrated away from it as it's not as actively maintained and JOSE contains a critical number of enhancements with additional options for security and compatibility.

Thanks for the PR, but I'm going to close this one as I don't think there is a defect here, just a case of confusion because of a gotcha currently presented in the API (that is known and an accepted limitation as not-trivial to resolve).

PR #1084, opened 2021-01-11
feat: set session to the context on a server side if available

lnikell:
**What**:
The problem described in the https://github.com/nextauthjs/next-auth/issues/844 issue. Currently using getServerSideProps and getSession don't set session context properly which results in having an extra call on the client-side even if we get it already on the server-side.

**Why**:

- Avoiding extra network calls has a good impact on a performance
- It prevents confusion during development when you retrieve a session on the server-side but then have again a call on the client-side

**How**:

The issue comes from the Provider, where context was always set by fetching the context with useSession. In the useSession however, there is a fallback to do a useSessionHook. This means even if we already have Session, we anyway set it only from the client-side.

https://github.com/nextauthjs/next-auth/blob/416d92c33f051c2a13219814e6deb6010e6462f5/src/client/index.js#L294-L297

The fix is simple, just set the value to the context which is an array of session + loading if we already have it on the server side. 

https://github.com/lnikell/next-auth/blob/1c7be413f56ee304b129c31763053a4597f8a131/src/client/index.js#L294-L298

**Before**:

https://user-images.githubusercontent.com/2697570/104178161-27daf800-540a-11eb-8536-a894dd7c96b5.mov

**After**:

https://user-images.githubusercontent.com/2697570/104177996-f19d7880-5409-11eb-9113-6b8924828587.mov

**Checklist**:

- [ ] Documentation n/a
- [ ] Tests n/a
- [x] Ready to be merged


balazsorban44:
I really appreciate the before/after videos on your contributions, very helpful! Also, thanks for digging into this problem.

I am not sure if this is the perfect solution though, as it breaks the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level). (It does so already, `_useSessionHook` is not exactly called right either). I am wondering if we could work this around another way.

lnikell:
@balazsorban44 thanks, good point about hooks rules,  I will refactor it  a little bit.

iaincollins:
@lnikell This is great work, thank you! @balazsorban44 good catch consideration for Rules of Hooks 

I think I tried to refactor this once but gave up because I couldn't work out how to do it "right". :-/

FWIW if we can't figure out how to do it without breaking the rules but we don't see any side effects in practice, personally I have no objection to going ahead anyway (although I agree, doing it without breaking them is the best scenario!).

balazsorban44:
I had a go at this at https://github.com/nextauthjs/next-auth/pull/943 but I wasn't satisfied, and got unsure if I broke anything so did not proceed. I might get back to it at a later point, but I would like to see if we come up with something smart on this PR.

lnikell:
@balazsorban44 @iaincollins I spent some time learning more the code that related to this issue and I think we can consider removing the "useSession" execution from the `Provider. 

My assumptions:

- `Provider` should  only keep `session` data  and don't make an  api call and also keep a loading state.
-  Having `useSession` inside Provider creates an extra "magic" logic and can confuse users(such as me). I was confused the first time when I get calls to the "session" endpoint without having anywhere in my code `useSession` method. 
- If we remove `useSession` call from Provider we will have to add extra state to Provider to set the session, which we will run from `useSession` in order to update the context.

So as a temprorary solution we can go with what we have in this PR, and later consider refactoring of the Provider. I can try to make those changes if you confirm.

iaincollins:
> I think this is causing me MANY problems! Is there any update on getting it into the main release? What can I do in the meantime?

I would suspect it's not the cause of any problems you are having as it's more an optimisation.

Can you elaborate on sort of problems are you having and how do you think this would resolve them?

PS: If you haven't already, it might be best to post in Discussions about them and link back here.

rozenmd:
FWIW: I did a few Lighthouse tests to compare before/after of the workaround described [here](https://github.com/nextauthjs/next-auth/issues/844#issuecomment-733770587) - the extra getSession request can increase speed index by up to 3.0 seconds. 

Pretty worthwhile optimisation if you ask me.

balazsorban44:
So I think I managed to work around this, without breaking the Rules of Hooks! Added in this commit: https://github.com/nextauthjs/next-auth/pull/1428/commits/231a66c01b3866ca76c13836dd7aed7eea1e14aa

We can do the state initialization smarter. The interesting part is:
```js
const [data, setData] = useState(context?.[0] ?? session)
const [loading, setLoading] = useState(!data)
...
useEffect(
```

So instead of calling the `useEffect` optionally, we set the state by first checking the context value.

One potential downside for this is that the useEffect has to be called on each render (notice in the commit, there is no dependency array at the end), but early bailout ensures that we still only fetch the session when it is needed, as before.

balazsorban44:
Closing in favor of #1428. Thank you so much for your examples @lnikell, they helped a lot while evaluating this!

PR #1105, opened 2021-01-14
feat: replace node-oauth

### LINKED ISSUE #1048: Replace node-oauth dependency
https://github.com/ciaranj/node-oauth Is a library we use to communicate with OAuth providers. Although it works, we have already patched it in many places, mainly because it uses callbacks over Promises, and the library itself looks abandoned. (Last commit was in 2017).

We should either get rid of it, or replace it fully.

[`node-openid-client`](https://github.com/panva/node-openid-client) looks like a nice alternative, but it does not seem to support OAuth 1.0.

We could add it now, and only utilize it for OAuth 2 and OIDC flows, and we could remove OAuth 1 support at a later point. (It would be a breaking change)

As far as I can tell, only the Twitter provider uses OAuth 1.

Ref: https://developer.twitter.com/en/docs/authentication/api-reference/token

balazsorban44:
**What**:

**Why**:

**How**:

https://github.com/panva/node-openid-client/blob/main/docs/README.md#
**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      

closes #1048

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1865,11 +1865,35 @@
         }
       }
     },
+    "@sindresorhus/is": {
+      "version": "4.0.0",
+      "resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-4.0.0.tgz",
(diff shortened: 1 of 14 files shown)


balazsorban44:
Continued in #1698

Issue #603, opened 2020-08-25
calling signOut function doesn't force page reload


Kozie1337:
**Describe the bug**
Calling the signOut function from the [nextauth Client API](https://next-auth.js.org/getting-started/client#signout) doesn't force a page-reload which makes it look like the user's still signed in.

**Steps to reproduce**
1. Sign in on https://next-auth-example.now.sh/ (with Google)
2. Press 'Sign out' 

**Expected behavior**
Page reloads after clicking 'Sign out' (calling signOut function)

as the Client AP states:
> [...] It reloads the page in the browser when complete.

* [ ] Found the documentation helpful
* [x] Found documentation but was incomplete
* [ ] Could not find relevant documentation
* [ ] Found the example project helpful
* [ ] Did not find the example project helpful

PaulKushch:
I got the same issue, what helped me is to remove e.preventDefault() from onClick handlers (commented out below):

`                onClick={() => {
                  // e.preventDefault();
                  signIn();
                }`

SharadKumar:
#532 this is a duplicate, I suppose.

Kozie1337:
I added `e.preventDefault(); ` to the onClick function and it still didn't work.

Then I noticed that I was getting the error from #552 during production.  

Adding
  `"peerOptionalDependencies": {
  "mongodb": "^3.6.0"
},`
to my package.json fixed both errors.

Everything seems to be working now.

samuelgoldenbaum:
This should not be closed as the issue still persists in Google and Facebook providers as per #532

PR #782, opened 2020-10-18
Update providers.md

angad777:

Issue #781, opened 2020-10-17
Client Side Page Transitions


HaNdTriX:
**Your question**

Why is `next-auth` not using [Next.js client page transitions](https://nextjs.org/docs/api-reference/next/router#routerreplace)?

**In Detail**

next-auth currently implements page transition using `window.location= <newURL>`. 
It uses this technique on login and on logout. I understand that on first oauth redirect request this doesn't make a difference, but on all other cases (Logout, api base oauth, CredentialsAuth, etc) is really makes a difference.

**Example**

https://github.com/nextauthjs/next-auth/blob/8115a7c66cdb1d86b0d2a0d76b9aa33c2bfaa33b/src/client/index.js#L230-L255

**Cons (full redirects)**

- Doesn't allow animated page transitions
- Is a lot slower
- Feels clunky
- Everything needs to be refetched when the auth state changes
- Is inconsistent with the tab share logic (see localstorage messaging)

**Pros (full redirects)**

- new state
- new env
- easier to get right

**Proposal**

The login flow could be much more fluent, if we switch or allow the usage of next client page transitions using `next.router.push()` and `next.router.replace()`, because we don't need to reinitialise the entire app.

**Whats needs to be done to make next client transitions work**

- [ ] Make next-auth context reactive (for all CRUD Actions) 
- [ ] Differentiate between internal, external and api redirects
- [ ] Move responsibility of redirects to the client
- [ ] Release major version (this will be a breaking change)

**What are you trying to do**

Improve the login flow.

**Can I help**

If the next-auth team would be interested in such a change I could work on this topic.

**Feedback**

*Documentation refers to searching through [online documentation](https://next-auth.js.org), code comments and issue history. The example project refers to [next-auth-example](https://github.com/iaincollins/next-auth-example).*

* [x] Found the documentation helpful
* [x] Found documentation but was incomplete
* [x] Could not find relevant documentation
* [x] Found the example project helpful
* [x] Did not find the example project helpful

iaincollins:
Thanks for raising this and describing it in detail.

Essentially "easier to get right" is the main reason for this - it reduces the amount of problems people run into and the amount of requests for help using it that get raised.

A secondary reason is keeping the methods decoupled from Next.js to try to avoid being impacted by breaking changes in Next.js and to be framework agnostic, even to React; with longer term goals of being suitable as a solution for Nuxt.js and other frameworks (currently there are some dependencies request/response middleware in a few places, but those are easily refactorable).

[...]

PR #1107, opened 2021-01-14
Client-side signOut force reload if redirect to the same page.

ThewBear:
**What**:

This pr adds additional check in client-side `signOut`  and forces reloading the page if final location is the same page.

**Why**:

Client-side `signOut` will not reload a page if it is the same page and url contains hash `#`. #603 

**How**:

It checks if all of url fragments are equal except hash.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged


balazsorban44:
Thanks for this PR!

I am not sure we would like to force reload the whole page or just make sure the session change triggers the useSession hook?

ThewBear:
I assume we would like to force reload the whole page for consistency with other url and following the documentation.

> It reloads the page in the browser when complete.

https://next-auth.js.org/getting-started/client#signout

balazsorban44:
I wanted to send you to https://github.com/nextauthjs/next-auth/issues/781, but I see you recently commented on that as well. Hmm. we have to discuss this @iaincollins

ThewBear:
I think #782 is a long-term feature but this PR solves a bug which produce inconsistent reloading when sign out.

src/client/index.js
```diff
@@ -287,7 +287,15 @@ export const signOut = async (args = {}) => {
   const res = await fetch(`${baseUrl}/signout`, fetchOptions)
   const data = await res.json()
   _sendMessage({ event: 'session', data: { trigger: 'signout' } })
-  window.location = data.url ?? callbackUrl
+  const finalLocation = new URL(data.url ?? callbackUrl)
```
balazsorban44, line 290: I wasn't sure myself if this would be a problem supporting IE11, or using it in Next.js would make sure that URL is polyfilled here?

src/client/index.js
```diff
@@ -287,7 +287,15 @@ export const signOut = async (args = {}) => {
...
+  window.location = finalLocation
+  if (
```
balazsorban44, line 292: I'm not sure I understand the need for this if statement. wouldn't setting window.location trigger a reload?

src/client/index.js
ThewBear: Nextjs includes URL polyfill. https://nextjs.org/docs/basic-features/supported-browsers-features

src/client/index.js
ThewBear: It won't trigger if the url is the same page and contains `#`.
Example:
| Current  | Final | window.location trigger reload |
| ------------- | ------------- | ------------- |
| /  | /  | yes |
| / | /#  | no |
| /# | /#  | no |
| / | /other  | yes |
| / | /other#  | yes |

src/client/index.js
balazsorban44: I'm aware, but will those apply for node_modules as well? I remember we had to transpile next-auth further down at work to be able to run in IE11. Could you please confirm that this would work in IE11 as well by trying it out?

src/client/index.js
balazsorban44: So I recently came across this at work and remembered your PR! Do you know if this is the intended browser behavior, or some kind of bug?

I can verify it very easily but opening the inspector on any page that has a hashtag in the url, it won't reload!

balazsorban44:
Closing this in favor of #1298. I do not wish you to fix the massive merge conflicts @ThewBear as it was our fault! I credited you in the new PR. Many thanks for discovering this! 🙏

PR #1124, opened 2021-01-16
feat: add native hkdf

### LINKED ISSUE #1112: use native node hkdf
Inspired by https://github.com/auth0/express-openid-connect/pull/177, we could also get rid of the `futoin-hkdf` dependency conditionally.

koolii:
**What**:

Closes #1112

Add using Node.js native `hkdf` instead of the third party's one.
I didn't remove the third party one (which is `futoin-hkdf`) 
because I think we can't restrict apps less than Node.js (v15.0.0).

**Why**:

Because Node.js(v15.0.0) supports `hkdf` by itself.

**How**:

Use Node.js `hkdf` if its version is bigger than v15.0.0.
otherwise third party one (which is `futoin-hkdf`) as usual.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      

I need your supports.
I would like to add some unit tests about this.
But I don't know how to change Node.js versions.

Could you give me some ideas about it?

### DIFF
--- a/src/lib/jwt.js
+++ b/src/lib/jwt.js
@@ -137,19 +137,19 @@ let DERIVED_SIGNING_KEY_WARNING = false
 let DERIVED_ENCRYPTION_KEY_WARNING = false
 
 // Do the better hkdf of Node.js one added in `v15.0.0` and Third Party one
-function hkdf (secret, { BYTE_LENGTH, DIGEST, ENCRYPTION_INFO }) {
+function hkdf (secret, { byteLength, encryptionInfo, digest = 'sha256' }) {
   if (crypto.hkdfSync) {
     return Buffer.from(
       crypto.hkdfSync(
-        DIGEST,
+        digest,
         secret,
         Buffer.alloc(0),
-        ENCRYPTION_INFO,
-        BYTE_LENGTH
+        encryptionInfo,
+        byteLength
       )
     )
   }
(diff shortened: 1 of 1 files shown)


src/lib/jwt.js
```diff
@@ -1,5 +1,6 @@
+import crypto from 'crypto'
 import jose from 'jose'
-import hkdf from 'futoin-hkdf'
+import thirdPartyHkdf from 'futoin-hkdf'
```
balazsorban44, line 3: maybe we could import this conditionally with require, when we don't use hkdf from crypto?

balazsorban44 (commented):
Thanks for this PR! I suggested a small change, and was also wondering, if you tested this in any way just to be sure it actually works?

EDIT: I see it's a draft right now, please ignore my review until you feel finished.. 😅

koolii:
Thank you for your reply :)
Let me turn it on after I check works by my self. Maybe tomorrow. Thank you.

src/lib/jwt.js
koolii: thank you for your reviews!

koolii:
### WHAT I did after reviews

- tweaked a bit about digest value and argument names.
- check the function by myself.

balazsorban44 (approved):
Thanks!

PR #1134, opened 2021-01-17
fix(adapter) -  fix cookie expiration in Fauna getSession

### LINKED ISSUE #1131: Session not being persisted when I close the browser
Hi, Is there a way to persist the session? As soon as I close the browser, the user is being logged out. 

When I saw the cookies that are set, I see `Expires/Max-Age` value of `next-auth.session-token` set to `Session`. I am using the default configuration. By default, I assumed the max-age will be set to 30 days as per documentation. But that doesn't seem to be the case. Am I missing something?

pbteja1998:
Fixes #1131 

**What**:
`session.data.expires` that we get from Fauna DB is in `Fauna.Time` format. In this PR, I am converting it into milliseconds.

**Why**:
Session not being persisted when browser is closed because of invalid `expiresAt` date

**Checklist**:

- [N/A] Documentation
- [N/A] Tests
- [x] Ready to be merged

### DIFF
--- a/src/adapters/fauna/index.js
+++ b/src/adapters/fauna/index.js
@@ -284,14 +284,17 @@ const Adapter = (config, options = {}) => {
       _debug('getSession', sessionToken)
 
       try {
-        const session = await faunaClient.query(
-          q.Get(
-            q.Match(
-              q.Index(indexes.Session),
-              sessionToken
-            )
-          )
-        )
+        const sessionFQL = q.Get(q.Match(q.Index(indexes.Session), sessionToken))
+
+        const session = await faunaClient.query({
+          id: q.Select(['ref', 'id'], sessionFQL),
+          userId: q.Select(['data', 'userId'], sessionFQL),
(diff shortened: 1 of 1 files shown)


balazsorban44 (changes requested):
Looks like there are linting errors, could you fix those at least so builds don't fail? :) @pbteja1998

ndom91 (approved):

balazsorban44:
So sorry for this @pbteja1998, but we decided to drop builtin support for Fauna DB for now. it will be removed soon in a later release. it is a bit of an overhead to make sure it works, while there aren't even tests for it. As you did as well so good, it is still possible to use it with the custom adapter option. 😉

pbteja1998:
> So sorry for this @pbteja1998, but we decided to drop builtin support for Fauna DB for now. it will be removed soon in a later release. it is a bit of an overhead to make sure it works, while there aren't even tests for it. As you did as well so good, it is still possible to use it with the custom adapter option. 😉

Not a problem. It works for my use-case. So it's OK.

PR #1137, opened 2021-01-17
fix: faunadb adapter, emailVerified value updated to fauna compatable…

adharshmk96:
… format.

**What**:

Updated query in faunadb adapter, email verified field

**Why**:

Faunadb wasn't updating emailVerified field

**How**:

Changed emailVerified ( of type new Date() ) to q.Time(emailVerified.toISOString())

**Checklist**:

- [ ] Documentation N/A
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/adapters/fauna/index.js
+++ b/src/adapters/fauna/index.js
@@ -43,7 +43,7 @@ const Adapter = (config, options = {}) => {
             email: profile.email,
             image: profile.image,
             emailVerified: profile.emailVerified
-              ? profile.emailVerified
+              ? q.Time(profile.emailVerified.toISOString())
               : false,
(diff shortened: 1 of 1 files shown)


balazsorban44:
So sorry for this @adharshmk96, but we decided to drop builtin support for Fauna DB for now. it will be removed soon in a later release. it is a bit of an overhead to make sure it works, while there aren't even tests for it. It is still possible to use it with the custom adapter option. 😉

adharshmk96:
No problem, I think it would be helpful for people if the adapter for fauna is kept as a reference or example though.

balazsorban44:
Check out #1146

PR #1168, opened 2021-01-21
feat: add option to define `NEXTAUTH_URL` via next runtime config

### LINKED ISSUE #1156: Adding options to define `NEXTAUTH_URL` not via env vars
Adding options to define `NEXTAUTH_URL` not via environment variables but via server handler creation parameters or client default parameters.

Allow custom and dynamic `NEXTAUTH_URL` value assignment via function parameters instead of environment variables, where in some cases cannot set environment variables or requires dynamic assignment.

Some edge cases (especially Vercel multi-domain deployments) requires different values for `NEXTAUTH_URL` (e.g. `example.com` and `staging.example.com`), but since `NEXTAUTH_URL` needs to be defined via environment variable and the value retrieval is mostly hardcoded inside `next-auth`, currently there is no way to use different values.

**Potential problems**
If adding a way to define `NEXTAUTH_URL` but still keep the option to define via environment variables, possible issues would be changing how assigning and using the default parameters work. Notable lines:

https://github.com/nextauthjs/next-auth/blob/0987f72acbbe655aedfcd8b9b6d18acf59207329/src/client/index.js#L25-L26

https://github.com/nextauthjs/next-auth/blob/0987f72acbbe655aedfcd8b9b6d18acf59207329/src/client/index.js#L313-L314

https://github.com/nextauthjs/next-auth/blob/0987f72acbbe655aedfcd8b9b6d18acf59207329/src/lib/jwt.js#L105-L108

https://github.com/nextauthjs/next-auth/blob/0987f72acbbe655aedfcd8b9b6d18acf59207329/src/server/index.js#L16-L20

https://github.com/nextauthjs/next-auth/blob/0987f72acbbe655aedfcd8b9b6d18acf59207329/src/server/index.js#L50

Some alternative approaches would be either automagically detect the URL, or use [Next.js runtime configuration](https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration) to store various default values including `NEXTAUTH_URL` (probably not the best approach since it's incompatible with automatic static optimization).

**Additional context**
This proposal issue was created because I am currently developing a Next.js based application which requires multiple subdomain and deployed using Vercel and we can't set dynamic env vars for `NEXTAUTH_URL` for different branches on Vercel.

Another context is that I've tried recreating the `next-auth` codebase using tsdx but since there's Docusaurus and various sources that I'm not familiar with, I didn't continue working on the rewrite (for now).

I am willing to help in any way, unless if https://github.com/nextauthjs/next-auth/pull/516#issuecomment-763126791 is on the way, I can help rewrite some TypeScript stuff and include this proposed feature if considered.

grikomsn:
**What**:

Resolves #1156, this PR adds an option to define `NEXTAUTH_URL` via [Next.js runtime configuration](https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration).

**Why**:

By design, NextAuth.js retrieves `NEXTAUTH_URL` value via environment variables. Some cases requires dynamically generated canonical URL (e.g. multi-stage environments and multi-tenancy websites) and cannot set environment variables dynamically (e.g. Vercel for prod/prev/dev environments).

With this proposed PR, NextAuth.js can read `NEXTAUTH_URL` value from Next.js runtime configurations via `publicRuntimeConfig.NEXTAUTH_URL` property on the project's `next.config.js` file.

**How**:

This is approached by creating a function which returns either from `NEXTAUTH_URL` environment variable or from the runtime configuration. If via env, the runtime config will be ignored.

**Checklist**:

- [X] Documentation
- [ ] Tests (see note below)
- [X] Ready to be merged

**Additional notes**:

- Building locally works and already being used in production deployment
- If needed, add another approach to compose `NEXTAUTH_URL` via [Vercel's system environment variables](https://vercel.com/docs/environment-variables#system-environment-variables)
- `next` is a dev dependency but is being used in server/client code, probably change as peer dependency?

**Note to maintainers:**

This PR is heavily opinionated and only serves as a temporary solution for those who really need a way to dynamically set `NEXTAUTH_URL`. If there is a better way to implement, or this approach is not aligned with NextAuth.js architecture, or anything, merge is optional. ✌🏻


williamluke4 (approved):

iaincollins:
I've left a comment on this in #1156.

I think I get where you are going with this but don't think this a direction we want to go in as I don't think it does what people would hope it does (in the case of most OAuth providers) and I think it would probably cause confusion.

I did considering building it this way originally, and explored this with v2.x (originally v2 used an option) but it turns out it creates some interesting problems; especially for API route handling on some hosting platforms; and a race condition that was hard to spot. I lamentably chose a single env var as the least worst option.

If you could frame the conceptional challenge - e.g. "signing in with Google on Pull Request instances" and include any constraints (e.g. must support JWT sessions / must support a database, etc. - perhaps we can some up with some approaches?

dpatrikeev (approved):

PR #1171, opened 2021-01-21
fix(prisma): replace findOne with findUnique

williamluke4:
Fixes the breaking change in Prisma **2.15** which removes the deprecated method `findOne` which was replaced with `findUnique` in **2.12**

### DIFF
--- a/src/adapters/prisma/index.js
+++ b/src/adapters/prisma/index.js
@@ -57,7 +57,7 @@ const Adapter = (config) => {
     async function getUser (id) {
       debug('GET_USER', id)
       try {
-        return prisma[User].findOne({ where: { id } })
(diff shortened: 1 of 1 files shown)


balazsorban44:
Thanks, but this has been resolved on https://github.com/nextauthjs/next-auth/blob/canary/src/adapters/prisma/index.js

you can try it out with `npm i next-auth@canary`

williamluke4:
@balazsorban44 This was intended as a patch for stable hence `main`. As you are about to do a release then closing this makes sense :smile:

oliviermtl:
In order to get next-auth to work with Prisma, we have to install next-auth like this `npm i next-auth@canary`.

I wonder why do I have to come here to get this info, or am I missing something here ?

PR #1174, opened 2021-01-21
fix(provider): use authed_user on slack instead of spotify

lukel97:
This fixes what looks like a typo from 3fcdd2265637e8c66adcbeaedf7d2a3335e8b661
Spotify doesn't nest the user inside an `authed_user` field


balazsorban44 (approved):
Thanks @bubba! 🎉 This is indeed a typo, which started in https://github.com/nextauthjs/next-auth/pull/895, and then I made a mistake in the refactoring. Very good catch!

Issue #989, opened 2020-12-24
Is it possible to avoid page reloading when using custom credentials?


danyel117:
**Your question**
Is it possible to avoid page reloading when using custom credentials? (like an SPA behaviour)

**What are you trying to do**
I'm using custom credentials because I'm using a customer's database which has its own api, and I've designed my own login page. When I hit login after entering credentials, I handle the authentication with my cusotmer's api inside the [...nextauth].js file. When authentication happens, the page goes to whichever page I've specified in the callback.

However, I would prefer to show a "loading" component and wait until authentication happens, listening the session inside a useEffect, for example. Once the session variable has data, I would then handle the logic without seing a reload on the page.

**Reproduction**
Implement the custom login page in the documentation and watch the redirect after hitting Sign In:
import React from 'react';
import { csrfToken } from 'next-auth/client';

export default function SignIn({ csrfToken }) {
  return (
    <form method='post' action='/api/auth/callback/credentials'>
      <input name='csrfToken' type='hidden' defaultValue={csrfToken} />
      <label>
        Username
        <input name='username' type='text' />
      </label>
      <label>
        Password
        <input name='password' type='text' />
      </label>
      <button type='submit'>Sign in</button>
    </form>
  );
}

SignIn.getInitialProps = async (context) => {
  return {
    csrfToken: await csrfToken(context),
  };
};

**Feedback**
*Documentation refers to searching through [online documentation](https://next-auth.js.org), code comments and issue history. The example project refers to [next-auth-example](https://github.com/iaincollins/next-auth-example).*

* [ ] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [x] Could not find relevant documentation
* [ ] Found the example project helpful
* [x] Did not find the example project helpful

YuriFontella:
I'm looking for that too.
It would be interesting if the singIn function had a return response instead of reloading the page, and we had to do something like this: 

Promise.reject ('/ login? error = true') 
The form data is lost and no user interaction occurs.

![image](https://user-images.githubusercontent.com/4920619/103154881-a59bd400-4779-11eb-9812-7a8c42788c25.png)

balazsorban44:
Hi there! Could someone tell me, ideally what would you like to be returned? Or would simply don't redirecting anywhere/reloading enough?

marco-koenen:
[...]

Issue #517, opened 2020-07-30
Add Custom Error Callback


BenjaminWFox:
**Summary of proposed feature**
Add a new callback `error` that allows for custom handling of errors.

**Purpose of proposed feature**
In Azure B2C there is a (strange, imo) implementation where, if the user clicks the 'password reset' link in the (Azure B2C controlled) authentication form it posts back to the callback URL with an error (example below). It would be nice to have a be able to write a handler that could check for this, or any, particular error code and act accordingly.

```
    error: 'access_denied',
    error_description: 'AADB2C90118: The user has forgotten their password.\r\n' +
      'Correlation ID: 40e2100f-c43f-4dc6-a01b-4b1002861c74\r\n' +
      ...
```

I'm sure there are other situations this might be handy. One I saw was #373 where this could be used with `signInCallbackResponse` to pass along a failure message `signIn`.

Note this would require changing the signature of the `signIn` callback, which I think would be a breaking change? But could look something like this:

```
const signInCallbackResponse = await callbacks.signIn(userOrProfile, account, OAuthProfile)
if (signInCallbackResponse.success === false) {
  return callbacks.error(error, redirect, `${baseUrl}${basePath}/error?error=${signInCallbackResponse.message}`)
}
```

**Detail about proposed feature**
This is a very bare-bones example of what implementation might look like:

https://github.com/nextauthjs/next-auth/compare/main...BenjaminWFox:enhancement/explicit-error-callback

Implemented it looks like this. As long as I can figure out the structure of `error` (see potential problems) I can handle it as needed:
```
callbacks: {
  signIn: async (user, account, profile) => {
    console.log('Callback: signIn: user/account/profile', user, account, profile)

    return Promise.resolve(true)
  },
  error: async (error, redirect, redirectUrl) => {
    console.log('NEW Error callback: ', redirectUrl)

    return redirect(`https://some.custom.redirect.url/.../...`)
  }
}
```

**Potential problems**
- Error standardization/what to pass as first prop `error`? There are [areas in the code](https://github.com/nextauthjs/next-auth/blob/main/src/server/routes/callback.js#L138) that call errors that are not triggered by caught or thrown errors. [Some places](https://github.com/nextauthjs/next-auth/blob/main/src/server/routes/callback.js#L37) have a variable `error` that may or may not be an instance of `Error`. [Some places](https://github.com/nextauthjs/next-auth/blob/main/src/server/routes/callback.js#L76) are also just a bit confusing.
[...]

PR #1180, opened 2021-01-22
Add: Tutorial for credentials provider using custom backend

twilink:
**What**:

An in-depth tutorial of how to use credentials provider with a custom backend including how to display error messages received from server (400-500 codes) on your custom login page

**Why**:

Was unable to find a detailed example of using the credentials provider with a custom backend and error handling. Also saw questions in stackoverflow (https://stackoverflow.com/questions/65297986/how-to-ensure-auth-with-axios-in-custom-backend-with-credentials-provider/65845618#65845618) and nextauth discussions requesting the same (https://github.com/nextauthjs/next-auth/discussions/661#discussioncomment-154364)

Hope it helps!

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [x] Ready to be merged

### DIFF
--- a/www/docs/tutorials.md
+++ b/www/docs/tutorials.md
@@ -74,6 +74,6 @@ This tutorial walks step by step on how to get Sign In with Apple working (both 
 
 ### [Using Credentials Provider with a Custom Backend](https://dev.to/twisha/using-credentials-provider-with-a-custom-backend-in-nextauth-js-43k4)
 
-This `dev.to` articles walks you through adding authentication using credentials provider with a custom backend. It also includes handling error responses from your server and displaying them on your custom login page. By Twisha M.
+This `dev.to` articles walks you through adding authentication using credentials provider with a custom backend. It also includes handling error responses from your server and displaying them on your custom login page. By Twisha Mankad.
 
 
--- a/www/docs/tutorials.md
+++ b/www/docs/tutorials.md
@@ -71,3 +71,9 @@ This `dev.to` tutorial walks one through adding NextAuth.js to an existing proje
 ### [Adding Sign in With Apple Next JS](https://thesiddd.com/blog/apple-auth)
 
 This tutorial walks step by step on how to get Sign In with Apple working (both locally and on a deployed website) using NextAuth.js.
+
+### [Using Credentials Provider with a Custom Backend](https://dev.to/twisha/using-credentials-provider-with-a-custom-backend-in-nextauth-js-43k4)
+
+This `dev.to` articles walks you through adding authentication using credentials provider with a custom backend. It also includes handling error responses from your server and displaying them on your custom login page.
+
+
--- a/www/docs/tutorials.md
+++ b/www/docs/tutorials.md
@@ -74,6 +74,6 @@ This tutorial walks step by step on how to get Sign In with Apple working (both 
 
 ### [Using Credentials Provider with a Custom Backend](https://dev.to/twisha/using-credentials-provider-with-a-custom-backend-in-nextauth-js-43k4)
 
-This `dev.to` articles walks you through adding authentication using credentials provider with a custom backend. It also includes handling error responses from your server and displaying them on your custom login page.
+This `dev.to` articles walks you through adding authentication using credentials provider with a custom backend. It also includes handling error responses from your server and displaying them on your custom login page. By Twisha M.
 
 


balazsorban44 (changes requested):
Thank you! I'll look at it in-depth sometime later. Could you in the meantime remove the `Promise.resolve` and `Promise.reject` wrappers in your code snippets? They are unnecessary, and only cause confusion amongst users.

Also, feel free to disagree, but could you change property arrow functions to methods like so:
this:
```js
jwt: async (token, user) => {
  if (user) {
    token.accessToken = user?.data.token
  }
  return token
},
```
to this (no need for optional chaining on user, you already checked in the if block if user is defined):
```js
async jwt(token, user) {
  if (user) {
    token.accessToken = user.data.token
  }
  return token
}
``` 
(Look for similarity defined property functions and replace those as well)

This might only be my personal preference, but then it would also better match with our code examples in the docs.

With our `canary` release (hopefully soon to be `latest`), you can also simplify the following:
```js
const options = {
  providers,
  callbacks
  pages: {
    error: '/login' // Changing the error redirect page to our custom login page
  }
}

```
to
```js
export default NextAuth({
  providers,
  callbacks
  pages: {
    error: '/login' // Changing the error redirect page to our custom login page
  }
})
```

You don't have to change this, since it is not yet in the stable release, but please keep an eye on when this gets to stable and change your article accordingly, so that we know the tutorials out there are up-to-date 🙂


Again, I have to read through the tutorial again, these small improvements/changes were from the top of my head only.

Nice work either way!

twilink:
Thank you for the comment! 🙂

I have updated the docs to reflect the changes you suggested and will keep an eye out for the stable release to update the export function.

Please let me know if any other changes are required. Appreciate the detailed review! 🙂

balazsorban44 (commented):
TL;DR

**Since we already have an issue/request to better handling this problem, I am not going to accept this PR, but I appreciate the effort anyway!**

Some further tips/suggestions for your article though:
1. `pages/api/[..next-auth].js` -> `pages/api/auth/[...next-auth].js`
2. could you maybe remove `axios` to keep the example code as simple as possible? The less external libraries, the better. A `fetch` polyfill is included in Next.js already, and it is barely more code.
3. The authorize callback could be as simple as
```js
async authorize(user) {
  return fetch("https://myapi.com/login", {
      headers: { "Content-Type": "application/json", method: "POST" },
      body: JSON.stringify({user})
  })
}
```
4. the `signIn` callback is unnecessary in the form you defined it, just remove it.
5. The `session` callback also needs an if block to check if `session` already have the accessToken, I think. Could you check if your example code actually returns an `accessToken` on `/api/auth/session`?
6. Would be nice to show the whole custom login page code, not just the `handleLogin` function.


Further explanation to the first sentence: To be honest, I think the best way would be if we just let the `signIn` function return something (like errors, and success), rather than requiring users to create custom pages and all that fuss. (see #517, #989)

twilink:
> TL;DR
> 
> **Since we already have an issue/request to better handling this problem, I am not going to accept this PR, but I appreciate the effort anyway!**
> 
> Some further tips/suggestions for your article though:
> 
>     1. `pages/api/[..next-auth].js` -> `pages/api/auth/[...next-auth].js`
> 
>     2. could you maybe remove `axios` to keep the example code as simple as possible? The less external libraries, the better. A `fetch` polyfill is included in Next.js already, and it is barely more code.
> 
>     3. The authorize callback could be as simple as
> 
> 
> ```js
> async authorize(user) {
>   return fetch("https://myapi.com/login", {
>       headers: { "Content-Type": "application/json", method: "POST" },
>       body: JSON.stringify({user})
>   })
> }
> ```
> 
>     1. the `signIn` callback is unnecessary in the form you defined it, just remove it.
> 
>     2. The `session` callback also needs an if block to check if `session` already have the accessToken, I think. Could you check if your example code actually returns an `accessToken` on `/api/auth/session`?
> 
>     3. Would be nice to show the whole custom login page code, not just the `handleLogin` function.
> 
> 
> Further explanation to the first sentence: To be honest, I think the best way would be if we just let the `signIn` function return something (like errors, and success), rather than requiring users to create custom pages and all that fuss. (see #517, #989)

Okay! Yeah I agree having the `signIn` function return errors would be the best way to handle these scenarios. Looking forward to having that option in future release.

Thank you for taking the time to go through the article and for the suggestions. I shall update the article accordingly.

PR #1184, opened 2021-01-22
feat(provider): re-add state, expand protection provider options 

### LINKED ISSUE #1172: New provider property: protection (support pkce, state, none)
We should let users define what protective measures should be taken while logging into an OAuth provider, even if it is a custom one. The options are `"state"` (using the `CSRF` token), `"pkce"` (see the [PKCE spec](https://tools.ietf.org/html/rfc7636), also, introduced in #941) or `"none"`.

So it should look like this:

```js
{
...
  protection: "pkce" // or "state" or "none"
...
}
```

We should decide on what should be the default value. Currently state is always checked, except if `state: false` is set in the options, so `protection: "state"` seems to be the natural choice here. But is PKCE something we would like to encourage/prefer over state?

**Potential problems**
As `state` has been removed in `canary`, strictly speaking, it MIGHT be a breaking change for custom providers. I can easily add the `state` option back, and show a deprecation warning in favour of `protection: "state"`.

In the current stable release of `next-auth`, a provider can pass an option called `state: boolean`. This has been removed in the `canary` in #1022 to simplify the provider options. I was under the assumption that only Apple does not support this, so I hardcoded an Apple specific solution:
https://github.com/nextauthjs/next-auth/blob/47a5c9d83037ad690784deef7bc24cfedd914224/src/server/lib/oauth/callback.js#L25-L30

In retrospect, I think it was not a good alternative, but with the new proposed option `protection: "none"`, should fix this. (question. Anyone knows if Apple supports PKCE?)

Interested in what the community thinks!

balazsorban44:
**What**:

An earlier PR (#1023) removed the `state` provider option, as I was under the assumption, that only Apple utilized it. This was wrong. This PR adds this option back, with one caveat. Users using it will get a warning from now on, that will tell them to use `protection: "state"` or `protection: "none"` instead. 

**Why**:

Unifies protection, one can now set either `"state" | "pkce" | "none"` to `protection`. If the property is not set, it will default to `"state"`, as earlier `next-auth` version defaulted to this. (@iaincollins do we want to default to something else?)

**How**:

Added a new state handler, docs, and a deprecation warning if the `state` provider option is present.

**Checklist**:

- [x] Documentation
- [x] Tests (Tested locally, with the Auth0 provider)
- [x] Ready to be merged
      

fixes #1172

### DIFF
--- a/src/providers/salesforce.js
+++ b/src/providers/salesforce.js
@@ -8,7 +8,7 @@ export default (options) => {
     accessTokenUrl: 'https://login.salesforce.com/services/oauth2/token',
     authorizationUrl: 'https://login.salesforce.com/services/oauth2/authorize?response_type=code',
     profileUrl: 'https://login.salesforce.com/services/oauth2/userinfo',
-    state: false,
+    protection: 'none', // REVIEW: Can we use "pkce" ?
     profile: (profile) => {
       return {
         ...profile,
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -12,7 +12,7 @@ import * as routes from './routes'
 import renderPage from './pages'
 import csrfTokenHandler from './lib/csrf-token-handler'
 import createSecret from './lib/create-secret'
-import * as pkce from './lib/pkce-handler'
+import * as pkce from './lib/oauth/pkce-handler'
 
 // To work properly in production with OAuth providers the NEXTAUTH_URL
 // environment variable must be set.
@@ -145,9 +145,7 @@ async function NextAuthHandler (req, res, userOptions) {
           return render.signout()
         case 'callback':
           if (provider) {
-            const error = await pkce.handleCallback(req, res)
-            if (error) return res.redirect(error)
-
+            if (await pkce.handleCallback(req, res)) return
             return routes.callback(req, res)
           }
           break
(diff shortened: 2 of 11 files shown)


src/providers/salesforce.js
```diff
@@ -8,7 +8,7 @@ export default (options) => {
     accessTokenUrl: 'https://login.salesforce.com/services/oauth2/token',
     authorizationUrl: 'https://login.salesforce.com/services/oauth2/authorize?response_type=code',
     profileUrl: 'https://login.salesforce.com/services/oauth2/userinfo',
-    state: false,
+    protection: 'none', // REVIEW: Can we use "pkce" ?
```
balazsorban44, line 11: @mmahalwy Do you know if Salesforce supports PKCE? https://tools.ietf.org/html/rfc7636

src/server/index.js
```diff
@@ -63,6 +64,13 @@ async function NextAuthHandler (req, res, userOptions) {
     const providers = parseProviders({ providers: userOptions.providers, baseUrl, basePath })
     const provider = providers.find(({ id }) => id === providerId)
 
+    if (provider &&
+      provider.type === 'oauth' && provider.version?.startsWith('2') &&
+       (!provider.protection && provider.state !== false)
+    ) {
+      provider.protection = 'state' // Default to state, as we did in 3.1 REVIEW: should we use "pkce" or "none" as default?
```
balazsorban44, line 71: @iaincollins should we default to something else than "state"?

PR #1198, opened 2021-01-27
Add a null check to the window 'storage' event listener

### LINKED ISSUE #1125: Runtime error "Cannot read property 'event' of null" when testing with Cypress
**Describe the bug**
When using NextAuth in conjunction with Cypress in next dev mode, occasionally an unhandled runtime error "TypeError: Cannot read property 'event' of null" shows up. 

**Steps to reproduce**
Run Cypress test with running local Next dev server (npm run dev). Sign in using Cypress ui handler with credentials on /api/auth/signin, then test protected api routes and/or protected pages. Tests are all passing, but in the visual representation of the UI, the error shows up, pointing to next-auth/dist/client/index.js 60:0 (see screenshot). This error is not thrown if Next has been started with next build / next start. 

**Expected behavior**
No error when running in dev mode. 

**Screenshots or error logs**

Bildschirmfoto 2021-01-16 um 11 07 40

**Additional context**
node v14.15.1
npm 6.14.10
next: "10.0.1",
next-auth: "^3.1.0",
and for Cypress:
Cypress package version: 6.2.1
Cypress binary version: 6.2.1
Electron version: 11.1.1
Bundled Node version: 12.18.3

**Feedback**

* [x ] Could not find relevant documentation

dmmulroy:
**What**:

Adding a null check to `message` (which is the `newValue` off of a [StorageEvent](https://developer.mozilla.org/en-US/docs/Web/API/StorageEvent)) to prevent `Runtime error "Cannot read property 'event' of null"` when testing with Cypress. This could likely also occur in other iframe applications that use the storage API. 
Fixes #1125 

**Why**:

It's currently throwing errors in Cypress and the additional null check on a field that _can_ be null adds extra resilience 

**How**:
Adding a null check on the value of `message`.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -48,7 +48,7 @@ if (typeof window !== 'undefined') {
     window.addEventListener('storage', async (event) => {
       if (event.key === 'nextauth.message') {
         const message = JSON.parse(event.newValue)
-        if (message && message.event && message.event === 'session' && message.data) {
+        if (message?.event=== 'session' && message.data) {
           // Ignore storage events fired from the same window that created them
           if (__NEXTAUTH._clientId === message.clientId) {
             return
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -48,7 +48,7 @@ if (typeof window !== 'undefined') {
     window.addEventListener('storage', async (event) => {
       if (event.key === 'nextauth.message') {
         const message = JSON.parse(event.newValue)
-        if (message.event && message.event === 'session' && message.data) {
+        if (messsage && message.event && message.event === 'session' && message.data) {
           // Ignore storage events fired from the same window that created them
           if (__NEXTAUTH._clientId === message.clientId) {
             return
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -48,7 +48,7 @@ if (typeof window !== 'undefined') {
     window.addEventListener('storage', async (event) => {
       if (event.key === 'nextauth.message') {
         const message = JSON.parse(event.newValue)
-        if (messsage && message.event && message.event === 'session' && message.data) {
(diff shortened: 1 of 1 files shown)


iaincollins:
I don't fully understand in what circumstance this occurs, but seems like a reasonable conditional check to add!

iaincollins (approved):

src/client/index.js
```diff
@@ -48,7 +48,7 @@ if (typeof window !== 'undefined') {
     window.addEventListener('storage', async (event) => {
       if (event.key === 'nextauth.message') {
         const message = JSON.parse(event.newValue)
-        if (message.event && message.event === 'session' && message.data) {
+        if (message && message.event && message.event === 'session' && message.data) {
```
balazsorban44, line 51:
```suggestion
        if (message?.event=== 'session' && message.data) {
```

dmmulroy:
> I don't fully understand in what circumstance this occurs, but seems like a reasonable conditional check to add!

@iaincollins I believe this occurring when Cypress is tearing down before/after tests and is clearing localStorage, and because it happens outside the context of the document that the listener is in the `storage` event is being fired.  On `clear` or the deletion of key from storage, the value of `newValue` will be `null` effectively throwing the error we're seeing. 

If this is true and what's going on, this error could occur at runtime in applications using `next-auth` that are also being used within an iframe in which the parent context clears storage or deletes keys from storage.

balazsorban44 (changes requested):
looks good, and simple, but could you maybe simplify it as I suggested?

balazsorban44 (approved):
thank you!

Issue #1132, opened 2021-01-16
Explain how to populate Provider in _app


balazsorban44:
In the current version of the docs https://next-auth.js.org/getting-started/client#provider we do not explain how the `pageProps.session` prop is populated. It is not something that comes magically by `next-auth` as explained here: https://github.com/nextauthjs/next-auth/issues/1130#issuecomment-761689414

Simply adding a `Provider` in the `_app` with a `session={pageProps.session}` file won't magically add any session. The props provided to the `_app` component must come from somewhere. In this case, either from `getInitialProps` or `getServerSideProps`. Do note that this does not have to be in the `_app` file, so you don't necessarily opt-out of any Next.js optimization. Add it only to those pages that need auth.

Eg.:

```js
// pages/index.js
import {getSession} from "next-auth/client"
...
export async function getServerSideProps(ctx) {
  return {
    props: {
      session: await getSession(ctx)
    }
  }
}
...
```

This then will be available through `pageProps.session`, and this is the value you have to provide to the `Provider`. Thanks to this, any subsequent `useSession()` hook calls will only get the client-cached session value, instead of re-visiting the backend. You can configure how long you want to cache the session before getting the new one from the backend. Documented here: https://next-auth.js.org/getting-started/client#options

It would be nice to incorporate this into the docs, to avoid further confusion.

kriscarle:
Hi, I just wanted to leave a note that the example in the NextJS repo appears to implement this incorrectly too? https://github.com/vercel/next.js/tree/canary/examples/with-next-auth

balazsorban44:
Absolutely, many projects does that, and I think it is easier for us to manage a single example, instead of two, where one is even in another repo.

But we definitely must add this to our docs as well, I'm glad that at least our example project does it correctly 😅:

https://github.com/nextauthjs/next-auth-example/blob/918ef6de41b56840ac4a7c4e95f542c06a6a135d/pages/server.js#L35

jeremyhaile:
Where does this come from in Next.JS 13 with appDir?

Issue #1078, opened 2021-01-10
Add tutorial on passing access token to session (good first issue)


balazsorban44:
**Summary of proposed feature**
A large number of issues are created that asks how to pass `access_token` to the session.

**Purpose of proposed feature**
Currently, we point the users to the `jwt` and `session` [callbacks documentation](https://next-auth.js.org/configuration/callbacks), but since it's such a highly requested thing, we might just want to create a specific tutorial under https://next-auth.js.org/tutorials that explains this in details. 

**Detail about proposed feature**
We could either add it directly to the tutorials as a standalone page, or add la link to an external post, preferably with code examples, maybe a working demo app as well, that is live on an accessible url.

**Describe any alternatives you've considered**
Docs are clearly not good enough, as we are still asked quite a lot.

*Please indicate if you are willing and able to help implement the proposed feature.*

This is up for grab, marked as a good first issue.

jackHedaya:
I'd be happy to tackle this issue once I figure out the solution in #1068 :)

PR #1207, opened 2021-01-28
client-side authentication in Next.js with NextAuth.js

ejirocodes:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged


balazsorban44 (commented):
Please add PRs against the `canary` branch next time.

It is so unfortunate that we document this badly, but I see your tutorial does not mention either that the `pageProps.session` value in `_app` must come from somewhere! See #1132

The "How NextAuth.js works" section seems to be not fully explained. Although you are right about the default session value (we keep it minimal to be secure by default), you can extend it as you like, with the help of the `jwt` and `session` callbacks (see: https://next-auth.js.org/configuration/callbacks)

I think it is pretty important to say here that this is where you can provide an `access_token` that can be used to call APIs from the client that needs authentication. Since your tutorial tries to focus on client-side features of NextAuth.js, I feel it is very important to name. (We should actually add this to the documentation. Related issue: #1078)

Apart from that, the tutorial looks nice, but I don't feel it explains anything else the documentation already tells you. The requested changes should more likely be a part of the official docs, rather than in an arbitrary tutorial.

If you are interested in adding those pinpointed changes to the official docs, that would be more than welcome!

ejirocodes:
> Please add PRs against the `canary` branch next time.
> 
> It is so unfortunate that we document this badly, but I see your tutorial does not mention either that the `pageProps.session` value in `_app` must come from somewhere! See #1132
> 
> The "How NextAuth.js works" section seems to be not fully explained. Although you are right about the default session value (we keep it minimal to be secure by default), you can extend it as you like, with the help of the `jwt` and `session` callbacks (see: https://next-auth.js.org/configuration/callbacks)
> 
> I think it is pretty important to say here that this is where you can provide an `access_token` that can be used to call APIs from the client that needs authentication. Since your tutorial tries to focus on client-side features of NextAuth.js, I feel it is very important to name. (We should actually add this to the documentation. Related issue: #1078)
> 
> Apart from that, the tutorial looks nice, but I don't feel it explains anything else the documentation already tells you. The requested changes should more likely be a part of the official docs, rather than in an arbitrary tutorial.
> 
> If you are interested in adding those pinpointed changes to the official docs, that would be more than welcome!



> Please add PRs against the `canary` branch next time.
> 
> It is so unfortunate that we document this badly, but I see your tutorial does not mention either that the `pageProps.session` value in `_app` must come from somewhere! See #1132
> 
> The "How NextAuth.js works" section seems to be not fully explained. Although you are right about the default session value (we keep it minimal to be secure by default), you can extend it as you like, with the help of the `jwt` and `session` callbacks (see: https://next-auth.js.org/configuration/callbacks)
> 
> I think it is pretty important to say here that this is where you can provide an `access_token` that can be used to call APIs from the client that needs authentication. Since your tutorial tries to focus on client-side features of NextAuth.js, I feel it is very important to name. (We should actually add this to the documentation. Related issue: #1078)
> 
> Apart from that, the tutorial looks nice, but I don't feel it explains anything else the documentation already tells you. The requested changes should more likely be a part of the official docs, rather than in an arbitrary tutorial.
> 
> If you are interested in adding those pinpointed changes to the official docs, that would be more than welcome!

@balazsorban44 I'll add these issues you've pointed out to the official docs and also update my article! Sounds cool?

balazsorban44:
Awesome!

PR #1209, opened 2021-01-28
Adding support for idToken storage

sponte:
This pull request introduces idToken support by adding idToken to account object. 

Ability to read idToken gives users ability to use the token as means of authenticating
with their backend services.

idToken is read from provider's response and persisted into account object.

**Checklist**:

- [ ] Documentation
  - I scanned through documentation but can't find the right place to document idToken's addition. Can you recommend a place?
- [ ] Tests
- [ ] Ready to be merged


balazsorban44:
Just a heads up that #1211 solves at least part of this PR

PR #1211, opened 2021-01-29
refactor: provide raw idToken through account object

balazsorban44:
**What**:

Move `idToken` from `profile` to `account` object in callbacks.

This PR also forwards the `expires_in` value for `accessTokenExpires`, if available. This has long been simply defined as `null`.

**Why**:
The `idToken` has been provided "as is" to the `signIn` and `jwt` callbacks since #1024 through the `profile` object. Although, both `accessToken` and `refreshToken` have been on the `account` object.
To better group the tokens, this PR simply moves `profile.idToken` to `account.idToken`.

### DIFF
--- a/src/server/lib/oauth/callback.js
+++ b/src/server/lib/oauth/callback.js
@@ -36,8 +36,7 @@ export default async function oAuthCallback (req) {
     }
 
     try {
-      const { accessToken, refreshToken, results } = await client.getOAuthAccessToken(code, provider, pkce.code_verifier)
-      const tokens = { accessToken, refreshToken, idToken: results.id_token }
+      const tokens = await client.getOAuthAccessToken(code, provider, pkce.code_verifier)
       let profileData
       if (provider.idToken) {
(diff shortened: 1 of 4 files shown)


src/server/lib/oauth/callback.js
```diff
@@ -121,9 +110,8 @@ async function _getProfile ({
         provider: provider.id,
         type: provider.type,
         id: profile.id,
-        refreshToken,
-        accessToken,
-        accessTokenExpires: null
+        ...tokens,
+        accessTokenExpires: results.expires_in ?? null
```
balazsorban44, line 114: @iaincollins any reason why the raw result of the tokens fetch call not simply spread onto the `account` object or added as a `tokens` property or something? Currently, It seems to only be remapping of only **some** of those values, plus the `accessTokenExpires` is null'ed out. Also, I don't see the need for `provider.id` being added twice(?) both as `provider` and `id`.

PR #1213, opened 2021-01-29
feat(provider): Add medium

defvova:
Add Medium as a provider.


balazsorban44 (approved):
It has been a long time since a provider was added so smoothly 😅 you added it to the documentation in all the necessary places, no funky Medium required specific behavior, it just uses OAuth 2.0 as it should. So bad that email is not returned though!

Thanks anyway 🙂

Issue #1192, opened 2021-01-24
Custom Login Page Error Handling?


Lightgaming:
I have a **custom login page** that uses the signIn method (LDAP Credential Provider) which works perfectly. 
**When the wrong credentials are entered the user is redirected to an Error page, which I don't want.**
I don't want this error redirect.
![image](https://user-images.githubusercontent.com/18364007/105645239-a8820580-5e9a-11eb-872d-b662eeb269b0.png)

Is there a way to use the signIn method like this?:

![image](https://user-images.githubusercontent.com/18364007/105645166-2d205400-5e9a-11eb-8582-a3956d20408b.png)

Or maybe something similar, so I can handle the error myself?

balazsorban44:
You could try creating a custom error page as well? you will receive the type of error as a query parameter to handle the different cases

Lightgaming:
I could do that, this would solve my redirect problem since it's just a route, but how would I get the error message back in my React Login component? 

I want it that way: When the user clicks login and the creds are wrong I get the error in my React Logic and show some nicely formatted error message in the form, so the user knows what's wrong

balazsorban44:
Duplicate of #989

balazsorban44:
@Lightgaming follow #1219 :wink:

Taha-mlaiki:
@Lightgaming  did you resolve your problem , because I'm facing it to right now
??

Lightgaming:
@Taha-mlaiki unfortunately I don't really know. It was a school project and I don't have access to the code anymore

Issue #944, opened 2020-12-10
When clicking sign in button, How can I check if the username and password is valid or not without redirecting the sign in page from modal?


wildorwilliamgit:
When clicking sign in button, How can I check if the username and password is valid or not without redirecting the sign in page from modal?

Thanks for your efforts.
After searching the auth features related to next.js, I think this is the best and complete auth library.
Current I had a problem working with this next-auth library in my frontend written in next.js.
I know this next-auth library makes the sign in page automatically from the providers in [...nextauth].js file.
**I would like to sign in by calling api when clicking sign in button, without redirecting the sign in page.**
In other words, now I am going to show the sign in page as modal or by poping up it.
So when clicking the sign in button, I am going to check if the credentials is valid or not without redirecting the new sign in page from the modal.
I have researched all of the next-auth.org pages, but there are no things related to this problem.
Any helps will be very thankful.

Issue #878, opened 2020-11-24
How to avoid redirects on failed sign in?


peculiarity:
**Your question**
How to avoid redirects on failed sign in?

**What are you trying to do**
I have custom sign in page. I would like to display an error message below the sign in form if the user provided wrong username and password. I create a `fetch` request `onSubmit` but the response is a redirect to `/?` which is weird. I have no idea why it redirects to `/?` and how to get the error.

**Feedback**
*Documentation refers to searching through [online documentation](https://next-auth.js.org), code comments and issue history. The example project refers to [next-auth-example](https://github.com/iaincollins/next-auth-example).*

* [ ] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [X] Could not find relevant documentation
* [ ] Found the example project helpful
* [ ] Did not find the example project helpful

Issue #754, opened 2020-10-07
Return an error message to client in the same page


natsathorn:
**Your question**
How to return the message to the login form?

**What are you trying to do**
I'm using Credentials provider which will just fire a request to another server for login. It's work fine when everything go as plan. In the case that login is not success, I can only send the error message via querystring which make the page reloaded.
Is there any other way to send the errorMessage back to the form without reload it.

FYI. I used `signIn('myAuth', { ...data })` on my login form.

**Feedback**
In this [section](https://next-auth.js.org/providers/credentials#example), it only shows these two options.
```
// You can also Reject this callback with an Error or with a URL:
return Promise.reject(new Error('error message')) // Redirect to error page
return Promise.reject('/path/to/redirect')        // Redirect to a URL
```

* [ ] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [x] Could not find relevant documentation
* [ ] Found the example project helpful
* [x] Did not find the example project helpful

bdub9:
I'm having the same issue.  Desired use case to skip a redirect:

1. Check if a user/password combo exists or password match.
2. They don't. Would like to Promise.reject with an error code or error message WITHOUT redirecting. Doesn't seem like there's a way to skip the redirect, even Promise.reject('/login?errorMessage=tryAgain') here reloads the page. How can we just send down a simple error message in a response so our client can handle it?

mdesilva:
I'm having the same issue as well. Would be nice if we can just do the login RESTfully and do the redirection ourselves.

MateuszWawrzynski:
*bump*
@iaincollins 

I'm looking for the answer for exactly the same problem.
Any ideas?

Is there any way to handle error code on the signin page without redirecting to "error" page and just show the error under the login form?

mdesilva:
You can return the error as a query parameter to the login page

PR #1219, opened 2021-01-31
feat(provider): option to disable client-side redirects (credentials)

### LINKED ISSUE #989: Is it possible to avoid page reloading when using custom credentials?

balazsorban44:
**What**:

In case the user is using the [Credentials](https://next-auth.js.org/providers/credentials) provider, make it possible to disable the redirection after the response from the server.

It also utilizes `URLSearchParams` to encode the body of `POST` requests, as Next.js will polyfill it for us.

Certain status codes were added to errors on the server-side to provide the user with more information about the error.

**Why**:
Example: In case the sign-in flow breaks because the user has given the wrong credentials, it may be desirable to handle it without a page reload.

**How**:

A new option has been added to the `signIn` and `signOut` client-side functions.

**Sign in**
> NOTE: The `redirect` option will do nothing unless you set the provider to `"credentials"`.

To preserve backward compatibility, the default will be to always redirect, so in theory, setting `{redirect: true}` statically won't be necessary.
```ts
interface SignInResponse {
  /** The reason for why the login process has stopped */
  error: string | null
  /** @see https://developer.mozilla.org/en-US/docs/Web/API/Response/status */
  status: number
  /** @see https://developer.mozilla.org/en-US/docs/Web/API/Response/ok */
  ok: boolean
}

const response: SignInResponse = await signIn("credentials", { redirect: false })
```
**Signout**
Provider agnostic setting. This is less likely to be used by users, only added to match the `signIn` behavior.
```ts
interface SignOutResponse {
  /** The URL the user should have been redirected to. (Same page by default) */
  url: string
}
const response: SignOutResponse = await signOut({ redirect: false })
```
**Checklist**:

- [ ] Documentation
- [x] Tests (Tested it locally so far, through the developer application, looking good there)
- [ ] Ready to be merged
      

This has been requested several times:

Resolves #989
Also related:
- Issues/PRs: #1180,#517, #1192, 
- Discussions: #944, #754, #878

### DIFF
--- a/components/header.js
+++ b/components/header.js
@@ -96,6 +96,11 @@ export default function Header () {
               <a>API</a>
             </Link>
           </li>
+          <li className={styles.navItem}>
+            <Link href='/credentials'>
+              <a>Credentials</a>
+            </Link>
+          </li>
         </ul>
       </nav>
     </header>
--- a/pages/api/auth/[...nextauth].js
+++ b/pages/api/auth/[...nextauth].js
@@ -2,7 +2,6 @@ import NextAuth from "next-auth"
 import Providers from "next-auth/providers"
 
 export default NextAuth({
-  // https://next-auth.js.org/configuration/providers
   providers: [
     Providers.GitHub({
       clientId: process.env.GITHUB_ID,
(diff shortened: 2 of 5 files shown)


cgerikj:
Thank you, this is perfect! :)

StefanSelfTaught:
This would work for email signin too so it won't redirect you to another page just to tell you 'go check you email' or for errors. @balazsorban44

balazsorban44:
thanks, I'll check that out as well!

yelnyaface:
hi, has the documentation for this been added?

balazsorban44:
See:

https://next-auth.js.org/getting-started/client#using-the-redirect-false-option-1

shozib13:
In the top of the example code const `response: SignInResponse = await signIn("credentials", { redirect: false })` response type  is SignInResponse  but typescript showing me error as `Object is possibly 'undefined'.ts(2532)`

image

PR #1223, opened 2021-01-31
feat(ts): preliminary TypeScript support

balazsorban44:
**What**:

Initial TypeScript support, without actually adding TS to the project, through [JSDoc](https://jsdoc.app/)

**Why**:

There are many requests to migrate the project to TypeScript and although it would be awesome to do it overnight, it would require a lot of effort to do it totally right. Instead, we could do an incremental migration that does not put pressure on the core contributors.

**How**:

Leveraging JSDoc interop with VSCode IntelliSense, one can have type suggestions in the IDE, without actually writing TypeScript code.

The way I see it, a good migration path for us could be something like this:

1. Add JSDoc comments to user-facing API functions/methods
2. Add .d.ts files, and use those for functions defined in 1. (eg. you can do * @param {import("next").NextApiRequest} req and at least VSCode IntelliSense will work)
3. Add JSDoc for all internal functions
4. Repeat 2. only for internal functions
5. enable // @ts-check at the top of the files, and fix remaining errors (use any for faster iterations)
6. add TS compilation, merge .d.ts and .js files into .ts files.
7. retire @types/next-auth with a deprecation message
8. expose the .d.ts types through the types package.json property

In this PR, I do stuff related to the above list, but I will be a bit all over the place in the beginning. I would like to do this migration without burning out, so I think after this PR is merged, migration will happen incrementally, meaning whenever I touch some code, I will see if I can improve the TS support just a bit every time. Eventually, we will be able to do the full conversion. Until then, please be patient 🙂

Suggestions ARE welcome! Cheers


benbender:
Hey @balazsorban44,

first let me thank you for your effort! Native TS-support will be much appreciated!

I've installed the latest canary and expected it, after seeing the changelog, to have (preliminary) typings for the package. But there doesn't seem to be any. Comments are stripped out in `dist/` and no `*.d.ts` in place.

After seeing the PR it is not quite obvious to me what the expected state of TS-support is like atm. These are only preparations and `@types/next-auth` is needed anyways. Correct?

I don't wan't to push anyone as this is open-source and I appreciate every minute someone puts into such a project to make my life easier! But there might be a (small) problem with expectation management in regarding to the changelog and the state of the implementation. It might be a good idea to clearify this a bit before the final release of v3.3!

Thanks again!

balazsorban44:
I see. It should have not been in the changelog, or said something like "preliminary work to to support TS later".

Sorry for the confusion! Currently I would recommend you using `@types/next-auth` as we have a long way to go before we can confidentially recommend using the built-in types.

benbender:
No big problem - as I said: more like a warning that it might be confusing if the final is released. Thanks again!

PR #1227, opened 2021-01-31
Adding EVEOnline provider

geraldm74:
Adding EVE Online Provider

I need to use EVE Online as a SSO provider for a project I am working on and I am sure other people in the future will use this Provider as well.

These changes were implemented using vs code and tested to work correctly.

**Checklist**:

- [X] Documentation
- [X] Tests
- [X] Ready to be merged
      

Adding a new provider for EVE Online SSO.

Documentation about EVE Online oauth2 authentication method can be found here:
https://developers.eveonline.com/blog/article/sso-to-authenticated-calls


balazsorban44 (changes requested):
Thank you! But before merging, it would be nice to add some minimal documentation about this provider, so people can start using it much easier.

You can see many examples here: https://github.com/nextauthjs/next-auth/tree/canary/www/docs/providers

It is mainly a link to how to configure a client on the platform, a simple code example, and maybe some tips/warnings users of this provider should be aware of.

You should then also add this new provider under https://github.com/nextauthjs/next-auth/blob/canary/www/providers.json

geraldm74:
> Thank you! But before merging, it would be nice to add some minimal documentation about this provider, so people can start using it much easier.
> 
> You can see many examples here: https://github.com/nextauthjs/next-auth/tree/canary/www/docs/providers
> 
> It is mainly a link to how to configure a client on the platform, a simple code example, and maybe some tips/warnings users of this provider should be aware of.
> 
> You should then also add this new provider under https://github.com/nextauthjs/next-auth/blob/canary/www/providers.json

Thank you for reply! I have created the documentation MD file and added the provider to the providers.json file.

www/docs/providers/eveonline.md
```diff
@@ -0,0 +1,63 @@
...
+:::tip When creating your application, make sure to select `Authentication Only` as the connection type.
+
+:::tip If using JWT for the session, you can add the `CharacterID`, `CharacterName`, and `image` to the JWT token and session. Example:
+
+```js
+...
+options: {
+  session: {
+    jwt: true,
+  },
```
balazsorban44, line 37:
```suggestion
```
This is the default value, so unnecessary to define

www/docs/providers/eveonline.md
```diff
@@ -0,0 +1,63 @@
...
+    },
+    session: async (session, token) => {
+      if (token) {
+        session.id = token.id;
+        session.name = token.name;
+        session.image = token.image;
+      }
+      return session;
+    }
+  }
```
balazsorban44, line 60:
```suggestion
```

The `profile` property method you defined is exactly there so if you don't set anything else on your JWT token, it is unnecessary to create a `jwt` callback.

You can see it being called here: https://github.com/nextauthjs/next-auth/blob/da01aa693a1bc5764662af82b6a48a8cd8c6b909/src/server/lib/oauth/callback.js#L110-L116
You can see it being sent to `jwt` here: https://github.com/nextauthjs/next-auth/blob/da01aa693a1bc5764662af82b6a48a8cd8c6b909/src/server/routes/callback.js#L82-L97

`session` will also include these fields by default, so you actually don't need any of these changes!
You can see it being sent by default here: https://github.com/nextauthjs/next-auth/blob/da01aa693a1bc5764662af82b6a48a8cd8c6b909/src/server/routes/session.js#L30-L43
And so the
> :::tip If using JWT for the session, you can add the `CharacterID`, `CharacterName`, and `image` to the JWT token and session. Example:

actually becomes redundant and not necessary.

www/providers.json
```diff
@@ -11,6 +11,7 @@
   "credentials": "Credentials",
   "discord": "Discord",
   "email": "Email",
+  "eveonline": "EVE Online",
```
balazsorban44, line 14: I saw this both as `EVEOnline` and `EVE Online`. Either change it here or everywhere else. I think it would be nice to keep it consistent. Do you agree?

balazsorban44 (changes requested):
Some small changes, and it will be good to go!

www/docs/providers/eveonline.md
geraldm74: Ah right, the main reason I did the jwt and session callbacks is because the `id` seems to be omitted from the default session payload?

www/providers.json
geraldm74: Yes, I'll change it.

www/docs/providers/eveonline.md
balazsorban44: Should be sent as `sub` https://github.com/nextauthjs/next-auth/blob/da01aa693a1bc5764662af82b6a48a8cd8c6b909/src/server/routes/callback.js#L86

geraldm74: I can't see `sub` ? The only data in the session object is:
`{"user":{"name":"*****","email":null,"image":"https://image.eveonline.com/Character/*********_128.jpg"},"expires":"2021-03-03T03:45:16.227Z"}` where is `sub` ?

geraldm74: JWT token also only has the following:
`{
  "name": "*******",
  "email": null,
  "picture": "https://image.eveonline.com/Character/*******_128.jpg",
  "iat": 1612151725,
  "exp": 1614743725
}`

balazsorban44: So again:

https://github.com/nextauthjs/next-auth/blob/da01aa693a1bc5764662af82b6a48a8cd8c6b909/src/server/lib/oauth/callback.js#L110-L116

Here you can see that the `provider.profile` function receives `profileData`, which is all the info we get from your provider. You can use the `profile` function to map it to what you need later on.

If you `console.log` the `profile` in the `profile()` callback, and you cannot find the id anywhere, then your provider does not send it somehow.

You could try the `debug: true` option and check the logs and see if you actually receive `CharacterID`

geraldm74: Hi,

Here is the debug response:

```
[next-auth][debug][profile_data] {
  CharacterID: *****,
  CharacterName: '*****',
  ExpiresOn: '2021-02-01T20:10:34',
  Scopes: '',
  TokenType: 'Character',
  CharacterOwnerHash: '*****',
  IntellectualProperty: 'EVE'
}
[next-auth][debug][oauth_callback_response] {
  profile: {
    name: '*****',
    email: null,
    image: 'https://image.eveonline.com/Character/*****_128.jpg'
  },
  account: {
    provider: 'eveonline',
    type: 'oauth',
    id: *****,
    refreshToken: null,
    accessToken: '*****',
    accessTokenExpires: null
  },
  OAuthProfile: {
    CharacterID: *****,
    CharacterName: '*****',
    ExpiresOn: '2021-02-01T20:10:34',
    Scopes: '',
    TokenType: 'Character',
    CharacterOwnerHash: '*****',
    IntellectualProperty: 'EVE'
  }
}
```

As you can see, the id or CharacterID is only in the `profile_data`, `account`, and `OAuthProfile` ... it is being omitted from the callback response, why? I couldn't work that out hence in the tips I showed how I explicitly had to add it to the JWT token from what was passed in from `profile_data`

geraldm74: So I'm assuming nextauth is expecting id in the response, but it's not there ... the provider only returns ` CharacterID`.

balazsorban44: Try returning `CharacterID` in the profile callback on the provider you added.

geraldm74: You mean like this? I have, but `id` just isn't passed back:

```js
profile: (profile) => {
	return {
		id: profile.CharacterID,
		name: profile.CharacterName,
		image: `https://image.eveonline.com/Character/{profile.CharacterID}_128.jpg`
	};
},
```

balazsorban44: can you verify that profile has `CharacterID`? what is `console.log(profile)` before return?

I really want to help you, went so far as to register myself on EVE Online, but it wants me to pay, and that's where I stopped, sorry :confused: Hope we will be able to figure this out¤

www/providers.json
balazsorban44: From what I can see on their website, it should be `EVE Online`, I guess.

www/docs/providers/eveonline.md
geraldm74: I console logged the json as shown:

```js
profile: (profile) => {
	const res = {
		id: profile.CharacterID,
		name: profile.CharacterName,
		image: `https://image.eveonline.com/Character/${profile.CharacterID}_128.jpg`
	}
	console.log(res);
	return res;
}
```

and what was logged was as follows:

```
{
  id: *********,
  name: '*********',
  image: 'https://image.eveonline.com/Character/*********_128.jpg'
}
```

So it's definitely assigning the `CharacterID` to `id` correctly and passing it back in the `return` but what ever handles it after that seems to be ignoring the `id` ?

balazsorban44: I think I know what is going on! 🤦 You probably still use `next-auth@3.1.0`, which by default did not return the `id` in the default jwt... `next-auth@3.2.0` came out today, and it contains the pr #784 which adds it... Could you update to the latest `next-auth` version and verify that this is the case?

balazsorban44: In any case, I just don't want to add an example code that will give the user the impression that they have to this to get the user id, when most new users will be on `next-auth@3.2.0` or higher when they install `next-auth`, and they don't actually have to anything. If you want to leave the jwt/session callbacks example that is fine but maybe add a non-standard field to the JWT instead! Sorry for this whole thread, really! 😅 You have done nice work, and we WILL get this merged in one way or another :wink:

geraldm74: ok, I just upgraded to 3.2.0:

```
"next-auth": "^3.2.0",
```

But I'm still having the same problem.  However I did noticed one question in the `[next-auth][debug][oauth_callback_response` ... it's now actually showing the `id` field which it wasn't doing before:

```
[next-auth][debug][oauth_callback_response] {
  profile: {
    id: *********,
    name: '*********',
    image: 'https://image.eveonline.com/Character/*********_128.jpg',
    email: null
  },
```

So I'm one step closer :)

however printing out what's in the `session` on in the browser, there is no `sub` or `id` field?

```js
{JSON.stringify(session)}
```

output:

```
{"user":{"name":"*********","email":null,"image":"https://image.eveonline.com/Character/*********_128.jpg"},"expires":"2021-03-03T21:59:49.164Z"}
```

balazsorban44: That's right! After some digging, I can see that the `session` callback will not expose `id` by default, for a more secure default:

https://github.com/nextauthjs/next-auth/blob/562bcf216c9c83263027cffad5511cdd36f41f48/src/server/routes/session.js#L30-L40

So I guess you can include the `session` callback in the example to forward `id` to the front-end!

Glad it has been sorted out!

balazsorban44:
@geraldm74 Just request a new review when you feel you are ready! And thank you so much for your patience in getting this merged!

www/providers.json
geraldm74: ok thanks.

However I also needed to add `it` to the jwt callback as well:

```js
callbacks: {
  jwt: async (token, user, account, profile, isNewUser) => {
    if (profile) {
      token = {
        ...token,
        id: profile.CharacterID,
      }
    } 
    return token;
  },
  session: async (session, token) => {
    if (token) {
      session.user.id = token.id;
    }
    return session;
  }
}
```

Now I see this in the browser for `session`:

```
{"user":{"name":"*********","email":null,"image":"https://image.eveonline.com/Character/*********_128.jpg","id":*********},"expires":"2021-03-03T22:41:25.212Z"}
```

This is exactly what I want, showing the character ID in the session :)

www/docs/providers/eveonline.md
geraldm74: ok thanks.

However I also needed to add `id` to the jwt callback as well:

```js
callbacks: {
  jwt: async (token, user, account, profile, isNewUser) => {
    if (profile) {
      token = {
        ...token,
        id: profile.CharacterID,
      }
    } 
    return token;
  },
  session: async (session, token) => {
    if (token) {
      session.user.id = token.id;
    }
    return session;
  }
}
```

Now I see this in the browser for `session`:

```
{"user":{"name":"*********","email":null,"image":"https://image.eveonline.com/Character/*********_128.jpg","id":*********},"expires":"2021-03-03T22:41:25.212Z"}
```

This is exactly what I want, showing the character ID in the session :)

balazsorban44 (approved):
Thanks!

PR #1246, opened 2021-02-03
improve(ts): colocate TS typings

jasonkuhrt:
**What**:

Colocate the TS typings originally found [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/next-auth/index.d.ts). 

**Why**:

Incremental migration to TS. Reboot/Redo of https://github.com/nextauthjs/next-auth/pull/516.

**How**:

Have started by copying what's currently published on npm. We can continue to iterate after an initial release with this colocated.

**Checklist**:

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/adapters.d.ts
+++ b/adapters.d.ts
@@ -43,15 +43,15 @@ export interface AdapterInstance<
   TSession,
   TVerificationRequest
 > {
-  createUser(profile: TProfile): Promise<TUser>
-  getUser(id: string): Promise<TUser | null>
-  getUserByEmail(email: string): Promise<TUser | null>
-  getUserByProviderAccountId(
+  createUser: (profile: TProfile) => Promise<TUser>
+  getUser: (id: string) => Promise<TUser | null>
+  getUserByEmail: (email: string) => Promise<TUser | null>
+  getUserByProviderAccountId: (
     providerId: string,
     providerAccountId: string
-  ): Promise<TUser | null>
-  updateUser(user: TUser): Promise<TUser>
-  linkAccount(
+  ) => Promise<TUser | null>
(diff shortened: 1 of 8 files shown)


adapters.d.ts
```diff
@@ -43,47 +43,47 @@ export interface AdapterInstance<
   TSession,
   TVerificationRequest
 > {
-  createUser(profile: TProfile): Promise<TUser>
-  getUser(id: string): Promise<TUser | null>
-  getUserByEmail(email: string): Promise<TUser | null>
-  getUserByProviderAccountId(
+  createUser: (profile: TProfile) => Promise<TUser>
```
jasonkuhrt, line 46: I disagree with this change strongly. It changes autocompletion. It makes these appear as properties instead of methods (the icon). The problem is that properties can be callable or not. It just makes the autocomplete less usable, because it is harder to tease apart callables from non-callables.

balazsorban44:
So after our internal discussion, I don't want to accept this PR in this form, but I am very happy if you would start typing params and return types on internal functions with `JSDoc` notation, and maybe some of the user-facing ones as well. If you want to be sure everything is typed in a file, just add a temporary `// @ts-check` comment at the top. :wink:

Interfaces can be added in .d.ts files, but I would like to wait with adding them to `package.json/types` just yet.

Also, I am very open to adding some kind of a script that extracts interfaces/JSDoc comments from the codebase and keeps our docs updated! Right now we have to explain the same functions over and over again, which can lead to inconsistencies really fast!

Thanks again for the chat!

jasonkuhrt:
@balazsorban44 yep! Thanks for your time as well!

PR #1257, opened 2021-02-04
fix(provider): okta client authentication

mdgraser:
**What**: Adding the bearer Authorization header here breaks the Okta provider flow.

**Why**: See [okta:client-authentication-methods](https://developer.okta.com/docs/reference/api/oidc/#client-authentication-methods). While the bearer token is not a supported client authentication method, posting the client secret is supported. So nothing need be added here.

**How**: Remove the Authorization header for `provider === 'okta'`.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/pages/api/auth/[...nextauth].js
+++ b/pages/api/auth/[...nextauth].js
@@ -17,11 +17,6 @@ export default NextAuth({
       clientId: process.env.TWITTER_ID,
       clientSecret: process.env.TWITTER_SECRET
     }),
-    Providers.Okta({
-      clientId: process.env.OKTA_ID,
-      clientSecret: process.env.OKTA_SECRET,
-      domain: process.env.OKTA_DOMAIN
-    }),
     Providers.Credentials({
       name: 'Credentials',
       credentials: {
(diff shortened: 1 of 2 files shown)


pages/api/auth/[...nextauth].js
```diff
@@ -17,6 +17,11 @@ export default NextAuth({
       clientId: process.env.TWITTER_ID,
       clientSecret: process.env.TWITTER_SECRET
     }),
+    Providers.Okta({
+      clientId: process.env.OKTA_ID,
+      clientSecret: process.env.OKTA_SECRET,
+      domain: process.env.OKTA_DOMAIN
+    }),
```
Comment (balazsorban44) [inline pages/api/auth/[...nextauth].js:24]: For now, I don't wish we start adding every provider to the dev app, it's totally fine if you do it for local development though, and I encourage you to actually test your changes locally (which I guess you did since it was added here)! Could you please remove this section from this PR?
```suggestion
```

balazsorban44 (changes requested):

balazsorban44 (approved):
That was on 🔥 😅

Issue #1264, opened 2021-02-05
No session after signin with redirect: false


cgerikj:
**Describe the bug**
When using the new redirect: false feature with credentials (https://github.com/nextauthjs/next-auth/pull/1219) in 3.3.0-canary.6, useSession returns null for the session and false for loading. Only a hard refresh/redirect updates the session.

**Steps to reproduce**
Steps to reproduce the behavior.
1. await a signIn with 'credentials' and { redirect: false }
2. router.replace('/profile')

**Expected behavior**
The session updates

**Sample code**
```
const signinResponse = await signIn('credentials', {
  email: email,
  password: password,
  redirect: false,
})
if (signinResponse.status === 200) {
  // getSession should not be needed here but even after this, useSession returns null in the /profile route
  getSession().then(session => {
    // session looks good here..!
    router.replace('/profile')
  })
} else {
  setError('Could not sign in!')
}

```

**Temporary solution**
```window.location.href = '/profile'```

**Comment by balazsorban44** [comment]
Hi there! The point of adding `redirect: false` is exactly stopping the page from redirecting to anywhere. Since it is the redirect that will update the session, disabling it will not update the session either.

I assumed this option has been asked from us to make it possible for users to handle errors on the same page.
You could do something similar to what we have in the dev app:

```jsx
...
const handleLogin = async () => {
  const response = await signIn('credentials', options)
  if (response.ok) return window.location.reload()
  // Handle errors here
  console.log(response.error)
}
...
return (
  <button onClick={handleLogin}>Login</button>
)
```

iaincollins:
Hmm, if this is confusing for folks I wonder if we should call getSession() from the signIn() and signOut() functions when they are not redirecting?

balazsorban44:
Yes, I had to think for a second about how to do this, but it makes sense. I created #1267 with a fix for this. I tested locally, and it looks fine for me.

balazsorban44:
I have to say, that `signOut({redirect: false})` is a very nice experience, if you don't need to redirect the user to any other page than the one you are on...

iaincollins:
Oh wow that was quick! 😀

Good point about the user experience.

I wonder if that should even be the default for signOut?

balazsorban44:
Was thinking the same, we can decide on that later. I updated the PR with some docs as a tip for this.

cgerikj:
Thank you, this is perfect! :)

its-kos:
[...]

PR #1267, opened 2021-02-05
feat(provider): update session when signIn/signOut successful

balazsorban44:
**What**:

Fixes #1264 

`signIn("credentials", {redirect: false})`
 will now also return `url` if there was no error, for convineance.

An added bonus of this change if you have `signOut({redirect: false})` and you don't have to redirect the user to a log-out page or similar, the session simply will disappear from `useSession()`, without any page reload! Or even if you have to redirect, you will get the URL returned by `signOut`, which you yan use together with Next.js `useRouter().push(url)`!

**Why**:

Until now, we relied on the page reload to update the `session`. When using `"credentials"` with `redirect: false`, we have to manually trigger a session update, so the `useSession` hook gets properly updated.

**How**:

Call `await __NEXTAUTH._getSession({ event: 'storage' })` if no error returned

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/components/header.js
+++ b/components/header.js
@@ -54,7 +54,7 @@ export default function Header () {
                 className={styles.button}
                 onClick={(e) => {
                   e.preventDefault()
-                  signOut()
+                  signOut({ redirect: false })
                 }}
               >
                 Sign out
--- a/pages/credentials.js
+++ b/pages/credentials.js
@@ -10,10 +10,6 @@ export default function Page () {
     }
     const response = await signIn('credentials', options)
     setResponse(response)
-    if (response.ok) {
-      window.alert('Manually refreshing to update session, if login was successful')
-      window.location.reload()
-    }
   }
 
   const handleLogout = (options) => async () => {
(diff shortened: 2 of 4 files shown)


Gabsys:
Anyone facing issue with this? I have `signIn()` with `redirect: false`. I use `router.push(result.url)` if no error is returned but the session isn't updated, I still have to manually refresh the page to see session changes

GvidasP:
Can confirm - running into the same problem, session is not updated after signIn called with `redirect: false`. Could this somehow be related that I am using `JWT`?

gabrielmlinassi:
Same problem here. For some reason just happens in the production build. Perhaps there's something in the dev mode that prevents that like `reactStrictMod: true`

minhchu:
I'm having the same problem too.

```
    "next": "12.2.3",
    "next-auth": "^4.12.2",
```

LucaPiccirillo:
Same problem, still need to reload the page to update the session.
Is it possible to have a method to manually update the session maybe?

alexwaddell97:
Having this issue on Next.js v13.3.1

balthazur:
same with next 14.0.2 and 4.24.5. router.push() working locally, but not in production.

hashirshoaeb:
do we have this in NextAuth v5?

PR #1280, opened 2021-02-08
Update azure-ad-b2c.md

rhufsky:
**What**:
Add hint for redirection URL

**Why**:
It is difficult to find out the correct value.

**How**:

**Checklist**:

- [x ] Documentation
- [x] Tests: Tested with my own Azure account
- [x] Ready to be merged


www/docs/providers/azure-ad-b2c.md
```diff
@@ -12,7 +12,20 @@ https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-c
...
 
 ## Example
+- In https://portal.azure.com/ -> Azure Active Directory create a new App Registration.
+- Make sure to remember / copy
+  - Application (client) ID
+  - Directory (tenant) ID
+- When asked for a redirection URL, use http://localhost:3000/api/auth/callback/azure-ad-b2c
+- Create a new secret and remember / copy its value immediately, it will disappear.
 
+In .env create the follwing entries
```
balazsorban44, line 22: According to the Next.js [Environment Variables](https://nextjs.org/docs/basic-features/environment-variables) guidelines, local env variables should be added to `.env.local`, since `.env` is supposed to be added to version control. (It is especially important in case of `AZURE_CLIENT_SECRET`)

balazsorban44 (changes requested):
Thank you, probably a nice addition, just a little comment!

www/docs/providers/azure-ad-b2c.md
rhufsky: Thanks for the hint, I filed another PR which covers that.

PR #1285, opened 2021-02-09
Fixes for email sign in

iaincollins:
**What**:

* Fixes for email sign in flow with Prisma adapter.
* Fixes for info text colour for error messages on the built-in pages.
* Includes improvements to the provided test/dev app to make it easier to test it with different adapters.

**Why**:

Prisma adapter uses delete() but should use deleteMany().

Additionally, error / info messages were using the wrong text colour on the light theme.

**How**:

Have now verified the Email sign in flow behaves as expected for:

* Prisma 2 adapter with Postgres and Email provider flow
* TypeORM adapter with Postgres and Email provider flow
* TypeORM adapter with MySQL and Email provider flow
* TypeORM adapter with MongoDB and Email provider flow

In each case I have checked the positive flow and negative flows, including for know issues in the current release. I have also regression tested OAuth flows (though they are not impacted by this change, or other recent changes).

It has not been possible to test the MSSQL release at this time, but no issues that impact it specifically are expected.

**Checklist**:

- [x] Documentation
- [x] Tests (manual regression testing, as above)
- [x] Ready to be merged


.env.local.example
```diff
@@ -1,8 +1,13 @@
-# Rename file to .env.local and populate values
+# Rename file to .env.local (or .env) and populate values
```
balazsorban44, line 1: Next.js lists `.env` as a file that should be added to version control: https://nextjs.org/docs/basic-features/environment-variables#default-environment-variables

so we should not encourage people to use it for adding secrets. (`.env.local.example` is for the dev app anyway, not for running the tests)

package.json
```diff
@@ -89,6 +90,7 @@
     "pg": "^8.2.1",
     "postcss-cli": "^7.1.1",
     "postcss-nested": "^4.2.1",
+    "prisma": "^2.16.1",
```
balazsorban44, line 93: `peerOptionalDependencies` still lists `  "@prisma/client": "^2.12.0"`. Should that be updated as well?

balazsorban44 (changes requested):

.env.local.example
iaincollins: The problem with this that Prisma doesn't pick up .env.local file.

iaincollins: BTW The NEXTAUTH_ prefixes are deliberate in this case (and don't want to change them).

I actually want to do more with them, but is probably best to expand on another time.

package.json
iaincollins: Hmm yes good catch thanks! Will update.

(At least now the instructions work for the current release of Prisma!)

.env.local.example
balazsorban44: Yeah, it's a bummer: https://github.com/prisma/prisma/issues/3865 Let it be as-is for now then. Will ping the Prisma folks

balazsorban44 (approved):

PR #1294, opened 2021-02-10
feat(logger): introduce user configurable logger

### LINKED ISSUE #869: Configurable logging

### LINKED ISSUE #907: How to log errors to file instead of console?

balazsorban44:
**What**:

Let the user configure logging.

**Why**:

This lets the user override certain log levels, in case they would like to forward the log messages to something else than `console`.

**How**:

Adds a new option to `NextAuth`, called `logger`.

It has three properties, `error`, `warn` and `debug`. Those not overridden will fall back to the basic logger function.

To cover client-side logs, we do make the assumption that your 3rd party logger does not support client-side logging. To mitigate this, we proxy all client-side log events to an endpoint (`/api/auth/_log`) that sends those further to your custom logger config.

We use [`sendBeacon`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) if available.

**Checklist**:

- [x] Documentation
- [x] Tests (locally)
- [x] Ready to be merged
      

closes #869
closes #907

continuation of #870, #1090 (cannot git 🙈)

### DIFF
--- a/www/docs/configuration/options.md
+++ b/www/docs/configuration/options.md
@@ -326,6 +326,9 @@ export default NextAuth({
     error(code, ...message) {
       log.error(code, message)
     },
+    warn(code, ...message) {
+      log.warn(code, message)
+    }
     debug(code, ...message) {
       log.debug(code, message)
     }
--- a/www/docs/configuration/options.md
+++ b/www/docs/configuration/options.md
@@ -307,6 +307,39 @@ Set debug to `true` to enable debug messages for authentication and database ope
 
 ---
 
+### logger
+
+* **Default value**: `console`
+* **Required**: *No*
+
+#### Description
+
+Override any of the logger levels (`undefined` levels will use the built-in logger), and intercept logs in NextAuth. You can use this to send NextAuth logs to a third-party logging service.
+
+Example:
+```js title="/pages/api/auth/[...nextauth].js"
+import log from "logging-service"
+
+export default NextAuth({
+  ...
+  logger: {
+    error(code, ...message) {
+      log.error(code, message)
+    },
+    debug(code, ...message) {
+      log.debug(code, message)
+    }
+  }
+  ...
+})
+```
+
+:::note
+If the `debug` level is defined by the user, it will be called regardless of the `debug: false` [option](#debug).
+:::
+
+---
+
 ### theme
 
 * **Default value**: `"auto"`
--- a/src/adapters/example/index.js
+++ b/src/adapters/example/index.js
@@ -1,84 +1,83 @@
 const Adapter = (config, options = {}) => {
   async function getAdapter (appOptions) {
+    const { logger } = appOptions
     // Display debug output if debug option enabled
-    function _debug (...args) {
-      if (appOptions.debug) {
-        console.log('[next-auth][debug]', ...args)
-      }
+    function debug (debugCode, ...args) {
+      logger.debug(`ADAPTER_${debugCode}`, ...args)
     }
 
     async function createUser (profile) {
-      _debug('createUser', profile)
+      debug('createUser', profile)
       return null
     }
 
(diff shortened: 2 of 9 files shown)


ndom91:
This looks awesome! Good docs example as well :+1:

ndom91 (approved):
This looks awesome! Good docs example as well :+1:

www/docs/configuration/options.md
```diff
@@ -307,6 +307,42 @@ Set debug to `true` to enable debug messages for authentication and database ope
...
+
+export default NextAuth({
+  ...
+  logger: {
+    error(code, ...message) {
+      log.error(code, message)
+    },
+    warn(code, ...message) {
+      log.warn(code, message)
+    }
```
followbl, line 331: @balazsorban44 missing comma 😺

balazsorban44: https://github.com/nextauthjs/next-auth/commit/5ae6f6118cf0b12c1ab38e39200291c4d919a10d 🎉

PR #1322, opened 2021-02-16
fix: add oauth_token_secret to requests

### LINKED ISSUE #855: OAuth 1.0 expects "oauth token secret" when requesting the access_token (for trello authenticatino)
**Describe the bug**
When trying to use Trello OAuth 1.0 flow, it returns: 
```
[next-auth][error][oauth_v1_get_access_token_error] [ { statusCode: 400, data: 'Invalid Signature' }, undefined ]
```

**Steps to reproduce**
Here is the custom provider that I'm using:
```
{
      id: "trello",
      name: "Trello",
      type: "oauth",
      version: "1.0",  // try both 1.0 and 1.0A
      scope: "read,write,account",
      accessTokenUrl: 'https://trello.com/1/OAuthGetAccessToken',
      requestTokenUrl: 'https://trello.com/1/OAuthGetRequestToken',
      authorizationUrl: 'https://trello.com/1/OAuthAuthorizeToken',
      profileUrl: 'https://api.trello.com/1/members/me/',
      params: { expiration: 'never' },
      profile: (profile) => ({}),
      clientId: process.env.TRELLO_CLIENT_ID,
      clientSecret: process.env.TRELLO_CLIENT_SECRET,
    }
```
You can get your own credentials get here: https://trello.com/app-key/

**Expected behavior**
Being able to connect with trello authentication

**Investigation**
After digging into the code, I think I found what is missing to make Trello OAuth 1.0 and probably other oauth1.0 working. It isn't clear why twitter works (knowing how every company interprets OAuth spec, that is probably why :) )

In order to make sure that my credentials and configuration were correct, I used: https://github.com/adunkman/node-trello to get a user token, and it worked! They both use `oauth` so that made it easy to spot the difference. Here it is:

https://github.com/nextauthjs/next-auth/blob/5126f4e3428f2b4b8ebdd38b10066edfdd4ea2fa/src/server/lib/oauth/callback.js#L108

When calling: `client.getOAuthAccessToken` the second argument needs to be the `oauth token secret` generated when calling: `client.getOauthRequestToken`
https://github.com/nextauthjs/next-auth/blob/5126f4e3428f2b4b8ebdd38b10066edfdd4ea2fa/src/server/lib/signin/oauth.js#L33
The 3rd argument of `getOauthRequestToken` is `oauth_token_secret`
 
https://github.com/ciaranj/node-oauth/blob/a7f8a1e21c362eb4ed2039431fb9ac2ae749f26a/lib/oauth.js#L551

Now I'm a bit unclear to know how I would fix that:
1. store that secret in the session? is that safe? do I even have access to the session in these 2 functions?
2. rely on a key-value store like redis to store that?

### LINKED ISSUE #1321: For oAuth1.0 NextAuth incorrectly refers to accessTokenSecret as refreshToken
**Describe the bug**
For oAuth1.0 there is no concept of refreshToken. Yet nextauth gives you a `refreshToken` as part of the token object even when used with oAuth1.0 provider. Other users have been looking for access_token_secret but haven't been able to becaue of this incorrect naming. #855 

**Steps to reproduce**
Can be reproduced with any oAuth1.0 provider. For example using the twitter provider. 

**Expected behavior** 
The token object should replace the `refreshToken` key with `accessTokenSecret` key. The documentation also needs to be updated. 

**Feedback**

* [x] Found documentation but was incomplete

balazsorban44:
**What**:

Fixing OAuth 1 flows

**Why**:

Multiple issues reported that we do not use `oauth_token_secret` as intended.
See spec: https://oauth.net/core/1.0a/

**How**:

As far as I can tell, we only checked with a single OAuth 1.0 provider (Twitter), which apparently worked by simply setting `oauth_token_secret` to `null`.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      

fixes #1321 
fixes #855


balazsorban44:
@iaincollins Would appreciate your eyes on this, as you might have a reason for dropping `oauth_token_secret` in certain places! I checked with the Twitter provider, and with these changes, it still works!

aksanoble (changes requested):
This may lead to some breaking changes as typeorm still depends on refreshToken. Can you please check if the necessary column and tokenSecret gets added to the DB as well?

balazsorban44:
Good catch @aksanoble! I am not good with the adapters though, so I might have to wait for help from Iain.

erhankaradeniz:
Any progress on this?

aksanoble:
> Any progress on this?

Looks like there's still work to do. In the meanwhile if you want to access oauth_token_secret just use the token available as refreshToken.

erhankaradeniz:
> > Any progress on this?
> 
> Looks like there's still work to do. In the meanwhile if you want to access oauth_token_secret just use the token available as refreshToken.

I'm not sure I'm understanding what you mean. I'm trying to combine next-auth with strapi. Followed this guide: [Authenticate with next-auth](https://strapi.io/blog/user-authentication-in-next-js-with-strapi). 

I'm seeing the refreshToken and accedToken, but when I use the accessToken or the refreshtoken to send to my backend, next-auth prompts me this error: 

```
https://next-auth.js.org/errors#oauth_callback_handler_error TypeError: Cannot read property 'id' of undefined
```

This is correct since my backend does not understand the callback with the accessToken, probably because the accessToken is not correct?

I'm sorry I do not want to hijack this PR with my issue.

Cryptoking74 (approved):
Cryptoking74 (bryanjones)(Administrator)approved

src/server/lib/oauth/callback.js
```diff
@@ -62,11 +62,13 @@ export default async function oAuthCallback (req) {
     const {
       oauth_token: oauthToken, oauth_verifier: oauthVerifier
     } = req.query
-    const tokens = await client.getOAuthAccessToken(oauthToken, null, oauthVerifier)
```
Cryptoking74, line 65: Cryptoking74 (bryanjones)
(Administrator) approved

Cryptoking74: Approved

Cryptoking74 (approved):

huynhminhcanh:
.github/ci-workflow.yml

huynhminhcanh:
.github/ci-workflow.yml

huynhminhcanh:
> .github/ci-workflow.yml

Cryptoking74 (approved):
Approved coin, CryptoKing💫(Bryan Jones)authorized docs.for coin

Cryptoking74:
> .github/ci-workflow.yml

What is this?

Cryptoking74 (changes requested):
Bryan jones

balazsorban44:
@iaincollins I wonder if we could just also return this value with a different name IN ADDITION to the current one? This wouldn't break anything, but people using OAuth 1.0 would know what to look for?

giuseppeg:
> @iaincollins I wonder if we could just also return this value with a different name oauth_token_secret IN ADDITION to the current one (refreshToken)

@balazsorban44 I second this. I just spent one hour trying to figure out how to get the `oauth_token_secret` and eventually landed here :)

balazsorban44:
@aksanoble as you suggested, I just also keep `refreshToken` for backward compatibility.

`oauth_token` and `oauth_token_secret` will now be available in the `account` object in both the `signIn` and `jwt` callbacks!

PR #1325, opened 2021-02-17
fix: twitter oauth authorize authorizationUrl

harshithpabbati:
**What**:

Changed the twitter oauth authorizationURL.

**Why**:

As twitter updated the endpoint.

Screenshot 2021-02-17 at 8 41 50 PM

Here is the [link](https://developer.twitter.com/en/docs/authentication/oauth-1-0a/obtaining-user-access-tokens) to the doc.

**How**:

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged

### DIFF
--- a/src/providers/twitter.js
+++ b/src/providers/twitter.js
@@ -7,7 +7,7 @@ export default (options) => {
     scope: '',
     accessTokenUrl: 'https://api.twitter.com/oauth/access_token',
     requestTokenUrl: 'https://api.twitter.com/oauth/request_token',
-    authorizationUrl: 'https://api.twitter.com/oauth/authenticate',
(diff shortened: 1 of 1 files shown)


balazsorban44:
As far as I can tell, the Twitter provider works as expected, returning all user data. If you have further issues accessing it, you can try creating an issue with a full reproduction, or open a [discussion](https://github.com/nextauthjs/next-auth/discussions).

From the docs:
> This method differs from GET oauth / authorize in that if the user has already granted the application permission, the redirect will occur without the user having to re-approve the application. To realize this behavior, you must enable the Use Sign in with Twitter setting on your application record.

https://developer.twitter.com/en/docs/authentication/api-reference/authenticate

Issue #1238, opened 2021-02-02
Get more data while creating user (with custom adapter) for GitHub and LinkedIn


pbteja1998:
I have set up `GitHub` and `LinkedIn` auth using `next-auth`. Even though, I am setting the scope to `read:user` for GitHub, I am only getting `name`, `email`, and `image` fields under profile. Is there a way to get other fields like `username` from GitHub?

Also, when I logged in with LinkedIn, I am unable to get a profile image.

Here are my providers with corresponding scopes:
```js
providers: [
    Providers.GitHub({
      clientId: process.env.GITHUB_ID,
      clientSecret: process.env.GITHUB_SECRET,
      scope: 'read:user',
    }),
    Providers.LinkedIn({
      clientId: process.env.LINKEDIN_ID,
      clientSecret: process.env.LINKEDIN_SECRET,
      scope: 'r_liteprofile r_emailaddress',
    }),
  ],
```

I have a custom adapter, and I am trying to get the details from `createUser` function.

```js
// createUser function in the custom adapter.
async function createUser(profile) {
    // I am expecting more information from this "profile" object
}
```

Any help related to this is highly appreaciated.

Thanks.

balazsorban44:
This is what the `profile` provider method property is for! It receives all the data we get from your provider, than you can map it over to whatever you want. That is then sent as the first argument in the `signIn` callback, and the as the second argument to the `jwt` callback. You can than use those callbacks to persist/send that info to the client.

pbteja1998:
Thanks a lot 🙏 @balazsorban44 
I was able to resolve this by using the above method. 

---

For anybody who comes across this issue, I also had to update to the latest version for it to work. 

Previous versions:
```
"next-auth": "3.1.0",
"@types/next-auth": "3.1.18"
```

Current versions: 
```
"next-auth": "3.2.0",
"@types/next-auth": "3.1.24"
```

balazsorban44:
@iaincollins, I don't remember changing anything with the adapter code in the new release? Do you have an idea why updating would fix this?

pbteja1998:
Also for getting the profile picture for LinkedIn, I also had to overwrite `profileUrl` field. 

My LinkedIn provider looks like below:

```javascript
Providers.LinkedIn({

      clientId: process.env.LINKEDIN_ID,

      clientSecret: process.env.LINKEDIN_SECRET,

      scope: 'r_liteprofile',

[...]

Issue #955, opened 2020-12-14
Expose accessToken from accounts database


jackmerrill:
**Summary of proposed feature**
I'm using the Discord OAuth sign in, and would like to interface with the API using the [Discord-OAuth2](https://npmjs.org/package/discord-oauth2) package on NPM, but I can't get the `accessToken` that's in the `accounts` database.

**Purpose of proposed feature**
Allow people to interface with the APIs that they use with this package.

**Potential problems**
None that I can think of.

**Describe any alternatives you've considered**
A hacky method of using an ORM and getting the token by finding by some unique identifier, in my case, find the user in the `users` database by email, use their `_id` (using MongoDB) to find their account in `accounts`, then using the `accessToken` from that.

*Please indicate if you are willing and able to help implement the proposed feature.*
Most likely not.

naillizard:
Take a look at TypeORM [query builder](https://typeorm.io/#/select-query-builder/joining-relations).

zvshka:
+1

almeidx:
I'm facing the same problem.

balazsorban44:
not sure about the OPs problem. You can get anything out from the account object in both the signIn and the jwt callback. Check the documentation.

https://next-auth.js.org/configuration/callbacks

do I miss something?

PR #1329, opened 2021-02-17
feat: make tokens available in profile callback

balazsorban44:
**What**:

Adds a new parameter to the `profile()` callback in provider options.

**Why**:
Since there is a lot of inconsistencies in OAuth provider implementations, the `profile` callback often provider an inconsistent result in its first parameter. The `profile` callback is called asynchronously, so it is a good place to fetch more data, but usually, that request should be authenticated by the logged-in user. Without an access_token, it is in most cases probably impossible.

**How**:

Forward `tokens` to `profile`, so it will receive these params:
```js
  const email = await getEmailForUser(tokens.access_token)
  return {
    id: profile.sub,
    name: profile.nickname,
    email,
    image: profile.picture
  }
}
```

We will forward all the tokens we have for a provider. Depending on which provider we talk about, you will have an `access_token`, `id_token`, and `refresh_token`.

**Checklist**:

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged
      

Discusses briefly: https://github.com/nextauthjs/next-auth/issues/1238#issuecomment-771796703

Multiple of our built-in providers also return `null` for email and image because of this. With an `access_token` available, it might be possible to fetch that info for those as well.

### DIFF
--- a/src/server/lib/oauth/callback.js
+++ b/src/server/lib/oauth/callback.js
@@ -108,7 +108,7 @@ async function getProfile ({ profileData, tokens, provider, user }) {
 
     logger.debug('PROFILE_DATA', profileData)
 
-    const profile = await provider.profile(profileData, tokens.access_token)
+    const profile = await provider.profile(profileData, tokens)
     // Return profile, raw profile and auth provider details
     return {
       profile: {
--- a/src/server/lib/oauth/callback.js
+++ b/src/server/lib/oauth/callback.js
@@ -108,7 +108,7 @@ async function getProfile ({ profileData, tokens, provider, user }) {
 
     logger.debug('PROFILE_DATA', profileData)
 
-    const profile = await provider.profile(profileData)
+    const profile = await provider.profile(profileData, tokens.access_token)
     // Return profile, raw profile and auth provider details
     return {
       profile: {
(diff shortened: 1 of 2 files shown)


jackmerrill:
Hoping this gets merged! I hope this also clears the path for #955. Too much work for something so simple.

ndom91 (approved):
That was easy enough of a change! Haha 

LGTM :+1:

iaincollins:
Sorry it's taken a few days to get round to this and thank you for tagging me! :-)

Instead of passing a string as the second argument, I wonder if it's worth passing the entire `tokens` object to future proof things?

IIRC, the `tokens` object may optionally contain a raw `id_token` which might also be helpful to be able to access in some cases (although I think of that as more useful for advanced usage, like someone creating a new provider).

I appreciate having type definitions here would be helpful and that the object format varies between providers makes it ambiguous what will be returned, but if we only return a specific key of `access_token`, I think we are going to run into cases where it won't work for some folks, and just passing the entire object seems like a helpful thing to do.

Maybe we could then log a FIXME / TODO / or an issue to create and test better defined types with a range of providers to document them and see what they actually return?

balazsorban44:
Thanks for the feedback! I can very easily send all the tokens as the second param, no problem. In v4, I would actually prefer if all our user facing apis could take a single object (aka named parameters) in case if they have more than one param. Easier to create types also. For now I can send the tokens and we see what we come up with later.

Issue #709, opened 2020-09-24
Magic Passcode


alex-cory:
**Your question**
I'm curious if there's currently a way to have the same behavior as the magic link, but instead send a 4-6 digit code to their email that the user can then enter in themselves on the site.

[![image](https://user-images.githubusercontent.com/5455859/94602783-2f570680-024a-11eb-9f37-2008b31703bc.png)](https://www.npmjs.com/package/react-code-input)

Also kind of like [this](https://support.apple.com/guide/iphone/automatically-fill-in-sms-passcodes-on-iphone-iphc89a3a3af/ios) but with email code instead of SMS code.

**What are you trying to do**
I'm working on a PWA and a magic link will take you directly to the mobile browser (ex: iOS Safari) instead of directly to the PWA. This is a way to get around that.

**Feedback**

* [x] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [x] Could not find relevant documentation
* [ ] Found the example project helpful
* [ ] Did not find the example project helpful

glenames:
Also experience the same user experience issue - another way to solve this would be to handle email auth the same way https://magic.link/ or Vercel do. Rather than opening a new session, the link authenticates the original session.

alex-cory:
@glenames I thought about this. I guess I should take a deeper look into that. I'm trying to think if there are any security concerns with that approach?

glenames:
@alex-cory Its not possible with the current email auth implementation - its a suggested alternate feature request I guess. I think it's a fair bit trickier to implement because it would require a subscription to the database from the client requesting access.

alex-cory:
How does this flow look? Also, what is the recommended way of adding the jwt or session token or csrf token to the frontend. I'm not quite sure how next-auth is setting these on the frontend so when calling `useSession` it will have the correct values to fetch the current session.

- [ ] `/login` frontend
  - [ ] type in email, click login/signup - POST `{ email, isAddedToHomescreen }` to `/api/auth/magic-link`
  - [ ] if isAddedToHomescreen, subscribe to channel `private-magic-link-me@gmail.com`
      - [ ] on trigger event from channel `private-magic-link-me@gmail.com` - set JWT from socket body
- [ ] `/api/auth/magic-link` POST
  - [ ] create/get user with post body.email
  - [ ] create JWT with `userId`, `email`, `isAddedToHomescreen` and whatever else should go into the JWT
  - [ ] email with link `/magic-link?jwt=<the-jwt>`
[...]

PR #1378, opened 2021-02-26
feat(provider): call generateVerificationToken async

balazsorban44:
**What**:

When using the "email" provider, you can supply your own `generateVerificationToken` function to create a token of your liking.

**Why**:

Until now, it was not possible to do async operations inside this function.

**How**:

When `generateVerificationToken` is called, we simply put an `await` in front of it, so you can do any async task inside it.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged
      

As mentioned by @iaincollins  in https://github.com/nextauthjs/next-auth/issues/709#issuecomment-784077142, this is something we are supposed to support. So here you go :slightly_smiling_face:

### DIFF
--- a/src/server/lib/signin/email.js
+++ b/src/server/lib/signin/email.js
@@ -10,7 +10,7 @@ export default async function email (email, provider, options) {
     const secret = provider.secret || options.secret
 
     // Generate token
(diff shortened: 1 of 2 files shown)


ndom91 (approved):
LGTM :+1:

Issue #1154, opened 2021-01-19
Multi language support


balazsorban44:
**Summary of proposed feature**
`next-auth` pages are currently only available in English: https://next-auth.js.org/configuration/pages

**Purpose of proposed feature**
These pages can be customized/replaced by the user to mitigate any requirements, but it would be nice if we provided a way to show these pages in multiple languages out-of-the-box.

**Detail about proposed feature**

Next.js 10 added new features for a better [i18n development experience](https://nextjs.org/blog/next-10#internationalized-routing) which we could [utilize](https://github.com/nextauthjs/next-auth/discussions/1129#discussioncomment-293562).

`next-auth` takes a `pages` property to configure/replace the built-in ones.

`pages` could take a new property, called `locales`, which would hold the translations for the different built-in pages.

Example:
```js
// /api/auth/[...nextauth].js
import NextAuth from "next-auth"
export default NextAuth({
  ...
  pages: {
    locales: {
      en: {
        "verify-request": {
          title: "Check your e-mail",
          subtitle: "A sign in link has been sent to your email address.",
        },
        signout: {
          title: "Are you sure you want to sign out?",
          button: "Sign out",
        },
        ...
      },
    },
  }
  ...
})
```

**Potential problems**
?

**Describe any alternatives you've considered**
The only way to achieve this today is through fully custom pages.

**Additional context**
Any other context, screenshots, etc.
Originally asked here: https://github.com/nextauthjs/next-auth/discussions/1129
In #831, there was an attempt to add support for language change in the sent e-mails.

stevedaniels-traintime:
@balazsorban44 did you manage to implement this through fully custom pages?

It would be awesome if you could give us some tips on how this was achieved 🙏

YuriGor:
Hi @stevedaniels-traintime currently I am using next/router to navigate to signIn custom page instead of `signin` method and it seems to work fine.

```js
import { useRouter } from 'next/router';
import { useSession } from 'next-auth/client';
import useTranslation from 'next-translate/useTranslation';
import { AlertTitle } from '@material-ui/lab';
import { useRouter } from 'next/router';
import Link from './link';
import Alert from './alert';

export default function withAuth(Component, { pageTitle }) {
  return (props) => {
    const router = useRouter();
    // below is instead of 'signin' next-auth method
    const signin = () => {
      router.push(`/auth/welcome?callbackUrl=${window.location.href}`);
    };
    const [session] = useSession();
[...]

PR #1380, opened 2021-02-26
fix(page): add character encoding and page titles

### LINKED ISSUE #1284: Add the character encoding of the signin page
**Describe the bug**
The sign in page did not have a character encoding.

**Steps to reproduce**
Just click a button with `signIn()` method on it.

**Expected behavior**
You'll see this error in your console. 
```
The character encoding of the HTML document was not declared. The document will render with garbled text in some browser configurations if the document contains characters from outside the US-ASCII range. The character encoding of the page must be declared in the document or in the transfer protocol.
```

**Screenshots or error [logs**]
image

**Additional context**
Just add a
```html
<meta charset="utf-8" />
```
in the `<Head>` of the page and the error will go away.  
And also please add favicon too.  
Thank You and have a great day. 

**Feedback**

* [x] Found the documentation helpful

ExpressGradient:
**What**:
Fixed the character encoding for the sign in page

**Why**:
Every HTML document needs to have a UTF-8 character encoding.

**How**:
Just added a simple 
```html
<meta charset="utf-8">
```
in `src/server/pages/index.js`
Also added a missing title
```html
<title>Next Auth Sign in</title>
```

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [x] Ready to be merged
      

Fixes #1284

### DIFF
--- a/src/server/pages/index.js
+++ b/src/server/pages/index.js
@@ -14,9 +14,29 @@ export default function renderPage (req, res) {
   }
 
   return {
-    signin (props) { send(signin({ csrfToken, providers, callbackUrl, ...req.query, ...props })) },
-    signout (props) { send(signout({ csrfToken, baseUrl, basePath, ...props })) },
-    verifyRequest (props) { send(verifyRequest({ baseUrl, ...props })) },
-    error (props) { send(error({ basePath, baseUrl, res, ...props })) }
+    signin (props) {
+      send({
+        html: signin({ csrfToken, providers, callbackUrl, ...req.query, ...props }),
+        title: 'Sign In'
+      })
+    },
+    signout (props) {
+      send({
+        html: signout({ csrfToken, baseUrl, basePath, ...props }),
+        title: 'Sign Out'
+      })
+    },
+    verifyRequest (props) {
+      send({
+        html: verifyRequest({ baseUrl, ...props }),
+        title: 'Verify Request'
+      })
+    },
+    error (props) {
+      send({
+        html: error({ basePath, baseUrl, res, ...props }),
+        title: 'Error'
+      })
+    }
   }
 }
--- a/src/server/pages/index.js
+++ b/src/server/pages/index.js
@@ -9,24 +9,8 @@ export default function renderPage (req, res) {
   const { baseUrl, basePath, callbackUrl, csrfToken, providers, theme } = req.options
 
   res.setHeader('Content-Type', 'text/html')
-  function send (html) {
-    res.send(`<!DOCTYPE html>
-      <html lang="en">
-      <head>
-          <meta charset="UTF-8">
-          <meta http-equiv="X-UA-Compatible" content="IE=edge">
-          <meta name="viewport" content="width=device-width, initial-scale=1.0">
-          <style>
-              ${css()}
-          </style>
-          <title>Next Auth Sign in</title>
-      </head>
-      <body class="__next-auth-theme-${theme}">
-          <div class="page">
-              ${html}
-          </div>
-      </body>
-      </html>`);
+  function send ({ html, title }) {
+    res.send(`<!DOCTYPE html><html lang="en"><head><meta charset="UTF-8"><meta http-equiv="X-UA-Compatible" content="IE=edge"><meta name="viewport" content="width=device-width, initial-scale=1.0"><style>${css()}</style><title>${title}</title></head><body class="__next-auth-theme-${theme}"><div class="page">${html}</div></body></html>`)
   }
 
   return {
--- a/src/server/pages/index.js
+++ b/src/server/pages/index.js
@@ -10,7 +10,23 @@ export default function renderPage (req, res) {
 
   res.setHeader('Content-Type', 'text/html')
   function send (html) {
-    res.send(`<!DOCTYPE html><head><style type="text/css">${css()}</style><meta name="viewport" content="width=device-width, initial-scale=1"></head><body class="__next-auth-theme-${theme}"><div class="page">${html}</div></body></html>`)
+    res.send(`<!DOCTYPE html>
+      <html lang="en">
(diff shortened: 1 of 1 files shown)


src/server/pages/index.js
```diff
@@ -10,7 +10,23 @@ export default function renderPage (req, res) {
 
   res.setHeader('Content-Type', 'text/html')
   function send (html) {
-    res.send(`<!DOCTYPE html><head><style type="text/css">${css()}</style><meta name="viewport" content="width=device-width, initial-scale=1"></head><body class="__next-auth-theme-${theme}"><div class="page">${html}</div></body></html>`)
+    res.send(`<!DOCTYPE html>
```
balazsorban44, line 13: So obviously we could generate this smarter than an inline string... But for now - since it is such simple code - I think we can live with it. The reason it was on a simple line is (I guess) to make the sent html as small as possible. Could you remove the white spaces to reflect the code that was here before?

src/server/pages/index.js
```diff
@@ -10,7 +10,23 @@ export default function renderPage (req, res) {
...
+      <html lang="en">
+      <head>
+          <meta charset="UTF-8">
+          <meta http-equiv="X-UA-Compatible" content="IE=edge">
+          <meta name="viewport" content="width=device-width, initial-scale=1.0">
+          <style>
+              ${css()}
+          </style>
+          <title>Next Auth Sign in</title>
```
balazsorban44, line 22: This one is probably a good idea to add, but if you check the lines 33-36 in this file, you can see that simply setting `Next Auth Sign in` is not sufficient. I think a nice way of setting a custom title for each of the page types would be if the `send` function took a `title` param in addition to html, and add individual titles for `signin`, `signout`, `verifyRequest` and `error`.

I would also prefer if the function signature becomes: `function send({html, title})` to future proof us. #1154 would require us to send `locale` or something similar here.

balazsorban44 (changes requested):
Thanks so much! Only some small change requests. :slightly_smiling_face:

Also note that linting has failed! Try running `npm run lint --fix` before commiting.

src/server/pages/index.js
ExpressGradient: Sure. I just thought it would be more readable if its multi-line.  
But as you wish.

src/server/pages/index.js
ExpressGradient: Ok, so should I change the function's signature to `function send({ html, title })`?

src/server/pages/index.js
balazsorban44: It is certainly, and we should maybe create a new .html file so we get syntax highlighting and so on, but this could be a different PR. Than we should also minify that file and inject variables into it. It's just more work not indeed for this PR, but ideas are welcome! :)

src/server/pages/index.js
balazsorban44: yes. Usually I have a rule of thumb, if you need more than one param I just like to name them for convinience.

balazsorban44 (changes requested):
Sorry if I explained myself poorly, my thoughts were that you also update this section:
https://github.com/nextauthjs/next-auth/blob/d8d497cc91a73adcc0eec2b1479a342f0b97f440/src/server/pages/index.js#L17-L20

So it becomes something like this. (Please run the linter before commiting)
```diff
- signin (props) { send(signin({ csrfToken, providers, callbackUrl, ...req.query, ...props })) },
+ signin (props) { send({
+   html: signin({ csrfToken, providers, callbackUrl, ...req.query, ...props }),
+   title: "Sign in
+ }) },
- signout (props) { send(signout({ csrfToken, baseUrl, basePath, ...props })) },
+ signout (props) { send({
+   html: signout({ csrfToken, baseUrl, basePath, ...props }),
+   title: "Sign out
+ }) },
- verifyRequest (props) { send(verifyRequest({ baseUrl, ...props })) },
+ verifyRequest (props) { send({
+   html: verifyRequest({ baseUrl, ...props }),
+   title: "Verify request"
+ }) },
- error (props) { send(error({ basePath, baseUrl, res, ...props })) }
+ error (props) { send({
+   html: error({ basePath, baseUrl, res, ...props }),
+   title: "Error
+ }) }
```

huynhminhcanh:
31410003781426

ExpressGradient:
> 
> 
> Sorry if I explained myself poorly, my thoughts were that you also update this section:
> 
> https://github.com/nextauthjs/next-auth/blob/d8d497cc91a73adcc0eec2b1479a342f0b97f440/src/server/pages/index.js#L17-L20
> 
> So it becomes something like this. (Please run the linter before commiting)
> 
> ```diff
> - signin (props) { send(signin({ csrfToken, providers, callbackUrl, ...req.query, ...props })) },
> + signin (props) { send({
> +   html: signin({ csrfToken, providers, callbackUrl, ...req.query, ...props }),
> +   title: "Sign in
> + }) },
> - signout (props) { send(signout({ csrfToken, baseUrl, basePath, ...props })) },
> + signout (props) { send({
> +   html: signout({ csrfToken, baseUrl, basePath, ...props }),
> +   title: "Sign out
> + }) },
> - verifyRequest (props) { send(verifyRequest({ baseUrl, ...props })) },
> + verifyRequest (props) { send({
> +   html: verifyRequest({ baseUrl, ...props }),
> +   title: "Verify request"
> + }) },
> - error (props) { send(error({ basePath, baseUrl, res, ...props })) }
> + error (props) { send({
> +   html: error({ basePath, baseUrl, res, ...props }),
> +   title: "Error
> + }) }
> ```

No no, you explained it right, it's my dumb brain couldn't interpret it.  
Anyways, is there anything else more?

balazsorban44 (approved):
LGTM! Thanks :slightly_smiling_face:

PR #1393, opened 2021-02-28
feat: improve and export typescript definitions

euxn23:
**What**:

Add typings, improved based on DefinitelyTyped's implementation, and expose from this Library.

**Why**:

Now `@types/next-auth` is provided by DefinitelyTyped, but that implementation is behinded from this JS implementation. (such as new NextAuth configration, Requests, Responses and Providers)
I improved type definitions and made pull request on DefinitelyTyped repository(https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51469), but as you see, result of CI is broken.
The reason is `next-auth` types depends on `next` types in `next` package, not in `@types/next`. Type definitions of next uses `ts-ignore` in a few files, but DefinitelyTyped not allow `ts-ignore` and there is no way to avoid `ts-ignore` banning in contributor's config.

As described above, now it is too difficult to pass all CI on DefinitelyTyped, maybe for everybody, without help and fixing DefinitelyTyped. This problem had been discussed in https://github.com/microsoft/dtslint/issues/297#issuecomment-733195100 , but discussion seems to be stoped.
So I think this is the time to provide type definitions in `next-auth` own package.

**How**:

I wrote types reading JS implementations honestly.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged
      

@lluia
Thanks to provide current type definitions!
I rewrited too much files, so I'll appreciate if you reviewed and checked there is no degrade.


balazsorban44:
Right now, our main priority is not built-in TypeScript support (although there are internal minor improvements sporadically, utilizing JSDoc and .d.ts files). If you are not satisfied with the type definitions provided from `@types/next-auth`, I suggest you open PRs against https://github.com/DefinitelyTyped/DefinitelyTyped instead.

Currently, it is just too much overhead for the maintainers to also take care of TS.

Regarding `@types/next`, it is [deprecated](https://www.npmjs.com/package/@types/next) and thus we don't use it:

image

PR #1428, opened 2021-03-03
refactor: client improvements

balazsorban44:
**What**:

Simplify client code, and add TypeScript+JSDoc declarations.

**Why**:

For better DX. Also, in the process, I have been learning a lot about the choices and implementation. The bundle is also 821 bytes leaner. 🥳

**How**:

Abstracted away the storage syncing into a handy `BroadcastChannel` function inspired by https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API (which does not work in Safari, unfortunately :slightly_frowning_face: )

Also removed unnecessary boolean checks in if statements, simplified if blocks, exported all user-facing API other than the `export default` keyword. (In local development, imports from "next-auth/client" did not work for some reason. Might be something with the dev app setup)

the concept of `clientId` was not necessary, since `storage` events already only listen to other window contexts. See https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      

#844, #1084

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -39,130 +39,47 @@ const __NEXTAUTH = {
 const logger = proxyLogger(_logger, __NEXTAUTH.basePath)
 
 // Add event listners on load
-if (typeof window !== 'undefined') {
-  if (__NEXTAUTH._eventListenersAdded === false) {
-    __NEXTAUTH._eventListenersAdded = true
+if (typeof window !== 'undefined' && !__NEXTAUTH._eventListenersAdded) {
+  __NEXTAUTH._eventListenersAdded = true
 
-    // Listen for storage events and update session if event fired from
-    // another window (but suppress firing another event to avoid a loop)
-    window.addEventListener('storage', async (event) => {
-      if (event.key === 'nextauth.message') {
-        const message = JSON.parse(event.newValue)
-        if (message?.event === 'session' && message.data) {
-          // Ignore storage events fired from the same window that created them
-          if (__NEXTAUTH._clientId === message.clientId) {
-            return
-          }
-
-          // Fetch new session data but pass 'true' to it not to fire an event to
-          // avoid an infinite loop.
-          //
-          // Note: We could pass session data through and do something like
-          // `setData(message.data)` but that can cause problems depending
-          // on how the session object is being used in the client; it is
-          // more robust to have each window/tab fetch it's own copy of the
-          // session object rather than share it across instances.
-          await __NEXTAUTH._getSession({ event: 'storage' })
+  // Listen for storage events and update session if event fired from
+  // another window (but suppress firing another event to avoid a loop)
+  window.addEventListener('storage', async (event) => {
+    if (event.key === 'nextauth.message') {
+      const message = JSON.parse(event.newValue)
+      if (message?.event === 'session' && message.data) {
+        // Ignore storage events fired from the same window that created them
+        if (__NEXTAUTH._clientId === message.clientId) {
+          return
         }
+
+        // Fetch new session data but pass 'true' to it not to fire an event to
+        // avoid an infinite loop.
+        //
+        // Note: We could pass session data through and do something like
+        // `setData(message.data)` but that can cause problems depending
+        // on how the session object is being used in the client; it is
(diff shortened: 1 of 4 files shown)


src/client/index.js
```diff
@@ -20,167 +18,65 @@ import parseUrl from '../lib/parse-url'
...
+  keepAlive: 0,
+  clientMaxAge: 0,
   // Properties starting with _ are used for tracking internal app state
-  _clientLastSync: 0, // used for timestamp since last sycned (in seconds)
-  _clientSyncTimer: null, // stores timer for poll interval
-  _eventListenersAdded: false, // tracks if event listeners have been added,
-  _clientSession: undefined, // stores last session response from hook,
-  // Generate a unique ID to make it possible to identify when a message
-  // was sent from this tab/window so it can be ignored to avoid event loops.
-  _clientId: Math.random().toString(36).substring(2) + Date.now().toString(36),
```
balazsorban44, line 35: Moved to BroadcastChannel's internals.

src/client/index.js
```diff
@@ -20,167 +18,65 @@ import parseUrl from '../lib/parse-url'
...
-  clientMaxAge,
-  keepAlive
-} = {}) => {
-  if (baseUrl) { __NEXTAUTH.baseUrl = baseUrl }
-  if (basePath) { __NEXTAUTH.basePath = basePath }
-  if (clientMaxAge) { __NEXTAUTH.clientMaxAge = clientMaxAge }
-  if (keepAlive) {
-    __NEXTAUTH.keepAlive = keepAlive
-
-    if (typeof window !== 'undefined' && keepAlive > 0) {
```
balazsorban44, line 103: since the `if(keepAlive)` check, `keepAlive` is always bigger than 0. (0 is falsy). No need to check.

src/client/index.js
```diff
@@ -345,40 +301,126 @@ export async function signOut (options = {}) {
...
+    receive (onReceive) {
+      if (typeof window === 'undefined') return
+      window.addEventListener('storage', async (event) => {
+        if (event.key !== name) return
+        /** @type {import(".").BroadcastMessage} */
+        const message = JSON.parse(event.newValue)
+        if (message?.event !== 'session' || !message?.data ||
+           // Ignore storage events fired from the same window that created them
+           message?.clientId === clientId
+        ) return
```
balazsorban44, line 412: Do we need this check? MDN says:
> The storage event of the Window interface fires when a storage area (localStorage) has been modified in the context of another document.

https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event

src/client/index.js
```diff
@@ -345,40 +301,126 @@ export async function signOut (options = {}) {
...
+     */
+    receive (onReceive) {
+      if (typeof window === 'undefined') return
+      window.addEventListener('storage', async (event) => {
+        if (event.key !== name) return
+        /** @type {import(".").BroadcastMessage} */
+        const message = JSON.parse(event.newValue)
+        if (message?.event !== 'session' || !message?.data ||
+           // Ignore storage events fired from the same window that created them
+           message?.clientId === clientId
```
balazsorban44, line 411: Do we need this check? MDN says:
> The storage event of the Window interface fires when a storage area (localStorage) has been modified in the context of another document.

https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event

src/client/index.js
```diff
@@ -345,40 +301,126 @@ export async function signOut (options = {}) {
...
+ * Only not using it directly, because Safari does not support it.
+ *
+ * https://caniuse.com/?search=broadcastchannel
+ */
+function BroadcastChannel (name = 'nextauth.message') {
+  /**
+   * Unique ID to make it possible to identify when a message
+   * was sent from this tab/window so it can be ignored to avoid event loops.
+   */
+  const clientId = Math.random().toString(36).substring(2) + Date.now().toString(36)
```
balazsorban44, line 397: Do we need this? MDN says:
> The storage event of the Window interface fires when a storage area (localStorage) has been modified in the context of another document.

https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event

PR #1468, opened 2021-03-07
Fix useSession loading state

### LINKED ISSUE #1467: Always redirected to /auth/signin after updating to 3.8.0
**Describe the bug**
I noticed an issue after upgrading to today's latest and after trying each version between 3.1.0 and 3.11.0, I managed to determine that the issue mentioned below started with 3.8.0. Even though the release notes only mentioned adding the Instagram provider, I noticed that there was also quite a large refactoring of the client code in #1428. I think the issue comes from this commit.

The issue I have is that if I go to a page different than `/auth/signin`, I'm always redirected to `/auth/signin` even though I'm already logged-in. Then, once the signin page opens, I'm redirected back to the original requested page as I'm already logged-in.

Then, when I land on the original page, something extremely slow is going on and I can see my chrome devtools using more than 100% of CPU. I have the same behaviour on Firefox though. (This slowness is only on 3.8.0 and not on the latest though, probably fixed by https://github.com/nextauthjs/next-auth/pull/1462).

**Steps to reproduce**
My application is quite large so I'm still in the process of trying to find why I get redirected on load.
 
**Expected behavior**
When I am logged-in and refresh a page, I should not be redirected to the signin page.

**Screenshots**
I cannot tell if this happens before or after the redirection but I actually have an error in the console:
image

It might be because the  `/session` call is cancelled: 
image

ValentinH:
Fixes #1467

The issue was due to doing the `setLoading(false)` in the finally:  as we can do an early return [here](https://github.com/nextauthjs/next-auth/blob/a7e08e2a3266efa9c82eb859e7141c798fcf07ae/src/client/index.js#L100-L100), we would still go to the finally and mark the session as being loaded.

I simply removed the `finally` block to only set the `loading` state to false when:
- the data is ready
- an error occurs

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -130,9 +130,9 @@ function _useSessionHook (session) {
         __NEXTAUTH._clientSession = newClientSessionData
 
         setData(newClientSessionData)
+        setLoading(false)
       } catch (error) {
         logger.error('CLIENT_USE_SESSION_ERROR', error)
-      } finally {
         setLoading(false)
       }
     }


balazsorban44 (approved):
You are right! Thanks for catching this 🙂

Issue #1461, opened 2021-03-06
Inconsistent session returns after 3.8.0


theycallmeswift:
**Describe the bug**

A bug was introduced starting in 3.8.0 that prevents the session from being read multiple times. If you have multiple components that rely on the `useSession` hook, only the first one will return a session object as expected.

**Steps to reproduce**

You can introduce this behavior to the [next-auth-example](https://github.com/nextauthjs/next-auth-example) with the following diff. Changes added to `pages/index.js` are for easy visibility.

```diff
diff --git a/package.json b/package.json
index c305f4a..9b0f580 100644
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "license": "ISC",
   "dependencies": {
     "next": "^10.0.7",
-    "next-auth": "^3.6.0",
+    "next-auth": "^3.8.0",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "sqlite3": "^5.0.2"
diff --git a/pages/index.js b/pages/index.js
index f4aa3e1..ec76748 100644
--- a/pages/index.js
+++ b/pages/index.js
@@ -1,12 +1,17 @@
 import Layout from '../components/layout'

+import { signIn, signOut, useSession  } from "next-auth/client";
+
 export default function Page () {
+  const [session, loading] = useSession();
+
   return (
     <Layout>
       <h1>NextAuth.js Example</h1>
       <p>
         This is an example site to demonstrate how to use <a href={`https://next-auth.js.org`}>NextAuth.js</a> for authentication.
       </p>
+      <pre>{JSON.stringify(session?.user || {}, 2)}</pre>
     </Layout>
```

**Expected behavior**

![image](https://user-images.githubusercontent.com/434373/110215203-8060cb80-7e76-11eb-8cd7-9193d7b630d4.png)

**Note:** This works as expected with 3.7.1 and below.

**Actual behavior**

![image](https://user-images.githubusercontent.com/434373/110215250-b3a35a80-7e76-11eb-9a02-f9d8683e31af.png)

**Note:** The session works as expected for the header but returns an empty object the second time the hook is used.

**Feedback**
*Documentation refers to searching through [online documentation](https://next-auth.js.org), code comments and issue history. The example project refers to [next-auth-example](https://github.com/iaincollins/next-auth-example).*

* [ ] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [X] Could not find relevant documentation
* [X] Found the example project helpful
[...]

PR #1473, opened 2021-03-08
feat(react): create client tailored to React

balazsorban44:
**What**:

These changes ensure that we work more tightly with React that can also result in unforeseen performance boosts. In case we would decide on expanding to other libraries/frameworks, a new file per framework could be added.

**Why**:

Some performance issues (https://github.com/nextauthjs/next-auth/issues/844) could only be fixed by moving more of the client code into the `Provider`.

**How**:

Refactoring `next-auth/client`

Related: #1461, #1084, #1462

BREAKING CHANGE:
**1.** `next-auth/client` is renamed to `next-auth/react`.

**2.** In the past, we exposed most of the functions with different names for convenience. To simplify our source code, the new React specific client code exports only the following functions, listed with the necessary changes:

- `setOptions`: Not exposed anymore, use `SessionProvider` props
- `options`: Not exposed anymore, use `SessionProvider` props
- `session`: Rename to `getSession`
- `providers`: Rename to `getProviders`
- `csrfToken`: Rename to `getCsrfToken`
- `signin`: Rename to `signIn`
- `signout`: Rename to `signOut`
- `Provider`: Rename to `SessionProvider`

**3.** `Provider` changes.
- `Provider` is renamed to `SessionProvider`
- The `options` prop is now flattened as the props of `SessionProvider`.
- `clientMaxAge` has been renamed to `staleTime`.
- `keepAlive` has been renamed to `refetchInterval`.
An example of the changes:
```diff
- <Provider options={{clientMaxAge: 0, keepAlive: 0}}>{children}</Provider>
+ <SessionProvider staleTime={0} refetchInterval={0}>{children}</SessionProvider> 
```

**4.** It is now **required** to wrap the part of your application that uses `useSession` into a `SessionProvider`.

Usually, the best place for this is in your `pages/_app.jsx` file:

```jsx
import { SessionProvider } from "next-auth/react"

export default function App({
  Component,
  pageProps: { session, ...pageProps }
}) {
  return (
    // `session` comes from `getServerSideProps` or `getInitialProps`.
    // Avoids flickering/session loading on first load.
    <SessionProvider session={session}>
      <Component {...pageProps} />
    </SessionProvider>
  )
}
```


www/docs/tutorials/securing-pages-and-api-routes.md
```diff
@@ -69,7 +69,7 @@ export async function getServerSideProps(context) {
 This example assumes you have configured `_app.js` to pass the `session` prop through so that it's immediately available on page load to `useSession`.
 
 ```js title="pages/_app.js"
-import { Provider } from 'next-auth/client'
+import { Provider } from "next-auth/react"
```
hboylan, line 72: Looks like you've renamed `Provider` to `SessionProvider` so all documentation references need to be updated as well.

hboylan (commented):
Looks good so far!

www/docs/tutorials/securing-pages-and-api-routes.md
balazsorban44: Exactly. This is still a WIP, but thanks for the heads up. :)

src/client/react.js
```diff
@@ -0,0 +1,351 @@
...
+      window.addEventListener("storage", handler)
+      return () => window.removeEventListener("storage", handler)
+    },
+    /** Notify other tabs/windows. */
+    post(message) {
+      localStorage.setItem(
+        name,
+        JSON.stringify({ ...message, timestamp: _now() })
+      )
+    },
```
hboylan, line 349:
```suggestion
    post(message) {
      if (typeof localStorage === "undefined") return
      localStorage.setItem(
        name,
        JSON.stringify({ ...message, timestamp: _now() })
      )
    },
```

hboylan (commented):
I used `npm link` for my existing project to confirm this solves the issue of the unnecessary client-side session call. 🎉 

I had to install `nodemailer`, which seems like it only be an `peerOptionalDependency` though.

src/client/react.js
balazsorban44: My experience was that this check wasn't needed. Did you run into problems without it?

balazsorban44:
@hboylan yes, from version 4, nodemailer will be optional, as it's only needed if someone uses the email provider. anyone else won't have to install it. smaller bundle size, faster install. 😊

0ubbe (approved):
Impressive work 💯 , did a pair programming review with @balazsorban44 and in my opinion, we're good to go as a base for **v4** 👍🏽

src/client/react.js
hboylan: I believe it's needed, but I may be doing something wrong. My `getSession` call on the server yields the error "ReferenceError: localStorage is not defined".
Screen Shot 2021-06-11 at 2 41 22 PM

balazsorban44: I see :grimacing: I just merged this PR, but I'll create another one with a fix.

hboylan: Ah, sorry for my delayed response! I'll give v4 a try now 😄

Issue #1341, opened 2021-02-19
Nextjs10 Internationalized Routing support


YuriGor:
**Summary of proposed feature**
Would be great if `signin` from next-auth/client will support [Nextjs10 Internationalized Routing](https://nextjs.org/docs/advanced-features/i18n-routing) feature.
Currently `signin` method directly opens signIn page as configured in `[...nextauth].js` in 'pages' section not considering current locale.

**Purpose of proposed feature**
Support internalization in custom signIn form.

**Detail about proposed feature**
Currently I am using next/router to navigate to signIn page instead of `signin` method, for example:
```js
import { useRouter } from 'next/router';
import { useSession } from 'next-auth/client';
import useTranslation from 'next-translate/useTranslation';
import { AlertTitle } from '@material-ui/lab';
import { useRouter } from 'next/router';
import Link from './link';
import Alert from './alert';

export default function withAuth(Component, { pageTitle }) {
  return (props) => {
    const router = useRouter();
    // below is instead of 'signin' next-auth method
    const signin = () => {
      router.push(`/auth/welcome?callbackUrl=${window.location.href}`);
    };
    const [session] = useSession();
    const { t } = useTranslation('common');
    if (!session) {
      return (
        <Alert severity="access">
          <AlertTitle>{t('welcomeBack')}</AlertTitle>
          {pageTitle && t('accessingPage', { pageTitle: t(pageTitle) })}
          <br />
          {t('pleaseSignInLabel')}
          <Link href="#" onClick={signin}>
            {t('pleaseSignInButton')}
          </Link>
          !
        </Alert>
      );
    }
    return <Component {...props} />;
  };
}
```

**Potential problems**
I don't know if it's easy for next-auth to use `next/router` internally and maybe I have some drawback when I use `next/router` instead of `signin`?

**Additional context**
I think the same should be implemented for the rest of custom pages, but I don't use them so can't say for sure.

I am using `next-translate` but I don't think it does matter, because this lib relies on [Nextjs10 Internationalized Routing](https://nextjs.org/docs/advanced-features/i18n-routing) mentioned above.

*Please indicate if you are willing and able to help implement the proposed feature.*
I am not familiar with next-auth internals, so I am not ready for PR, but I can assist in testing.

YuriGor:
oh sorry, dup of #1154

PR #1489, opened 2021-03-09
feat: i18n routing & custom translations support

### LINKED ISSUE #1154: Multi language support

iffa:
**What**:

adds locale / internationalized routing support to default pages + proper redirection to locale-specific custom pages

**Why**:

#1154, #1341

i18n should be supported out of the box & easily configurable. 

**How**:

user locale is kept track of, and passed along with all requests to allow showing potential translated strings in default pages. user locale is prepended to custom pages as well, so these should work out of the box with the proper locale.

**Checklist**:

- [x] Locale included in all requests / pages, not lost in any error scenario
- [x] Allow passing of translations to default pages **and** default verify request email
- [x] Cleanup of debug locale info everywhere it doesn't belong (or show if debug=true?)
- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

**Notes, thoughts, questions**:

This is still a work-in-progress. Open to any feedback/code review or contributions. 

I don't know if the approach I've taken here is the right one, but it works well with minimal changes to the existing code base, and is (**should be**) backwards compatible.

> *For some reason the divider with "or" text is implemented with css pseudo elements, which needs to be refactored so it can be passed translations to.*
> -> css attributes to the rescue.

> *[Credentials provider](https://next-auth.js.org/providers/credentials) has hardcoded labels for fields, how should we handle translations for those labels in a non-breaking way? Allow passing object instead of string for credential labels?*
> -> yep, now accepts string or object with translations for all defined locales

> *Although a niche feature, does this work with i18n domain routing?*
> -> no idea.

Closes #1154


src/css/index.css
```diff
@@ -189,7 +189,7 @@ a.site {
     overflow: visible;
 
     &::before {
-      content: "or";
+      content: attr(divider-text);;
```
iffa, line 192: https://accessibleweb.com/question-answer/how-is-css-pseudo-content-treated-by-screen-readers/

may be nitpicking but this probably shouldn't be a pseudo element for accessibility reasons

components/layout.js
```diff
@@ -1,11 +1,16 @@
...
 
 export default function Layout ({ children }) {
+  const { locale, locales } = useRouter()
+
   return (
     <>
       <Header />
       <main>
+        <p>Current locale: {locale || 'N/A'}</p>
+        <p>Available locales: {locales ? locales.join(', ') : 'N/A'}</p>
```
balazsorban44, line 13:
```suggestion
        <p>Available locales: {locales?.join(', ') ?? 'N/A'}</p>
```

package.json
```diff
@@ -46,6 +46,7 @@
     "futoin-hkdf": "^1.3.2",
     "jose": "^1.27.2",
     "jsonwebtoken": "^8.5.1",
+    "lodash.merge": "^4.6.2",
```
balazsorban44, line 49: Would like to avoid any unnecessary packages, please remove this.

src/client/index.js
```diff
@@ -11,6 +11,7 @@
 import { useState, useEffect, useContext, createContext, createElement } from 'react'
```
balazsorban44, line 11: I have to go through the client code more thoroughly, but I don't think we should make it configurable by the client. I actually think we shouldn't make this configurable at all if possible, we should just follow the language Next.js provides. Haven't used Next.js's i18n features yet, so I'm not totally sure just yet if that is possible. Will check.

src/providers/email.js
```diff
@@ -1,5 +1,6 @@
 import nodemailer from 'nodemailer'
 import logger from '../lib/logger'
+const merge = require('lodash.merge')
```
balazsorban44, line 3:
```suggestion
```

src/css/index.css
huynhminhcanh: https://guides.github.com/activities/hello-world/readme-edits.gif

src/providers/email.js
```diff
@@ -22,7 +23,20 @@ export default (options) => {
...
+  html: {
+    signInAs: 'Sign in as <strong>%s</strong>',
+    signIn: 'Sign In',
+    didNotRequestHint: 'If you did not request this email you can safely ignore it.'
+  }
+}
+
+const sendVerificationRequest = ({ identifier: email, url, baseUrl, provider, locale, translations }) => {
+  // merge default and client provided translations
+  const texts = merge({}, defaultTranslations, translations?.[locale]?.email)
```
balazsorban44, line 38: As long as translations are flat, I think we get away really nicely with only using the spread `...` operator. If we have to deep merge objects, I would still prefer if we wrote a simple utility in another file rather than adding a new dependency.
```suggestion
  const texts = {...defaultTranslations, ...translations?.[locale]?.email }
```

huynhminhcanh (approved):

src/server/index.d.ts
```diff
@@ -33,14 +43,15 @@ export interface NextAuthOptions {
...
-    signIn?: string
-    signOut?: string
-    /** Error code passed in query string as ?error= */
-    error?: string
-    verifyRequest?: string
-    /** If set, new users will be directed here on first sign in */
-    newUser?: string
+  pages?: NextAuthPages
+  /** @docs https://next-auth.js.org/configuration/options#translations  */
+  translations?: {
```
balazsorban44, line 48: think we should just call it `locales` all the way
```suggestion
  locales?: {
```

src/server/index.d.ts
```diff
@@ -62,10 +73,12 @@ export interface NextAuthOptions {
   cookies?: CookiesOptions
   /** @docs https://next-auth.js.org/configuration/options#logger */
   logger: LoggerInstance
+  /** @docs https://next-auth.js.org/configuration/options#locale */
+  locale?: string
 }
 
 /** Options that are the same both in internal and user provided options. */
-export type NextAuthSharedOptions = 'pages' | 'jwt' | 'events' | 'callbacks' | 'cookies' | 'secret' | 'adapter' | 'theme' | 'debug' | 'logger'
+export type NextAuthSharedOptions = 'pages' | 'jwt' | 'events' | 'callbacks' | 'cookies' | 'secret' | 'adapter' | 'theme' | 'debug' | 'logger' | 'translations'
```
balazsorban44, line 81:
```suggestion
export type NextAuthSharedOptions = 'pages' | 'jwt' | 'events' | 'callbacks' | 'cookies' | 'secret' | 'adapter' | 'theme' | 'debug' | 'logger' | 'locales'
```

src/server/index.js
```diff
@@ -69,7 +69,9 @@ async function NextAuthHandler (req, res, userOptions) {
 
     const { csrfToken, csrfTokenVerified } = csrfTokenHandler(req, res, cookies, secret)
 
-    const providers = parseProviders({ providers: userOptions.providers, baseUrl, basePath })
+    const locale = req.query.locale || undefined
```
balazsorban44, line 72: Think this one needs to be moved to https://github.com/nextauthjs/next-auth/blob/2d9711eeefd2c5826192a0a5e9e30f39264bb1bf/src/server/index.js#L52-L57

As we use a catch-all route.

How would the URLs look like by the way? `/api/auth/en/signin` or something? I have to think more about this.

huynhminhcanh:
[**__**]()

src/server/index.js
```diff
@@ -149,16 +152,18 @@ async function NextAuthHandler (req, res, userOptions) {
           return res.json({ csrfToken })
         case 'signin':
           if (pages.signIn) {
-            let signinUrl = `${pages.signIn}${pages.signIn.includes('?') ? '&' : '?'}callbackUrl=${req.options.callbackUrl}`
+            // prepend locale to custom page path if available
+            let signinUrl = `${locale ? `/${locale}` : ''}${pages.signIn}${pages.signIn.includes('?') ? '&' : '?'}callbackUrl=${req.options.callbackUrl}`
```
balazsorban44, line 156: What a mess with the URLs... 😅 Not your fault though, but I think we need to clean it up a bit somehow...

Have been thinking of something like this lately:

```js
class URLRelative extends URL {
  constructor (path) {
    super(new URL(path, 'http://n'))
  }

  toString () {
    return `${this.pathname}${this.search}`
  }
}
```

Then we could utilize the URL API, also for relative URLs.

src/server/pages/error.js
```diff
@@ -59,6 +61,8 @@ export default function error ({ baseUrl, basePath, error = 'default', res }) {
   return (
     <div className='error'>
       <h1>{heading}</h1>
+      {locale &&
+        <p>Locale: {locale || 'N/A'}</p>}
```
balazsorban44, line 65: No use for the `locale &&` check if you fall back to `N/A` anyway
```suggestion
      <p>Locale: {locale || 'N/A'}</p>
```

src/server/pages/index.js
```diff
@@ -4,39 +4,95 @@ import signout from './signout'
 import verifyRequest from './verify-request'
 import error from './error'
 import css from '../../css'
+const merge = require('lodash.merge')
+
+export const defaultTranslations = {
```
balazsorban44, line 9: Think we should move this into its own file like `default-events.js` and `default-callbacks.js` are. I think this could actually be just a `default-translations.json` file as well.

src/server/pages/signin.js
```diff
@@ -14,23 +14,12 @@ export default function signin ({ csrfToken, providers, callbackUrl, email, erro
...
-    default: 'Unable to sign in.'
-  }
-
-  const error = errorType && (errors[errorType] ?? errors.default)
+  const error = errorType && (texts.errors[errorType] ?? texts.errors.default)
 
   return (
     <div className='signin'>
+      {locale &&
+        <p>Locale: {locale || 'N/A'}</p>}
```
balazsorban44, line 22: I don't think we would actually need this. Especially with a hardcoded `Locale:` English text. What was the intention here?

src/server/routes/callback.js
```diff
@@ -110,34 +111,34 @@ export default async function callback (req, res) {
         // e.g. option to send users to a new account landing page on initial login
         // Note that the callback URL is preserved, so the journey can still be resumed
         if (isNewUser && pages.newUser) {
-          return res.redirect(`${pages.newUser}${pages.newUser.includes('?') ? '&' : '?'}callbackUrl=${encodeURIComponent(callbackUrl)}`)
+          return res.redirect(`${locale ? `/${locale}` : ''}${pages.newUser}${pages.newUser.includes('?') ? '&' : '?'}callbackUrl=${encodeURIComponent(callbackUrl)}`)
```
balazsorban44, line 114: OMG... 😅
Maybe we should make `res.redirect` smarter to take something like:

```res.redirect(`/${locale ?? ""}${pages.newUser}`, { callbackUrl } )```

And merge query params and a normalized (remove duplicate "/" for example. Node has [`path.normalize`](https://nodejs.org/api/path.html#path_path_normalize_path)) path to the correct URL.

What do you think?

**Review by balazsorban44 — CHANGES_REQUESTED** [review]
Overall a very nice start, thank you so much for the thorough work! I think we will take some rounds on this to get there, but I think it's on the right path!

**Comment by iffa** [comment]
I'll go through your notes later this week! Sorry about some of the spaghetti I introduced, it was a late night 😵

File: src/server/pages/signin.js (hunk 13 above)
Comment (iffa) [reply src/server/pages/signin.js:22]: Added these just for testing purposes, will remove them.

File: src/client/index.js (hunk 4 above)
Comment (iffa) [reply src/client/index.js:11]: The implementation I've got going here uses the locale provided by the Next.js router & assumes that it is not changed manually, so yes agree on that front.

File: src/providers/email.js (hunk 6 above)
Comment (iffa) [reply src/providers/email.js:38]: I think it's better to keep the nested/scoped translation structure, as its clearer to the end user as to what translations are used where. I'll rework this to avoid the extra dependency.

File: src/server/index.js (hunk 9 above)
Comment (iffa) [reply src/server/index.js:72]: This is what api routes look like with locale data: `http://localhost:3000/api/auth/signin?callbackUrl=url&locale=en`

Locale is tracked with a query param for all /api/auth/ routes. I think that is a better option than implementing it as part of the path.

File: src/server/pages/index.js (hunk 12 above)
Comment (iffa) [reply src/server/pages/index.js:9]: Agree, will do.

File: src/server/routes/callback.js (hunk 14 above)
Comment (iffa) [reply src/server/routes/callback.js:114]: I'm not that familiar (yet) with the whole codebase, but I assume you mean extending .redirect like its done for other res.* methods in `extend-req.js`? (which has a misleading name by the way :p)

But yeah this isn't the only instance of url-forming spaghetti that is made worse by including locale.

File: src/css/index.css (hunk 1 above)
Comment (iffa) [reply src/css/index.css:192]: Yeah never hurts to make sure the built-in pages are accessibility-friendly. definitely not in scope of this PR - we should probably open a new issue to track this.

**Comment by iffa** [comment]
Addressed most of the comments, other than the URL ones for now. Will try some things out to see what works best!

**Comment by iffa** [comment]
Made some changes: added cookie for locale info (like callbackUrl) so it won't be lost even if the query param is missing. Also cleaned up the URL forming spaghetti, it now uses an extended URL utility class so its easier to work with in the future & expand.

File: src/server/lib/locale-handler.js [hunk 15]
```diff
@@ -0,0 +1,18 @@
...
+  // get locale from body, query or cookie
+  const localeCookieValue = req.cookies[cookieOptions.locale.name] || null
+  const locale = body.locale || query.locale || localeCookieValue || null
+
+  // Save locale in a cookie so that can be used for subsequent requests in signin/signout/callback flow
+  if (locale && (locale !== localeCookieValue)) {
+    cookie.set(res, cookieOptions.locale.name, locale, cookieOptions.locale.options)
+  }
+
+  return locale
```
balazsorban44, line 17: We could instead extend `req.options.locale` to be `mergeDeep({}, defaultTranslations, locales[locale])`, so the internal state `req.options.locale` would only carry around the object containing the actual translations, instead of the locale string and the entire locales object. This would centralize this logic to `localeHandler` and wouldn't need another cookie.

src/server/pages/index.js
```diff
@@ -4,39 +4,44 @@ import signout from './signout'
...
-  const { baseUrl, basePath, callbackUrl, csrfToken, providers, theme } = req.options
+  const { baseUrl, basePath, callbackUrl, csrfToken, providers, theme, locale, locales } = req.options
 
   res.setHeader('Content-Type', 'text/html')
   function send ({ html, title }) {
     res.send(`<!DOCTYPE html><html lang="en"><head><meta charset="UTF-8"><meta http-equiv="X-UA-Compatible" content="IE=edge"><meta name="viewport" content="width=device-width, initial-scale=1.0"><style>${css()}</style><title>${title}</title></head><body class="__next-auth-theme-${theme}"><div class="page">${renderToString(html)}</div></body></html>`)
   }
 
+  // merge default and client provided translations
+  const texts = mergeDeep({}, defaultTranslations, locales[locale])
```
balazsorban44, line 20: This could be removed, see my comment on the `locale-handler` file.

src/providers/email.js
```diff
@@ -22,7 +24,10 @@ export default (options) => {
   }
 }
 
-const sendVerificationRequest = ({ identifier: email, url, baseUrl, provider }) => {
+const sendVerificationRequest = ({ identifier: email, url, baseUrl, provider, locale, locales }) => {
+  // merge default and client provided translations
+  const texts = mergeDeep({}, defaultTranslations.email, locales[locale]?.email)
```
balazsorban44, line 29: This could be removed, see my comment on the `locale-handler` file.

balazsorban44 (changes requested):
Another round with suggestions! We will get there! 😉

src/server/lib/locale-handler.js
```diff
@@ -0,0 +1,18 @@
+import * as cookie from '../lib/cookie'
+
+/**
+ * Get locale based on body / query param / cookie
+ */
+export default function localeHandler (req, res, cookieOptions) {
+  const { body, query } = req
+  // get locale from body, query or cookie
+  const localeCookieValue = req.cookies[cookieOptions.locale.name] || null
```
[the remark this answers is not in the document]
iffa: Seeing as the cookie in question changes routing behavior and is actually supposed to be set by language switching logic on the website itself, I don't think next-auth should set a value to this cookie as it may cause behavior changes the user may not expect. We could use it as one of the fallbacks if it is defined however?

src/server/lib/locale-handler.js
```diff
@@ -0,0 +1,18 @@
...
+  const locale = body.locale || query.locale || localeCookieValue || null
```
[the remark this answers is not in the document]
iffa: There should never be a scenario where this would fallback all the way to null, unless the app is single-locale. Can we even access the next config js file? I am not that familiar with the inner workings of NextJS 😄

src/server/lib/locale-handler.js
iffa: We do however need the locale string to keep track of the active locale, no? I'm not quite sure what you mean by this recommended change

src/server/routes/callback.js
balazsorban44: You brought my attention to this, so I did fix `extend-req`->`extend-res`: https://github.com/nextauthjs/next-auth/commit/dbe283f0fa5b0521b6a4be1f63bf916891c0d5fe  But I DID write it wrong in the commit message... 🤦 😅

danestves:
Any updates on this 👀?

iffa:
I haven't had time outside my day job to work on this or other projects. There are a lot of conflicts but surely nothing too extreme. Also got some hanging questions ^ from myself and @balazsorban44 . I'll have more time in a month or so, if there's still demand for this functionality!

balazsorban44:
Yeah, my focus have been elsewhere as well. We can come back to this a while later I think.

Kakise:
Hey there ! Thanks for your work on this feature, we need it for a company project and would love to help with dev. What are the points that need to be fixed before a merge ?

emrysal:
Hi everyone! This feature was just marked stale and I wonder if there is still a desire to have this being worked on? I would love this feature myself.

balazsorban44:
it certainly has to wait until v4 lands in stable. I don't have an ETA.

balazsorban44:
Instead of fixing all the conflicts, let's revisit this after v4 has landed in stable if there is still interest then.

I believe we are less than a month away from a stable release.

medihack:
Any chance to reopen this issue? I am especially interested in the support of i18n routes with next-auth.

addlistener:
Is this still being worked on?

PR #1516, opened 2021-03-13
feat(provider): add Zoho provider

jkwiz:
**What**: I'm working on a project with Zoho so I figured I would contribute to the library.

**Why**: Others may find it helpful if they also use Zoho.

**How**: Following the documentation and examples to add a provider.

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [x] Ready to be merged


balazsorban44 (approved):

www/docs/providers/zoho.md
```diff
@@ -0,0 +1,26 @@
+---
```
balazsorban44, line 1: So there seems to be a very strange issue here. Our script that should generate some of the docs reads these .md files with some RegEx to extract the title and id. Since this file was created on Windows, newline sequences are created as CRLF, in contrast to what our regex reads (LF).

I kindly ask you to commit this file with LF formatting, so our build can succeed:

image


We should probably improve the script here https://github.com/nextauthjs/next-auth/blob/main/www/scripts/generate-providers.js to not to trip up in the future.

balazsorban44 (changes requested):

www/docs/providers/zoho.md
jkwiz: Thanks for the quick review. I was wondering why the build failed 🙂

I've committed this change and it appears to be passing.

balazsorban44 (commented):
Awesome, thanks!

balazsorban44 (approved):

PR #1523, opened 2021-03-15
[WIP] add email as query param to verify request

sualko:
**What**: It would be nice to show the user which email address he has been entered.

**Why**: Sometimes users are mistyping there email address and are complaining that they are not getting any sign in email. This could help to mitigate the issue.

**How**: We add the email as query string.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      

I will add documentation and tests to the pr as soon as I get feedback.


balazsorban44:
For this usecase, check out the `redirect: false` option of `signIn`

PR #1544, opened 2021-03-18
fix: reject access token if slack login flow was canceled

### LINKED ISSUE #1513: Crash when user cancels login flow for Slack
**Describe the bug**
If the user cancels (does not accept) the permissions that's asked for on the Slack login page the app crashes.

**Steps to reproduce**
- Use the Slack provider with only `clientId` and `clientSecret` set.
- Perfrom login using `signIn()` or `signIn('slack')`.
- When on the Slack permissions page, press Cancel.

**Expected behavior**
Get redirected back with error in login flow. I.e. `/api/auth/signin?error=Callback`

**Screenshots or error logs**
```
/.../node_modules/next-auth/dist/server/lib/oauth/client.js:191
        var accessToken = provider.id === 'slack' ? raw.authed_user.access_token : raw.access_token;
                                                                    ^

TypeError: Cannot read property 'access_token' of undefined
    at /.../node_modules/next-auth/dist/server/lib/oauth/client.js:191:69
    at passBackControl (/.../node_modules/oauth/lib/oauth2.js:134:9)
    at IncomingMessage.<anonymous> (/.../node_modules/oauth/lib/oauth2.js:157:7)
    at IncomingMessage.emit (events.js:327:22)
    at endReadableNT (internal/streams/readable.js:1327:12)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
```

**Additional context**
The body from Slack is formatted as the following object:
```js
{
  ok: false,
  error: 'invalid_code'
}
```
Slack Application is a fresh created app with no special OAuth scopes requested and distributed. 

**Feedback**

mrmurb:
**What**:
Access token parsing when the provider is Slack.

**Why**:
If the user canceled the login flow when on Slack's consent page the application would crash when trying to read access token from response body.  
Closes #1513.

**How**:
Slack responds with an `ok: boolean` property in the body object. If it's false we reject the auth flow. Else continue as before.

**Checklist**:

- [ ] Documentation - N/A
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/src/server/lib/oauth/client.js
+++ b/src/server/lib/oauth/client.js
@@ -167,9 +167,18 @@ async function getOAuth2AccessToken (code, provider, codeVerifier) {
           raw = querystring.parse(data)
         }
 
-        const accessToken = provider.id === 'slack'
-          ? raw.authed_user.access_token
-          : raw.access_token
+        let accessToken;
+        if (provider.id === 'slack') {
+          const { ok, error } = raw;
+          if (!ok) {
+            return reject(error);
+          }
+
+          accessToken = raw.authed_user.access_token;
+        }
+        else {
+          accessToken = raw.access_token;
+        }
 
         resolve({
           accessToken,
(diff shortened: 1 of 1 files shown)


balazsorban44 (approved):
Thank you so much! Looks good to me! 🙂

PR #1558, opened 2021-03-21
Added tutorial on testing pages with authentication by setting next-a…

joostvanwollingen:
**What**:
This PR adds a tutorial page about testing pages with authentication by setting a session-token cookie.

**Why**:
The current docs/tutorials only mention testing with https://www.npmjs.com/package/cypress-social-logins, which doesn't support all authentication providers supported by NextAuth.js yet, and, in my experience, makes test scenario's unnecessarily slow/long. I'm not sure if this is the best way to go about it, but in my workflow the approach of seeding a session-token to the DB and setting that in Cypress before visiting any page worked faster and circumvented setting up credentials with an authentication provider. For tests that are not focused on the authentication flow itself, this allows for writing focused functional tests.

If this doesn't belong here, or is not in-line with the vision of the project, feel free to close the PR, just figured I'd share my approach.

**Checklist**:

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/www/docs/tutorials/testing-pages-with-authentication.md
+++ b/www/docs/tutorials/testing-pages-with-authentication.md
@@ -1,0 +1,39 @@
+---
+id: testing-pages-with-authentication
+title: Testing pages with authentication
+---
+
+NOTE: This approach requires that you are using [a database](/configuration/databases.md) and are not using JWT.
+
+You may want to functionally test pages of your application that require authentication. In order to circumvent a login with a real authentication provider, you can create a session in the database and set a cookie inside your browser to use that session.
+
+## Create a session 
+Add a record to the `Sessions`-table, and make sure that the userId exists. A good place to do this is in your [database seed script](https://www.prisma.io/docs/guides/application-lifecycle/seed-database).
+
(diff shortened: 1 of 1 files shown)


balazsorban44:
@iaincollins is working on the built-in tests, so I assume he has some thoughts around this.

djejaquino:
I am very interested in the ability to set the session token using from my IDP's  `/connect/token` response `access_token`. Is that doable already? Because I tried that and NextAuth just clears it during load.

timohermans:
> I am very interested in the ability to set the session token using from my IDP's `/connect/token` response `access_token`. Is that doable already? Because I tried that and NextAuth just clears it during load.

The reason for this is that you must set a authorization flow token. You cannot, say, put a token you receive from a grant_type password token.

I'm actually curious how to get a token programmatically that enables testing. @joostvanwollingen do you have an example for this?

balazsorban44:
Why can't you just mock out the `/api/auth/session` request with something like [`Mock Service Worker`](https://mswjs.io/)? That is the endpoint that functions like `useSession` and `getSession` contacting to get a valid session from the backend.

Issue #1530, opened 2021-03-16
PKCE flow does not work


viniciuscr:
**Describe the bug**
Using provider with specify PKCE auth flow does not work.
A provider such Okta expects a PKCE flow to work, this flow expect the follow params:

> client_id=0oapu4btsL2xI0y8y356
> &redirect_uri=http://localhost:3000/api/auth/callback/okta
> &response_type=id_token token
> &response_mode=fragment
> &state=SU8nskju26XowSCg3bx2LeZq7MwKcwnQ7h6vQY8twd9QJECHRKs14OwXPdpNBI58
> &nonce=Ypo4cVlv0spQN2KTFo3W4cgMIDn6sLcZpInyC40U5ff3iqwUGLpee7D4XcVGCVco
> &scope=openid profile email

Using the `protection: "pkce"` config, does not seems to work, it also removes the param `state`.
I tried to create a custom provider, but I could not find any logic where the nonce param is set.

**Steps to reproduce**
 I've used the Okta provider and tried with and without the `protection: "pkce"`.

**Expected behavior**
protection: "pkce" should have the follow changes:
- keep the `state` param
- add `nonce` param
- `response_type` changed to **id_token token code**
- add `response_mode` = **fragment**

**Screenshots or error logs**
Error provided on next console does not match the errors returned by the okta api.
The real erros from the api would be complaining  about missing params.

**Feedback**
there is a great doc about PKCE with okta here: 
https://developer.okta.com/blog/2019/08/22/okta-authjs-pkce

* [ ] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [x] Could not find relevant documentation
* [ ] Found the example project helpful
* [x] Did not find the example project helpful

balazsorban44:
please provide a full reproduction. we are using PKCE at work (not with Okta) without any problems

viniciuscr:
without given my credentials it's impossible to you to reproduce, unless you have a okta account.
Same problems was reported here, 9days ago:
https://github.com/nextauthjs/next-auth/issues/685#issuecomment-792331149

balazsorban44:
Some kind of information must be given, or we are as helpless as you are. As mentioned, I cannot see this issue occurring on our end, so there must be something special with your usecase/config. Providing more information is the key to resolve the problem. We could start by looking at your `[...nextauth].js` file and how you initiate the login flow from your code. Node, next-auth, Next.js versions, what kind of OS/Browser it is, etc.

[...]

Issue #1367, opened 2021-02-24
Error The code_verifier cookie was not found when trying to use with PKCE


vtrphan:
**Describe the bug**

I got an error : The code_verifier cookie was not found when trying to use NextAuth with protection option set to 'pkce' and identity provider is Azure B2C. I check the cookie but it was there under the name next-auth.pkce.code_verifier. Why does it say code_verifier_cookie not found ?

**Steps to reproduce**
Steps to reproduce the behavior.

Include a link to public repository which can be used to reproduce the behaviour.

**Expected behavior**
No error happens

**Screenshots or error logs**
If applicable add screenshots or error logs to help explain the problem.

**Additional context**

**Feedback**
*Documentation refers to searching through [online documentation](https://next-auth.js.org), code comments and issue history. The example project refers to [next-auth-example](https://github.com/iaincollins/next-auth-example).*

* [ ] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [ ] Could not find relevant documentation
* [ ] Found the example project helpful
* [ ] Did not find the example project helpful

balazsorban44:
Could you please provide a reproduction/some code?

Also what browser/device do you use? We have seen similar problems on iOS devices/Safari

vtrphan:
a full reproduction i cannot post, but i use a Mac with Chrome, and use Azure B2C with the option "protection" set to 'pkce', then comes the error. Using "protection" set to "state" also not works, i got error that the State value is invalid. So i have to set "protection" to "none".

vtrphan:
i was trying to use it with Azure B2C (not Azure AD) , and with "protection" set to 'pkce' i got that error. I don't know from where it comes

balazsorban44:
could you confirm if the problem persists on a non-Apple device?

vtrphan:
i don't have any windows machine so i can't really test it. But you think this is related to Apple devices?

balazsorban44:
https://github.com/nextauthjs/next-auth/issues/952 https://github.com/nextauthjs/next-auth/issues/656 suggest that

viniciuscr:
> a full reproduction i cannot post, but i use a Mac with Chrome, and use Azure B2C with the option "protection" set to 'pkce', then comes the error. Using "protection" set to "state" also not works, i got error that the State value is invalid. So i have to set "protection" to "none".

@balazsorban44 it seems to be the same problem that I'm facing with Okta, setting the `protection: pkce` wont send the state param.

[...]

PR #1565, opened 2021-03-22
fix: update protection to accept array with multiples options

viniciuscr:
**What**:
Add new protection value: "both", to use both handlers (state and pkce).

**Why**:
Some provider require the state param when using the pkce protection as reported on this issues:
https://github.com/nextauthjs/next-auth/issues/1530
https://github.com/nextauthjs/next-auth/issues/1367

**How**:
Just editing the handlers validations to handle in case of "both" is set on protection.

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged
      
It's note ready, actually its more a suggestion than a PR.

Note: I called it both, but it could have other names, such as: pkce-state or pkce-with-state

### DIFF
--- a/src/server/lib/oauth/client.js
+++ b/src/server/lib/oauth/client.js
@@ -167,9 +167,17 @@ async function getOAuth2AccessToken (code, provider, codeVerifier) {
           raw = querystring.parse(data)
         }
 
-        const accessToken = provider.id === 'slack'
-          ? raw.authed_user.access_token
-          : raw.access_token
+        let accessToken
+        if (provider.id === 'slack') {
+          const { ok, error } = raw
+          if (!ok) {
+            return reject(error)
+          }
+
+          accessToken = raw.authed_user.access_token
+        } else {
+          accessToken = raw.access_token
+        }
 
         resolve({
           accessToken,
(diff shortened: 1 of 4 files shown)


balazsorban44:
I have been thinking about this, and I am curious what do you think if we implement it such that protection could take an array as well? so `protection: ["pkce", "state"]` so it's room for the future.

In any way, have you tried/verified that this resolves your Okta problem?

viniciuscr:
Any update here?
I see more people are getting stuck with the same problem and I kind of need it to use next-auth, otherwise I will have to jump to the okta sdk.

balazsorban44:
Sorry if I haven't been explicit about it, but if you could change the implementation to be able to provide an array instead of `"both"` for `protection` to have more room for the future, I could proceed to merge this.

viniciuscr:
> Sorry if I haven't been explicit about it, but if you could change the implementation to be able to provide an array instead of `"both"` for `protection` to have more room for the future, I could proceed to merge this.

No worries, it was an easy change.
Now it accepts an array but also keeps compatibility with string.

LuisValgoi:
Looks good to me

LuisValgoi (approved):

viniciuscr:
@iaincollins @balazsorban44 
any advance to merge this PR?

balazsorban44 (approved):
Thanks for your patience! I finally had the chance to check this out locally, and it seems to work nicely! Tested with Auth0.

PR #1570, opened 2021-03-23
fix(provider): added options in instagram provider

shubhamnshukla:
**What**: No clientId in instagram url.

**Why**: In the Instagram provider, the URL was not taking clientId as we are not adding passed options to the object.

**How**: By spreading the options object

**Checklist**: 

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/providers/instagram.js
+++ b/src/providers/instagram.js
@@ -45,6 +45,7 @@ export default function Instagram (options) {
         email: null,
         image: null
       }
-    }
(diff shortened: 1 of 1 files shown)


balazsorban44 (approved):
🤦‍♂️ I think this was my mistake. Thanks for catching and providing a fix!

PR #1576, opened 2021-03-25
Add orcid provider

rezze:
**What**:

Orcid OAuth2 provider

**Why**:

Orcid is a commonly used identity service for researchers.

**How**:

Added orcid provider

**Checklist**:

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged


src/server/lib/oauth/client.js
```diff
@@ -225,6 +225,10 @@ async function getOAuth2 (provider, accessToken, results) {
     url = prepareProfileUrl({ provider, url, results })
   }
 
+  if (provider.id === 'orcid') {
```
lluia, line 228: I think is ok to pass this case, however, this `if ()` branching won't scale well as we add more of them.

In case you have time, the best would be to refactor below `prepareProfileUrl` to be **provider** agnostic ( _right now only works for the Bungie case_ ) 💭
```js
const url = prepareProfileUrl({ url, results, providerId: provider.id  })
// ...
function prepareProfileUrl({ results, url, providerId }) {
  const providersKeyMap = {
    orcid: 'orcid',
    bungie: 'membership_id'
  }
  
  if (!Object.keys(providersKeyMap).includes(providerId)) return url
  
  const keyToReplace = providersKeyMap[providerId]

  if (!Object.keys(results).includes(keyToReplace)) {
    throw new Error(`[next-auth]: Expected ${keyToReplace} to be passed for ${providerId} provider`)
  }

  return url.replace(`{${keyToReplace}}', results[keyToReplace])
}

0ubbe (commented):
Thanks a lot 💚 , few small suggestions 💭

src/server/lib/oauth/client.js
balazsorban44: #1846

balazsorban44:
Haven't been activity in 2 months here, so I am closing it. If the OP would like to pick this up again, I am happy to re-open!

Issue #1586, opened 2021-03-25
Provider "Azure Active Directory B2C" does not implement Azure AD B2C


BenjaminWFox:
**Describe the bug**
This is really just a naming issue. The provider called "Azure Active Directory B2C" implements authentication against Azure AD which is not the same thing as Azure AD B2C.

Azure AD B2C is [a specific product](https://docs.microsoft.com/en-us/azure/active-directory-b2c/overview) & subset of Azure AD, and comes with additional setup and requirements, including using base URLs and specifying User Flows that are different than used and supported in this provider.

This is extra confusing (blaming MSFT here 😊) because creating an Azure AD B2C tenant ([as in the link in the NextAuth docs](https://docs.microsoft.com/en-us/azure/active-directory-b2c/tutorial-create-tenant)) creates both an Azure AD instance and allows the creation of App Registrations, so while the intention is to use the Azure AD B2C functionality you could _**also**_ just use the tenant that's been created for basic Azure AD authentication, which is what's happening with this provider.

**Steps to reproduce**
I can validate this by going into an Azure AD instance not associated with a B2C tenant and creating an app registration, which I am then able to successfully authenticate against.

**Additional context**
I'd propose just renaming the provider and adding a different link for the 'Configuration' example, although I guess this would be a breaking change if people have already implemented the provider as-is? That said, for someone who things they've implemented Azure AD B2C there will likely be a lot of confusion as to why certain B2C features aren't working as expected (including User Flows and the B2C Logout url).

I can submit a PR that either renames the current provider or creates a new provider for 'Azure AD' separately. At some point, though, the 'Azure AD B2C' provider should either be removed or updated to work with B2C functionality.

balazsorban44:
https://github.com/nextauthjs/next-auth/discussions/1326#discussioncomment-391403

BenjaminWFox:
@balazsorban44 submitted https://github.com/nextauthjs/next-auth/pull/1591

PR #1591, opened 2021-03-26
Apply Azure AD, B2C changes on next to fix provider semantics

BenjaminWFox:
**What**: 
1. Created new provider `azure-ad` containing what was previously `azure-ad-b2c`

2. Updated content for provider `azure-ad-b2c` to reflect correct configuration for that provider

**Why**:

https://github.com/nextauthjs/next-auth/issues/1586

**How**:

Changes to Provider JS & MD files, and updating the provider index.

**Checklist**:

- [x] Documentation
- [ ] Tests (May be N/A)
- [ ] Ready to be merged
  - Should be reviewed for correctness
  - Do I need to do something specific to generate types?
  - Doesn't pass deployment, unsure if that's expected at this point on `next`
      

# * This will probably be a breaking change *
Anyone who previously has implemented what is currently provider `azure-ad-b2c`/`Providers.AzureADB2C` will need to update their provider to be `azure-ad`/`Providers.AzureAD`. There should be no other changes needed.


balazsorban44:
Not sure why the Vercel deployment fails, I switched to your branch locally, and the build was successful. 🤷

src/providers/index.js
```diff
@@ -0,0 +1,81 @@
...
+  Reddit,
+  Salesforce,
+  Slack,
+  Spotify,
+  Strava,
+  Twitch,
+  Twitter,
+  VK,
+  Yandex
+}
```
BenjaminWFox, line 81: @balazsorban44 Fixing merge conflicts I saw this file was removed on the `next` branch but it is still referenced in the dev app config file [https://github.com/nextauthjs/next-auth/blob/next/app/pages/api/auth/%5B...nextauth%5D.js](https://github.com/nextauthjs/next-auth/blob/next/app/pages/api/auth/%5B...nextauth%5D.js)

Is the removal of the providers `index.js` intentional?

BenjaminWFox:
@balazsorban44 also is this Types lint error a known issue? Unsure if I should do something in my code to fix that or if it will be fixed elsewhere on `next`.

```
Error: .../next-auth/types/adapters.d.ts:2:49
ERROR: 2:49  expect  TypeScript@4.3 compile error:
Cannot find module 'typeorm' or its corresponding type declarations.

.../next-auth/types/index.d.ts:5:35
ERROR: 5:35  expect  TypeScript@4.3 compile error:
Cannot find module 'typeorm' or its corresponding type declarations.
```

src/providers/index.js
balazsorban44: Yes, it is autogenerated now! See #1782

balazsorban44:
Yes, I guess it is "expected"... I am experimenting with cutting the `typeorm` package out of the core (making it optional to those actually use it because takes up around 70% of the bundle...), but we still reference stuff in types from it. I'll have a look next time I am working on the next branch!

src/providers/azure-ad-b2c.js
```diff
@@ -1,5 +1,7 @@
-export default function AzureADB2C(options) {
-  const tenant = options.tenantId ? options.tenantId : "common"
+export default (options) => {
```
balazsorban44, line 1: Could you make this a named default export function again? Goes together with auto-generating the index.js file

src/providers/azure-ad.js
```diff
@@ -0,0 +1,24 @@
+export default (options) => {
```
balazsorban44, line 1: Same as above, must be a named default export now :slightly_smiling_face:

src/providers/azure-ad.js
```diff
@@ -0,0 +1,24 @@
...
+  const tenant = options.tenantId ? options.tenantId : 'common'
```
balazsorban44, line 2: How about:
```suggestion
  const tenant = options.tenantId ?? 'common'
```

balazsorban44 (changes requested):

src/providers/azure-ad.js
BenjaminWFox: I had no idea that 'suggested change' functionality existed 🤯

balazsorban44: you can even batch my suggestions into a single commit and apply them 😉

https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request#applying-suggested-changes

balazsorban44 (approved):

balazsorban44:
NOTE to other maintainers: When merging, we should to point out the breaking changes here, following the https://www.conventionalcommits.org/en/v1.0.0/ `BREAKING CHANGE` format for the commit message.

www/docs/providers/azure-ad-b2c.md
```diff
@@ -5,11 +5,7 @@ title: Azure Active Directory B2C
 
 ## Documentation
 
-https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow
-
-## Configuration
-
-https://docs.microsoft.com/en-us/azure/active-directory-b2c/tutorial-create-tenant
+<!-- https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow -->
```
balazsorban44, line 8: Why was this commented out? Should we remove the whole `Documentation` section if it is empty or have a better link?

balazsorban44 (changes requested):
Hi, @BenjaminWFox! Sorry for being slow on this, but I would like to pick this up now and get this merged! I had a simple comment, could you address that? If everything is fine, then I'll merge this ASAP, to get things ready for `next-auth@4`... :eyes:

PR #1644, opened 2021-04-04
Convert sub to int for getUser()

sir-dunxalot:
**What**: Make sure an int is passed to `getUser()` when checking if the user is logged in the server callback handler.

**Why**: I cannot link accounts to a single user without this code change... The adapters' `getUser()` requires a number as the sole param. Per the [JWT spec (sub section)](https://tools.ietf.org/html/rfc7519#section-4.1.2), `sub` is be a string. Thus, if you're calling `getUser(sub)`, you're passing a string to `getUser()`, which will throw an error in the adapter (e.g. prisma adapter).

Notably, [the `catch` block for this code](https://github.com/nextauthjs/next-auth/blob/main/src/server/lib/callback-handler.js#L67) does not throw an error, which made this issue hard to track down.

**How**: Calling `getUser(parseInt(sub))` instead of `getUser(sub)`.

**Checklist**:

- [N/A] Documentation
- [ ] Tests
- [ ] Ready to be merged
      

I am submitting this PR with the following questions, which will help us improve the PR before its merged (or an alternative fix is approved):

- Do all adapters require the param passed to `getUser()` to be an int? I have only used the Prisma adapter though, upon reading the source for other adapters, believe all adapters `getUser()` methods require a param passed as an int, not string, to function properly.
- Why is an error not thrown in [the catch block](https://github.com/nextauthjs/next-auth/blob/main/src/server/lib/callback-handler.js#L67)?
- I presume a test for linking OAuth accounts to a single user would help catch this issue. However, I do not see any tests for linking multiple OAuth accounts to a single user. Is this correct? If so, adding those tests seems valuable but outside the scope of this PR. Do you agree?

### DIFF
--- a/src/server/lib/callback-handler.js
+++ b/src/server/lib/callback-handler.js
@@ -60,7 +60,7 @@ export default async function callbackHandler (sessionToken, profile, providerAc
       try {
         session = await jwt.decode({ ...jwt, token: sessionToken })
         if (session?.sub) {
-          user = await getUser(session.sub)
+          user = await getUser(parseInt(session.sub))
           isSignedIn = !!user
         }
       } catch {


jcdogo:
This change would definitely break the MongoDB adapter since MongoDB does not use int id's. If a string value is causing an error to be thrown with the prisma adapter and all prisma id's are supposed to be int values, then the `parseInt` change should go into the prisma adapter code and not here, where it affects all adapaters.

balazsorban44:
Have you checked the https://github.com/nextauthjs/adapters/tree/canary/packages/prisma? It seems to use ints as an id. (available through `@next-auth/prisma-adapter`)

cc: @williamluke4 @ndom91 could you help here?

sir-dunxalot:
Hi @balazsorban44. Yes, I use `@next-auth/prisma-adapter`. That adapter's `getUser()` method ([source here](https://github.com/nextauthjs/adapters/blob/canary/packages/prisma/src/index.ts#L106)) requires user id to be passed as a number. So I believe the issues described in this PR should be fixed in order to be compatible to the new Prisma adapter.

Per @jcdogo 's comment, maybe a change should be made to the adapter instead of this repo. Alternatively, the core problem may be that some adapter's `getUser()` methods require strings but others require ints.

For example, from what I can tell reading the adapter source code:

- The [new Prisma adapter requires user ID as an int](https://github.com/nextauthjs/adapters/blob/canary/packages/prisma/src/index.ts#L106)
- [The TypeORM adapter requires user ID as int](https://github.com/nextauthjs/adapters/blob/canary/packages/typeorm/models/user.js#L20)
- [The Fauna adapter requires user ID as string](https://github.com/nextauthjs/adapters/blob/canary/packages/fauna/src/index.js#L67) ([further reading here](https://docs.fauna.com/fauna/current/api/fql/functions/ref?lang=javascript))

jcdogo:
@sir-dunxalot TypeORM doesn't always use int id's, but rather it depends on the underlying database being used with TypeORM. As I mentioned earlier TypeORM+MongoDB would use string id's.

Since `session.sub` is always a string, and we cannot guarantee that id's are always int's, then the fix should be made in the prisma adapter.  

Change https://github.com/nextauthjs/adapters/blob/canary/packages/prisma/src/index.ts
```typescript
    async function getUser(id: number) {
      debug("GET_USER", id);
      try {
        const cachedUser = userCache.get(id);
```
to
```typescript
    async function getUser(id: number | string) {
      debug("GET_USER", id);
      try {
        const cachedUser = userCache.get(Number(id));
```

Similar changes would also have to be made to `deleteUser`, `linkAccount`, and `unlinkAccount` since they all take in the user id as a number.

sir-dunxalot:
I've opened nextauthjs/adapters#46 to supersede this PR.

FWIW, I still worry that having an inconsistent API between adapters is prone to issues like the one I identified in this PR. For example, the new Prisma adapter accepts `userId` as string but according to @jcdogo, using the TypeORM adapter for MongoDB requires `userId` as a number. If the adapters are an abstraction layer between NextAuth and a database, then shouldn't the adapters have a consistent API? For example, shouldn't the `getUser` method of every adapter have the same arg types and return type?

balazsorban44:
@sir-dunxalot I don't disagree, the public API should be as similar as possible. As I haven't used databases yet, I cannot tell if it is fully possible, or not.

This has been an issue with the OAuth providers in this repo as well. I actually advocated against one-off properties for single provider configuration before and tried to remove as many as I could (#1022).

In any case, I am happy if we move this conversation to the adapters repo, as ultimately, all adapters will live there (including the built-in TypeORM one.)

sir-dunxalot:
@balazsorban44 Sounds good. Thank you for the hard work you and the team put into this library!

Issue #1649, opened 2021-04-04
Move types from DefinitivelyTyped to here 🙇🏽‍♂️


lluia:
## Summary 💭

We've been maintaining the types for this library on [DefinitivelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/af3a9697e41e3594f012525909beef1812588389/types/next-auth). It's been working ok so far as there was no interest in moving the source code to TS to get the types automatically generated. 

However, it's challenging 🧯 to keep the types up to date, given they live in a separate repository and rely on updates from contributors. After speaking with @balazsorban44, we think we can spike to move the types to a folder here (`types/`), get them linted through [`dtslint`](https://github.com/Microsoft/dtslint) and integrated with the build pipeline 🎊

## Tasks ✅

- [x] Move types to this repo, to a folder called `types` in the root _(suggestions welcome)_
- [x] Add `dtslint` and add a command to lint the types
  - [x] add it to the CI
- [x] Modify `package.json` to declare types for the package
- [x] Modify build script to get types into `/dist` _(suggestions welcome)_
- [x] Make a **beta release** and verify everything works correctly
- [x] Update the `README` and website to explain types are now part of the `next-auth` package
- [x] Test whether the new types take precedence over `@types/next-auth`
- [x] fix typos
- [x] Sort-out module augmentation issues
- [ ] Deprecate `@types/next-auth`
- [ ] Release new types on `next-auth`

## Additional context 📓

@JuanM04 @euxn23, you've been actively maintaining the types in **DefitnivelyTyped**. Would you be interested in joining/supporting this effort? (any other contributors welcome 💚)

( cc. @iaincollins )

- Current type definitions on [DefinitivelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/af3a9697e41e3594f012525909beef1812588389/types/next-auth) _(what you get when installing `@types/next-auth`)_
- [`dtslint`](https://github.com/Microsoft/dtslint)

lluia:
Relates to #1650

JuanM04:
I'll be glad to help. Should we make PRs to `main` or to another branch?

balazsorban44:
@JuanM04, awesome! I think we should create a `beta` branch.

JuanM04:
I was looking at the repo and there are lots of `.d.ts` files around. I was wondering, migrating all the repo to TypeScript isn't an option, right?

balazsorban44:
[...]

PR #1666, opened 2021-04-07
feat(ts): Export types related to email and credential providers

jgeurts:
**What**: Export types for Email and Credential providers

**Why**: Simplify custom adapter code

**How**: Added `export` keyword before types/interfaces in providers.d.ts

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged
      

Related: #1649

### DIFF
--- a/types/providers.d.ts
+++ b/types/providers.d.ts
@@ -88,7 +88,7 @@ type Email = (
   options: ProviderEmailOptions
 ) => NonNullParams<ProviderEmailOptions> & { id: "email"; type: "email" }
 
-interface VerificationRequestParams extends Provider {
+export interface VerificationRequestParams extends Provider {
   identifier: string
   url: string
   baseUrl: string
@@ -96,7 +96,7 @@ interface VerificationRequestParams extends Provider {
   provider: ProviderEmailOptions
 }
 
-interface ProviderEmailOptions {
+export interface ProviderEmailOptions {
   name?: string
   server?: string | ProviderEmailServer
   from?: string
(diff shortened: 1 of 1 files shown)


0ubbe (commented):
Thanks for the PR @jgeurts 💚 . To have a bit more context, could you explain what are the problems you're trying to address when exporting these types ( _as maybe there's no need_ )?

Bear in mind that the actual types still live at [**DefinitivelyTyped**](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/next-auth) and the types here are part of a [migration that is WIP](https://github.com/nextauthjs/next-auth/issues/1649) (_so even if we merge these updates, you won't be able to benefit from them just yet_ 😌)

jgeurts:
I have a [custom adapter](https://github.com/minarets/website/blob/main/minarets-api/auth/index.ts) that defines functions for verification requests. Currently, I duplicate the type definitions, but it would be handy to import them from next-auth. I’ll add a PR to DefinitelyTyped in the coming week, but figured I’d start here in case this goes through before I make the other pr.  Hope that helps!

0ubbe (approved):
Sounds good 😊 , I think we can merge here and no need to raise the PR on **DefinitivelyTyped**.

balazsorban44:
Would wait to merge until we reviewed #1690

balazsorban44:
As far as I can tell, all these will be redundant/are addressed in #1690, so closing. Tell me if something is missing still!

Issue #497, opened 2020-07-28
How to setup NEXTAUTH_URL for preview deployments?


ghoshnirmalya:
Vercel has the option to preview deployments of your pull requests before merging them. As a result, setting `NEXTAUTH_URL` would be tricky as the deployment url of a preview pull request might be different from the production url.

**Your question**
Is it possible to set the value of `NEXTAUTH_URL` dynamically? Also, why is this environment variable necessary to be set?

* [ ] Found the documentation helpful
* [x] Found documentation but was incomplete
* [ ] Could not find relevant documentation
* [ ] Found the example project helpful
* [ ] Did not find the example project helpful

iaincollins:
Great question!

There are multiple contributing reasons as to why this needs to be set - and why it is an environment variable in v3.

API routes and Server Side Rendered functions (e.g. `getServerSideProps` and `getInitialProps`) need to know the site URL as they make a request back to the server - passing though the session token in the request object - to determine if a user is signed in.

This allows API routes and Server Side Rendered Pages to be lightweight and not have to pull in all the database code and other logic, which stays in the authentication route. Of course, using JSON Web Tokens sessions and the `getToken` helper is even faster as you can just verify the token without having to make a network call.

In v2 the site name and base path were options on NextAuth.js configuration, and this was passed to API routes and Server Side Pages using a Secure server side only configuration cookie. This has some security challenges, even a secure, server only cookie and was problematic in practice.

It is especially problematic in development mode or when rendering Server Side Pages. On first hit of a new deploy (a particular problem on local development instances) the server rendered page would not have had access to the cookie yet and so wouldn't know how to contact the server. Even requesting the page was enough for it to see and grab the value from the cookie for later reuse (as even Serverless functions are memory resident, e.g. for ~2 hours on AWS Lambda) but it was still an edge case there was no other obvious solution for.

It's possible to grab the host name from the `req` object, but this isn't a great solution as it has security problems; anyone can spoof the `host` header and exploit that (e.g. in a DDoS attack against another host).

Another consideration is that the hostnames in callback URLs for most OAuth providers need to *exactly* match the specified domains with the OAuth provider. In the vast majority of cases you can't use a wildcard with an OAuth provider either; you have to explicitly configure each Full Qualified Domain Name (e.g. `example.com`, `www.example.com`) so attempting to auto-discover it it doesn't work for a lot of folks in practice (security issues aside).

[...]

PR #1672, opened 2021-04-08
Use secure cookie prefix in Vercel preview apps

cathykc:
**What**:

Use secure prefix for cookie name if application is a Vercel preview app.

**Why**:

For NextAuth to work in Vercel preview apps we need to leave `NEXTAUTH_URL` unset so we can fall back on `VERCEL_URL` - this changes for every preview deployment so we can't statically set it.

When `NEXTAUTH_URL` is unset, the previous logic would have `secureCookie = false` which results in an error when trying to decode the JWT in preview apps. This change fixes that.

**How**:

Updated the default value for `secureCookie` - no other changes.

**Checklist**:

- [ ] Documentation N/A
- [ ] Tests N/A
- [X] Ready to be merged


balazsorban44:
Couldn't you just call `getToken` like so?:

```js
getToken ({
  secureCookie: process.env.NEXTAUTH_URL?.startsWith('https://') ?? process.env.VERCEL_ENV === "preview"
  ... your own config here
})
```

I don't think we should optimize defaults for any services too much. (We already use VERCEL_URL as a fallback in some places, which honestly I don't really understand)

I might feel too strongly about this, but I think this is easy to fix in user-land and so I don't want to accept this PR. You could maybe open an issues/discussion, and if this gets traction, we could add it anyway, but I want to hear what other think.

cathykc:
That's exactly what I did to get past this in my project but it took a while of digging to figure out that `secureCookie` parameter needed to be customized for preview apps and it seems from reading through issues and documentation that `next-auth` does make it easy to use with Vercel out of the box. @iaincollins talks about defaults and preview apps in Vercel here https://github.com/nextauthjs/next-auth/issues/497. 

Because there are already a lot of `VERCEL_URL` fallbacks in this library it makes sense to me to also have this default to a Vercel deployment environment but I'm happy to create an issue and see if there's interest! :)

jonlow:
For anyone looking at this, I can confirm 

> Couldn't you just call `getToken` like so?:
> 
> ```js
> getToken ({
>   secureCookie: process.env.NEXTAUTH_URL?.startsWith('https://') ?? process.env.VERCEL_ENV === "preview"
>   ... your own config here
> })
> ```
> 
> I don't think we should optimize defaults for any services too much. (We already use VERCEL_URL as a fallback in some places, which honestly I don't really understand)
> 
> I might feel too strongly about this, but I think this is easy to fix in user-land and so I don't want to accept this PR. You could maybe open an issues/discussion, and if this gets traction, we could add it anyway, but I want to hear what others think.

I can confirm that I can use next-auth in vercel preview deploys using this  ☝️ and omitting  `VERCEL_URL` from previews

aakash14goplani:
@jonlow - If possible, can you provide a working example of how have you integrated `getToken` in your code flow.
Context - I am using OAuth provider with NextAuth and things are breaking in Vercel preview environment due to mismatch in callback urls.

PR #1681, opened 2021-04-09
feat(ts): support module augmentation

balazsorban44:
**What**:

Support Module Augmentation. Also adds a documentation section explaining how to do it.

**Why**:

This way, certain interfaces like `JWT` or `Session` can be defined in a single location and will be shared across the entire app.

For example [`useSession`](https://next-auth.js.org/getting-started/client#usesession) and the [`callbacks.session`](https://next-auth.js.org/configuration/callbacks#session-callback) will share the same type! 

**How**:

It only required some cleanup in the `CallbacksOptions` interface

**Checklist**:

- [x] Documentation
- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -7,7 +7,7 @@ import { Adapter } from "./adapters"
 import { JWTOptions, JWT } from "./jwt"
 import { AppProvider, Providers } from "./providers"
 import { NextApiRequest, NextApiResponse, NextApiHandler } from "./_next"
-import { NonNullParams, WithAdditionalParams } from "./_utils"
+import { Awaitable, NonNullParams, WithAdditionalParams } from "./_utils"
 
 export interface NextAuthOptions {
   providers: Providers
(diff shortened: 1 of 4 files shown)


types/index.d.ts
```diff
@@ -62,30 +62,34 @@ export interface AppOptions
...
+  refresh_token?: string
+  id_token?: string
+  id: string
+  provider: string
+  type: string
+}
+export interface Profile extends Record<string, unknown> {}
+
+export interface CallbacksOptions<
+  P extends Record<string, unknown> = Profile,
```
lluia, line 80: Given `Profile` is just `Record<string, unknown>`, maybe we can simplify this line to 💭
```ts
export interface CallbacksOptions<
  P = Record<string, unknown>,
  A extends Record<string, unknown> = Account
```

types/index.d.ts
```diff
@@ -62,30 +62,34 @@ export interface AppOptions
...
+  A extends Record<string, unknown> = Account
+> {
+  signIn?(user: User, account: A, profile: P): Awaitable<string | boolean>
+  redirect?(url: string, baseUrl: string): Awaitable<string>
+  session?(session: Session, userOrToken: JWT | User): Awaitable<Session>
```
lluia, line 85: This type will force the `session` callback to always take 2 parameters `session, userOrToken`. Are we coolio with that? I  &nbsp;👀&nbsp;  the default callback for the session [is just using one](https://github.com/nextauthjs/next-auth/blob/main/src/server/lib/default-callbacks.js#L46-L48), maybe we should change it to:
```ts
session?:(session: Session, userOrToken?: JWT | User): Awaitable<Session>

File: www/docs/getting-started/typescript.md [hunk 3]
```diff
@@ -1,22 +1,77 @@
 ---
 id: typescript
-title: TypeScript Support
+title: TypeScript
 ---
 
-Currently, NextAuth.js relies on the community to provide TypeScript types. You can download it from [DefinitelyTyped](https://www.npmjs.com/package/@types/next-auth).
+TypeScript comes with its own types, so you can safely use it in your TypeScript projects. Even if you don't use TypeScript, IDEs like VSCode will pick this up, to provide you with a better developer experience. While you are typing, you will get suggestions of what certain objects are, and sometimes also links to documentation, and examples.
```
lluia, line 6: Should we change:
```
TypeScript comes with its own types, so you can safely use it in your TypeScript projects.
```
to:
```
NextAuth comes with its own types, so you can safely use it in your TypeScript projects.
```

0ubbe (approved):
Very neat 💯, much easier to read the `CallbacksOptions` type now 💚

I leave 3 small nitpicks 💬 ; none of them prevents merging this 😊

types/index.d.ts
balazsorban44: Yes, so the defulat session callback actually just an identity function and does nothing than return the first param. In any other case, when the user defines/overrides this, they will use the second param to extend session in some way. So I think it's good. 🙂

www/docs/getting-started/typescript.md
balazsorban44: 🤦‍♂️

types/index.d.ts
balazsorban44: It is because it's kind of unfinished, as we may want to make it extensible as the other interfaces. We could either try to somehow infer it from providers, or add it as a generic to NextAuth (same goes for Account, they are not necessarily used anywhere else than NextAuth, so generic could work just fine. Or we could go all in on Module Augmentation. Haven't decided yet. 🤔) In any case, I think that could be a subsequent PR

PR #1684, opened 2021-04-11
fix .flat is not a function when running node version lower than 11

### LINKED ISSUE #1675: `v3.14.0` bug: TypeError: [provider.protection].flat is not a function
**Describe the bug**

`provider.protection` is not a required property, and not providing it breaks the package since `.flat()` is called without first verifying it is an array.

**Steps to reproduce**

Add a custom provider without specifying `protection` property

**Additional context**

Here are all the places this is done:
https://github.com/nextauthjs/next-auth/search?q=%5Bprovider.protection%5D.flat%28%29

fxamauri:
**What**:
Fix .flat is not a function when running node version lower than 11

**Why**:
This problem started in #1565 

**How**:
A check was made to check if the provider.protection is a string and if this is true, the provider.protection is rewritten to array and the string is placed inside and the handlers were  edited again adjusted

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      

Fixes #1675

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -79,6 +79,10 @@ async function NextAuthHandler (req, res, userOptions) {
       provider.protection = 'state' // Default to state, as we did in 3.1 REVIEW: should we use "pkce" or "none" as default?
     }
 
+    if (typeof provider.protection === 'string') {
+      provider.protection = [provider.protection]
+    }
+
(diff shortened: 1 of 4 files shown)


balazsorban44 (approved):
Thanks!

PR #1689, opened 2021-04-11
Update callbacks.md

AustinKirby:
This update fixes an issue with @types/next-auth. Since the Session type doesn't have an accessToken value you need to set it in the return statement. Since the return is WithAdditionalParams<Session> we can tack on accessToken there.

**What**:

This update fixes an issue with @types/next-auth. Since the Session type doesn't have an accessToken value you need to set it in the return statement. Since the return is WithAdditionalParams<Session> we can tack on accessToken there.

**Why**:

@types/next-auth Session type doesn't contain a accessToken value

**How**:

By submiting the token in the return statement you are utilizing WithAdditionalParams type

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/www/docs/configuration/callbacks.md
+++ b/www/docs/configuration/callbacks.md
@@ -181,8 +181,10 @@ callbacks: {
    */
   async session(session, token) {
     // Add property to session, like an access_token from a provider.
-    session.accessToken = token.accessToken
-    return session
+    return {
+      ...session,
+      accessToken: (token as JWT).accessToken
+    } as WithAdditionalParams<Session>
   }
 }
 ...


lluia:
Thanks for raising a PR 💚

I think the examples on the websites are currently just in plain **JS**, hence I don't think we should be using **TS** syntax on them.

There's [an epic going on](https://github.com/nextauthjs/next-auth/issues/1649) to move the types to this repo to make them easier to maintain. When we ship the new types 🚀 ( _hopefully around end of next week_ ) [they'll come with support for **module agumentation**](https://github.com/nextauthjs/next-auth/pull/1681) which I think will help in this specific issue.

balazsorban44:
Came here to say the same as @lluia. Going to close this. Look out for #1649

PR #1698, opened 2021-04-12
refactor: replace node-oauth with openid-client

### LINKED ISSUE #1048: Replace node-oauth dependency

### LINKED ISSUE #1305: Expose session_state to allow checking identity session using iframe
Add a way to pass `session_state` from identity provider response to jwt callback

`session_state` parameter is a required part to use in check_session_iframe scenario. By forwarding it to the jwt callback it can be saved to client session and used in the browser to find out if identity session has changed

Minimal required change would probably be passing `req.query.session_state` to the jwt callback, although I'm not quite sure in which object or parameter it should be put  

**Potential problems**
`session_state` is not useful for all scenarios and may not be supported by certain oidc providers 

Polling identity provider with `prompt=none`, but this causes extra network traffic
Polling introspection endpoint to find if `access_token` is still valid, but this requires using `access_token` and having access to introspection

**Additional context**
More info about [oidc session management](https://openid.net/specs/openid-connect-session-1_0.html)

I am willing to implement the feature, but I might require some additional guidance as to where to actually put the parameter

balazsorban44:
Continuation of #1105

We rely on a package that hasn't been updated in a very long time. We kind of depend on it even after this, as our new dependency doesn't support OAuth 1.0 anymore, but we will treat it as a legacy solution and will only provide minimal maintenance to it. There are a few bigger providers still using the old spec like Twitter, but new providers will hopefully rather rely on newer specs.

Our new choice `openid-client` is maintained properly by awesome people, and will help us with PKCE, state and other things that until now we had to do manually.

**Checklist**:

- [x] handle PKCE
- [x] handle State
~- [ ] decide on new provider API (#1846)~ will address in another PR
~- [ ] add migration to docs~ will address in another PR
- [ ] Tests
- [x] Ready to be merged
      

Related issues:
Closes #1048, closes #1305


src/server/lib/signin/oauth.js
```diff
@@ -1,53 +1,47 @@
...
-      oauth_token_secret: tokens.oauth_token_secret,
-      ...tokens.params,
-    })}`
-    logger.debug("GET_AUTHORIZATION_URL", url)
-    return url
-  } catch (error) {
-    logger.error("GET_AUTHORIZATION_URL_ERROR", error)
-    throw error
-  }
+  // TODO: authorizationParams vs params. What's the difference?
```
balazsorban44, line 39: Figure this out before merge

ndom91 (approved):
LGTM :+1:

PR #1724, opened 2021-04-14
style: replace ts-standard with eslint/prettier

ndom91:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged
      

Moved to _more_ standard tools, i.e. eslint/prettier, instead of ts-standard.


balazsorban44 (approved):

balazsorban44:
For anyone reading this, we are still using the rules suggested by https://standardjs.com/, but having it in an ESLint configuration makes it possible to combine it with Prettier, which takes away a lot of pain from manually having to format while writing code.

PR #1736, opened 2021-04-15
fix(build): use optional-require dependency

balazsorban44:
**What**:

Use `optional-require` to require certain modules optionally.

**Why**:

To my sadness, after trying out #1682 on a local project, I discovered that Next.js Webpack will try to include modules in the dev/prod builds that were added through dynamic imports, **even though they are wrapped in a conditional that is never `true`.*. That just seem to be a limitation/behaviour I do not understand fully, to be honest. If someone visits this PR, please enlighten me 💡, or show me to some documentation/reasoning! In the meantime, let's see that using a wrapper for importing helps.

I am trying `optional-require` instead of `require_optional` first, as the latter looks to be unmaintained (with still a number of downloads/week though), and even the `mondodb` module (that we used as an example when implementing optional modules) switched from it, as it introduced a bug in Windows.

**How**:

Switch out the failed dynamic imports to `optionalRequire` calls, and see if solves the problem.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14081,8 +14081,7 @@
     "optional-require": {
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/optional-require/-/optional-require-1.0.3.tgz",
-      "integrity": "sha512-RV2Zp2MY2aeYK5G+B/Sps8lW5NHAzE5QClbFP15j+PWmP+T9PxlJXBOOLoSAdgwFvS4t0aMR4vpedMkbHfh0nA==",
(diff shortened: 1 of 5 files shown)


balazsorban44:
DISCLAIMER: Did not seem to work as expected... I am going to try to find a better way that will make `typeorm` and `nodemailer` truly optional, so we can drastically reduce bundle size for those not using any of them.

Issue #736, opened 2020-10-02
How do you prevent new signups?


jesperordrup:
How do you prevent new signups?

Im creating a setup where you should not be able to create a new user. Im using email magic links provider.

**Feedback**

* [ ] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [x ] Could not find relevant documentation
* [x] Found the example project helpful
* [ ] Did not find the example project helpful

Xodarap:
> TIP
If you only want to allow users who already have accounts in the database to sign in, you can check for the existance of a user.id property and reject any sign in attempts from accounts that do not have one.

> If you are using NextAuth.js without database and want to control who can sign in, you can check their email address or profile against a hard coded list in the signIn() callback.

https://next-auth.js.org/configuration/callbacks

LoriKarikari:
@jesperordrup were you able to make it work?

jschuur:
Based on the update in #1737, that checking user.id is not reliable, it looks like you can look for the presence of any custom field that is in your local user schema, but not any of the providers you use. Just set a default for that field in your schema definition. I'm using the recommended [Prisma schema](https://next-auth.js.org/schemas/adapters#prisma-schema) as a basis, so I could look for createdAt, as long as that doesn't get returned by the provider I use.

If you wanted to avoid possible conflicts better, you can set a custom field in your user schema that is so unique, it's not likely to be used by any external provider.

Not ideal. It would be great if isNewUser() was supported by database logins. This has been brought up before (#558). but it looks like that issue became stale and was closed.

dan-kwiat:
My use case is different but worth noting here as it's related and @jschuur 's suggestion doesn't apply.

My app allows authentication (sign in + sign up) from certain providers only (`email` and `google`). Once a user is signed in however, they can authenticate with other providers (e.g. `strava`) to link those accounts to the existing user in the db.

There are two problems with the above approach:

1. When a signed in user tries to link an account for the first time, the `user` object does not contain the custom fields.
2. If an account is already linked, the `user` object will always have the custom fields, even if `signIn` is called by an un-authenticated user

For this use case I think we have to resort to getting the session:

```js
// [...nextauth].js

const getOptions = (req, res) => ({
  callbacks: {
    async signIn(user, account, profile) {
      if (account.provider === "strava") {
        // Prevent signin/signup, but allow linking
        const session = await getSession({ req })
        return session ? true : false
      }
      return true
    },
  },
  ...
})

[...]

PR #1737, opened 2021-04-15
Remove claim that new users do not have an ID

Xodarap:
I'm not sure when this changed, but it's no longer true. If the person logging in doesn't have a stored user account, the ID will be the provider_account_id

The answer to #736 is no longer accurate. I'm not sure if there's a different field people can look at to tell if the user has an account already.


jschuur:
I verified this today after I came across @Xodarap's original comment. I created a user via GitHub auth and in my signIn() callback, the user.id field was my GitHub user ID.

balazsorban44 (approved):

PR #1743, opened 2021-04-16
fix: use upgraded require optional

balazsorban44:
**What**:

Having another stab at #1682

**Why**:

The end goal is to not bundle/require `typeorm` and `nodemailer`, if someone doesn't use any of these in their project.

**How**:

I took [`require_optional`](https://github.com/christkv/require_optional) as a base, but published my own version under [`@balazsorban/require-optional`](https://www.npmjs.com/package/@balazsorban/require-optional). The difference is that it will use the now seemingly established fields `peerDependencies` and `peerDependenciesMeta`, instead of `peerOptionalDependencies`, which doesn't appear to be anywhere. In fact, googling it will point to one of this repository's issue with mongodb!

Side note, mongodb themselves decided to turn away from `peerOptionalDependencies`: https://jira.mongodb.org/browse/NODE-2867

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2259,6 +2259,30 @@
         }
       }
     },
+    "@balazsorban/require-optional": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/@balazsorban/require-optional/-/require-optional-1.0.0.tgz",
+      "integrity": "sha512-LD5glsU1iRvfIpv+1HGjvN7IhUjN4WM6Fe+hvSHOZPY3XsQ59117b5cvozA1CHzdod6XeYl3J54UICap9AWM1Q==",
+      "requires": {
+        "resolve-from": "^5.0.0",
+        "semver": "^7.3.5"
+      },
+      "dependencies": {
+        "resolve-from": {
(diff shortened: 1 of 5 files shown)


ndom91:
Yeah strange, right? I couldn't find `peerOptionalDependencies` referenced anywhere online really. Not sure where that originated from haha.

Anyway, is your `require-optional` fork doing the trick?

balazsorban44:
`peerOptionalDependencies` comes from the  `require_optional` package. It was written BEFORE `peerDependenciesMeta` was a thing. And no, unfortunately my solution did not work. Searching for `peerDependenciesMeta` up in the node_modules tree stopped looking for `nodemailer` in my `next`'s package.json.. 😵 I start to think that the best option is to drop the `database` config, asking the user to explicitly choose a adapter, and use the form `next-auth/providers/your-provider` for importing built-in providers. These two things would mean that we don't need any additional package for requiring optionally. Having imports like this would make it explicit and opt-in which packages you will have to use. I could use your help @ndom91 with testing.

PR #1756, opened 2021-04-18
feat(provider): add Dropbox

manishrc:
**What**:

- Add dropbox provider
- Add documentation for dropbox

**How**:

Add provider similar to other providers.  
Since dropbox does not support GET for Profile, conditionally used POST if provider is Dropbox. 

**Checklist**:

- [x] Documentation
- [ ] Tests N/A
- [x] Ready to be merged


src/server/lib/oauth/client.js
```diff
@@ -238,8 +239,15 @@ async function getOAuth2 (provider, accessToken, results) {
     url = prepareProfileUrl({ provider, url, results })
   }
 
+  /** Dropbox requires POST instead of GET
+   * Read more: https://www.dropbox.com/developers/reference/auth-types#user
+   */
+  if (provider.id === 'dropbox') {
+    httpMethod = 'POST'
+  }
```
balazsorban44, line 247: So I really push against provider-specific handling, unless it is absolutely necessary. At the same time, I try to keep the provider options surface minimal. When #1698 is done, we will be able to loan some of their API, and I can see they have something for this exact situation: https://github.com/panva/node-openid-client/blob/main/docs/README.md#clientuserinfoaccesstoken-options, where you can tell the library what method to use for the `userinfo` endpoint. I think we could go with something like this, so this change would also be useful for future providers that support POST in this case. What do you think?

src/providers/dropbox.js
```diff
@@ -0,0 +1,26 @@
+export default (options) => {
+  console.log('> Dropbox')
```
balazsorban44, line 2: You forgot a `console.log` :slightly_smiling_face:

Also, could you maybe use a named function with some JSDoc comments above? I have plans about extracting that information from these files, so we could auto-generate the documentation page for them!

See: https://github.com/nextauthjs/next-auth/blob/beta/src/providers/instagram.js

balazsorban44 (changes requested):
Just a few small comments, otherwise looking good! So I assume you have tested these all locally, right? Currently, we don't have any robust test setup, so we kind of rely on the ones implementing providers that they test it out before committing. 🙈

src/providers/index.js
```diff
@@ -9,6 +9,7 @@ import Bungie from './bungie'
 import Cognito from './cognito'
 import Credentials from './credentials'
 import Discord from './discord'
+import Dropbox from './dropbox'
```
balazsorban44, line 12: Could you revert this file change to get rid of the merge conflict? This file is now autogenerated!

michalbundyra:
Just to note here - we need to add the new provider also to types - https://github.com/nextauthjs/next-auth/blob/55a29329739614b1c8c0922a536b42a655bc1c6f/types/providers.d.ts#L56 (per #1773 
Thanks! :)

src/providers/dropbox.js
```diff
@@ -0,0 +1,26 @@
...
+    params: { grant_type: 'authorization_code' },
+    accessTokenUrl: 'https://api.dropboxapi.com/oauth2/token',
+    authorizationUrl:
+      'https://www.dropbox.com/oauth2/authorize?token_access_type=offline&response_type=code',
+    profileUrl: 'https://api.dropboxapi.com/2/users/get_current_account',
+    profile: (profile) => {
+      return {
+        id: profile.account_id,
+        name: profile.name.display_name,
+        email: profile.login,
```
adambrgmn, line 18: Thanks for this PR! It solves my immediate issues, hopefully it can be merged 👏 But I think this should be `profile.email`, at least from me testing this out locally.

```suggestion
        email: profile.email,
```

src/providers/dropbox.js
```diff
@@ -0,0 +1,26 @@
...
+        image: profile.profile_photo_url,
+        email_verified: profile.email_verified
+      }
+    },
+    protection: 'none',
```
balazsorban44, line 23:
```suggestion
    protection: ["state", "pkce"]
```

From the docs you linked, it looks like Dropbox supports both state and pkce:

image

I think we should add them by default

balazsorban44 (changes requested):

src/server/lib/oauth/client.js
manishrc: @balazsorban44 
I couldn't see any other way around this currently. But, agree that the `method` option in the `userinfo` would be perfect. I'd be happen to update once this is supported.

src/providers/dropbox.js
manishrc: Oh! Thanks you for catching this. Fixed.

manishrc:
I've all the suggested changes. But I'm now seeing the below error after merging main to this branch:

image

### Unhandled Runtime Error
TypeError: Cannot use 'in' operator to search for 'undefined' in null

```
next-auth/client/index.js (166:16) @ _callee7$

  164 | 
  165 |   // Redirect to sign in page if no valid provider specified
> 166 |   if (!(provider in providers)) {
      |                ^
  167 |     // If Provider not recognized, redirect to sign in page
  168 |     window.location = `${baseUrl}/signin?callbackUrl=${encodeURIComponent(callbackUrl)}`
  169 |     return

```

#### Call Stack
```
tryCatch
node_modules/regenerator-runtime/runtime.js (63:14)
Generator.invoke [as _invoke]
node_modules/regenerator-runtime/runtime.js (293:0)
Generator.eval [as next]
node_modules/regenerator-runtime/runtime.js (118:0)
asyncGeneratorStep
node_modules/@babel/runtime/helpers/esm/asyncToGenerator.js (3:0)
_next
node_modules/@babel/runtime/helpers/esm/asyncToGenerator.js (25:0)
```

src/providers/dropbox.js
```diff
@@ -0,0 +1,54 @@
...
+ * import { signIn } from "next-auth/client"
+ * ...
+ * <button onClick={() => signIn("dropbox")}>
+ *   Sign in
+ * </button>
+ * ...
+ * ```
+ * *Resources:*
+ * - [NextAuth.js Documentation](https://next-auth.js.org/providers/dropbox)
+ * - [Dropbox Documentation](ttps://developers.dropbox.com/oauth-guide)
```
adambrgmn, line 27: Minor typo 😊

```suggestion
 * - [Dropbox Documentation](https://developers.dropbox.com/oauth-guide)
```

balazsorban44 (approved):

PR #1758, opened 2021-04-18
feat: pass all providerAccount options in linkAccount cb

ndom91:
**What**: Passed all `providerAccount` tokens to the `linkAccount` cb

**Why**: https://github.com/nextauthjs/adapters/issues/48

**Checklist**:

- [ ] Documentation ?
- [ ] Tests 
- [ ] Ready to be merged No
      

@balazsorban44  @RazerMoon - so this PR passes the rest of the providerAccount object along in the linkAccount cb. However, they are not persisted in the Account table yet obviously as that would require breaking changes to the db model. I also didn't do the named parameter part yet as that would also be a pretty big breaking change. 

What do you guys think? Is the named parameter thing a breaking change we can live with in 4.x, for example? If so, I can do that here as well.


balazsorban44:
Just a small comment. If it's not breaking, it should have been opened against `main`. Also seeing that Prettier formatted the file. should we wait with such changes until `beta` is in `main`?

to sum it up, try to branch of from main, and re-open this PR against main for now. other than that, I think it is OK.

I really wish we could use named params for ALL our public facing apis where there is more than one parameter in a function.

ndom91:
Yeah i explicitly Branched from beta to use prettier, etc. 

Do you not except beta to merge back into main for a while?

balazsorban44:
yes I do. @lluia should look at my last PR regarding TS , soon. But still, I wanted us to use beta for TS migration because it seemed a bigger task than that can fit into a single PR. simple, non-brraking features/fixes can be done against main directly.

ndom91:
@balazsorban44 okay gotcha, makes sense. Should i rebased that PR against `main`?

balazsorban44:
I'm just not sure how rebase would work with `semantic-release`. Since it is such a small change, maybe just create a new branch from `main` and re-implement your small change in a new PR.

balazsorban44 (commented):
Yes, as I expected, simply changing the base from `beta` to `main` caused a lot of conflicts. Please just open a new PR with your changes and create the branch from `main`! Sorry. :see_no_evil:

PR #1760, opened 2021-04-18
chore(docs): update contributing.md

ndom91:
Regarding ESLint / Prettier use and link to their VSCode extensions

**What**: Updated `CONTRIBUTING.md` with `ESLint`/`Prettier` notes and links to their VSCode extensions, in place of the existing `ts-standard` info.

**Why**:

**How**:

**Checklist**:

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ Anyone can be a contributor. Either you found a typo, or you have an awesome fea
 * The latest changes are always in `main`, so please make your Pull Request against that branch.
 * Pull Requests should be raised for any change
 * Pull Requests need approval of a [core contributor](https://next-auth.js.org/contributors#core-team) before merging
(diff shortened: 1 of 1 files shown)


balazsorban44 (approved):
This looks good! I think it is fine for now, but maybe we could set this up with a commit hook in the future, so the users don't have to think about it:

https://prettier.io/docs/en/precommit.html

ndom91:
@balazsorban44 yeah I use husky and lint-staged with prettier on most of my side projects lately. Integrates nicely with commitlint and conventional commits as well. 

Want me to take a stab at adding that here?

balazsorban44:
yeah, I think it would make sense. Although I don't want us to enforce conventional commits on individual contributiors, only linting/formatting. These will cause our github actions fail anyway, so it's just nice if the users can fix it before they commit, or we can even fix it for them automatically.

PR #1781, opened 2021-04-20
feat(provider): add Mailchimp provider

michalbundyra:
**What**: add new provider - Mailchimp

**Why**: working in project where I am using this provider, so it might be useful also for others

**How**:

**Checklist**:

- [x] Documentation
- [ ] Tests N/A ?
- [x] Ready to be merged

### DIFF
--- a/src/providers/mailchimp.js
+++ b/src/providers/mailchimp.js
@@ -1,0 +1,22 @@
+export default function Mailchimp(options) {
+  return {
+    id: 'mailchimp',
+    name: 'Mailchimp',
+    type: 'oauth',
+    version: '2.0',
+    scope: '',
+    params: { grant_type: 'authorization_code' },
+    accessTokenUrl: 'https://login.mailchimp.com/oauth2/token',
+    authorizationUrl: 'https://login.mailchimp.com/oauth2/authorize?response_type=code',
+    profileUrl: 'https://login.mailchimp.com/oauth2/metadata',
+    profile: (profile) => {
+      return {
+        id: profile.login.login_id,
+        name: profile.accountname,
+        email: profile.login.email,
+        image: null
+      }
+    },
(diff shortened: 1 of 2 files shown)


src/providers/mailchimp.js
```diff
@@ -0,0 +1,22 @@
+export default (options) => {
+  return {
+    id: 'mailchimp',
+    name: 'Mailchimp',
+    type: 'oauth',
+    version: '2.0',
+    scope: '',
```
balazsorban44, line 7: Do you really not need to set any scopes here? 🤔

balazsorban44 (changes requested):
Gotta love providers that implement OAuth in a way that "just works". 😁🎉 Nice PR, had a tiny comment, could you please check?

src/providers/mailchimp.js
```diff
@@ -0,0 +1,22 @@
+export default (options) => {
```
balazsorban44, line 1:
```suggestion
export default Mailchimp(options) {
```

balazsorban44:
@michalbundyra my question stands about the scope. I wasn't curious if you can remove it but rather if it really works for you locally without providing any? I assumed that you would need something as basic as openid or something like that

michalbundyra:
@balazsorban44 in my code I have empty string for scope and it works, but I'll verify it again and will be back here. Thanks

balazsorban44:
thanks! could you please revert the src/providers/index.js file change? that file is auto-generated now

michalbundyra:
@balazsorban44 I checked again documentation and there is no single word about the scope. It works for me with empty scope tho. 

It's now rebased, so merge conflicts are resolved. Thanks

balazsorban44 (approved):
Thank you so much!

PR #1794, opened 2021-04-21
fix(ts): adjust properties on default interfaces

balazsorban44:
**What**:

Adjusting the `JWT` and `Session` interfaces

**Why**:

Here is the shape of the default JWT payload:
https://github.com/nextauthjs/next-auth/blob/cb1e5a717459d2b975f7f157d1d99d60ae0c4e39/src/server/routes/callback.js#L86-L91

Here is the default Session payload:
https://github.com/nextauthjs/next-auth/blob/cb1e5a717459d2b975f7f157d1d99d60ae0c4e39/src/server/routes/session.js#L32-L39

We should make it possible for users to make their own assumptions about the default properties. Currently, because of how type augmentation works, even if you made for example the `user` field required, it would have been shown as optional, because our default said it was optional. By extracting the default session and JWT interfaces, the user will have full control of those interfaces

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      

Thanks to @kripod for the heads-up!

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -375,7 +375,6 @@ export interface PagesOptions {
  */
 export interface Session extends Record<string, unknown> {
   user?: User
-  accessToken?: string
   expires: string
 }
 
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -365,12 +365,12 @@ export interface PagesOptions {
 }
 
 export interface DefaultSession extends Record<string, unknown> {
-  user: {
-    name: string | null
-    email: string | null
-    picture: string | null
+  user?: {
+    name?: string | null
+    email?: string | null
+    image?: string | null
   }
-  expires: string
+  expires?: string
 }
 
 /**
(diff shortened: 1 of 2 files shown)


0ubbe (approved):

types/index.d.ts
```diff
@@ -373,11 +382,7 @@ export interface PagesOptions {
  * [`Provider`](https://next-auth.js.org/getting-started/client#provider) |
  * [`session` callback](https://next-auth.js.org/configuration/callbacks#jwt-callback)
  */
-export interface Session extends Record<string, unknown> {
-  user?: User
-  accessToken?: string
-  expires: string
-}
+export interface Session extends Record<string, unknown>, DefaultSession {}
```
kripod, line 385: I think that the `Session` and `JWT` interfaces shouldn’t extend `Default{Session,JWT}` as that nullifies the point of this change, unfortunately.

balazsorban44: image
So here is how it looks like for me, when I override `Session` (in the middle)

kripod: That worked like charm, thank you! 🙌

PR #1820, opened 2021-04-23
fix(oauth): handle errors when callback is GET

### LINKED ISSUE #1819: Handle errors on GET requests to callback url
For more info, see https://github.com/nextauthjs/next-auth/discussions/1740#discussioncomment-648097

balazsorban44:
**What**:

Add error handling to `/api/auth/callback/:provider`, when the method is `GET`

**Why**:

As pointed out by @ptorkko in https://github.com/nextauthjs/next-auth/discussions/1740#discussioncomment-648097, we only handled errors when the method was `POST`

**How**

Checking for error on `req.body` AND `req.query` uniformly.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      

fixes #1819

### DIFF
--- a/src/server/lib/oauth/callback.js
+++ b/src/server/lib/oauth/callback.js
@@ -9,37 +9,24 @@ export default async function oAuthCallback(req) {
   const client = oAuthClient(provider)
 
   if (provider.version?.startsWith("2.")) {
-    // The "user" object is specific to the Apple provider and is provided on first sign in
-    // e.g. {"name":{"firstName":"Johnny","lastName":"Appleseed"},"email":"johnny.appleseed@nextauth.com"}
-    let { code, user } = req.query // eslint-disable-line camelcase
+    const params = req.method === "GET" ? req.query : req.body
+    let { code, user = null, error, ...other } = params
 
-    if (req.method === "POST") {
-      try {
-        const body = JSON.parse(JSON.stringify(req.body))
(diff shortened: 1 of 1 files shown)


ptorkko:
the oauth2 spec has an optional `error_description` parameter too, which would be nice to include in the Error as at least for Azure AD it can contain a trivial explanation of the issue.

balazsorban44:
@ptorkko On the following line you can see that I forward *ALL* the parameters to the logger function, in case a provider does not follow the spec well and sent extra information in another way. The front-end will still receive a simplified error (I believe `error=Callback` or `error=OAuthCallback`) for security, but you will now be able to see the reason for failure in the logs.

https://github.com/nextauthjs/next-auth/blob/9b2d007132014c697b90c3c6be5667161f8937e7/src/server/lib/oauth/callback.js#L19

ptorkko:
yup the problem is the trivial error page showing non trivial errors, having to go to the logs when the user could see that `error_description` themselves is kind of annoying

not directly related to the issue at hand but maybe there could be a callback to format the error message with given data

balazsorban44:
I THINK it is a deliberate design choice from @iaincollins' part, so we don't show potentially sensitive error messages to the user, so I am not totally sure. I hope Iain will chip in on this, sorry.

PR #1828, opened 2021-04-23
feat: pass all of providerAccount in linkAccount cb

ndom91:
**What**: Passed all `providerAccount` tokens to the `linkAccount` cb

**Why**: https://github.com/nextauthjs/adapters/issues/48

**How**:

**Checklist**:

- [ ] Documentation ?
- [ ] Tests
- [x] Ready to be merged
      

New PR based off the correct branch (`main`) - replaces #1758.

Also applies new eslint/prettier formatting to this `callback-handler.js` file.

Original PR comment: 

@balazsorban44  @RazerMoon - so this PR passes the rest of the providerAccount object along in the linkAccount cb. However, they are not persisted in the Account table yet obviously as that would require breaking changes to the db model. I also didn't do the named parameter part yet as that would also be a pretty big breaking change. 

What do you guys think? Is the named parameter thing a breaking change we can live with in 4.x, for example? If so, I can do that here as well.

### DIFF
--- a/src/server/lib/callback-handler.js
+++ b/src/server/lib/callback-handler.js
@@ -161,8 +161,7 @@ export default async function callbackHandler(
           providerAccount.id,
           providerAccount.refreshToken,
           providerAccount.accessToken,
-          providerAccount.accessTokenExpires,
-          ...providerAccount
+          providerAccount.accessTokenExpires
         )
         await dispatchEvent(events.linkAccount, {
           user,
@@ -223,8 +222,7 @@ export default async function callbackHandler(
         providerAccount.id,
         providerAccount.refreshToken,
         providerAccount.accessToken,
-        providerAccount.accessTokenExpires,
-        ...providerAccount
+        providerAccount.accessTokenExpires
       )
       await dispatchEvent(events.linkAccount, {
         user,
--- a/src/server/lib/callback-handler.js
+++ b/src/server/lib/callback-handler.js
@@ -1,5 +1,5 @@
-import { AccountNotLinkedError } from '../../lib/errors'
-import dispatchEvent from '../lib/dispatch-event'
+import { AccountNotLinkedError } from "../../lib/errors"
+import dispatchEvent from "../lib/dispatch-event"
 
 /**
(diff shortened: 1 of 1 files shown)


src/server/lib/callback-handler.js
```diff
@@ -205,16 +223,20 @@ export default async function callbackHandler (sessionToken, profile, providerAc
         providerAccount.id,
         providerAccount.refreshToken,
         providerAccount.accessToken,
-        providerAccount.accessTokenExpires
+        providerAccount.accessTokenExpires,
+        ...providerAccount
```
balazsorban44, line 227: So actually I don't think you can spread an object like this onto a function's arguments.
It will throw something like:
> Uncaught TypeError: Found non-callable @@​iterator
    at <anonymous>:1:1

Maybe `...Object.values(providerAccount)` would work, but since it is a temporary solution (We **DEFINITELY** have to make this a named parameters function in 4.0... ), I think just pass it as `providerAccount`. If `providerAccount` has many items (I think it has like 10), it will be a nightmare to get anything out of it otherwise.

We will clean this up in an upcoming breaking release. What do you think?

balazsorban44 (changes requested):

src/server/lib/callback-handler.js
ndom91: Hmm yeah you're right, I thought for sure you could do that. Maybe its only with an array then. 

Okay well lets scrap that part of this PR then, and maybe we can use this PR to do the aforementioned migration to named parameters for `linkAccount()`. What do you think?

balazsorban44: Yes, you can do it with an array, that is why I mentioned `Object.values`.

So in the end, we won't try to support this case in 3.x then? (Which I am fine with personally, but those who use the adapters could benefit anyway.)

balazsorban44: I don't think a simple file formatting is worthy enough change to make it to `main`, so I'll close this. Feel free to reopen, if you actually make the change discussed! :slightly_smiling_face:

(This is only to don't keep hanging PRs!)

PR #1841, opened 2021-04-24
Fixes #1840

### LINKED ISSUE #1840: Signing in using magic link with Outlook email when SafeLinks enabled does not work
**Describe the bug**
Signing in with an Outlook email which is linked to an account where the `SafeLinks` premium feature is enabled does not work.

Another authentication library seems to have the same issue: https://github.com/FusionAuth/fusionauth-issues/issues/629.
When this Outlook feature is enabled, the magic link is encapsulated by a `SafeLinks` Outlook protection URL, e.g.:
```
https://eur02.safelinks.protection.outlook.com/?url=<magic-link-url>
```

From the issue in `FusionAuth`, it seems that the token gets invalidated because of some request that `SafeLinks` is being made beforehand. (Source: https://github.com/FusionAuth/fusionauth-issues/issues/629#issuecomment-629157742)

**Steps to reproduce**
1. Sign in with Outlook email which has `SafeLinks` enabled
2. Click on Sign link from email

**Expected behavior**
User should be able to login

**Screenshots or error logs**
None. The user gets redirected to `/auth/sign-in?error=Verification`

**Additional context**
/

**Feedback**

dweemx:
**What**:

Signing in with an Outlook email which has `SafeLinks` enabled does not work. This bug has been issued #1840.

**Why**:

The token gets invalidated because of some request that `SafeLinks` outlook protection link is making before the validation.

**How**:

When the provider is `email` and the request is a `HEAD` request, it gets redirected to the base URL. This avoids the token to be invalidated.

**Checklist**:

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/server/routes/callback.js
+++ b/src/server/routes/callback.js
@@ -133,6 +133,8 @@ export default async function callback (req, res) {
       logger.error('OAUTH_CALLBACK_ERROR', error)
       return res.redirect(`${baseUrl}${basePath}/error?error=Callback`)
     }
+  } else if (provider.type === 'email' && req.method === 'HEAD') {
+    return res.redirect(baseUrl)
   } else if (provider.type === 'email') {
     try {
       if (!adapter) {


lluia:
Thanks for PR @dweemx 💚

This seems quite a tricky specific case, could you explain more in-depth your solution, why re-directing to `baseUrl` will prevent the token from being invalidated. Could you as well explain how the `HEAD` method on the request comes into play here? 💭

balazsorban44 (commented):
This seems to be a very case-specific change, not to mention that it won't work anyway. The `rotues/callback.js` default exported function is used in `src/server/index.js`, which only handles `POST` and `GET` itself.

We will need more context and maybe a reproduction of how and what this is supposed to solve. Let's discuss this further on #1840, before we have a PR for this.

PR #1851, opened 2021-04-25
Made `GetSessionOptions` optional

typedashutosh:
As it was confusing on client side to put in an empty object

**What**: Made `GetSessionOptions` optional in `getSession()`

**Why**: Because in client it was unnecessary, but when used with typescript, TS makes unnecessary complaints.

**How**: Just adding a `?` at param of the function. `getSession(options?: GetSessionOptions)`

**Checklist**:

- [x] Documentation
- [x] Tests
- [x] Ready to be merged


0ubbe (approved):
Thank you ❤️

PR #1853, opened 2021-04-25
fix(ts): improve `events` handlers' types

### LINKED ISSUE #1639: Proper Typing for Events
Since the arguments to each event in the `events` object are known, the typing should include that information.

To making coding easier and prevent errors when using `events`

Currently the events object has the following typing:
```typescript
export type EventCallback = (message: any) => Promise<void>;
export type EventsOptions = Partial<Record<EventType, EventCallback>>;
```

I think it should have an actual shape defining the [event names supported](https://next-auth.js.org/configuration/events) and their message types. As an example:

```typescript
export type EventCallback<MessageType = any> = (message: MessageType) => Promise<void>;

type SignInMessage = {
  user: User;
  account: Record<string, unknown>;
  isNewUser?: boolean;
};

export type EventsOptions = Partial<{
  signIn: EventCallback<SignInMessage>;
  // And so on...
}>;
```

**Potential problems**
I don't foresee any issues with this. The only thing I'm unsure about is if your .d.ts files are autogenerated or manually written.

Of course, just let me know if your .d.ts files are autogenerated or manually written.

zackdotcomputer:
**What**:

This change introduces typing for the `events` object so that developers using Typescript with next-auth will receiving hinting as to the type of the incoming arguments to events callback functions.

Resolves #1639 

**Why**:

Because previously developers didn't have any idea what arguments they'd receive when listening to events even though it was relatively well-defined in the source code.

**How**:

By analyzing the source code for the possible paths to trigger `events` and then editing the typing files to match the discovered pathways.

**Checklist**:

- [x] Documentation (I think N/A since this basically _is_ documentation?)
- [X] Tests
- [X] Ready to be merged


balazsorban44 (changes requested):
Thank you!


Could you please also update the docs in these places:

https://github.com/nextauthjs/next-auth/blob/main/www/docs/configuration/options.md#description-7

https://github.com/nextauthjs/next-auth/blob/main/www/docs/configuration/events.md

types/index.d.ts
```diff
@@ -340,20 +352,59 @@ export interface CookiesOptions {
...
 /** [Documentation](https://next-auth.js.org/configuration/events) */
-export type EventType =
-  | "signIn"
-  | "signOut"
-  | "createUser"
-  | "updateUser"
-  | "linkAccount"
-  | "session"
-  | "error"
+export type EventCallback<MessageType = any> = (
```
lluia, line 355: Could we mark the default type for `MessageType` as [`unknown`](https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type) to be more type-safe? 💭

types/index.d.ts
```diff
@@ -340,20 +352,59 @@ export interface CookiesOptions {
...
+  message: MessageType
+) => Promise<void>
 
-/** [Documentation](https://next-auth.js.org/configuration/events) */
-export type EventCallback = (message: any) => Promise<void>
+export interface SessionEventMessage {
+  session: Session & Record<string, any>
+  jwt?: JWT
+}
```
lluia, line 362: This `interface` doesn't seem to be used; if needed, could you let us know the use-case for it? 🧐

types/index.d.ts
```diff
@@ -340,20 +352,59 @@ export interface CookiesOptions {
...
+  account: Record<string, unknown>
+  isNewUser?: boolean
+}
+
+export interface LinkAccountEventMessage<TUser> {
+  user: TUser
+  providerAccount: Record<string, unknown>
+}
+
+export interface EventOptions<TUser> {
```
lluia, line 381: Could we rename this interface to `EventCallbacks`? 

Given all these are callback functions, I think the name is more explicit for developers less familiar with the library 💆🏽‍♂️

I'm aware this will be a breaking change, as before it was also called `EventOptions`, but I think we would gain some clarity renaming it. 

We're treating breaking changes on the types as patch releases of the lib, so I think it would be ok 👍🏽

0ubbe (changes requested):
Thanks 💯 , brilliant work; I leave a couple of suggestions 👍🏽

types/index.d.ts
zackdotcomputer: Yup - my bad, missed that clean-up from an earlier draft.

types/index.d.ts
zackdotcomputer: I'm in favor of that rename and agree that breaking the types makes a breaking change overall.

types/index.d.ts
zackdotcomputer: This required tweaking how the callback object's types were constructed - now the type for the signOut and session events is passed as a generic and the JWT and Session constructions are type aliases.

balazsorban44 (changes requested):
Could you please fix the merge conflicts?

types/index.d.ts
```diff
@@ -18,7 +23,12 @@ import {
  *
  * [Documentation](https://next-auth.js.org/configuration/options#options)
  */
-export interface NextAuthOptions {
+export interface NextAuthOptions<
+  TUser extends User = User,
+  TProfile extends AdapterProfile = AdapterProfile,
+  TSession extends AdapterSession = AdapterSession,
+  TVerificationRequest extends VerificationRequest = VerificationRequest
+> {
```
balazsorban44, line 31: I don't see why this was added here? Could you explain? If you want to override any of these, check out https://next-auth.js.org/getting-started/typescript#module-augmentation

types/index.d.ts
```diff
@@ -127,7 +139,7 @@ export interface NextAuthOptions {
    * [Default adapter](https://next-auth.js.org/schemas/adapters#typeorm-adapter) |
    * [Community adapters](https://github.com/nextauthjs/adapters)
    */
-  adapter?: Adapter
+  adapter?: Adapter<TUser, TProfile, TSession, TVerificationRequest>
```
balazsorban44, line 142: Won't be needed when you fix the merge conflicts

balazsorban44 (changes requested):

balazsorban44:
Also my previous comment still stands: https://github.com/nextauthjs/next-auth/pull/1853#pullrequestreview-644163820

zackdotcomputer:
Sure thing @balazsorban44 - I was just starting on the code comments before tackling the documentation so I was sure to be documenting my final draft. Will tackle the merge and your code comments and then wrap up the docs changes.

types/index.d.ts
zackdotcomputer: This is required so that we can demand conformance between the adapter and the events - by pulling values like `TUser` up to be a generic on the options object we enable Typescript to make sure that the User object our adapter expects is the same as the User object our events expect.

types/index.d.ts
zackdotcomputer: See above for why this is still included.

types/index.d.ts
balazsorban44: They will be, if you simply reference the same `User` interface anywhere! Check out the module augmentation stuff! It is used instead of generics, because it works across submodules as well! (For example overriding the `Session` interface with module augmentation, you will get the same types on the `session()` callbacks first param, and the returned valued of `useSession()`, even though one comes from `next-auth` and the other in `next-auth/client`)

zackdotcomputer: Oh huh... Ok on second read through I am starting to actually grok what module augmentation is.

Ok if this is how you want to do things throughout then I'll just ditch the generics. I will just 2 cents that as a user of the library this is unexpected to me since I can't think of another library where you customize the base interfaces using module augmentation rather than just passing generics - even if you would have to pass them twice (server + client). I don't love it.

balazsorban44: We do document this here, so anyone using this library should expect it 😊:

https://next-auth.js.org/getting-started/typescript#module-augmentation

I first saw it on the `@material-ui` documentation page:

https://material-ui.com/guides/typescript/#customization-of-theme

it is very powerful, and works very neatly! you will only have to defined your interfaces in a single location.

zackdotcomputer: Yeah reading the linked TS handbook page from your docs page was what finally got it into my head. My instincts about this are still that it's too similar to hard-casting and scala-like `implicit` declarations - two things that I feel are very dangerous - for me to feel comfortable with it.  I'm going to have to sit on it for a day to really chew through why I don't like it, though. (I think it's something about the fact that next-auth is sandwiched between code you own on both sides.)

But I super disagree with the idea that there is less magic with this than generics - with generics it's very clear right in place what is being passed around. And if TS infers the wrong generic you can manually specify it. With this it feels much more like "ah yes you have told TS over in this one file that it should cast every single User coming from this module to be type MyUser and it will just silently and magically do that cast."

zackdotcomputer: BTW why did you leave the generics in Adapter at all? Seems like if you're going in on this, go whole hog.

zackdotcomputer:
All comments responded to and documentation updated.

types/index.d.ts
balazsorban44: The thing is, those places where it is referenced, they HAVE TO match in ALL cases. There is just no place between for example what the session callaback returns and what the useSession hook gives you, where the value could change. Providing it in a single location and referencing it implicitly makes it impossible for the user to screw it up.

zackdotcomputer: @balazsorban44 hmm ok yeah that makes sense - generic typing with hooks is very very painful. I still don't love it but since React hasn't given us a better way I can't think of something that would make that more graceful.

balazsorban44: The `Adapters` interface will be used by third party, and I guess it's harder for us to forsee what users want to do. Maybe they have a database that won't much our interfaces 100%.

I'm also not quite there yet with the adapters/databases, but both @lluia and @kriood have seen most of the types/interfaces redesigns or even contributed and I got positive feedback from them.

If in the future we can actually simplify the Adapters interface as well, I'm all for it!

zackdotcomputer: Aha well I think there's the big difference in how we were approaching this - I've been coming at this from the adapter working upwards. I was trying to get the TUser in the options and callbacks to infer a type match to the user object in the adapter.

Basically I was thinking of next-auth as mediating with my ORM (Prisma in my case, but same with TypeORM) and the cookies/session for me, but that I would want to still use the underlying ORM object so that I could stick all sorts of extra values on my user object. So I was using the generics piping to run the specific user type that Prisma had generated in and out from the other server functions.

But you're quite right that that does nothing for the tidy typing on the client-side of things. Tbh, I'm not using the client side in practice other than plucking out the user's id from the session, so I had a bit of a blindspot there. I get better the motivation now and I gotta go to sleep. We'll tackle the adapters another time 😉

balazsorban44: Glad we came to an understanding! 

I'm gonna 😴 as well. 😁

www/docs/configuration/options.md
```diff
@@ -5,7 +5,7 @@ title: Options
 
 ## Environment Variables
 
-### NEXTAUTH_URL 
+### NEXTAUTH_URL
```
lluia, line 8: Could we revert this file? 

I have a [PR open](https://github.com/nextauthjs/next-auth/pull/1900) updating it extensively and I think it's gonna end a bit in conflict hell 😬 due to the `prettier` changes. I'll add the bit you updated pointing to the `events` documentation  on my PR 👍🏽

0ubbe (changes requested):
Just one small thing to prevent conflicts on the documentation 🙏🏽 , thanks a lot for this massive work 💯

www/docs/configuration/options.md
lluia: In fact, NVM I confused the file name 🤦🏽‍♂️ ☕️ , ha ha we're all good 💚

0ubbe (approved):
💪🏽

types/index.d.ts
```diff
@@ -342,20 +342,69 @@ export interface CookiesOptions {
...
-export type EventCallback = (message: any) => Promise<void>
+/**
+ * If using a `credentials` type auth, the user is the raw response from your
+ * credential provider.
+ * For other providers, you'll get the User object from your adapter, the account,
+ * and an indicator if the user was new to your Adapter.
+ */
+export interface SignInEventMessage<TUser> {
+  user: TUser
+  account: Record<string, unknown>
```
balazsorban44, line 357: We have an `Account` interface for this.
```suggestion
  account: Account
```

types/index.d.ts
```diff
@@ -342,20 +342,69 @@ export interface CookiesOptions {
...
-/** [Documentation](https://next-auth.js.org/configuration/events) */
-export type EventCallback = (message: any) => Promise<void>
+/**
+ * If using a `credentials` type auth, the user is the raw response from your
+ * credential provider.
+ * For other providers, you'll get the User object from your adapter, the account,
+ * and an indicator if the user was new to your Adapter.
+ */
+export interface SignInEventMessage<TUser> {
+  user: TUser
```
balazsorban44, line 356:
```suggestion
export interface SignInEventMessage {
  user: User
```

No need for generics, as per our Module Augmentation discussion. The `User` here will never defer from the `User` elsewhere.

types/index.d.ts
```diff
@@ -342,20 +342,69 @@ export interface CookiesOptions {
...
+  account: Record<string, unknown>
+  isNewUser?: boolean
+}
 
-/** [Documentation](https://next-auth.js.org/configuration/events) */
-export type EventsOptions = Partial<Record<EventType, EventCallback>>
+export interface LinkAccountEventMessage<TUser> {
+  user: TUser
+  providerAccount: Record<string, unknown>
```
balazsorban44, line 363:
```suggestion
export interface LinkAccountEventMessage {
  user: User
  account: Account
```

types/index.d.ts
```diff
@@ -342,20 +342,69 @@ export interface CookiesOptions {
...
+/**
+ * The event callbacks will take this form if you are using Sessions
+ * and not using JWTs:
+ * signOut will receive the underlying db adapter's session object, and session
+ * will receive the NextAuth client session with extra data.
+ */
+export type SessionEventCallbacks = EventCallbacks<
+  Session | null,
+  {
+    session: Session & Record<string, any>
```
lluia, line 403: Here and in line `387` we don't need `Record<string, any>`, these interfaces can be extended through [module augmentation](https://next-auth.js.org/getting-started/typescript) as explained before 👍🏽

types/index.d.ts
```diff
@@ -342,20 +342,69 @@ export interface CookiesOptions {
...
-  | "signIn"
-  | "signOut"
-  | "createUser"
-  | "updateUser"
-  | "linkAccount"
-  | "session"
-  | "error"
+export type EventCallback<MessageType = unknown> = (
+  message: MessageType
+) => Promise<void>
```
balazsorban44, line 347:
```suggestion
export type EventCallback<MessageType> = (
  message: MessageType
) => Promise<void>
```

I think here it is OK to enforce a generic, as otherwise this type is quite useless.

types/index.d.ts
```diff
@@ -342,20 +342,69 @@ export interface CookiesOptions {
...
+ * The event callbacks will take this form if you are using Sessions
+ * and not using JWTs:
+ * signOut will receive the underlying db adapter's session object, and session
+ * will receive the NextAuth client session with extra data.
+ */
+export type SessionEventCallbacks = EventCallbacks<
+  Session | null,
+  {
+    session: Session & Record<string, any>
+  }
```
balazsorban44, line 404:
```suggestion
export interface CommonEventCallbacks {
  signIn: EventCallback<SignInEventMessage>
  error: EventCallback<unknown>
}

/**
 * The event callbacks will take this form if you are using JWTs:
 * signOut will receive the JWT and session will receive the session and JWT.
 */
export interface JWTEventCallbacks extends CommonEventCallbacks{
  signOut: EventCallback<JWT>
  session: EventCallback<{
    session: Session
    jwt: JWT
  }>
}

/**
 * The event callbacks will take this form if you are using Sessions
 * and not using JWTs:
 * signOut will receive the underlying DB adapter's session object, and session
 * will receive the NextAuth client session with extra data.
 */
export interface SessionEventCallbacks extends CommonEventCallbacks {
  signOut: EventCallback<Session | null>
  session: EventCallback<{ session: Session }>
  createUser: EventCallback<User>
  updateUser: EventCallback<User>
  linkAccount: EventCallback<LinkAccountEventMessage>
}

export type EventCallbacks = JWTEventCallbacks | SessionEventCallbacks
```

So I also removed all the generics that were not needed here, set better defaults, and separated `JWTEventCallbacks` and `SessionEventCallbacks` into their interfaces, according to when they can be present. (Eg.: updateUser is not an event when you use JWT)

types/index.d.ts
```diff
@@ -342,20 +342,69 @@ export interface CookiesOptions {
...
+>
+
+export type EventType = keyof EventCallbacks<any>
```
balazsorban44, line 407:
```suggestion
export type EventType = keyof EventCallbacks
```

balazsorban44 (changes requested):
Some feedback from me! Please remove the extraneous generics, as we discussed Module Augmentation. Otherwise, it is looking good and will be a nice improvement, so thank you! :pray: 

I am also thinking that most of these should be moved to `internals/events.d.ts` maybe to keep down the `index.d.ts` file size as well. What do you think @lluia?

types/index.d.ts
zackdotcomputer: Done 👍

types/index.d.ts
zackdotcomputer: Swapped 👍

types/index.d.ts
zackdotcomputer: Swapped 👍

types/index.d.ts
zackdotcomputer: Swapped 👍

types/index.d.ts
zackdotcomputer: Yeah this type is purely to avoid having to copy-paste the extra characters of `=> Promise<void>`

types/index.d.ts
zackdotcomputer: Swapped 👍

types/index.d.ts
zackdotcomputer: Let me know if I'm misunderstanding the code here but I think `createUser`, `updateUser`, and `linkAccount` will all still be sent even if you're using JWTs?

zackdotcomputer:
Resolved merge conflict, updated to main repo's `main` branch, applied all suggested changes except for keeping the `createUser`, `updateUser`, and `linkAccount` in the common events interface since I think those are sent even if you are using JWTs.

types/index.d.ts
zackdotcomputer: Applied the rest of this, but left those three back in the common events interface

balazsorban44:
They aren't, you can see it in the source code:

https://github.com/nextauthjs/next-auth/blob/17b789822de64eb845e1e8e49ea83dbff56344f4/src/server/lib/callback-handler.js

See the following lines: 95, 100, 154, 199, 210

zackdotcomputer:
Ah no worries, it sounds like we're actually reading this the same way @balazsorban44 that those events are sent based on whether or not you have an adapter connected, which as far as I can tell is separate from whether you are using JWTs (since you could configure an adapter and also set `session: { jwt: true }` I think). If we want to get **precise** then, I think we wind up needing `interface HasAdapterEventCallbacks` with these three methods and then the final type is `type EventCallbacks = JWTEventCallbacks | ((JWTEventCallbacks | SessionEventCallbacks) & HasAdapterEventCallbacks)`

But are we now splitting hairs? Let me know if you think that's excessive or not - if not I can restructure the type like that pretty easily.

balazsorban44:
Right! I think it's a big improvement! I'll approve this in its current form, and if for some reason it won't be good enough, we can open a new PR. Thank you for your contribution!

balazsorban44 (approved):

balazsorban44:
@zackdotcomputer sorry, this is the last thing I'll ask, bu before merging, could you please fix the failing tests?

balazsorban44 (changes requested):
There are some failing tests

zackdotcomputer:
Okay - we'll see how the build goes but that should have fixed it.

balazsorban44:
@zackdotcomputer Unfortunately it still fails. I recommend running `npm run test` locally before you commit and push.

zackdotcomputer:
Ah yup my bad @balazsorban44 - I was running the tests but not the lint. Fixed the lint now too - run test is green locally.

balazsorban44:
Yeah, we don't really have tests at the moment. 🥲. Working on it though! #1937

balazsorban44 (approved):

Issue #1086, opened 2021-01-11
Improve contributing instructions on using tests


balazsorban44:
As per @iaincollins's request https://github.com/nextauthjs/next-auth/pull/1085#issuecomment-757903043

We should better document the tests, and why/how are they useful.

iaincollins:
Thanks for the reminder!

balazsorban44:
Closing this as the there is lack of activity on this matter. Once we have a better strategy for testing we can join forces and create sufficient documentation around it together.

PR #1862, opened 2021-04-26
feat(adapter): split out adapters

balazsorban44:
## Reasoning 💡

**IMPORTANT :warning::** With this change, **you don't have to do anything**, this is an internal change only, and should not affect you in any way. If it does, it is most possibly a bug and not intended. Please create a [bug report :bug:](https://github.com/nextauthjs/next-auth/issues/new?assignees=&labels=bug&template=bug_report.md)

We are migrating the built-in providers to their own repository, they can now be found at https://github.com/nextauthjs/adapters/tree/canary/packages

TypeORM Adapter : `@next-auth/typeorm-legacy-adapter`
Prisma Adapter: `@next-auth/prisma-legacy-adapter`

NOTE: for new projects, you might be more interested in using `@next-auth/prisma-adapter` which has been created in a collaboration with folks from Prisma (@williamluke4) and with the help of our awesome community members like @kripod! :tada:

These adapters are included by default so we can provide an easier migration to 4.0. Our plans are that in `next-auth@4` and above we might not have any built-in providers, you might have to install them explicitly, or we might have a single "default" adapter.

Another major change is the removal of tests... This is not done lightly, and I feel bad, but in its current form, tests are dysfunctional and do not bring anything to the table. There hasn't been a single successful test run in a very long time: https://github.com/nextauthjs/next-auth/actions/workflows/integration.yml They either fail or just time out. There is a better testing strategy in work by @iaincollins and once it's in place, it will increase the quality of this package by margins. Until then we just have to rely on the recently added TypeScript type tests and good community relations (which we have done so for months now, anyway) so we can get bugs reported earlier.

## Checklist 🧢

~- [ ] Documentation~
~- [ ] Tests~
- [x] Ready to be merged
      

## Affected issues 🎟

https://github.com/nextauthjs/adapters/issues/37, #1086


package.json
```diff
@@ -37,22 +37,9 @@
...
-    "test:app:rebuild": "npm run build && docker-compose -f test/docker/app.yml up -d --build",
-    "test:app:stop": "docker-compose -f test/docker/app.yml down",
-    "test": "npm run test:app:rebuild && npm run test:integration && npm run test:app:stop && npm run test:types",
-    "test:db": "npm run test:db:mysql && npm run test:db:postgres && npm run test:db:mongodb && npm run test:db:mssql",
-    "test:db:mysql": "node test/mysql.js",
-    "test:db:postgres": "node test/postgres.js",
-    "test:db:mongodb": "node test/mongodb.js",
-    "test:db:mssql": "node test/mssql.js",
-    "test:integration": "mocha test/integration",
+    "test": "echo \"Write some tests...\"; npm run test:types",
```
lluia, line 40: 😢 , I was thinking about this, though. 

The best IMO would be to have integration/e2e tests covering the sig-in/sign-out user paths provided by **NextAuth**. 

However, I think we could still even benefit from unit tests, and they could serve as documentation as well. We could, for instance, aim to have all the exports of a given module unit tested. 🤔

0ubbe (approved):
Coolio 💯 💪🏽

package.json
balazsorban44: Yes, we have options how to create some basic but still useful tests that could benefit us, and would like to chat about them with @iaincollins.

Issue #48, opened 2018-09-05
Bug: `sessionResave` default value should be `true`


iaincollins:
Currently the default value for `sessionResave` is `false`. The default for this option should be `true`.

Having it set to false is causes sessions to eventually expire after `sessionMaxAge` (which has a default of 7 days) instead of rotating (the expiry time being bumped every time the session is active).

Some people will prefer to be able to control this option so it should still be exposed, but the default should be to automatically rotate sessions once active.

A fix will be coming this week. To work around this for now, just specify `sessionResave : true` in the options in `next-auth.config.js`.

iaincollins:
Fixed in 1.12.1

PR #1866, opened 2021-04-27
fix(types): accessTokenExpires can only be null

kripod:
## Reasoning 💡

In adapters, the value of `accessTokenExpires` can either be `number` or `null`.

## Checklist 🧢

~- [] Documentation~
~- [] Tests~
- [x] Ready to be merged


kripod:
Resolves the confusion pointed out in #48.

balazsorban44 (changes requested):
Please make sure tests are also passing, thank you!

Issue #1832, opened 2021-04-23
Adapter Method Promise Rejections are Unhandled, causes Unreliable Upstream Promise Resolutions


WillTheVideoMan:
**Describe the bug**
Promises that reject with errors from within adapter methods are in some cases unhandled, leading to `UnhandledPromiseRejectionWarning` and potentially leaving pending upstream promises that never resolve.

An example of such an unhandled promise rejection is during the email sign-in flow. In this case, if the adapter method `getUserByEmail` rejects with an error, then an `UnhandledPromiseRejectionWarning` is logged to the `node` console.

In development, upstream promises (such as the one returned by the client `signIn` method when `redirect` is `false`) that rely on the `getUserByEmail` promise resolving or rejecting are left indefinitely pending.

In production, upstream promises are eventually rejected by virtue of a `500` response from the server as it deals with the unhandled rejection exception.

**Steps to reproduce**
Either clone repro:
https://github.com/WillTheVideoMan/nextauth-repro-unhandled-promise-rejection-warning

Or:
1) Configure `next-auth` in a development environment with an adapter. 
2) To test this case, configure the adapter such that the adapter methods always reject with an error e.g. provide incorrect database credentials. 
3) Make a request to an endpoint which contains an unhandled adapter promise e.g.
https://github.com/nextauthjs/next-auth/blob/17b789822de64eb845e1e8e49ea83dbff56344f4/src/server/routes/signin.js#L42
4) Observe an `UnhandledPromiseRejectionWarning` warning in the `node` console.
5) Observe that any upstream promises such as the one making the request are indefinitely pending. 
6) Deploy instance to a production environment.
7) Observe that the upstream promises are resolved eventually by virtue of a `500` response.

**Expected behavior**
Expect all promises which may reject to be handled explicitly.

I think the overall intention here should be to avoid unhandled promise rejections entirely. This would ensure errors more consistently cascade in predictable ways, and importantly, would move in line with the upcoming `node v15` revisions: https://github.com/nodejs/node/pull/33021

**Screenshots or error logs**
```console
(node:21664) UnhandledPromiseRejectionWarning: Error: get_user_by_email_error
    at getUserByEmail (webpack-internal:///./auth/fauna-adapter.js:135:31)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
[...]

PR #1871, opened 2021-04-27
feat(adapter): take away error handling from adapters

balazsorban44:
## Reasoning 💡

## Checklist 🧢

- [ ] Documentation
- [ ] Tests 
- [x] Ready to be merged
- [x] add debugging
      

## Affected issues 🎟

#1832


balazsorban44:
TODO add logger.debug

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "next-auth",
-  "version": "0.0.0-semantically-released",
+  "version": "3.20.0-canary.3",
```
balazsorban44, line 3: need to change this back

src/adapters/error-handler.js
```diff
@@ -0,0 +1,22 @@
...
+ * Handles adapter induced errors.
+ * @param {import("types/adapters").AdapterInstance} adapter
+ * @return {import("types/adapters").AdapterInstance}
+ */
+export default function adapterErrorHandler(adapter) {
+  return Object.keys(adapter).reduce((acc, method) => {
+    const adapterMethod = adapter[method]
+    acc[method] = async (...args) => {
+      try {
+        return await adapterMethod(...args)
```
balazsorban44, line 13: should log out args with debug, and an adapter prefix(?)

@kripod, currently `logger.debug` is rewritten in adapters to include a prefix for  `code`. Do you think we could return a name property next to getAdapter for this purpose?

any thoughts?

src/adapters/error-handler.js
```diff
@@ -0,0 +1,22 @@
...
+      } catch (error) {
+        const e = new UnknownError(error)
+        e.name = `${method[0].toUpperCase()}${method.slice(1)}Error`
```
balazsorban44, line 16: turns out custom errors may not need to be defined as they only have  different name. this works great!

src/adapters/error-handler.js
kripod: We may require a `{ displayName: string }` as the minimum required `config` param when instantiating adapters. By doing so, we’ll also solve the typing issue of `Adapter(requiredParam, …)`.

balazsorban44: After my recent commit, it will be possible for adapters to return a `displayName` property in `getAdapter`. Unfortunately we don't have the adapter configuration when we invoke the `.getAdapter()` method.

balazsorban44:
Currently waiting for @kripod and his results after testing these changes in production. We expect fewer unhandled errors.

kripod (approved):

Issue #1752, opened 2021-04-17
Typescript errors when using customer provider


bhaskarmurthy:
**Describe the bug**
I'm getting a type error when passing a custom `Provider` to the `NextAuth` method.

**Error 1**
When passing an `id` to the provider config, there's an error about `Type 'string' is not assignable to type 'undefined'.`

**Error 2**
However, if I don't set the `id` property, I get a different error about `Object literal may only specify known properties, and 'protection' does not exist in type 'Provider<undefined, "oauth">'`

**Error 3**
I'm also getting another type error if for `domain` if I set the Provider type using the template:
`const customProvider: Provider<'custom'> = {...}`

Error:
`Property 'domain' is missing in type 'Provider<"custom", "oauth">' but required in type '{ domain: string; }'`

```
import NextAuth from "next-auth";
import { Provider } from "next-auth/providers";

const customProvider: Provider = {
  // required
  id: "custom",
  name: "Custom Provider",
  type: "oauth",
  version: "2.0",
  accessTokenUrl: process.env.OAUTH_TOKEN_URL,
  authorizationUrl: `${process.env.OAUTH_AUTHORIZE_URL}?response_type=code`,
  clientId: process.env.OAUTH_CLIENT_ID,
  // optional
  params: { grant_type: "authorization_code" },
  protection: ["pkce", "state"],
};

export default NextAuth({
  providers: [customProvider],
  // debug: true,
});
```

**Steps to reproduce**
Steps to reproduce the behavior.

Include a link to public repository which can be used to reproduce the behaviour.

**Expected behavior**
If I set the `id` to be a string constant using the `Provider<'custom'>` type, I should be able to set the `protection` optional value (as per the docs)

**Screenshots or error logs**
If applicable add screenshots or error logs to help explain the problem.

**Additional context**

**Feedback**
*Documentation refers to searching through [online documentation](https://next-auth.js.org), code comments and issue history. The example project refers to [next-auth-example](https://github.com/iaincollins/next-auth-example).*

* [ ] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [ ] Could not find relevant documentation
* [ ] Found the example project helpful
* [ ] Did not find the example project helpful

balazsorban44:
a big rewrite and built-in types are coming soon, watch out for 

https://github.com/nextauthjs/next-auth/pull/1690

it will first be merged to beta, then soon after to main.

hopefully that should address all your problems.

if not, feel free to repoen this.

Issue #1846, opened 2021-04-25
RFC: Improve OAuth provider configuration


balazsorban44:
**Summary of proposed feature**
Introducing a new way of configuring OAuth providers. With this proposal, we can provide granular control from basic (with the best defaults we can think of) to completely controlled hooks into the login flow. (The user would be able to do the actual fetch calls themselves. Hopefully, this will almost never would be necessary)

Here is an example:
```js
export default function Auth0(options) {
  return {
    id: "auth0",
    name: "Auth0",
    authorization: `https://${options.domain}/authorize?grant_type=authorization_code&response_type=code&scope=openid%20email%20profile`,
    // Alternatively, you could pass an url and params
    // which will be combined for you for convinience.
    // authorization: {
    //   url: `https://${options.domain}/authorize`,
    //   params: {
    //     grant_type: "authorization_code",
    //     response_type: "code",
    //     scope: "openid email profile",
    //   },
    // },
    // Skips fetching user data. In some rare situations, you may not need this. See #1065
    userinfo: null,
    token: {
      url: `https://${options.domain}/oauth/token`,
      method: "POST",
      via: "header",
      // Complete control over how to get the userinfo.
      async request({ tokens }) {
        const response = await fetch("custom/endpoint/token", {
          headers: {
            Authorization: `Bearer ${tokens.access_token}`,
          },
        })
        return await response.json()
      },
    },
    profile(profile) {
      return {
        id: profile.sub,
        name: profile.nickname,
        email: profile.email,
        image: profile.picture,
      }
    },
    ...options,
  }
}
```

**Purpose of proposed feature**
To simplify the configuration by reducing the number of required fields, but give full control at the same time, for advanced use cases.

**Detail about proposed feature**

So to sum it up, 3 new methods could be introduced, using a similar shape:

```ts
/** Any contextual information that makes the request useful, eg.: tokens in case of the token endpoint */
type RequestContext = any

type OAuthEndpoint = string | null | {
  url?: string
  method?: "POST" | "GET"
  /** How to pass the access token */
  via?: "header" | "body" | "query"
  /** Will be used as URL params if method is GET, or the body as x-url-form-encoded if POST*/
  params?: URLSearchParams | Record<string, unknown>
  request(context: RequestContext)?: Promise<any>
}
```

**Potential problems**
As there are no conflicting configuration options, this could be implemented in a non-breaking way, meaning in theory we don't have to wait until the next major version.

**Describe any alternatives you've considered**
[...]

Issue #1065, opened 2021-01-10
Error on missing `profileUrl` and/or `profile`


jackHedaya:
**Describe the bug**
In the documentation, `profileUrl` and `profile` for custom providers are said to be not required in documentation but the application will crash if they are not given. This error is also the root of issue #209 as discovered by @ospira.

**Steps to reproduce**
Steps to reproduce the behavior.
+ Create custom provider
+ Omit `profileUrl` and/or `profile`

**Expected behavior**
Authentication to succeed and `profile` to be set to undefined.

**Screenshots or error logs**
```
[next-auth][error][oauth_get_access_token_error] [
  TypeError [ERR_INVALID_ARG_TYPE]: The "url" argument must be of type string. Received undefined
      at new NodeError (node:internal/errors:278:15)
      at validateString (node:internal/validators:123:11)
      at Url.parse (node:url:160:3)
      at Object.urlParse [as parse] (node:url:155:13)
      at exports.OAuth2._request (/Users/jackhedaya/Desktop/thoughtful-fish/node_modules/oauth/lib/oauth2.js:72:22)
      at /Users/jackhedaya/Desktop/thoughtful-fish/node_modules/next-auth/dist/server/lib/oauth/client.js:240:12
      at new Promise (<anonymous>)
      at exports.OAuth2.<anonymous> (/Users/jackhedaya/Desktop/thoughtful-fish/node_modules/next-auth/dist/server/lib/oauth/client.js:239:12)
     ...
```

**Additional context**
Original custom Provider:
```ts
NextAuth(req, res, {
    providers: [
      {
        id: '[redacted]',
        name: '[redacted]',
        type: 'oauth',
        version: '2.0',
        accessTokenUrl: '[redacted]',
        requestTokenUrl: '[redacted]',
        authorizationUrl: `[redacted]`,
        clientId: process.env.CONSUMER_KEY,
        params: {
          grant_type: 'authorization_code',
        },
      },
    ],
    debug: true,
  })
```

Working custom provider:

```ts
NextAuth(req, res, {
    providers: [
      {
        id: '[redacted]',
        name: '[redacted]',
        type: 'oauth',
        version: '2.0',
        accessTokenUrl: '[redacted]',
        requestTokenUrl: '[redacted]',
        authorizationUrl: `[redacted]`,
        clientId: process.env.CONSUMER_KEY,
        params: {
          grant_type: 'authorization_code',
        },
        profileUrl: '[redacted]',
[...]

PR #1876, opened 2021-04-28
fix(ts): optional variables for custom provider options

leeoocca:
## Reasoning 💡

I modified the provider interface and made some of the options optional.

I followed the table available in the [docs](https://next-auth.js.org/configuration/providers#oauth-provider-options). Some were required even though the table says they aren't.

## Checklist 🧢

- [ ] ~Documentation~
- [x] Tests (local)
- [x] Ready to be merged

## Affected issues 🎟

#1752 was similar but has already been fixed.

### DIFF
--- a/types/providers.d.ts
+++ b/types/providers.d.ts
@@ -29,13 +29,13 @@ export interface OAuthConfig<P extends Record<string, unknown> = Profile>
   scope: string
   params: { grant_type: string }
   accessTokenUrl: string
-  requestTokenUrl: string
+  requestTokenUrl?: string
   authorizationUrl: string
-  profileUrl: string
-  profile(profile: P, tokens: TokenSet): Awaitable<User & { id: string }>
+  profileUrl?: string
+  profile?(profile: P, tokens: TokenSet): Awaitable<User & { id: string }>
   protection?: ProtectionType | ProtectionType[]
   clientId: string
-  clientSecret:
+  clientSecret?:
     | string
     // TODO: only allow for Apple
     | Record<"appleId" | "teamId" | "privateKey" | "keyId", string>
--- a/types/providers.d.ts
+++ b/types/providers.d.ts
@@ -31,11 +31,11 @@ export interface OAuthConfig<P extends Record<string, unknown> = Profile>
   accessTokenUrl: string
   requestTokenUrl?: string
   authorizationUrl: string
-  profileUrl?: string
-  profile?(profile: P, tokens: TokenSet): Awaitable<User & { id: string }>
+  profileUrl: string
+  profile(profile: P, tokens: TokenSet): Awaitable<User & { id: string }>
   protection?: ProtectionType | ProtectionType[]
   clientId: string
-  clientSecret?:
+  clientSecret:
     | string
     // TODO: only allow for Apple
     | Record<"appleId" | "teamId" | "privateKey" | "keyId", string>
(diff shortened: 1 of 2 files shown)


types/providers.d.ts
```diff
@@ -29,13 +29,13 @@ export interface OAuthConfig<P extends Record<string, unknown> = Profile>
   scope: string
   params: { grant_type: string }
   accessTokenUrl: string
-  requestTokenUrl: string
+  requestTokenUrl?: string
```
balazsorban44, line 32: This one is true. Only OAuth 1.0 providers will need this

types/providers.d.ts
```diff
@@ -29,13 +29,13 @@ export interface OAuthConfig<P extends Record<string, unknown> = Profile>
...
   authorizationUrl: string
-  profileUrl: string
-  profile(profile: P, tokens: TokenSet): Awaitable<User & { id: string }>
+  profileUrl?: string
+  profile?(profile: P, tokens: TokenSet): Awaitable<User & { id: string }>
```
balazsorban44, line 35: While technically you are right and the docs does mention that these are optional, not providing them will cause the app to crash. See #1065 and #1846

types/providers.d.ts
```diff
@@ -29,13 +29,13 @@ export interface OAuthConfig<P extends Record<string, unknown> = Profile>
...
   protection?: ProtectionType | ProtectionType[]
   clientId: string
-  clientSecret:
+  clientSecret?:
```
balazsorban44, line 38: I cannot think of a single OAuth provider where this could be optional. Do you have an example? If it says in the docs that this is optional, I think it should be fixed there!

balazsorban44 (changes requested):
Thank you! Added some small comments

types/providers.d.ts
leeoocca: Yes, you're right, it didn't make much sense to me either 😅
But then I just trusted the documentation...

I'm going to reverse this and correct the docs, if that's ok.

types/providers.d.ts
leeoocca: Yes, you're right. It's happened to me as well.

Is it a bug or do you think we should update the docs?
I'm going to revert the changes for now.

types/providers.d.ts
leeoocca: Ok, I'm gonna keep this one 👍🏼

types/providers.d.ts
balazsorban44: it is a bug, - although seems to be a very rare case when you don't need the user data - and I would like to address it in the above mentioned RFC. so for now just leave it as is

www/docs/configuration/providers.md
```diff
@@ -125,7 +125,7 @@ providers: [
 |   accessTokenUrl    |               Endpoint to retrieve an access token               |            `string`             |   Yes    |
 |  authorizationUrl   |         Endpoint to request authorization from the user          |            `string`             |   Yes    |
 |      clientId       |                 Client ID of the OAuth provider                  |            `string`             |   Yes    |
-|    clientSecret     |               Client Secret of the OAuth provider                |            `string`             |    No    |
+|    clientSecret     |               Client Secret of the OAuth provider                |            `string`             |   Yes    |
```
lluia, line 128: @balazsorban44 I see there's a mismatch between what's marked as required in `CommonProviderOptions` and [our documentation](https://next-auth.js.org/configuration/providers#oauth-provider-options) ( not just for the `clientId` field but also `profile`, `params`, etc... ) 🧐

I think it might be because what's a required parameter for an "OAuth" provider might not apply equally for the "Email" or "Credentials" providers 💭

0ubbe (commented):
Thanks 💚 

We need an internal clarification on this before we can merge 😊

www/docs/configuration/providers.md
lluia: I discussed this yesterday with @balazsorban44; basically, this table is out of date, and the source of truth is the types on what options are optional for which type of provider. 💆🏽‍♂️

@leeoocca, if you can revert this change for now and keep the ones for the types, I'll be opening a PR soon to fix this part of the documentation 👍🏽

0ubbe (approved):
Thanks 😊

Issue #1792, opened 2021-04-20
Housekeeping


balazsorban44:
## Issues :ticket: :
  - [x] Add TypeScript issue template
  - [x] Reword current templates
  - [x] Require bug reports to have a reproduction
  - [ ] Check out Beta issue templates
## Automation :robot: :
  - [x] introduce [CODEOWNERS](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners), add @lluia and @balazsorban44 for TypeScript
  - [x] Add Node 16 to workflows (#1790)
  - [x] Remove Node 10 from workflows (#1790)
  - [x] auto label TypeScript PRs with labeler
## Bundle :package: 
  - [ ] check out how to build different bundles (
![image](https://user-images.githubusercontent.com/18369201/115469852-aa5b2b00-a235-11eb-890e-3a4b2f9ada4a.png)
, https://docs.skypack.dev/package-authors/package-checks#esm)
## Docs :books:
  - [x] improve mobile view (
![image](https://user-images.githubusercontent.com/18369201/115469980-e098aa80-a235-11eb-8a77-8f92f9331660.png)
)
 - [x] upgrade Docosaurus to latest (#1826)
 - [ ] Discuss possible restructurings
 - [ ] improve branding (@lluia? )

## Other repos
  - [x] improve/fix/automate https://github.com/nextauthjs/adapters releases
  - [x] gitignore package-lock.json in https://github.com/nextauthjs/next-auth-example

PR #1900, opened 2021-04-30
docs(providers): update providers documentation

0ubbe:
## Reasoning 💡

[We realised](https://github.com/nextauthjs/next-auth/pull/1876/files/6df72b692e3e6427df3967be949b5265b8c76d4f#diff-7a1af8aab4c41d70f23eaf05d5d4fcf5f969a20f515122828f970ab3b008bd69) that the documentation for the different available provider options was out of date.

I took the opportunity to re-work a bit the **Providers** page in our docs to delineate better the 3 ways to sign-in

- `OAuth`
- `email`
- `credentials`

and the steps and options for each. When listing the options for each type of provider, it took [our types](https://github.com/nextauthjs/next-auth/blob/main/types/providers.d.ts) as the source of truth, as agreed with @balazsorban44. I also added emojis 🙈

Screenshot 2021-04-30 at 18 07 23

## Checklist 🧢

- [x] Documentation
- [ ] ~Tests~
- [x] Ready to be merged

## Affected issues 🎟

#1792

### DIFF
--- a/www/docs/configuration/providers.md
+++ b/www/docs/configuration/providers.md
@@ -96,6 +96,25 @@ providers: [
 |       domain        |                Only when using certain Providers                 |             `string`              |    No    |
 |      tenantId       |     Only when using Azure, Active Directory, B2C, FusionAuth     |             `string`              |    No    |
 
+:::tip
+Even if you are using a built-in provider, you can override any of these options to tweak the default configuration.
+
+```js title=[...nextauth].js
+import Providers from "next-auth/providers"
+
+Providers.Auth0({
+  clientId: process.env.CLIENT_ID,
+  clientSecret: process.env.CLIENT_SECRET,
+  domain: process.env.DOMAIN,
+  scope: "openid your_custom_scope", // We do provide a default, but this will override it if defined
+  profile(profile) {
+    return {} // Return the profile in a shape that is different from the built-in one.
+  },
+})
+```
+
+:::
+
 ### Using a custom provider
 
 You can use an OAuth provider that isn't built-in by using a custom object.
--- a/www/docs/configuration/providers.md
+++ b/www/docs/configuration/providers.md
@@ -130,7 +130,7 @@ As an example of what this looks like, this is the provider object returned for 
 }
 ```
 
-You can replace all the options in this JSON object with the ones from your custom provider - be sure to give it a unique ID and specify the correct OAuth version - and add it to the providers option when initializing the library:
+Replace all the options in this JSON object with the ones from your custom provider - be sure to give it a unique ID and specify the correct OAuth version - and add it to the providers option when initializing the library:
 
 ```js title="pages/api/auth/[...nextauth].js"
 import Providers from `next-auth/providers`
--- a/www/docs/configuration/providers.md
+++ b/www/docs/configuration/providers.md
@@ -18,7 +18,7 @@ NextAuth.js is designed to work with any OAuth service, it supports **OAuth 1.0*
 
 ## OAuth Providers
 
-### 📓 Available providers
+### Available providers
 
 <ul>
 {Object.entries(require("../../providers.json"))
@@ -29,7 +29,7 @@ NextAuth.js is designed to work with any OAuth service, it supports **OAuth 1.0*
 )}
 </ul>
 
-### 💭 How to
+### How to
 
 1. Register your application at the developer portal of your provider. There are links above to the developer docs for most supported providers with details on how to register your application.
 
@@ -70,7 +70,7 @@ providers: [
 
 <Image src="/img/signin.png" alt="Signin Screenshot" />
 
-### ⚙️ Options
+### Options
 
 |        Name         |                           Description                            |               Type                | Required |
 | :-----------------: | :--------------------------------------------------------------: | :-------------------------------: | :------: |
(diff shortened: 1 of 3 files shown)


balazsorban44 (changes requested):
https://next-auth-git-docs-1792-providers-params-table-nextauthjs.vercel.app/configuration/providers#-adding-a-new-provider

could you mention adding the type in types/providers to be able to see it in TS as step 3? (or we could figure out how to generate that type out of the src/providers folder, as we generate the index.js file as well)

www/docs/configuration/providers.md
```diff
@@ -3,13 +3,22 @@ id: providers
...
+:::note
+NextAuth.js is designed to work with any OAuth service, it supports **OAuth 1.0**, **1.0A** and **2.0** and has built-in support for most popular sign-in services.
+:::
+
+## OAuth Providers
+
+### Available providers
 
 <ul>
 {Object.entries(require("../../providers.json"))
```
balazsorban44, line 24: This list at
https://next-auth-git-docs-1792-providers-params-table-nextauthjs.vercel.app/configuration/providers#available-providers

is getting pretty long. Maybe instead of `ul>li`, we should just have a comma-separated list of the providers similar to how it is done here: https://next-auth-git-docs-1792-providers-params-table-nextauthjs.vercel.app/faq#what-authentication-services-does-nextauthjs-support

balazsorban44 (changes requested):

www/docs/configuration/providers.md
```diff
@@ -20,42 +29,73 @@ NextAuth.js is designed to work with any OAuth service, it supports OAuth 1.0, 1
...
+### Options
+
+|        Name         |                           Description                            |               Type                | Required |
+| :-----------------: | :--------------------------------------------------------------: | :-------------------------------: | :------: |
+|         id          |                    Unique ID for the provider                    |             `string`              |   Yes    |
+|        name         |                Descriptive name for the provider                 |             `string`              |   Yes    |
+|        type         |              Type of provider, in this case `oauth`              |              `oauth`              |   Yes    |
+|       version       |            OAuth version (e.g. '1.0', '1.0a', '2.0')             |             `string`              |   Yes    |
+|        scope        |          OAuth access scopes (expects array or string)           |      `string` or `string[]`       |   Yes    |
+|       params        |             Additional authorization URL parameters              |     `{ grant_type: string }`      |   Yes    |
```
balazsorban44, line 82: I think the description is wrong on this one (that said, the name is also too generic). I believe this is used to send extra params to the `/token` endpoint when getting an access_token from the OAuth provider. The type is not limited to `grant_type` either.

We do some magic :rainbow: with the params (open the link and scroll, there is more than in this snippet.): https://github.com/nextauthjs/next-auth/blob/dfe4620056b6ba3f72d1dd5ab166efb7f29db5c7/src/server/lib/oauth/client.js#L102-L115

We do some provider-specific things here that shouldn't be in the core, and I hope that #1846 will clear this up because right now it is quite confusing and not flexible enough.

www/docs/configuration/providers.md
```diff
@@ -89,7 +129,8 @@ As an example of what this looks like, this is the provider object returned for
   clientSecret: ""
 }
 ```
-You can replace all the options in this JSON object with the ones from your custom provider - be sure to give it a unique ID and specify the correct OAuth version - and add it to the providers option:
+
+You can replace all the options in this JSON object with the ones from your custom provider - be sure to give it a unique ID and specify the correct OAuth version - and add it to the providers option when initializing the library:
```
balazsorban44, line 133:
```suggestion
Replace all the options in this JSON object with the ones from your custom provider - be sure to give it a unique ID and specify the correct OAuth version - and add it to the providers option when initializing the library:
```

www/docs/configuration/providers.md
```diff
@@ -20,42 +29,73 @@ NextAuth.js is designed to work with any OAuth service, it supports OAuth 1.0, 1
...
+|     protection      | Additional security for OAuth login flows (defaults to `state`)  | `[pkce]`,`[state]`,`[pkce,state]` |    No    |
+|        state        | Same as `protection: "state"`. Being deprecated, use protection. |             `boolean`             |    No    |
+|       headers       |      Any headers that should be sent to the OAuth provider       |             `Object`              |    No    |
+| authorizationParams |    Additional params to be sent to the authorization endpoint    |             `Object`              |    No    |
+|       idToken       |   Set to `true` for services that use ID Tokens (e.g. OpenID)    |             `boolean`             |    No    |
+|       region        |                    Only when using BattleNet                     |             `string`              |    No    |
+|       domain        |                Only when using certain Providers                 |             `string`              |    No    |
+|      tenantId       |     Only when using Azure, Active Directory, B2C, FusionAuth     |             `string`              |    No    |
+
```
balazsorban44, line 98:
```suggestion
:::tip
Even if you are using a built-in provider, you can override any of these options to tweak the default configuration.

```js title=[...nextauth].js
...
Providers.Auth0({
  clientId: process.env.CLIENT_ID,
  clientSecret: process.env.CLIENT_SECRET,
  domain: process.env.DOMAIN,
  scope: "openid your_custom_scope", // We do provide a default, but this will override it if defined
  profile(profile) {
    return {} // Return the profile in a shape that is different from the built-in one.
  }
})
...
:::
```

balazsorban44 (changes requested):

www/docs/configuration/providers.md
lluia: Nice 💚 , this is how it's gonna look:

Screenshot 2021-05-02 at 21 14 31

lluia:
@balazsorban44 addressed all your suggestions 🙇🏽‍♂️

balazsorban44 (approved):
Awesome!

Issue #1910, opened 2021-05-01
Custom signout page url has 'error=undefined' as a querystring param


anubisoft:
## Description 🐜
Any custom page that overrides the default signout page shows an "?error=undefined" querystring param when the errors value is undefined (meaning there are no errors)

http://localhost:3000/auth/signout?error=undefined

Looking at the next-auth codebase, specifically the GET handler for 'signout' doesn't appear to check the error variable before outputting it as a querystring param. The GET signin handler does do this.

balazsorban44:
please fill out the template correctly!

anubisoft:
No error is thrown. The bug is about the signout url having an error querystring param for the case where "error" is undefined. The signin case below has logic that checks "error" but the signout case does not (nor does it append callbackUrl).

https://github.com/nextauthjs/next-auth/blob/main/src/server/index.js

```
 case 'signin':
  if (pages.signIn) {
    let signinUrl = `${pages.signIn}${pages.signIn.includes('?') ? '&' : '?'}callbackUrl=${req.options.callbackUrl}`
    if (error) { signinUrl = `${signinUrl}&error=${error}` }
    return res.redirect(signinUrl)
  }
  return render.signin()
case 'signout':
  if (pages.signOut) {
    return res.redirect(`${pages.signOut}${pages.signOut.includes('?') ? '&' : '?'}error=${error}`)
  }
  return render.signout()
```

anubisoft:
The default signout page handling via "render.signout()" does not have the 'error=undefined' param added.

balazsorban44:
I see. Would you be interested in opening a PR with a fix?

anubisoft:
Yes I'll submit a PR with a proposed fix.

PR #1912, opened 2021-05-02
fix(page): don't pass params to custom signout page

anubisoft:
Modified the handling of two querystring params on the custom signout page url. A conditional check on an "error" value is now made and an error query param is no longer added when 'error' is 'undefined'. Also added a conditional check on the callbackUrl in the options and if one is present a matching query param for that gets added to the signout api call. 

## Reasoning 💡

https://github.com/nextauthjs/next-auth/issues/1910

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged
      
The fix for the "error=undefined" on custom signout page url seems straightforward but I'm not entirely certain that adding the callbackUrl param to the custom signout page url won't have side effects. Because the default signout page handler references the callbackUrl my reasoning was that the custom page should get this param passed to it as well.

## Affected issues 🎟

If you write `"Fixes"` or `"Closes"` before the issue link like so:

```
Fixes #359
```

the connected issue will be automatically closed once the PR is merged and hence help with maintenance of the library 😊

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -159,7 +159,16 @@ async function NextAuthHandler (req, res, userOptions) {
           return render.signin()
         case 'signout':
           if (pages.signOut) {
-            return res.redirect(`${pages.signOut}${pages.signOut.includes('?') ? '&' : '?'}error=${error}`)
+            let signoutUrl = `${pages.signOut}`
+
+            // Append a callback param if one was specified in the request options.
+            if (req.options.callbackUrl)
+              signoutUrl = `${signoutUrl}${pages.signOut.includes('?') ? '&' : '?'}callbackUrl=${req.options.callbackUrl}`
+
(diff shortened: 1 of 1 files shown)


balazsorban44:
Thanks. So why not just remove any parameters? The `callbackUrl` is not something asent to the built-in signout page either as you can see here: https://github.com/nextauthjs/next-auth/blob/main/src/server/pages/signout.js

And as we established before, the `signout` page will never have an `error` parameter either.

anubisoft:
If it's a given that no errors will ever be passed in then the errors param should never be present in the url. As for the callbackUrl param, I would remove that from my proposed fix upon further thought. So I agree the best fix is to have no params in the custom signout page url.

balazsorban44 (approved):

0ubbe (approved):

PR #1914, opened 2021-05-03
fix(ts): tweak Adapter related types

balazsorban44:
## Reasoning 💡

1. The first parameter of an adapter will always be something like a client that is used to communicate with a database, so we can require it. The second parameter will be reserved for any additional configuration for the adapter.

2. some parameters in the `AppOptions` interface were nullable, but in reality, we set a default for almost all of them here:
https://github.com/nextauthjs/next-auth/blob/66473054f5f2f2f117504148d467f42d15aa473e/src/server/index.js#L93-L136

This means that the internal interface for `AppOptions` can be assumed to have much more non-nullable fields and thus playing better with the `Adapter` interface and when it is used to type custom adapters.

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/types/adapters.d.ts
+++ b/types/adapters.d.ts
@@ -118,13 +118,13 @@ export interface AdapterInstance<U = User, P = Profile, S = Session> {
  * [Create a custom adapter](https://next-auth.js.org/tutorials/creating-a-database-adapter)
  */
 export type Adapter<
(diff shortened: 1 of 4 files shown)


0ubbe (approved):
Coolios 📐

PR #1939, opened 2021-05-06
feat(provider): add WorkOS provider

i-palindrome-i:
## Reasoning 💡

Add WorkOS provider.

### What is WorkOS?
WorkOS is a set of building blocks for quickly adding enterprise features to your app (including SSO), for more information: [https://workos.com/](https://workos.com/)

## Checklist 🧢
- [x] Documentation
- [ ] Tests N/A (manually)
- [x] Ready to be merged

### DIFF
--- a/www/docs/providers/workos.md
+++ b/www/docs/providers/workos.md
@@ -18,20 +18,6 @@ providers: [
     clientSecret: process.env.WORKOS_SECRET,
     domain: process.env.WORKOS_DOMAIN
   }),
-  callbacks: {
-    async jwt(token, user, account, profile, isNewUser) {
-      if (profile?.first_name) {
-        // add any WorkOS profile properties you want to access client side here
-        token.firstName = profile.first_name
-      }
-      return token
-    },
-    async session(session, user) {
-      // add any WorkOS profile properties you want to access client side here
-      session.user.firstName = user.firstName
-      return session
-    }
-  },
 ],
 ...
 ```
--- a/www/docs/providers/workos.md
+++ b/www/docs/providers/workos.md
@@ -1,0 +1,37 @@
+---
+id: workos
+title: WorkOS
+---
+
+## Documentation
+
+https://workos.com/docs/sso/guide
+
+## Example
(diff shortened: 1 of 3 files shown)


www/docs/providers/workos.md
```diff
@@ -0,0 +1,37 @@
...
+        token.firstName = profile.first_name
+      }
+      return token
+    },
+    async session(session, user) {
+      // add any WorkOS profile properties you want to access client side here
+      session.user.firstName = user.firstName
+      return session
+    }
+  },
```
balazsorban44, line 34:
```suggestion
```

This is misplaced, should not be inside the providers array, but in any way, I would like the provider examples be specific for the provider. We document the callbacks here already: https://next-auth.js.org/configuration/callbacks

balazsorban44 (changes requested):
Nice PR, but had a small comment. Otherwise looks good!

www/docs/providers/workos.md
i-palindrome-i: Makes sense, updated.

www/docs/providers/workos.md
```diff
@@ -0,0 +1,23 @@
+---
+id: workos
+title: WorkOS
+---
+
+## Documentation
+
+https://workos.com/docs/sso/guide
+
```
lluia, line 9: Following #1955, can you add this section here, so we don't need to come back and update this file once your PR is merged? 😄 
```suggestion
## Options

The **42 School Provider** comes with a set of default options:

- [42 School Provider options](https://github.com/nextauthjs/next-auth/blob/main/src/providers/42.js)

You can override any of the options to suit your own use case.
```

0ubbe (changes requested):
Nice ❤️ , one small suggestion 👍🏽

www/docs/providers/workos.md
i-palindrome-i: Done!

balazsorban44 (approved):

PR #1949, opened 2021-05-06
docs: max cookie size information

### LINKED ISSUE #1746: Clarify cookie size limitations
Explain that the size of a single cookie in most browsers is 4 kilobytes. 

In our current FAQ page, we falsely claim that all the cookies combined on a single domain cannot reach 4096 bytes.
https://next-auth.js.org/faq#what-are-the-disadvantages-of-json-web-tokens

This is not true. 4096 bytes is per cookie, and different browsers support different number of cookies. 

Here is a source: 
http://browsercookielimits.iain.guru/ (not the most up-to-date either, but better than the one we used before.)

This is up for grab to anynone! 

The relevant file is https://github.com/nextauthjs/next-auth/blob/main/www/docs/faq.md

ndom91:
## Reasoning 💡

Fix incorrect information about max cookie size.

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged
      

## Affected issues 🎟

Closes #1746


balazsorban44 (changes requested):
Thanks! Found another place where it needs clarification

https://next-auth.js.org/configuration/callbacks#jwt-callback

www/docs/faq.md
```diff
@@ -196,9 +196,9 @@ JSON Web Tokens can be used for session tokens, but are also used for lots of ot
...
 
-* As with database session tokens, JSON Web Tokens are limited in the amount of data you can store in them. There is typically a limit of around 4096 bytes in total for all cookies on a domain, though the exact limit varies between browsers, proxies and hosting services.
+* As with database session tokens, JSON Web Tokens are limited in the amount of data you can store in them. There is typically a limit of around 4096 bytes in total for a single domain's cookies,  though the exact limit varies between browsers, proxies and hosting services. If you want to support most browsers, then do not exceed 50 cookies per domain and do not exceed 4096 bytes per domain. (Source: [browsercookielimits.iain.guru](http://browsercookielimits.iain.guru/))
```
balazsorban44, line 199: 4096 bytes per COOKIE, at least how I understand it. the important take here is that our session cookie can store up to 4096 bytes

balazsorban44 (changes requested):

www/docs/faq.md
ndom91: The `browsercookielimits` page says the following:

>If you want to support most browsers, then don't exceed 50 cookies per domain, and don't exceed 4093 bytes per domain (i.e. total size of all cookies <= 4093 bytes)

ndom91:
> Thanks! Found another place where it needs clarification
> 
> https://next-auth.js.org/configuration/callbacks#jwt-callback

This looks alright as well, if we're agreeing on the 4096 bytes per domain tip.

www/docs/faq.md
ndom91: Ah alright awesome, good eye! 

I'll adjust the text then shortly

balazsorban44 (changes requested):

ndom91:
Yup makes sense! Sorry it took me so long, just got to this now. I think I got all the places you'd mentioned.

www/docs/faq.md
```diff
@@ -196,9 +196,9 @@ JSON Web Tokens can be used for session tokens, but are also used for lots of ot
...
+* As with database session tokens, JSON Web Tokens are limited in the amount of data you can store in them. There is typically a limit of around 4096 bytes per cookie, though the exact limit varies between browsers, proxies and hosting services. If you want to support most browsers, then do not exceed 4096 bytes per cookie. If you want to save more data, you will need to persist your sessions in a database (Source: [browsercookielimits.iain.guru](http://browsercookielimits.iain.guru/))
 
-  The more data you try to store in a token and the more other cookies you set, the closer you will come to this limit. If you wish to store more than ~2 KB of data you probably at the point where you need to store a unique ID in the token and persist the data elsewhere (e.g. in a server side key/value store).
+  The more data you try to store in a token and the more other cookies you set, the closer you will come to this limit. If you wish to store more than ~2 KB of data you're probably at the point where you need to store a unique ID in the token and persist the data elsewhere (e.g. in a server side key/value store).
```
balazsorban44, line 201: Should be saying `~4 KB` here as well, I guess. 😁

balazsorban44 (changes requested):

balazsorban44 (approved):

www/docs/faq.md
ndom91: I think this is meaning 2kb of _additional_ user data. i.e. we're using ~2kb of data, so the user shouldn't put more than another 2kb in. Right?

balazsorban44: We don't have anything else than what the user puts there. At work, we store around 3.8kb in the session cookie.

PR #1950, opened 2021-05-06
docs(provider): clarify where user is created in db with email provider

### LINKED ISSUE #1313: Email Provider authentication does not create user account in account table in database after successful authentication
**Describe the bug**

Email Provider authentication does not create user account in `account` table in database after successful authentication (only if new user)

Do not report bugs with your own project here, ask from help by raising a question instead - this helps us a lot with administration overhead.

**Steps to reproduce**
1. Sign in using Email
2. Verify email will be sent to specified email
3. Go to email and click on the verify your email link
4. You should be logged in
5. Check DB. The documentation `https://next-auth.js.org/providers/email` says 

> An account will not be created for the user until the first time they verify their email address. If an email address already associated with an account, the user will be signed in to that account when they use the link in the email.

So I would expect to see an account entry in the database table account once successfully logged in.

Here is an example of the issue. 

Here is an example with the bug `https://github.com/JunayedMizan/next-auth-bug`

**Expected behavior**
I would expect to see an account entry in the database table `account` once successfully logged in.

ndom91:
## Reasoning 💡

Clarifying information about where exactly the User is being created in the DB once they verify their email based account.

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged
      

## Affected issues 🎟

Closes #1313

### DIFF
--- a/www/docs/providers/email.md
+++ b/www/docs/providers/email.md
@@ -78,7 +78,7 @@ The Email Provider can be used with both JSON Web Tokens and database sessions, 
   ```
 3. You can now sign in with an email address at `/api/auth/signin`.
 
-  An account will not be created for the user until the first time they verify their email address. If an email address already associated with an account, the user will be signed in to that account when they use the link in the email.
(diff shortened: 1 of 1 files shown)


balazsorban44 (approved):
LGTM 👍

PR #1955, opened 2021-05-08
docs(providers): link to providers' source code

### LINKED ISSUE #1942: Add link to provider source code
Provider pages under `https://next-auth.js.org/providers/provider` should have a subsection with a link to their source code on GitHub `https://github.com/nextauthjs/next-auth/blob/main/src/providers/provider-name.js`. We should also mention in those sections, that you will only have to override values that are different from the default config.

Still seeing use cases where people provide more configuration than they need.

0ubbe:
## Reasoning 💡

Captura de pantalla 2021-05-08 a las 19 03 43

This PR adds a new section on each of the **Provider pages** in our documentation linking to the Provider's source code and stating that users can override the options to suit their use case. 

## Checklist 🧢

- [x] Documentation
- [ ] ~Tests~
- [ ] ~Ready to be merged~

## Affected issues 🎟

Fixes #1942

### DIFF
--- a/www/docs/providers/42.md
+++ b/www/docs/providers/42.md
@@ -11,6 +11,14 @@ https://api.intra.42.fr/apidoc/guides/web_application_flow
 
 https://profile.intra.42.fr/oauth/applications/new
 
+## Options
+
+The **42 School Provider** comes with a set of default options:
+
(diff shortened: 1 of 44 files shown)


balazsorban44 (approved):
Awesome! Thanks!

PR #1826, opened 2021-04-23
chore: Test Docusaurus webpack 5 support

slorber:
Use an unreleased version of Docusaurus to test the webpack5 support works for this site.

See also https://github.com/facebook/docusaurus/pull/4089

PR #1989, opened 2021-05-12
Revert "docs: revert "Docusaurus webpack 5""

balazsorban44:
Reverts nextauthjs/next-auth#1982

Having another go at this, it was suggested (https://github.com/nextauthjs/next-auth/pull/1826#issuecomment-839869469) that the problem might gave been some initial CDN caching issues, as we (as far as I know) don't have any special cache config


balazsorban44:
Preview URL works fine at https://next-auth-git-revert-1982-revert-1826-slorber-docu-nextauthjs.vercel.app/

PR #1993, opened 2021-05-13
docs(www): upgrade docusaurus + docs search style

ndom91:
## Reasoning 💡

Upgraded docusaurus to their beta version of 2.0.0 which dropped recently.

## Checklist 🧢

- [x] Documentation
- ~[ ] Tests~
- [x] Ready to be merged
      

## Screenshots 📷

Before: 
image

After:
image

## Changes 

0. Updated `@docusaurus/core` and `@docusaurus/preset-classic`

1. Added the `/` character as a shortcut to focus the search `input` + added icon

2. Also changed the primary bg color from `#000` to something a littleeee bit lighter (`#090909`). That `#000` is pretty damn hard on the eyes - don't you think? I like its look on the homepage / hero section, etc. but the actual documentation part, reading long text, I found it a bit difficult. 

3. (No change yet) I noticed we were using a pretty old version of that Algolia DocSearch, so I applied on their website (its free for opensource projects) for their newer DocSearch implementation. You've probably seen / used it before. Ex: https://relay.dev/ and https://docsearch.algolia.com/.  What do you think, should we upgrade if we get in?

### DIFF
--- a/www/package-lock.json
+++ b/www/package-lock.json
@@ -1,8 +1,15170 @@
 {
   "name": "next-auth-docs",
   "version": "0.1.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "next-auth-docs",
+      "version": "0.1.1",
+      "dependencies": {
+        "@docusaurus/core": "2.0.0-beta.0",
+        "@docusaurus/preset-classic": "2.0.0-beta.0",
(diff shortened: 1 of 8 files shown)


balazsorban44:
Awesome! Thanks! Yeah, upgrade all the things!!!! I remember I went through the same process with Algolia to get Search working for the Hungarian React docs at https://hu.reactjs.org/

balazsorban44 (approved):

PR #2009, opened 2021-05-16
fix(deps): pin down legacy adapter versions

### LINKED ISSUE #2008: Non-Semver dependency on legacy-adapter packages caused upgrade headache
## Description 🐜 and Cause

> Edit: initially, this issue was about the documentation regarding the use of legacy-adapters because I believed the prisma legacy adapter was broken. Now that Balasz pointed me in the right direction for debugging it, I've updated it to reflect the actual issue.

Upgrading across a boundary where the dependency on the legacy adapter packages was also upgraded can cause build failures for end-developers. An example of this occurred for people upgrading from 3.20.0 to 3.20.1 who depended on the legacy prisma adapter. Here's how:

This `next-auth` package depends on `"@next-auth/prisma-legacy-adapter": "canary"` however between the two versions listed above, the package.lock updated _which_ canary from `canary-115` to `canary-127`. Because the dependency in the package.json just states "canary", though, downstream yarn consumers who upgraded across that boundary encountered a failure. Their dependency graph appeared to satisfy the requirements (because "canary" was installed), but the versions of canary had slightly different APIs and so were incompatible.

A clean install of 3.21.1 will succeed in all situations, but users who upgrade got a broken experience unless they manually deleted that one dependency from their yarn.lock (or deleted the lock file entirely).

The proposed solution to this is to move the `@next-auth/*-adapter` packages over to using a proper semver rather than the tag `canary`. That way, upgrades to the packages would properly flow to us end-developers.

## How to reproduce ☕️

See [this comment](https://github.com/nextauthjs/next-auth/issues/2008#issuecomment-841797746) for investigation into how to cause this issue.

## Screenshots / Logs 📽

Calls to the [...nextauth].ts endpoint produce the following error:

```
TypeError: next_auth_adapters__WEBPACK_IMPORTED_MODULE_2___default(...).Prisma.Adapter is not a function
```

## Environment 🖥

```
System:
    OS: macOS 11.3.1
    CPU: (8) x64 Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
    Memory: 706.36 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Yarn: 1.22.10 - ~/code/traveler/tweetsweep/node_modules/.bin/yarn
    npm: 7.9.0 - ~/.nvm/versions/node/v14.16.1/bin/npm
  Browsers:
    Chrome: 90.0.4430.212
    Firefox: 88.0
    Safari: 14.1
```

balazsorban44:
## Reasoning 💡

See #2008 for more info

## Affected issues 🎟

fixes #2008

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,8 +62,8 @@
   "license": "ISC",
   "dependencies": {
     "@babel/runtime": "^7.14.0",
-    "@next-auth/prisma-legacy-adapter": "canary",
-    "@next-auth/typeorm-legacy-adapter": "canary",
+    "@next-auth/prisma-legacy-adapter": "0.0.1-canary.127",
+    "@next-auth/typeorm-legacy-adapter": "0.0.2-canary.129",
     "futoin-hkdf": "^1.3.2",
     "jose": "^1.27.2",
     "jsonwebtoken": "^8.5.1",


balazsorban44:
@zackdotcomputer have a look, do you think this would help?

zackdotcomputer (approved):
Yup - I think that this will fix it too! I'm surprised it didn't cause any changes to package-lock.json too, did you run `npm i`?

0ubbe (approved):

balazsorban44:
I ran it, it was probably already saved as the newest canary versions.

Issue #1210, opened 2021-01-29
useSession({ required: true }), which always returns a non-null Session object


kripod:
**Summary of proposed feature**

A living session could be a requirement for specific pages. If it doesn’t exist, then the user should be redirected to the sign in page with an error like "Session expired, please try signing in again".

**Purpose of proposed feature**

Sometimes, a user might log out by accident, or by deleting cookies on purpose. If that happens (e.g. on a separate tab), then `useSession({ required: true })` should detect the absence of a session cookie and always return a non-nullable `Session` object type.

**Detail about proposed feature**

If the `required` option is specified, then an effect should be registered to redirect the user to the sign in page as soon as no session is available.

**Potential problems**

The session object could not only become nonexistent, but might even change over time. That edge case should be handled separately.

**Describe any alternatives you've considered**

Creating a hook in userland, e.g.:

```ts
function useSessionRequired() {
  const [session, loading] = useSession();
  const router = useRouter();

  React.useEffect(() => {
    if (!session && !loading) {
      router.push(`${process.env.NEXTAUTH_URL}/auth/sign-in?error=SessionExpired`);
    }
  }, [loading, router, session]);

  return [session, loading];
}
```

**Additional context**

As noticed in #1081, NextAuth.js already listens to page visibility changes. Session emptiness checks should be done each time the page becomes visible after hiding it.

balazsorban44:
So I think I found a very neat solution/pattern for this, let me know what you think!

```jsx
// pages/admin.jsx
export default function AdminDashboard () {
  const [session] = useSession() 
  // session is always defined here
  return "Some super secret dashboard"
}

AdminDashboard.auth = true
```
```jsx
//pages/_app.jsx
export default function App({ Component, pageProps }) {
  return (
    <SessionProvider session={pageProps.session}>
      {Component.auth
        ? <Auth><Component {...pageProps} /></Auth>
        : <Component {...pageProps} />
      }
    </SessionProvider>
  )
}

function Auth({ children }) {
  const [session, loading] = useSession()
  const isUser = !!session?.user
  React.useEffect(() => {
    if (loading) return // Do nothing while loading
    if (!isUser) signIn() // If not authenticated, force log in
  }, [isUser, loading])

  if (isUser) {
    return children
  }
  
  // Session is being fetched, or no user.
  // If no user, useEffect() will redirect.
  return <div>Loading...</div>
}
```

It can be easily be extended to support something like an options object:

```jsx
[...]

PR #2016, opened 2021-05-17
docs: explain where pageProps come from in Provider docs

### LINKED ISSUE #1132: Explain how to populate Provider in _app

ndom91:
## Reasoning 💡

Further explain how the `pageProps` get into the page component and how the session prop actually gets passed through, i.e. through the use of Next.js `getInitialProps` and our `getSession` function, for example. 

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged
      

## Affected issues 🎟

Closes #1132


www/docs/getting-started/client.md
```diff
@@ -328,6 +328,24 @@ export default function App ({ Component, pageProps }) {
...
+export async function getServerSideProps(ctx) {
+  return {
+    props: {
+      session: await getSession(ctx)
+    }
+  }
+}
+```
+
+If every one of your pages needs to be protected, you can do this in `_app`, otherwise you can do it on a page-by-page basis.
```
balazsorban44, line 347: Don't know if it's a true alternative, but the method described here https://github.com/nextauthjs/next-auth/issues/1210#issuecomment-782630909 makes it possible to do per page auth much easier without gSSP, and you can even show a fallback ui or extend it with roles or anything really.

gSSP in _app always felt very harsh to me, but I can see it's easier. What do you think?

ndom91: So the method you've linked looks really easy to implement with some nice conveniences tacked on, but I have to admit im not familiar enough with our internal session handling currently to fully appreciate the difference haha

I can definitely mention it in the docs though :+1:

balazsorban44: So it boils down to that if you use gSSP on the _app component, it means EVERY page transition will ALWAYS be blocking (meaning SSR), versus my suggestion, first you show a skeleton while loading/checking auth, and every page transition will be client-side afterwards.

ndom91:
@balazsorban44 I've added an "Alternatives" section to the `getting-started/client.md` and referenced (1) your aforementioned custom client side auth check and (2) your `nextauthjs/react-query` library. 

I hope I got all the details right :sweat_smile: haha

PR #2023, opened 2021-05-18
docs(search): add new algolia docsearch

ndom91:
## Reasoning 💡

Added new Algolia DocSearch :mag_right: :tada: 

## Checklist 🧢

- [x] Documentation
- ~[ ] Tests~
- [x] Ready to be merged
      

Algolia responded with our api key! They assigned the "app" to my preexisting Algolia account, so if you want access to the analytics (thats basically the only thing in the dashboard) I can invite you to the team!

Otherwise, here's some other important stuff regarding this new search:

- we can (and I did in this PR) remove the entire custom theme / search subdirectory under `www`
- algolias config for our docs page is available on their github and we can make changes to it simply via PR (https://github.com/algolia/docsearch-configs/blob/master/configs/next-auth.json)
- if we want to style the search a bit differently, theres some more docs here: https://docsearch.algolia.com/docs/styling/
- the docsearch stuff is automatically included in the docusaurus classic-preset, including a light/dark version. So no work needs to be done in that regard.
- their crawler goes over our site 1x every 24hrs, so it'll pick up new stuff for the search index at the latest 24hrs later

And I think thats it! How do you like it?

I think it works super well, is easily navigable via keyboard, did I mention its fast? Haha I like it a whole lot. It'll save users recent searches as well, and allow you to "favorite" search terms in their modal.


balazsorban44:
There seems to be an issue in mobile. The search bar is behind the navbar for me.

ndom91:
> 
> 
> There seems to be an issue in mobile view. The search bar is behind the navbar for me.
> 
> Other than that, it is just awesome!

Ah good eye! Thats no good. Let me see if we can't adjust that!

balazsorban44 (approved):

PR #2025, opened 2021-05-18
docs: update github PR template comments

ndom91:
## Reasoning 💡

@balazsorban44 what do you think about making more of the PR template comments?

I've kept all the same instructions as before, but made them almost all non-visible by default so that none of our "boilerplate" language is visible in the actual PR. I feel like I'm always commenting them out / deleting them anyway, this would save a decent amount of time / effort I feel like. Whats your opinion here?

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged


balazsorban44 (approved):
Agree 👍

PR #2038, opened 2021-05-20
fix(provider): make WorkOS domain configurable from signIn

maxdeviant:
## Reasoning 💡

This PR fixes some issues with the implementation of the WorkOS provider.

The `domain` parameter in the authorization URL needs to be set dynamically based on the user signing in. Previously the domain was being passed in as part of the static WorkOS provider configuration.

The `domain` can now be passed in using the `authorizationParams` on the `signIn` function, and the WorkOS provider docs have been updated to reflect this.

## Checklist 🧢

- [x] Documentation
- [ ] ~Tests~
- [x] Ready to be merged

Happy to answer any additional questions on the intended WorkOS provider usage if anyone has them.

### DIFF
--- a/src/providers/workos.js
+++ b/src/providers/workos.js
@@ -1,4 +1,6 @@
 export default function WorkOS(options) {
+  const apiUrl = options.apiUrl || 'api.workos.com';
+
   return {
     id: 'workos',
     name: 'WorkOS',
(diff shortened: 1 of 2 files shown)


src/providers/workos.js
```diff
@@ -1,4 +1,6 @@
 export default function WorkOS(options) {
+  const apiUrl = options.apiUrl || 'api.workos.com';
```
lluia, line 2: Could we call the new option `domain`? We already have it [reflected on the types](https://github.com/nextauthjs/next-auth/blob/main/types/providers.d.ts#L51) for other providers; once we figure out the way to express this via TS it's then easy for us to refactor.

0ubbe (changes requested):
Thanks 💚 ; one small request, otherwise great 💯

src/providers/workos.js
maxdeviant: Done!

balazsorban44 (approved):

PR #2046, opened 2021-05-22
Call user() to get user details to check in condition

kiranvj:
We were using `if(user)` which will be always true as `user` is a function. We need to call user() and use the return value in if condition

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged


balazsorban44 (changes requested):
Thanks! So if we want to clean this up a bit, I would suggest something like this:

```js
async authorize(credentials, req) {
  // You need to provide your own logic here that takes the credentials
  // submitted and returns either a object representing a user or value
  // that is false/null if the credentials are invalid.
  // e.g. return { id: 1, name: 'J Smith', email: 'jsmith@example.com' }
  // You can also use the `req` object to obtain additional parameters
  // (i.e., the request IP address) 
  const res = await fetch("/your/endpoint", {
    method: 'POST',
    body: JSON.stringify(credentials),
    headers: { "Content-Type": "application/json" }
  })
  const user = await res.json()
  
  // If no error and we have user data, return it
  if (res.ok && user) {
    return user
  }
  // Return null if user data could not be retrieved
  return null
}
```

I feel that this is a much more usual pattern. What do you think?

kiranvj:
> I feel that this is a much more usual pattern. What do you think?

I totally agree with you. The proposed pattern is more usual in the project.  I have update the code as suggested and pushed. Kindly check.

balazsorban44 (approved):

balazsorban44:
Glad you agreed! Thanks for the PR

PR #2062, opened 2021-05-25
feat: token endpoint

balazsorban44:
**EXPERIMENTAL**

## Reasoning 💡

A new take on #951 

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -198,6 +198,12 @@ async function NextAuthHandler(req, res, userOptions) {
             return res.redirect(pages.verifyRequest)
           }
           return render.verifyRequest()
+        case "token": {
+          if (provider) {
+            return routes.token(req, res)
+          }
+          break
+        }
         case "error":
           if (pages.error) {
             return res.redirect(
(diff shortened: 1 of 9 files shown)


balazsorban44:
OK, so the main issue here is if an `access_token` expires and let's say we start the token renewal by contacting the provider's /token endpoint, how do we make sure that subsequent requests under the renewal period does not cause a problem? It is especially hard, if the `refresh_token` is sliding (only usable once.).

My idea is that we need some kind of caching in the front-end that de-duplicates requests to this new `/api/auth/token` endpoint and rather subscribes to the result.

balazsorban44:
Sorry for those who are eagerly waiting for this, and only see me keep closing it... My hunch about **this PR** fixing some problems wasn't right, so I have to close this again, as I prioritize other work for now. Could be an addition later in `next-auth@4.x`. This is a new endpoint, so hopefully, it can be added in a non-breaking manner.

PR #2072, opened 2021-05-26
Disable broadcasting events

### LINKED ISSUE #2071: Disable broadcasting as it creates issues with refresh token rotation
## Description 🐜

Took me a bit of time but I think I finally understood why we were seeing so many errors related to refreshing the token in our app.

I'm following the suggested setup to refresh tokens https://next-auth.js.org/tutorials/refresh-token-rotation which works fine.

The issue is the broadcasting that if you have more than 1 tab open with your web app, it will trigger a session check for EVERY tab and then you end up triggering the `refreshAccessToken()` function multiple times simultaneously with the same set of parameters (same `refresh_token`)

This creates an issue because the first request will be successful and obtain a new `access_token` paired with a `refresh_token` while all the other calls that were made will get a 400 since the old `refresh_token` has just been invalidated by the first successful call.

Simplest way to fix this would be to have the option to disable broadcasting and let each tab try to refresh the session only when focused.

## How to reproduce ☕️

In the app try to set a `clientMaxAge` of 10 seconds
```js
      <AuthProvider
        session={pageProps.session}
        options={{
          clientMaxAge: 10,
        }}
      >
```
Then add a log in `[...nextauth].js` `jwt` callback
```js
    async jwt(token, user, account, profile) {
      console.log('JWT')
```

Open 2 tabs and move between them, you will notice that every 10 seconds there are 2 calls being made to refresh the session and 2 console logs
```
JWT
JWT
```

I can create a codesandbox later after I finish work if needed but it's pretty simple to verify I believe 😄 

I'm using `"next-auth": "3.23.3"`

mtt87:
Added an option to disable broadcasting events.

Just showing how it could possibly be done, maintainers feel free to take this as inspiration and add it yourself if needed.

I tested manually locally and it seems to be fine but I noticed my prettier on VSCode formatted a lot of things differently not sure why, quite annoying.

I saw there is prettier added as a devDependency but in fact there is no npm script to run prettier, I would suggest to add it so it's easier for people to run it and have the same output 😄 

## Reasoning 💡

https://github.com/nextauthjs/next-auth/issues/2071

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## Affected issues 🎟

Fixes #2071


balazsorban44:
Regarding prettier, I think most of us just have it as a VSCode extension, and the `formtOnSave` option is set to `true`. I Will have a look at the rest later on, thank you for this PR!

www/docs/tutorials/refresh-token-rotation.md
```diff
@@ -91,32 +91,53 @@ export default NextAuth({
...
   },
-});
+})
 ```
 
 ### Client Side
 
+Make sure to disable broadcasting the session event or you will end up in a situatin where if you have your app open on multiple tabs they will all try to refresh the token at the same time causing issues.
+
+You can do it by setting `disableSessionEventBroadcast` to `false` when adding the `Provider`.
```
balazsorban44, line 122:
```suggestion
You can do it by setting `disableSessionEventBroadcast` to `true` when adding the `Provider`.
```

balazsorban44:
I don't know yet if I like the naming, and also if we need to differentiate between the broadcast events. Maybe

```ts
syncAcrossTabs: boolean
```

would be enough/better?

mtt87:
I think there is value in being able to disable only a specific event, I've explained here https://github.com/nextauthjs/next-auth/issues/2071#issuecomment-851643572 
but if it's easier for you to start with disabling all events, then sure why not 😄 

Naming is hard I agree but `syncAcrossTabs` seems quite generic doesn't really tell the user what's going on, perhaps `syncSessionAcrossTabs` more precise and informative?

balazsorban44:
Liked the explanation, and that way I can accept the current names. Will discuss this with some of the other maintainers, and come back to this, hopefully soon! Thank you for the meaningful conversations so far!

mtt87:
You are welcome 😄  thanks for reading my long explanations ahahah, keep me updated and if you need help I'm happy to contribute wherever I can.

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "next-auth",
-  "version": "0.0.0-semantically-released",
+  "name": "@mtt/next-auth",
+  "version": "0.0.0",
```
balazsorban44, line 3:
```suggestion
  "name": "next-auth",
  "version": "0.0.0-semantically-released",
```

balazsorban44 (commented):
Just an update, I have discussed this briefly with @lluia while reviewing #1473, and we are thinking that it would make more sense if we waited with this change until the next major version (which is hopefully not that far away, aiming for this summer). I did a bunch of changes in #1473 which would make this a much nicer implementation, and it would be easier to test it as well.

balazsorban44:
Closing this in favor of #2171. Feel free to add feedback on that PR!

PR #2093, opened 2021-05-29
Signin: Allow client to override redirect_uri

benoror:
## Reasoning 💡

Allow client to override `redirect_uri` via query params, so getting a new separate set of tokens (potentially for different scopes) for an account is possible

## Checklist 🧢

- [x] Documentation
- [ ] Tests (*couldn't find relevant tests for this override)
- [x] Ready to be merged

## Affected issues 🎟

Discussion with context here: https://github.com/nextauthjs/next-auth/discussions/2068#discussioncomment-801705


balazsorban44:
Let's discuss this further first: https://github.com/nextauthjs/next-auth/discussions/2068#discussioncomment-801780

happy to reopen if we find it a valid case.

PR #2103, opened 2021-05-31
feat: simplify verification token flow

balazsorban44:
## Reasoning 💡
To utilize some of these changes, an effort from the user is required, but everything here is implemented in a backwards compatible way, so nothing needs to change.

1. Making `deleteVerificationRequest` optional.
When a user completes their email signin flow, a series of adapter calls will read (`getVerificationRequest`) and then delete `deleteVerificationRequest` the token from the database. These read/delete operations can be combined to make one less call to the db. For now, to utilize this, your adapter should return an object that has a `{token_deleted: true}` property. This is required for backwards compatibility. In the future, (in a major version bump), we can remove the deleteVerificationRequest method and require any adapters upgrading to delete the token in the `getVerificationRequest` method instead.
2. Sending a hashed token to the adapters. Instead of requiring the adapters to provide a secure hashing algorithm on tokens, we can do it at the core level, so we have a uniform/secure token for all adapters, taking away that responsibility. The methods in question now receive a new argument (see the file changes). In a future major release, we can drop some of the redundant params sent to the adapter methods.

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged


balazsorban44:
cc: @kripod it touches on our previous discussion where we should aim to reduce the number of DB calls, if possible.

src/server/routes/signin.js
```diff
@@ -21,15 +21,15 @@ export default async function signin(req, res) {
...
+  if (provider.type === "oauth") {
     try {
       const authorizationUrl = await getAuthorizationUrl(req)
       return res.redirect(authorizationUrl)
     } catch (error) {
       logger.error("SIGNIN_OAUTH_ERROR", error)
       return res.redirect(`${baseUrl}${basePath}/error?error=OAuthSignin`)
     }
-  } else if (provider.type === "email" && req.method === "POST") {
+  } else if (provider.type === "email") {
```
kripod, line 32: I think these changes may introduce a vulnerability by letting the user visit URLs directly from within their browsers. POST requests mitigate this issue.

src/server/routes/signin.js
```diff
@@ -73,17 +73,18 @@ export default async function signin(req, res) {
...
-      `${baseUrl}${basePath}/verify-request?provider=${encodeURIComponent(
-        provider.id
-      )}&type=${encodeURIComponent(provider.type)}`
-    )
+    const params = new URLSearchParams({
+      provider: provider.id,
+      type: provider.type,
+    })
+    const url = `${baseUrl}${basePath}/verify-request?${params}`
+    return res.redirect(url)
```
kripod, line 87: This is a nice one 🙌

kripod (changes requested):

src/server/routes/signin.js
balazsorban44: So the reason I think it's not a problem is that we already check if the req is a `POST` here:

https://github.com/nextauthjs/next-auth/blob/613c3033156652b1a93ba3376f99c7ee3267cdb2/src/server/index.js#L230-L237

This is similar to https://github.com/nextauthjs/next-auth/blob/main/src/server/routes/callback.js which will only run with `POST` but doesn't do this check either.

In the long term, the less references we have to `req` outside of `src/server/index.js`, the easier it will be to lift out these methods and make them pure, ultimately contributing to functionality like fetch-less `getSession`, etc.

kripod: Oh, that’s right, thanks for the explanation! This looks good to me, then 😊

src/server/routes/signin.js
```diff
@@ -44,7 +44,7 @@ export default async function signin(req, res) {
     // according to RFC 2821, but in practice this causes more problems than
     // it solves. We treat email addresses as all lower case. If anyone
     // complains about this we can make strict RFC 2821 compliance an option.
-    const email = req.body.email?.toLowerCase() ?? null
+    const email = req.body.email.toLowerCase()
```
lluia, line 47: There's an `await` below on **line 49** that is not part of a `try {} catch {}`. Maybe it makes sense to extend the `try {} catch {}` below to also cover these two lines:
```ts
// If is an existing user return a user object (otherwise use placeholder)
const profile = (await getUserByEmail(email)) || { email }
const account = { id: provider.id, type: "email", providerAccountId: email }
```
In this way, if there's an error here, we can do the redirect as well rather than let the error just bubble up 👍🏽

0ubbe (approved):
Very neat 💯 , just a nitpick 💭 about existing code 💅🏽

src/server/routes/signin.js
balazsorban44: This whole email signin function is encapsulated in a try..catch at a higher level, so I *think* it's unnecessary:

https://github.com/nextauthjs/next-auth/blob/1fb308a6f4c5d6b01cd3ba9f104264d59a270e4f/src/server/routes/signin.js#L75-L80

balazsorban44:
Decided to wait with this until v4. Hopefully, it is not that far away, then we don't require yet another thing from users to migrate. We will need a stable adapter release first that is compatible with v3, so we can start migrating the adapters for v4 in a pre-release.

PR #2108, opened 2021-06-01
Fix: Convert Github profile.id from int to string

tom-richter:
## Reasoning 💡

This Github profile.id should be converted from integer to string as described here: [https://github.com/nextauthjs/adapters/issues/111](https://github.com/nextauthjs/adapters/issues/111). We should check if other adapters/functions still rely on profile.id being an integer.

BREAKING CHANGE:

For consistency around our built-in providers, the profile id is now converted to a string. If you had assumptions about the id being a number, you can do `parseInt(profile.id, 10)`

### DIFF
--- a/src/providers/github.js
+++ b/src/providers/github.js
@@ -10,7 +10,7 @@ export default function GitHub(options) {
     profileUrl: "https://api.github.com/user",
     profile(profile) {
       return {
-        id: profile.id,
+        id: profile.id.toString(),
         name: profile.name || profile.login,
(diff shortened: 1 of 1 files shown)


balazsorban44:
Yeah.. :confused: I am a bit afraid that it might cause problems for people who currently expect this to be an integer actually. :fearful: We might just want to hold onto this PR until v4.0 and you can use your workaround locally until then?

If it only were TS, I wouldn't mind, but here we actually change the code.

PR #2117, opened 2021-06-02
Add `pretty-quick` ⚡️ 

0ubbe:
## Reasoning 💡

This project didn't have `prettier` initially when it was created; in the long run, this has caused some noise in the PRs 🎧

Screenshot 2021-06-02 at 13 46 43

We recently added `prettier` to the project and updated our `eslint` config. This PR builds upon that, making sure we format **all staged files** before a commit. In this way, we prevent people from committing files without having been formatted first. Documentation links:

- [`pretty-quick`](https://github.com/azz/pretty-quick)
- [`husky`](https://github.com/typicode/husky)

## Checklist 🧢

- [ ] ~~Documentation~~
- [ ] ~~Tests~~
- [x] Ready to be merged

## Affected issues 🎟

None.


ndom91 (approved):
Nice, big fan of all of the above! Haha

I think I read Balazs mentioned in another issue recently, that we'd just relied on people to have the prettier vscode plugin, for example, instead of forcing the formatting upon them via something like husky+pretty-quick.

So forcing this upon PRs should reduce the noise a lot!

src/client/index.js
```diff
@@ -211,10 +211,8 @@ export async function signIn(provider, options = {}, authorizationParams = {}) {
       callbackUrl,
       json: true,
     }),
-  }
+  })
```
lluia, line 214: This change was to check the new tooling worked, happy to revert it if you think before it read better 👍🏽

PR #2130, opened 2021-06-04
Fix husky 🐶

0ubbe:
## Reasoning 💡

Following the addition of [`husky`](https://typicode.github.io/husky/#/) #2117, I committed a `pre-commit` file with that a line that tries to require a file that doesn't exist, and hence makes the pre-commit check fail...

## Checklist 🧢

- [ ] ~~Documentation~~
- [ ] ~~Tests~~
- [ ] ~~Ready to be merged~~

## Affected issues 🎟

None

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,1 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
 npx pretty-quick --staged
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
 npx pretty-quick --staged
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "./errors": "./dist/lib/errors.js"
   },
   "scripts": {
+    "postinstall": "npx husky install",
     "build": "npm run build:js && npm run build:css",
     "build:js": "node ./config/build.js && babel --config-file ./config/babel.config.js src --out-dir dist",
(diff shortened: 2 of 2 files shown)


ndom91:
Why did you remove that? 

Husky 6.0 changed the install steps a bit. You have to run "npx husky install" after cloning the repo / the first time. Then it'll create the necessary scripts, etc. 

Im pretty sure it's alright the way it is

lluia:
@ndom91 interesting, I still worked for me by removing that line (maybe some cache?!) but you're right, now:
```bash
$ npx husky install
```
is an extra step when setting up the repo 🥲 , I wonder if there's a better way to avoid that though 🤔 , in the very least we'll need to update `CONTRIBUTING.md` to reflect that 👍🏽

ndom91:
Yeah we could definitely add that to the `postinstall` npm script. Or is there maybe a `postclone` type script that gets auto run first time you clone? Haha

lluia:
@ndom91 `postinstall` is actually a nice idea 💯

ndom91 (approved):
LGTM 👍

balazsorban44:
This needs more tinkering, but since it caused an install issue, for now, I just reverted this in #2158 

See: https://github.com/typicode/husky/blob/28d9b2dfae5d3da7d3a98ff6581d90846e18815c/docs/README.md#install-1

We already have a `npm run dev:setup` command, we could add `husky install` there instead:

https://github.com/nextauthjs/next-auth/blob/29862ac887d76fa936bd0a8236acdce81133529f/package.json#L35

Described in `CONTRIBUTING.md`:
https://github.com/nextauthjs/next-auth/blob/main/CONTRIBUTING.md#setting-up-local-environment

PR #2141, opened 2021-06-08
Corrected Prisma Adapter import statement

ajones55555:
The prisma adapter uses a default export, not a named export (unless that's a bug in the prisma adapter)

## Reasoning 💡
There's a bug in the example code for the prisma adapter - a named import is used when it should be a default import.

## Checklist 🧢

- [x] Documentation
~- [ ] Tests~
- [x] Ready to be merged

## Affected issues 🎟

Could not find a relevant issue.

### DIFF
--- a/www/docs/adapters/prisma.md
+++ b/www/docs/adapters/prisma.md
@@ -23,7 +23,7 @@ Configure your NextAuth.js to use the Prisma Adapter:
 ```javascript title="pages/api/auth/[...nextauth].js"
 import NextAuth from "next-auth"
 import Providers from "next-auth/providers"
-import { PrismaAdapter } from "@next-auth/prisma-adapter"
+import PrismaAdapter from "@next-auth/prisma-adapter"
 import { PrismaClient } from "@prisma/client"
 
 const prisma = new PrismaClient()


www/docs/adapters/prisma.md
```diff
@@ -23,7 +23,7 @@ Configure your NextAuth.js to use the Prisma Adapter:
 ```javascript title="pages/api/auth/[...nextauth].js"
-import { PrismaAdapter } from "@next-auth/prisma-adapter"
+import PrismaAdapter from "@next-auth/prisma-adapter"
```
lluia, line 26: In the [source code of the adapters repo](https://github.com/nextauthjs/adapters/blob/canary/packages/prisma/src/index.ts#L6), this is a named export. So I think the docs are correct 💭

0ubbe (changes requested):
Thanks for pointing it out 💅🏽 , I think though the documentation is correct.

www/docs/adapters/prisma.md
balazsorban44: Yes, @lluia is right. If you experience otherwise, please open a bug report and attach a reproduction.

aej11a: Ah, I had the wrong version installed, thanks for the info and sorry for the spam PR

PR #2150, opened 2021-06-09
docs(callbacks): don't use `signIn` for redirects

nicholaschiang:
Updates the callback documentation to specify that you shouldn't use the `signIn` callback for arbitrary redirects. Instead, use the `callbackUrl` option or the redirect callback.

## Reasoning 💡

I was previously using the `signIn` callback to redirect to a relative path (as mentioned [here](https://next-auth.js.org/getting-started/client#specifying-a-callbackurl)). I didn't realize that doing so canceled the current sign in and that I should instead use [the redirect callback](https://next-auth.js.org/configuration/callbacks#redirect-callback).

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/www/docs/configuration/callbacks.md
+++ b/www/docs/configuration/callbacks.md
@@ -78,6 +78,11 @@ When using NextAuth.js with a database, the User object will be either a user ob
 When using NextAuth.js without a database, the user object it will always be a prototype user object, with information extracted from the profile.
 :::
 
+:::note
+Redirects returned by this callback cancel the authentication flow. Only redirect to error pages that, for example, tell the user why they're not allowed to sign in.
+
+To redirect to a page after a successful sign in, please use [the `callbackUrl` option](/getting-started/client#specifying-a-callbackurl) or [the redirect callback](/configuration/callbacks#redirect-callback).
+:::
 
 ## Redirect callback
 


balazsorban44 (approved):
I guess it could have been more clear from the get-go. Thanks for the PR!

nicholaschiang:
@balazsorban44 Is there a way that I can decide where to redirect the user (after a successful sign-in) based on their profile data? For example, I might want to redirect users with a certain email address domain name to a specific page.

balazsorban44:
in case if it's OAuth, you could redirect them to a page or api route like /dashboard or /api/user. 

then use useSession or getSession respectively to redirect them to your profile specific url?

For email or credentials:

then do `router.push` to wherever you want.

nicholaschiang:
@balazsorban44 Yeah, that's the current workaround (just redirect to the `/feed` page and then in the `/feed` page, detect if the user has any subscriptions selected, and if not, redirect to `/subscriptions`). I was wondering if there was a more performant fix that avoids the double redirect (i.e. redirect directly to `/subscriptions` from the `pages/api/auth/[...nextauth].ts` route).

balazsorban44:
we don't have the user info before the OAuth flow is executed, so I'm not even sure how or what a good way should be. feel free to open a feature request, possibly with a suggested solution and the intent of helping to implement it.

PR #2168, opened 2021-06-11
fix(provider) dynamically require nodemailer for email provider

hboylan:
I still see a warning so there might be a better solution:
```
warn  - ../next-auth/dist/providers/email.js
Module not found: Can't resolve 'nodemailer' in '/Users/hughboylan/Development/next-auth/dist/providers'
```

## Reasoning 💡

The `nodemailer` dependency should be optional in case the email provider is not being used.

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/src/providers/email.js
+++ b/src/providers/email.js
@@ -1,5 +1,4 @@
 import logger from '../lib/logger'
-import nodemailer from "nodemailer"
 
 export default function Email(options) {
   return {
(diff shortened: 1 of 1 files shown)


balazsorban44:
This won't be necessary when we start recommending to import providers like this:

`import EmailProvider from "next-auth/providers/email"`

some work needs to be done still, but it's on the roadmap 😊

PR #2171, opened 2021-06-12
feat(react): add more control for session syncing

balazsorban44:
## Reasoning 💡

By default `next-auth` makes sure that the user session is synced across all tabs/windows (of the same browser instance). This provides a nicer experience for the user, but there are cases where this can cause problems. (See the linked issues under).

This PR makes it possible for the developer to opt-out certain certain session syncing events.

A new `SessionProvider` prop - `broadcast` - is introduced.

It can either be a `boolean` to toggle or broadcasting, or an object for more granular settings:
```jsx
broadcast={{
  session: false,
  signOut: true
}}
```

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [ ] Ready to be merged

## Affected issues 🎟

See also #2071, #2072,


mtt87:
Thanks for the PR, I only had a moment now to look at it.
I would suggest to avoid using the param `broadcast` as it's too generic. 
In fact you are only disabling the session broadcast not other events.
It's confusing and doesn't communicate explicitly to the developer what's happening.

I would propose `enableSessionBroadcast` that defaults to `true` so people can turn off just that broadcast, otherwise an object with the single properties and defaults to true
```js
broadcast: {
  session: true,
  signout: true,
}
```

balazsorban44:
Thank you for the comments @mtt87, I agreed to your suggestions and updated this PR accordingly.

Before I can mark it as ready for review, I need to make sure that tests are passing again.

mtt87:
That's fantastic, I wonder if we have an integration test for the scenario where we have a refresh token?
Otherwise I can probably write something

mtt87:
I had to write e2e tests for our web app a couple of months ago and I spent some time on this subject, this was my experience:
- `Cypress` is nice, I've already used it in the past successfully but has some weird limitation for example you can't visit more than 2 origins in a single test and our signup/login setup needed that because when running locally I have the login service running on `localhost:xxx1` and then the signup service running on `localhost:xxx2` and then the webapp running on `localhost:xxx3` so in the end it wasn't suitable.
- `Puppeteer` quite honestly I've used it like 3 years ago and it looked nice but I don't know what happened with it nowadays I feel like it's garbage. The API documentation is horrible there are 3 different ways to achieve the same function and most of the features I tested didn't work out of the box and required me to google and find a solution which was hard because things changed over the years and you find answers from 2 years ago that are not relevant anymore. 
- `Playwright` was my choice in the end: I think Microsoft did an amazing work, the documentation is well done and clear but most of all it **works** out of the box, I didn't have any single issue with it. I read what a command does in the doc, I try it and it works as expected, no need to waste hours of time googling a solution. It seems well maintained and given what Microsoft did in the past few years I would bet on it being the future standard for e2e testing. They recently released their own test runner, looks promising but I have not tried it yet, I'm using it with Jest.

Happy to share more details about the setup if you want 👍 

I will also look into writing a test separately for the refresh token issue in the coming days

balazsorban44:
Awesome! I did my best and I think this PR is ready for review! @mtt87 if you wish to go through the code and give feedback as well, please do! I am going to go through this with @lluia in any case.

src/client/react.js
```diff
@@ -160,7 +162,24 @@ export async function signOut(options = {}) {
...
+    baseUrl,
+    basePath,
+    staleTime = 0,
+    refetchOnWindowFocus = true,
+    refetchInterval,
+  } = props
+
+  let broadcastSession
+  let broadcastSignOut
+  if (typeof props.broadcast === "boolean") {
```
mtt87, line 176: I would not support both `boolean` and `{}`, I think it's ambiguous and confusing, if people want to disable all events they can do it manually, they are just 2 events and I don't think this number will grow much further 😄 
```js
broadcast: {
  session: false,
  signout: false
}
```

You are destructuring all the `props` above but not this one? Perhaps to keep consistency I would destructure also `broadcast` on line 172 😄

src/client/react.js
```diff
@@ -225,26 +244,25 @@ export function SessionProvider(props) {
...
-    const unsubscribe = broadcast.receive(
-      async () => await __NEXTAUTH._getSession({ event: "storage" })
-    )
-
-    return () => unsubscribe()
-  }, [])
+    return broadcast.receive(async ({ reason }) => {
+      if (
+        (reason !== "signOut" && broadcastSession) ||
+        (reason === "signOut" && broadcastSignOut)
```
mtt87, line 257: The logic here seems error prone.
`reason !== 'signOut'` means that it will be true when the reason is `getSession` AND every other reason we might add later.
Should it be instead `reason === 'getSession' && broadcastSession` so we know it's only for the `getSession` event and we don't introduce accidental bugs if we add more reasons later?

That said, when using these constant values I think it would be nice to declare them as constant in order to avoid little mistakes like `signout` vs `signOut`
```js
export const SIGNOUT = 'signOut'
export const GET_SESSION = 'getSession'

// and then always use the constants rather than the string
```

types/react-client.d.ts
```diff
@@ -164,6 +164,46 @@ export interface SessionProviderProps {
...
+         */
+        session?: boolean
+        /**
+         * By default, when you invoke the `signOut` method, this will broadcast
+         * a signal to other tabs/windows, to delete their copy of the session.
+         * Set this option to `false` if you do not need this behaviour.
+         */
+        signOut?: boolean
+      }
+    | false
```
mtt87, line 206: As mentioned above I would not support the boolean as well because of ambiguous behaviour

src/client/react.js
lluia: I agree with @mtt87; it is good to keep consistency on the type the prop expects. 

For now, if we only have two options is not necessary, but in the future, we could enable an extra switch to disable all at once:
```js
broadcast: { 
  disabled: true
}

```

types/internals/react.d.ts
```diff
@@ -23,7 +30,7 @@ export interface NextAuthConfig {
    * Stores the `SessionProvider`'s session update method to be able to
    * trigger session updates from places like `signIn` or `signOut`
    */
-  _getSession: any
+  _getSession(params: { event?: string }): void
```
lluia, line 33: I think we could type `event` further reflecting is known variations? 💭 
```ts
type SessionEventNames = 'storage' | 'refetch'
_getSession(params: { event?: SessionEventNames }): void
```

0ubbe (commented):
Impressive 💯 , small nitpick about the prop API for `broadcast` 🎨

mtt87:
@lluia @balazsorban44 I've done some testing and found something interesting, I'm using the code in this PR and somehow now it works fine even with the broadcasting.
Not sure if you changed something in the behaviour but before the broadcasting was being sent immediately to the other tabs, now it's being sent AFTER the first tab successfully refreshed the session (and by that time it doesn't have any issue because the session cookie in the browser has been updated)
I made a video with some explanation
https://www.youtube.com/watch?v=tfJiMLRWh0I

There is still a possible edge case when people are switching tabs very fast before the oauth server had the time to reply with the new refresh token but in order to fix this it requires some sort of sync between all the tabs and a lock mechanism which might be more complex to achieve.

balazsorban44:
Awesome reproduction/explanation @mtt87 I really value it! As you mentioned, handling a lock in mechanism might be a bit harder to achieve.

Something like Tanner Linsley's React hook to determine the "leading" tab might help https://twitter.com/tannerlinsley/status/1400824695746793478

mtt87:
👍  next days just to double check I'll repeat the same test with the current version of next-auth (without your changes) 

That gist looks nice, I wonder how difficult it would be to implement. 
The size of the dependency is not too bad (4kb gzipped and tree shakeable) https://bundlephobia.com/package/broadcast-channel@3.7.0

I still want to write some tests to handle these scenarios so we have them covered.

I've started to setup playwright and the mock server, I will continue tomorrow probably.

balazsorban44:
I wonder if you could work together with @lluia on setting up the test environment, he focuses mainly in our tests and it would probably benefit you both to cooperate. 😊

mtt87:
Sure sounds good 👍  if you use discord I'm `mtt#8345` or let me know what to do, I'm kinda new to this type of open source collaboration 😄 
I can push something to a branch in my fork tomorrow so you can see my code

mtt87:
I was reading the code again and I think I've found an issue @balazsorban44 
```js
  /**
   * If session was `null`, there was an attempt to fetch it,
   * but it failed, but we still treat it as a valid initial value.
   */
  const hasInitialSession = props.session !== undefined

  /** If session was passed, initialize as not loading */
  const [loading, setLoading] = React.useState(hasInitialSession)
```

According to the comment, that should be `!hasInitialSession` otherwise it's being initialized as loading `true`

mtt87:
In other news I wrote an e2e test to check the refresh token scenario with tabs in sync
https://github.com/mtt87/next-auth/blob/feat/client-config/app/e2e/refresh_token.test.ts
It spins `oauth2-mock-server` and I'm checking that no `refresh_token` is used twice (like a real behaviour)
Then I trigger the event `visibilitychange` on `document` that triggers the broadcast of `getSession` event and I check that there are no errors.

@lluia Did you have a chance to look at my test setup?

balazsorban44:
`props.session !== undefined` means that the value is either a valid session, or `null`, which is correct.

mtt87:
> `props.session !== undefined` means that the value is either a valid session, or `null`, which is correct for the value of `hasInitialSession`. We treat `null` as a valid attempt.

That's fine, the problem is in the second part
```js
/** If session was passed, initialize as not loading */
const [loading, setLoading] = React.useState(hasInitialSession)
```

You say "If session was passed, initialize as NOT loading" but this code when `hasInitialSession` is `true`, it will initialize as `loading` = `true` instead of `false` as it should be according to the comment.

balazsorban44:
That is true! My bad.

mtt87:
We should have a test for this 😃

mtt87:
@balazsorban44 @lluia  sorry to bother you, did you have a chance to look at this?

I've spent some time in the past couple of weeks:
- I've added a solution to fix the refresh token issues when using `getSession` on the client side by using the Web Locks API  https://github.com/nextauthjs/next-auth/issues/2071#issuecomment-874746217
- I've created an example of `e2e` test with `Playwright` to test the multi tab broadcast scenario + refresh token which includes `oauth2-mock-server` to make it simple to test different scenarios https://github.com/mtt87/next-auth/blob/feat/client-config/app/e2e/refresh_token.test.ts

If there is anything else I can do to help you land improvements faster please let me know 😄 🙏

balazsorban44:
So first we should focus the discussion here kn the PR at hand. what is needed to get this merged? appreciate your work, but maybe open different PRs if you have any feature/code suggestions so we can discuss them separately. haven't had the time yet to work on this.

mtt87:
> So first we should focus the discussion here on the PR at hand. what is needed to get this merged? appreciate your work, but maybe open different PRs if you have any feature/code suggestions so we can discuss them separately. haven't had the time yet to work on this.

Sorry you are right better not mix stuff together apologies 😄 

This PR only needs you addressing a couple of comments and then I think it's ready 👍 , I opened a separate PR for the other part.

saby22:
Any update on this, or #3940. Is this solved in v4?

PR #2173, opened 2021-06-13
feat(core): use named params in `callbacks`

balazsorban44:
Some of our user-facing callbacks come with a bunch of parameters, and it is not always the case that a user needs all of them. Picking out certain parameters from the end of the list would require the user to define params that they wouldn't even need.

Therefore this PR changes such callbacks so the user can only pick the necessary parameters.

This comes with the bonus of better TS support on the `session` and `signIn` callbacks, where some parameters historically could have been different types.

In the `session` callback, the second param could have been `token` (when using JWT sessions) or `user` (when using DB persisted sessions). Now they are separate parameters.

In the `signIn` callback, we now separate `profile` (OAuth), `email` (Email) and `credentials` (Credentials) provider params.

BREAKING CHANGE:

The `callbacks` method signatures are changing the following way:

```diff
- signIn(user, account, profileOrEmailOrCredentials)
+ signIn({ user, account, profile, email, credentials })
```
```diff
- redirect(url, baseUrl)
+ redirect({ url, baseUrl })
```
```diff
- session(session, tokenOrUser)
+ session({ session, token, user })
```
```diff
- jwt(token, user, account, OAuthProfile, isNewUser)
+ jwt({ token, user, account, profile, isNewUser })
```

> NOTE: You only need to define the params that you actually need (no more need  for `_` params.)

This way, if you only need `token` and `account` in the `jwt` callback, you can write:

```js
jwt({ token, account }) {
  if(account) {
    token.accessToken = account.access_token
  }
  return token
}
```


balazsorban44:
Note to self, incorporate #2103

src/server/lib/callback-url-handler.js
```diff
@@ -1,32 +1,42 @@
...
+      url: callbackUrlCookieValue,
+      baseUrl,
+    })
   }
 
-  // Save callback URL in a cookie so that can be used for subsequent requests in signin/signout/callback flow
-  if (callbackUrl && (callbackUrl !== callbackUrlCookieValue)) {
-    cookie.set(res, cookies.callbackUrl.name, callbackUrl, cookies.callbackUrl.options)
+  // Save callback URL in a cookie so that it can be used for subsequent requests in signin/signout/callback flow
+  if (callbackUrl && callbackUrl !== callbackUrlCookieValue) {
```
lluia, line 33: I have the impression we don't need the first boolean check on `callbackUrl`, as it's [going to be always truthy](https://github.com/nextauthjs/next-auth/blob/main/src/lib/parse-url.js#L23).

src/server/routes/signin.js
```diff
@@ -38,17 +38,21 @@ export default async function signin(req, res) {
     // according to RFC 2821, but in practice this causes more problems than
     // it solves. We treat email addresses as all lower case. If anyone
     // complains about this we can make strict RFC 2821 compliance an option.
-    const email = req.body.email?.toLowerCase() ?? null
+    const email = req.body.email.toLowerCase()
```
lluia, line 41: We skip `?`  here, which means we always expect `email` to be always supplied in the `POST` request. Are we sure it won't create any side effects 🙇🏽‍♂️

0ubbe (commented):
Very impressive 🤯 💯 , it seems like a big improvement. I leave 4 minor 💬 comments/questions ✨ 

I'm correct seeing some new features added? I think it would be helpful to present them on separate PRs to be able to focus better on the review 💡

types/index.d.ts
```diff
@@ -328,10 +354,10 @@ export interface CookieOption {
...
 /** [Documentation](https://next-auth.js.org/configuration/options#cookies) */
 export interface CookiesOptions {
-  sessionToken?: CookieOption
-  callbackUrl?: CookieOption
-  csrfToken?: CookieOption
-  pkceCodeVerifier?: CookieOption
+  sessionToken: CookieOption
+  callbackUrl: CookieOption
+  csrfToken: CookieOption
+  pkceCodeVerifier: CookieOption
```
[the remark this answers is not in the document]
balazsorban44: So the `CookiesOptions` interface is marked this way now, but on line 199 you can see it is `Partial`. This is so that `InternalOptions` can still use `CookieOptions` where all properties are present (as we merge with a default object).

src/server/lib/callback-url-handler.js
```diff
@@ -1,32 +1,42 @@
...
+    cookie.set(
+      res,
+      cookies.callbackUrl.name,
+      callbackUrl,
+      cookies.callbackUrl.options
+    )
```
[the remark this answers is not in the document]
balazsorban44: This is internal code, so I would do it in a different PR, we can also do it against `main` if we want to. This does not affect anything.

src/server/lib/callback-url-handler.js
balazsorban44: This is internal code, I would do it in a different PR. I think the whole `callbackUrlHandler` could be refactored. I'll have more confidence when we have tests at place.

balazsorban44:
I will be breaking down the initial PR into several PRs for easier reviews. First is the `callbacks`.

www/docs/configuration/callbacks.md
```diff
@@ -113,31 +98,23 @@ The redirect callback may be invoked more than once in the same flow.
...
-- The arguments _user_, _account_, _profile_ and _isNewUser_ are only passed the first time this callback is called on a new session, after the user signs in.
+- As with database persisted session expiry times, token expiry time is extended whenever a session is active.
+- The arguments _user_, _account_, _profile_ and _isNewUser_ are only passed the first time this callback is called on a new session, after the user signs in. In subsequent calls, only `token` will be available.
 
-The contents _user_, _account_, _profile_ and _isNewUser_ will vary depending on the provider and on if you are using a database or not. If you want to pass data such as User ID, OAuth Access Token, etc. to the browser, you can persist it in the token and use the `session()` callback to return it.
+The contents _user_, _account_, _profile_ and _isNewUser_ will vary depending on the provider and on if you are using a database or not. If you want, you can to persist data such as User ID, OAuth Access Token in this token. To make it available in the browser, check out the [`session()` callback](#session-callback) as well.
```
ndom91, line 109: "... If you want, you can ~to~ persist data"

www/docs/configuration/callbacks.md
```diff
@@ -146,18 +123,14 @@ callbacks: {
 ```
 
 :::tip
-Use an if branch in jwt with checking for existence of any other params than token. If any of those exist, you call jwt for the first time.
-This is a good place to add for example an `access_token` to your jwt, if you want to.
-:::
-
-:::tip
-Check out the content of all the params in addition `token`, to see what info you have available on signin.
+Use an if branch in jwt with checking for existence of any other params than `token`. If any of those exist, you call jwt for the first time.
```
ndom91, line 126: Bit confusingly worded imo, how about something like, "Use an if branch in the jwt callback to check for the existence of any other params in addition to `token`. If any exist, you can call jwt for the first time." ?

ndom91 (approved):
A few small nits regarding wording, otherwise LGTM :+1:

www/docs/configuration/callbacks.md
balazsorban44: Thanks, I now reworded this section!

PR #2188, opened 2021-06-15
docs(provider): scope expects space separated string

Cryt1c:
## Reasoning 💡

Currently the docs list string[] as a possible type for scope. 
However, It only accepts a string (with space as separator).

scope

## Checklist 🧢

- [x] Documentation
- [x] Ready to be merged

## Affected issues 🎟

-


balazsorban44:
Thanks for the PR! I went down the rabbit hole to find the reasoning and what I found is that at the end of the day, `string[]` is supported in a way by our parser that translates the scope to a url query parameter, but in a wrong way that probably some oauth providers wouldn't support. See

https://nodejs.org/api/querystring.html#querystring_querystring_stringify_obj_sep_eq_options

So I'm happy to accept this PR! one thing I changed is the PR target branch. We are working on our next major release, and so any non-critical changes will preferably be made in the `next` branch. This is only temporary, so I don't wish to change our contribution docs on that.

What that means for you that until v4 docs is officially out, this change will only be visible at https://next-auth-git-next-nextauthjs.vercel.app/ for now. Hope it's not a problem for you!

Cryt1c:
Thanks for your review! 
Makes sense. :+1: I have resolved all merge conflicts.

balazsorban44:
I was going to, but thank you! 😁

PR #2202, opened 2021-06-17
fix(provider): Set idToken for cognito provider as true by default

shriah:
Set `idToken` for cognito provider as true by default

## Reasoning 💡
Cognito provider currently does not set `idToken` option to `true` even though Cognito supports it. This limits the `profileData` being passed to `profile` mapper function. 

## Checklist 🧢

~- [ ] Documentation~
~- [ ] Tests~
- [x] Ready to be merged


balazsorban44:
it is very easy to set this in your code yourself, so I think this will need more validation to pass. This is also a breaking change, so we couldn't accept it yet.

Please open an issue and let's discuss, what others think, if there is interest for this change. until then, I'll close this PR.

PR #2227, opened 2021-06-21
fix(ts): improve `authorize` typing on Credentials provider

### LINKED ISSUE #2080: Incorrect typing for credentials parameter in Credentials provider's `authorize` method
## Question 💬

First of all, thank you for a great library! I've never worked with auth in JS/TS that was so plug&play before! I migrated my Next.js project from Firebase Auth to NextAuth.js, and all my auth headaches were solved. I was able to greatly simpify auth logic in my app and delete many lines of code. :100: 

As for my issue: I suspect the typing for the `authorize` method for `Providers.Credentials` is wrong/bugged. For me, the `credentials` parameter is typed as `Record<never, string>`, which seems wrong:
image

I'm on the latest published version at the time of writing, _3.23.2_, and Typescript _4.3.2_. This is the relevant part of my code:
```typescript 
  ...
  providers: [
    Providers.Credentials({
      name: "username and password",
      credentials: {
        email: { label: "E-mail", type: "email", placeholder: "example@user.com" },
        password: { label: "Password", type: "password", placeholder: "password" },
      },
      async authorize(credentials, req) { // `credentials` is typed as an empty object
        console.log("CREDS", credentials);

        const user = await prisma.user.findUnique({ where: { email: credentials.email } }); // TS2339: Property 'email' does not exist on type 'Record'.
        if (!user) {
          return null;
        }

        const correctPassword = await compare(credentials.password, user.passwordHash);
        if (!correctPassword) {
          return null;
        }

        return {
          id: user.id,
          email: user.email,
        };
      },
    }),
  ],
  ...
```

Here, the `console.log` statement logs the following:
```
CREDS [Object: null prototype] {
  redirect: 'false',
  email: 'test@test.com',
  password: 'test',
  csrfToken: 'fe99993acda0720143a1732e3d9f37faf77ea3451cfbe82306147bf7f393204b',
  callbackUrl: 'http://localhost:3000/login',
  json: 'true'
}
```

So the various properties are correctly set, but types report the object as empty.

This is the interface in question:
```
interface CredentialsConfig<C extends Record<string, CredentialInput> = {}>
  extends CommonProviderOptions {
  type: "credentials"
  credentials: C
  authorize(credentials: Record<keyof C, string>, req: NextApiRequest): Awaitable<User | null>
}
```

Now, this looks to be a bit above my Typescript skill level. `Record<keyof C, string>` looks to me like it should work, but it doesn't. I'd be happy to submit a PR if I knew what the fix was, but unfortunately that's not the case, and I'd love some input here. :slightly_smiling_face: 

I've simply added a `// @ts-ignore` above the `authorize` signature to work around the issue so far.

akomm:
## Reasoning 💡

Fixes typing error on `CredentialsConfig` and `CredentialsProvider`. Reasoning explained in my [comment](https://github.com/nextauthjs/next-auth/issues/2080#issuecomment-864930543)

## Checklist 🧢

~- [ ] Documentation~
- [X] Tests
- [X] Ready to be merged

## Affected issues 🎟

Fixes #2080


types/providers.d.ts
```diff
@@ -114,15 +114,15 @@ interface CredentialInput {
   placeholder?: string
 }
 
-interface CredentialsConfig<C extends Record<string, CredentialInput> = {}>
+interface CredentialsConfig<C extends Record<string, CredentialInput> = {}, K extends keyof C = keyof C>
```
balazsorban44, line 117:
```suggestion
interface CredentialsConfig<C extends Record<string, CredentialInput> = {}>
```

types/providers.d.ts
```diff
@@ -114,15 +114,15 @@ interface CredentialInput {
...
   extends CommonProviderOptions {
   type: "credentials"
   credentials: C
-  authorize(credentials: Record<keyof C, string>, req: NextApiRequest): Awaitable<User | null>
+  authorize(credentials: Record<K, string>, req: NextApiRequest): Awaitable<User | null>
```
balazsorban44, line 121:
```suggestion
  authorize(credentials: Record<keyof C, string>, req: NextApiRequest): Awaitable<User | null>
```

types/providers.d.ts
```diff
@@ -114,15 +114,15 @@ interface CredentialInput {
...
 }
 
-export type CredentialsProvider = (
-  options: Partial<CredentialsConfig>
+export type CredentialsProvider = <C extends Record<string, CredentialInput>>(
+  options: Partial<CredentialsConfig<C, keyof C>>
```
balazsorban44, line 125:
```suggestion
export type CredentialsProvider = <C extends Record<string, CredentialInput>>(
  options: Partial<CredentialsConfig<C>>
) => CredentialsConfig<C>
```

types/tests/providers.test.ts
```diff
@@ -24,7 +24,7 @@ Providers.Credentials({
   id: "login",
   name: "account",
   credentials: {
-    user: {
+    username: {
       label: "Password",
       type: "password",
```
balazsorban44, line 29:
```suggestion
    username: {
      label: "Username",
      type: "text"
```

balazsorban44 (changes requested):
So I added some suggestions to simplify this.
As you can see on the image below, it works:
image

Could you please have a look?

types/providers.d.ts
akomm: I'm really surprised. Instantiate here this way was first step I did, as its the main problem and it was not enough... But if you say it works, then it works... :+1:

balazsorban44 (approved):
Thank you!

types/providers.d.ts
```diff
@@ -121,8 +121,9 @@ interface CredentialsConfig<C extends Record<string, CredentialInput> = {}>
   authorize(credentials: Record<keyof C, string>, req: NextApiRequest): Awaitable<User | null>
 }
 
-export type CredentialsProvider = (
-  options: Partial<CredentialsConfig>
+export type CredentialsProvider = <C extends Record<string, CredentialInput>>(
+  options: Partial<CredentialsConfig<C>>
+) => CredentialsConfig<C>
 ) => CredentialsConfig
```
balazsorban44, line 127:
```suggestion
) => CredentialsConfig<C>
```

balazsorban44 (changes requested):
Sorry, I was a bit too eager to approve.

You forgot an extra line while adding my suggestion, and also some of the type checking tests are failing

types/providers.d.ts
akomm: I can not commit the suggestion

akomm:
I'm going to clean this up and force-update my branch, this is why:

1. I can not confirm your last suggestion, because there is supposedly no change. I don't really like fixing stuff directly on github, its flacky and messy (even though it improves). 
2. CONTRIBUTING.md does not mention there is actually a separate test for types `npm run test:types`. This is why I did not see that the type test fails. While the fix was correct, the assertion `$ExpectType` in the test was stale. 
3. Many new commits to merge, I think its simpler for you if everything is up to date and you don't need to clean up the commit & merge mess.

Push commin soon...

balazsorban44 (approved):

PR #2228, opened 2021-06-21
Feat: getServerSession() and getServerProvider()

### LINKED ISSUE #1535: Server-side HTTP requests via `getSession` usually get extremely slow on Vercel
**Describe the bug**

Before further ado, here’s a comparison between a `/api/user/{id}` request which [fetches `/api/auth/session`](https://github.com/nextauthjs/next-auth/blob/e5535734f83d2719597be57c90e4315e4d437e74/src/client/index.js#L146-L160) *through the network* even on the server side, versus a `/api/site/{id}` call, which isn’t protected via `getSession`. Unfortunately, there’s a quite large margin between response times.

image

**Steps to reproduce**
Call `getSession` on the server side and observe that it makes an HTTP request instead of just calling functions locally.

**Expected behavior**
Calls to `getSession` should be executed locally in a server environment. A separate `getServerSession` method could be added as a non-isomorphic solution without any HTTP calls.

**Additional context**
According [to the official Next.js docs](https://nextjs.org/docs/basic-features/data-fetching#getserversideprops-server-side-rendering):

> You should not use [`fetch()`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) to call an API route in `getServerSideProps`. Instead, directly import the logic used inside your API route. You may need to slightly refactor your code for this approach.
> 
> Fetching from an external API is fine!

This seems to hurt performance badly.

**Feedback**

* [x] Found the documentation helpful

Thank you for maintaining this wonderful library – it really helps a lot! 🙏

walid-mokrani:
## Reasoning 💡

Allow to reuse the API logic in `getServerSideProps` to avoid additional network request.

It adds two functions `getServerSession` and `getServerProviders` which reuse the logic of `getSession` and `getProviders` respectively.

Developers are still able use `getSession` and `getProviders` if they want to.

This is more a proof of concept to be able to test performance before making decisions.

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Types
- [ ] Lambda performance
- [x] Ready to be merged


walid-mokrani:
I m not really satisfied about the way I organized/named things. I find that adding a config file `next-auth.config.js` to share options is a bit cumbersome.

lluia:
Hi @walid-mokrani, thanks a lot for your suggestions 💯 ; it seems a bit delicate as it's changing core parts of our source code; hence we need to really review this with care. We'll try to find some time, we're currently a bit busy at the moment 👍🏽

PR #2236, opened 2021-06-23
feat(react): make session requireable in `useSession`

### LINKED ISSUE #1210: useSession({ required: true }), which always returns a non-null Session object

balazsorban44:
## Reasoning 💡

A living session could be a requirement for specific pages (like dashboards). If it doesn’t exist, then the user should be redirected to a page that asks them to sign in again.

Sometimes, a user might log out by accident, or by deleting cookies on purpose. If that happens (e.g. on a separate tab), then `useSession({ required: true })` should detect the absence of a session cookie and always return a non-nullable Session object type.

When `required: true` is set, the default behavior will be to redirect the user to the sign-in page. This can be overridden by an `action()` callback:

```js
const session = useSession({
  required: true,
  action() {
    // ....
  }
})
if (session.status === "Loading") return "Loading or not authenticated..."

// session.data is always defined here.
```
## Checklist 🧢

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged

## Affected issues 🎟

Fixes #1210
Related: #2269 

BREAKING CHANGE:

The `useSession` hook now returns an object. Here is how to accommodate for this change:

```diff
- const [ session, loading ] = useSession()
+ const { data: session, loading } = useSession()
```


balazsorban44:
**NOTE** to reviewers:
I tried updating the documentation everywhere, but there might be still something that I forgot. Therefore it should be scanned with extreme care, as `useSession` is one of our most important APIs :eyes:

www/docs/getting-started/client.md
```diff
@@ -34,24 +34,51 @@ You can use the [session callback](/configuration/callbacks#session-callback) to
...
 
 ```jsx
 import { useSession } from "next-auth/react"
 
 export default function Component() {
-  const [session, loading] = useSession()
+  const { data: session, status } = useSession()
 
-  if (session) {
+  if (status === "authenticated) {
```
kripod, line 47:
```suggestion
if (status === "authenticated") {
```

kripod (approved):
Great work, keep it up! 🙌

src/client/react.js
```diff
@@ -43,8 +43,33 @@ const logger = proxyLogger(_logger, __NEXTAUTH.basePath)
...
+        error: "SessionRequired",
+        callbackUrl: window.location.href,
+      })}`
+      if (action) action()
+      else window.location.replace(url)
+    }
+  }, [requiredAndNotLoading, action])
+
+  if (requiredAndNotLoading) {
+    return { data: value.data, loading: true, status: "loading" }
```
lluia, line 69: I think we should have just `status: loading` and drop boolean flags, in this case `loading`. 

`data` seems a bit generic too, I would call it `session`.

Usually, boolean flags [just end creating technical debt](https://medium.com/geekculture/why-you-shouldnt-use-booleans-in-rest-apis-2747b187876c). 📐

types/internals/react.d.ts
```diff
@@ -26,4 +26,12 @@ export interface NextAuthConfig {
...
 }
 
-export type SessionContext = React.Context<Session>
+export type SessionContextValue<R extends boolean = false> = R extends true
+  ?
+      | { data: Session; status: "authenticated"; loading: false }
+      | { data: null; status: "loading"; loading: true }
+  :
+      | { data: Session; status: "authenticated"; loading: false }
+      | { data: null; status: "unauthenticated" | "loading"; loading: boolean }
```
lluia, line 35: Let's have a call to discuss this one, I'm a bit confused about why the ternary here 🧐

types/tests/react.test.ts
```diff
@@ -11,9 +11,25 @@ const clientSession = {
...
 
+// $ExpectType { data: Session; status: "authenticated"; loading: false; } | { data: null; status: "loading"; loading: true; }
+const session = client.useSession({ required: true })
+if (session.status === "loading") {
+  // $ExpectType null
+  session.data
+} else {
+  // $ExpectType Session
+  session.data
+}
```
lluia, line 31: Should we have a third case here?
```ts
const session = client.useSession({ required: false })
// $ExpectType Session
session.data

File: www/docs/getting-started/client.md [hunk 5]
```diff
@@ -34,24 +34,51 @@ You can use the [session callback](/configuration/callbacks#session-callback) to
...
   return <a href="/api/auth/signin">Sign in</a>
 }
 ```
 
+### Require session
+
+Due to the way how Next.js handles `getServerSideProps` and `getInitialProps`, every protected page load has to make a server-side request to check if the session is valid and then generate the requested page (SSR). This increases server load, and if you are good with making the requests from the client, there is an alternative. You can use `useSession` in a way that makes sure you always have a valid session. If after the initial loading state there was no session found, you can define the appropriate action to respond.
+
+The default behavior is to redirect the user to the sign-in page, from where - after a successful login - they will be sent back to the page they started on. You can also define an `action()` callback, if you would like to do something else.
+
```
lluia, line 60: We must document here the different values of `status` for non TS users 💅🏽

www/docs/getting-started/example.md
```diff
@@ -72,11 +69,12 @@ export default function Page() {
 You can use the `useSession` hook from anywhere in your application (e.g. in a header component).
 :::
 
-### Add session state
+### Share/configure session state
 
-To allow session state to be shared between pages - which improves performance, reduces network traffic and avoids component state changes while rendering - you can use the NextAuth.js Provider in `pages/_app.js`.
+Use the `<SessionProvider>` to allows instances of `useSession()` to share the session object across components. It also takes care of keeping the session updated and synced between tabs/windows.
```
lluia, line 74:
```suggestion
Use the `<SessionProvider>` to allow instances of `useSession()` to share the session data across components. `<SessionProvider>` will also take care of keeping the session updated and synced between tabs/windows.
```
I'm not sure about this though given is not possible anymore to use `useSession` without `<SessionProvider />` 🤔

www/docs/tutorials/securing-pages-and-api-routes.md
```diff
@@ -60,15 +64,16 @@ export default function Page() {
...
+    props: {
+      session: await getSession(context)
+    },
   }
 }
 ```
 
 :::tip
-This example assumes you have configured `_app.js` to pass the `session` prop through so that it's immediately available on page load to `useSession`.
+You have to configure `_app.js` to pass it the `session` prop. This way `useSession` won't show a loading state, because it already has the session available.
```
lluia, line 76:
```suggestion
When you supply a `session` prop in `_app.js`, `useSession` won't show a loading state, as it'll already have the session available. In this way, you can provide a more seamless user experience.
```

0ubbe (changes requested):
Wow, very neat 🎸 💯 ✨ 

I'm leaving quite a lot of suggestions 💬 , in an effort to further polish the API, but I consider it's already very nice 💚

Let's have a call to go through it together 👍🏽

src/client/react.js
```diff
@@ -43,8 +43,33 @@ const logger = proxyLogger(_logger, __NEXTAUTH.basePath)
 /** @type {import("types/internals/react").SessionContext} */
 const SessionContext = React.createContext()
 
-export function useSession() {
-  return React.useContext(SessionContext)
+export function useSession(options = {}) {
+  const value = React.useContext(SessionContext)
```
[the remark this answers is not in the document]
balazsorban44: As explained earlier today, this helps in development.

src/client/react.js
```diff
@@ -43,8 +43,33 @@ const logger = proxyLogger(_logger, __NEXTAUTH.basePath)
...
+
+  if (process.env.NODE_ENV !== "production" && !value) {
+    throw new Error("useSession must be wrapped in a SessionProvider")
+  }
+
+  const { required, action } = options
+
+  const requiredAndNotLoading = required && value.status === "unauthenticated"
```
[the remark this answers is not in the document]
balazsorban44: I would wait with this until TS conversion

types/internals/react.d.ts
balazsorban44: This is to enforce stricter types when using `required: true`

types/tests/react.test.ts
balazsorban44: In the case of `required: true`, there is only two valid status values. `loading` and `authenticated`.

www/docs/getting-started/client.md
balazsorban44: Agree! Good catch.

www/docs/getting-started/example.md
balazsorban44: Yes, we should rephrase this to be explicit that users *must* use `SessionProvider` from now on.

src/client/react.js
balazsorban44: Yeah, I was a bit hesitant, as migration would have been simpler. But I am on board to drop the `loading` flag anyway.

I think of v4 as quite a few breaking changes but in the name of creating a good base for future improvements. Refactoring the APIs to use objects in most situations (returned values or function parameters)` will make it possible to extend almost anything with new options/fields without requiring the user to change their APIs.

ndom91 (approved):
LGTM :+1:

balazsorban44:
@lluia had some requests which first need to be reviewed again, when this PR is updated. 😊

0ubbe (approved):

PR #2237, opened 2021-06-23
fix(client): handle keepalive network failures (#2215)

### LINKED ISSUE #2215: NextAuth should not invalidate the client-side session when the browser is offline
## Description 🐜

NextAuth invalidates the client-side session when the browser is offline and the NextAuth client cannot communicate with the NextAuth backend. From an application perspective, the expectation is that the NextAuth library must provide the last-known session state unless session TTL runs out or the NextAuth client receives an authoritative response from the NextAuth backend.

## How to reproduce ☕️

```
git clone https://github.com/nextauthjs/next-auth.git
cd next-auth 
cp app/.env.local.example app/.env.local 
npm install
npm run dev:setup
npm run dev
```

- Open http://localhost:3000/protected in Chrome
- Sign In with credentials using "password" as password
- Observe that the page says "This is protected content. You can access this content because you are signed in."
- Open Chrome dev tools Network tab and select the "Offline" mode
- Minimize the browser window
- Restore the browser window
- Observe that the page says "Access Denied. You must be signed in to view this page"

## Screenshots / Logs 📽

Logs in browser console:
```
react_devtools_backend.js:2560 [next-auth][error][client_fetch_error] 
https://next-auth.js.org/errors#client_fetch_error session TypeError: Failed to fetch

asyncToGenerator.js?1da1:6 Uncaught (in promise) TypeError: Failed to fetch
```

## Environment 🖥

```
  System:
    OS: macOS 11.4
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
    Memory: 1.37 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.17.1 - /usr/local/bin/node
    npm: 6.14.13 - ~/Temp/next-auth/node_modules/.bin/npm
  Browsers:
    Chrome: 91.0.4472.106
    Firefox: 89.0
    Safari: 14.1.1
```

## Contributing 🙌🏽

I'm more than happy to author a PR but I'm just not sure about the logic behind https://github.com/nextauthjs/next-auth/blob/5a89ab69d3ceee2f84b683b6fa67455126f80815/src/client/index.js#L106 Is there any specific reason to force clientMaxAge=0 session re-validation on keep-alive timer or visibilitychange events? IMO clientMaxAge=0 session must be presumed valid unless the backend says the session is no longer valid (which would be triggered by 'storage' event).

igordanchenko:
## Reasoning 💡

The current implementation does not tolerate network failures in keep-alive calls (see #2215 for a detailed scenario). Even a single keep-alive call that fails due to network failure or browser being offline results in client-side session eviction. Proposed change: continue using the existing session object if the underlying fetch call fails.

## Checklist 🧢

- [ ] ~Documentation~
- [ ] ~Tests~
- [x] Ready to be merged

## Affected issues 🎟

Fixes #2215

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -138,6 +138,8 @@ function _useSessionHook(session) {
         // infinate loop.
         const newClientSessionData = await getSession({
           triggerEvent: !triggeredByStorageEvent,
+          currentSession: !triggeredByStorageEvent && (clientMaxAge === 0 || currentTime < clientLastSync + clientMaxAge)
+            && clientSession ? { ...clientSession } : null
         })
 
         // Save session state internally, just so we can track that we've checked
@@ -159,7 +161,7 @@ function _useSessionHook(session) {
 }
 
 export async function getSession(ctx) {
-  const session = await _fetchData("session", ctx)
+  const session = await _fetchData("session", ctx, ctx?.currentSession)
   if (ctx?.triggerEvent ?? true) {
     broadcast.post({ event: "session", data: { trigger: "getSession" } })
   }
@@ -314,7 +316,7 @@ export function Provider({ children, session, options }) {
  * work seemlessly in getInitialProps() on server side
  * pages *and* in _app.js.
  */
-async function _fetchData(path, { ctx, req = ctx?.req } = {}) {
+async function _fetchData(path, { ctx, req = ctx?.req } = {}, fallback = null) {
   try {
     const baseUrl = await _apiBaseUrl()
     const options = req ? { headers: { cookie: req.headers.cookie } } : {}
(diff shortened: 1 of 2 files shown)


balazsorban44:
Thank you for the PR, although I am going to close this for now.

First of all, we are in the process of releasing a new major version (v4), and the client code has been refactored significantly on the `next` branch. Any significant changes should therefore be made against the `next` branch. (See the new client code at: https://github.com/nextauthjs/next-auth/blob/next/src/client/react.js)

Furthermore, I am not really sure about this proposal. If there is a network error while fetching the session, then there is a big chance that this network error will propagate through the app, thus keeping the old session wouldn't be that useful, in my opinion.

When a request is made to the `/api/auth/session` endpoint, it is probably requested with a reason, like a user set up session polling (currently called `keepAlive`). If that session request fails, then it should be assumed that the current session is not valid anymore (that must have been the whole reason for the new session request).

I saw the point you made in https://github.com/nextauthjs/next-auth/issues/2215#issuecomment-864406330, but I would call this a userland problem, as in my opinion any failed request for a new session would assume the invalidness of the old session as well. I am up for debate on it and would like to hear what others think, so let us continue in #2215 a bit longer.

PR #2241, opened 2021-06-24
fix(provider): convert id to string

SagnikPradhan:
Converting id to string before returning profile

## Reasoning 💡

Github uses int ids but internally next-auth expects ids to be string

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

## Affected issues 🎟

Not sure.


balazsorban44:
This is already fixed in the `next` branch.

https://github.com/nextauthjs/next-auth/pull/2108

PR #2254, opened 2021-06-25
feat: bump dependencies

balazsorban44:
Node 10 has reached end-of-life, Next.js stopped supporting it in Next 11. Since we are a Next.js library, it doesn't make sense for us to support anything lower than that either.

I also upgraded a bunch of dependencies and dropped some that weren't necessary anymore.

BREAKING CHANGE:

The lowest supported Node version is *12*. (We still support IE11 in browsers, until that is not dropped by Next.js itself)


ndom91 (approved):
LGTM 👍

PR #2255, opened 2021-06-25
feat: rename `protection` to `checks`

balazsorban44:
This will better align with `openid-client`'s `checks` https://github.com/panva/node-openid-client/blob/main/docs/README.md#clientcallbackredirecturi-parameters-checks-extras, a library which we intend to migrate to in the future. Aligning our API early, so people get used to it.

Also, objectively the name `protection` might not have been as clear as I first thought. `checks` better describes the intention.

BREAKING CHANGE:

The `state` option on OAuth providers is now deprecated. Use `checks: ["state"]` instead.
`protections` is renamed to `checks`, here is an example:
```diff
- protection: ["pkce"]
+ checks: ["pkece"]
```

Furthermore, string values are not supported anymore. This is to be able to handle fewer cases internally.
```diff
- checks: "state"
+ checks: ["state"]
```


ndom91 (commented):
LGTM :+1:

0ubbe (approved):
Nice 💚

PR #2268, opened 2021-06-28
fix(ts): tweak pages option types for newUser

yokinist:
## Reasoning 💡

https://next-auth.js.org/configuration/pages
Because set `null` to `newUser` on pages options make type error...

image

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -414,7 +414,7 @@ export interface PagesOptions {
   error?: string
   verifyRequest?: string
   /** If set, new users will be directed here on first sign in */
-  newUser?: string
+  newUser?: string | null
 }
 
 export interface DefaultSession extends Record<string, unknown> {


balazsorban44 (changes requested):
Thank you! I think I would rather change the documentation around this. Currently, we actually don't care about the falsy value, as long as it is falsy. See:

https://github.com/nextauthjs/next-auth/blob/0475964a0f47ed5d323d0bcb779f208515ea9782/src/server/routes/callback.js#L147

or

https://github.com/nextauthjs/next-auth/blob/0475964a0f47ed5d323d0bcb779f208515ea9782/src/server/routes/callback.js#L294

If you don't care about `newUser`, just don't leave it out, and TS will be happy.

Setting `null` specifically doesn't have a real meaning, so I think we should drop it.

Not sure if it is a breaking change or not :thinking: 

What do you think @lluia, @ndom91?

ndom91:
Yeah I think just a docs update would be more appropriate here. 

Just leave it as an empty string, aka a falsy value, and well handle it appropriately 👍

yokinist:
@balazsorban44 @ndom91 Thank you for confirming! I understand
I'll send the PR to the documentation 🙌

balazsorban44:
@ndom91 not even an empty string is necessary. leaving out the field from the `pages` object is enough and TS is already setting it as an optional property. 😁

PR #2276, opened 2021-06-29
feat(client): follow 302 redirects

### LINKED ISSUE #1642: 302 redirects not followed
**Describe the bug**

302 redirects are not followed for the `profileUrl` when authenticating with OAuth 2. The root cause is due to the underlying library used for authentication, [oauth](https://github.com/ciaranj/node-oauth). However, this library is unmaintained and has had an open issue for it for 6+ years: https://github.com/ciaranj/node-oauth/issues/201.

This issue causes `profileData` to be an empty string. Then in [getProfile](https://github.com/nextauthjs/next-auth/blob/073da60c3d6c06cfa0ee0918cba0d65b55faf898/src/server/lib/oauth/callback.js#L97), an error is raised due to `JSON.parse`.

**Steps to reproduce**

I'm creating a custom provider where the `profileUrl` returns a redirect to the actual url. In this instance, `/api/users/current` redirects to `/api/users/{current_user_id}`.

**Expected behavior**

The `Location` header is followed for `302` redirects. Alternatively, the library includes a way for developers to follow the redirects themselves.

**Additional context**

If help is needed, I can open a PR for this. I think the callback to `_request` [here](https://github.com/nextauthjs/next-auth/blob/073da60c3d6c06cfa0ee0918cba0d65b55faf898/src/server/lib/oauth/client.js#L228-L236) can take in the `response` and issue another request if it's a response with a `Location` header and `3xx` status code.

**Feedback**

denkristoffer:
## Reasoning 💡

Some OAuth providers redirect clients with a `302 Found` and the current server client breaks in such cases.

There's a RFC for improved customisation capabilities in #1846 which can solve this in a more robust way, so I understand if you're not interested in adding this capability like this.

## Checklist 🧢

- [ ] ~Documentation~
- [ ] Tests
- [x] Ready to be merged

## Affected issues 🎟

Closes #1642

### DIFF
--- a/src/server/lib/oauth/client.js
+++ b/src/server/lib/oauth/client.js
@@ -246,14 +246,36 @@ async function getOAuth2 (provider, accessToken, results) {
     httpMethod = 'POST'
   }
 
-  return new Promise((resolve, reject) => {
-    this._request(httpMethod, url, headers, null, accessToken, (error, profileData) => {
-      if (error) {
-        return reject(error)
-      }
-      resolve(profileData)
+  const request = (httpMethod, url, headers, accessToken) =>
+    new Promise((resolve, reject) => {
+      this._request(
+        httpMethod,
+        url,
+        headers,
+        null,
+        accessToken,
+        (error, profileData, response) => {
+          if (error) {
+            return reject(error)
+          }
+
+          if (response.statusCode === 302 && response.headers.location) {
+            return resolve(
+              request(
+                httpMethod,
+                response.headers.location,
+                headers,
+                accessToken
+              )
+            )
+          }
+
+          return resolve(profileData)
+        }
+      )
     })
-  })
+
+  return request(httpMethod, url, headers, accessToken)
 }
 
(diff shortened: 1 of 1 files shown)


0ubbe (commented):
The quick fix looks good to me, thanks for the contribution 💚 💪🏽

The best would be to have this handled on the **RFC** as proposed with @balazsorban44, I need to sync with him on what's the timeline for implementing that (let's wait for his feedback before merging) 👍🏽

renanrodrigueszup:
Believe I'm running onto this exact same issue. @denkristoffer have you solved this?

denkristoffer:
@renanrodrigueszup With v3 you can add my code with a patch (which is what I was doing), but in v4 you don't need to do that, instead you can use custom HTTP options through `openid-client` in your provider: https://github.com/panva/node-openid-client/tree/main/docs#customizing-individual-http-requests – this can be set up to follow redirects.

Issue #159, opened 2020-05-25
RFC: Allow pasting of email verify token into 'Check Email' form


TimNZ:
Proof of concept experimenting.

Added option 'tokenInputCheckEmail' setting to Email provider.

If true:

* An input is presented on the Check Email page
* Token is displayed in the verification email
* Token can be pasted into the Check Email input to verify, instead of clicking the link

Compare branches:
https://github.com/iaincollins/next-auth/compare/master...TimNZ:email-token-form

I prefer a shorter token, but that's easy to support.

![image](https://user-images.githubusercontent.com/1051701/82792894-2c2ac300-9ec4-11ea-806f-18995ca42ea3.png)

![image](https://user-images.githubusercontent.com/1051701/82792911-3482fe00-9ec4-11ea-9b25-b045e65a06e6.png)

![image](https://user-images.githubusercontent.com/1051701/82793049-6a27e700-9ec4-11ea-8cc4-dc1c5f63073f.png)

iaincollins:
Thanks for this!

I think the option of a short code is a really good idea, especially useful in scenarios where you email gets sent to a different device to the one you are logging into.

e.g. logging into a desktop / work computer and getting a code on your phone (or vice versa).

I run into systems that work this way a lot (it seems a lot of video game related services use that mechanism) and it makes it easier to sign in as I don't have to fire up Gmail in a tab, go through 2FA with Google, then copy and paste the link, so I appreciate it.

Off hand, I think the way I would handle supporting would be by refactoring and keeping what we have (long tokens, that are very secure) and introducing this as secondary option (with nicely formatted emails displaying the short codes, but also still with the link option).

I've been thinking about refactoring 'email verification' tables, etc to just 'verification' (e.g. so people could use SMS based services for sending tokens like this).

Having them as separate providers - e.g. magiclink and short code - (basically the same flow, but with different callbacks, and the short code could have a link too) might be a good way to handle it, as it would make it very easy for people to choose the one they wanted with no additional config, and I could define a callback in providers for generating codes so it's a function inside the provider profile (like the email sending is). 

Both tokens could default to the same expiry time, but then this could be overridden by setting a different  expiry time on the provider, as it might make sense to set a shorter expiry time on shorter codes.

I think this might be one to leave till 2.1 but it may make sense to do some refactoring to make that possible before release - of course if the above makes sense to you and you have time to work on it, I'd be happy to include something like this at any time!

TimNZ:
Cool.

[...]

PR #2293, opened 2021-07-01
feat(provider): SMS provider

### LINKED ISSUE #1562: Add an example/tutorial for passwordless with SMS?

I'd like to use NextAuth to add passwordless authentication to an app using SMS rather than email to deliver the tokens. It would look like:

1. overriding `generateVerificationToken` to create a shorter token that a user can feasibly type
2. override `sendVerificationRequest` to use Twilio or similar to send an SMS message with the token
3. create a page where the user can type the token, similar to entering a 2FA code

It would be good to have an official tutorial or example showing how to implement this securely. I might be able to contribute one in the future.

Some groups of users, particularly those with poor digital skills, will find it easier to verify their identity with an SMS rather than an email.

Sometimes (for example, a community group of volunteers) an organiser may only have phone numbers for the users, rather than email addresses.

**Potential problems**
Not sure about the security impact of shortening the tokens.

A clear and concise description of any alternative options you've considered.

**Additional context**
Related to these three issues:
https://github.com/nextauthjs/next-auth/issues/159
https://github.com/nextauthjs/next-auth/issues/709
https://github.com/nextauthjs/next-auth/issues/414

Particularly this response: https://github.com/nextauthjs/next-auth/issues/709#issuecomment-784077142

Yes!

mahieyin-rahmun:
## Reasoning 💡
This PR attempts to add the functionality for a SMS provider as per #1562. An SMS provider will work almost the same way as Email provider. However, instead of sending the login link via email, it will make use of OTP and send it over to the provided phone number of the user.

In the current state, the actual implementation depends on the choice of SMS/Verification provider selected by the developer, such as MessageBird, Twilio or equivalent. The general principle is:

1. User initiates sign in flow by navigating to the app and entering his/her phone number.
2. An implementation for a `sendSmsVerificationRequest` function must be provided in the SMS Provider options which takes the user's provided phone number as input and calls the appropriate API (MessageBird, Twilio etc) to create a verification request object and send the OTP to user's phone. This function should return the `id` of the created verification request object.
3. Simultaneously, user is redirected to the `verify-request` page to enter the OTP. An `HttpOnly` cookie is set that contains the `id` returned in Step 2. This cookie should only be sent along with the submitted OTP to the `api/auth/callback` endpoint and it should be set to expire as soon as the OTP itself has become invalid.
4. An implementation for `verifySmsVerificationRequest` function must be provided in the SMS provided options, which will take a  verification request ID and the OTP entered by the user as parameters. Then, the appropriate API (MessageBird, Twilio) should be called to check if the verification ID and token pair are valid. If everything checks out, the user is now logged in using his/her phone.

Caveats:
1. Currently, this only works with JWT sessions. No data is persisted in the database.
2. No other provider so far has made use of phone numbers. As such, in order to be able to store e.g. phone number of the user in database, a change in the database schema is required.
3. Since user enters phone number instead of email, unless `jwt` and `session` callbacks are overridden manually, session object will end up being empty.
4. Dependency on 3rd party API to send SMS means certain customization options such as length of OTP, OTP expiry etc can only be done if the API in question supports it.

Interested to hear everyone's opinions regarding this implementation and possible ways to improve it further.

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [ ] Ready to be merged

## Affected issues 🎟

Closes #1562. Related to #159, #709, #1465

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "next-auth",
       "version": "0.0.0-semantically-released",
       "funding": [
         {
@@ -84,6 +85,8 @@
       "integrity": "sha512-poegjhRvXHWO0EAsnYajwYZuqcz7gyfxwfaecUESxDujrqOivf3zrjFbub8IJkrqEaz3fvJWh001EzxBub54fg==",
       "dev": true,
       "dependencies": {
+        "@nicolo-ribaudo/chokidar-2": "2.1.8-no-fsevents.2",
+        "chokidar": "^3.4.0",
         "commander": "^4.0.1",
         "convert-source-map": "^1.1.0",
         "fs-readdir-recursive": "^1.1.0",
@@ -1866,6 +1869,9 @@
       "resolved": "https://registry.npmjs.org/jsonfile/-/jsonfile-4.0.0.tgz",
       "integrity": "sha1-h3Gq4HmbZAdrdmQPygWPnBDjPss=",
       "dev": true,
+      "dependencies": {
+        "graceful-fs": "^4.1.6"
+      },
       "optionalDependencies": {
         "graceful-fs": "^4.1.6"
       }
@@ -2298,6 +2304,7 @@
         "jest-resolve": "^26.6.2",
         "jest-util": "^26.6.2",
         "jest-worker": "^26.6.2",
+        "node-notifier": "^8.0.0",
         "slash": "^3.0.0",
         "source-map": "^0.6.0",
         "string-length": "^4.0.1",
@@ -5663,6 +5670,7 @@
       "dependencies": {
         "anymatch": "~3.1.2",
         "braces": "~3.0.2",
+        "fsevents": "~2.3.2",
         "glob-parent": "~5.1.2",
         "is-binary-path": "~2.1.0",
         "is-glob": "~4.0.1",
@@ -7206,6 +7214,9 @@
       "resolved": "https://registry.npmjs.org/jsonfile/-/jsonfile-4.0.0.tgz",
       "integrity": "sha1-h3Gq4HmbZAdrdmQPygWPnBDjPss=",
       "dev": true,
+      "dependencies": {
+        "graceful-fs": "^4.1.6"
+      },
       "optionalDependencies": {
         "graceful-fs": "^4.1.6"
       }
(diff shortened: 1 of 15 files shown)


balazsorban44 (commented):
Very interested to see this when I have more time! Just some quick suggestions.

I think instead of focusing only on SMS, we should call this type of provider Token, or Passwordless instead. once the basics are laid out (most of it already is thanks to the Email provider), we could create multiple providers for multiple services you mentioned. In v4 I intend to separate provider imports for smaller bundle size to the end user. any additional dependencies could be optional peer dependencies.

I would like that this will be based on the `next` branch though, as most of the work on v3 is done from my part. aiming for v4 release this summer.

mahieyin-rahmun:
> I think instead of focusing only on SMS, we should call this type of provider Token, or Passwordless instead. once the basics are laid out (most of it already is thanks to the Email provider), we could create multiple providers for multiple services you mentioned. 

Yeah, it is quite similar to the Email provider, so these type of providers could be very easily refactored under a single "Passwordless" type.

> I would like that this will be based on the `next` branch though, as most of the work on v3 is done from my part. aiming for v4 release this summer.

Great! Looking forward to v4, then. I will resume working on this once v4 is released. I also cannot wait to try out the features of v4 in my own projects :D

mtt87:
👋  Just saw this PR and I think it's really interesting!

Completely agree with @balazsorban44 
> I think instead of focusing only on SMS, we should call this type of provider Token, or Passwordless instead

I was actually thinking a couple of weeks ago to build a prototype for this functionality as I wanted to see if it's possible to use it with our product.


Self promotion:  (but it's relevant to the subject) 😄 
@mahieyin-rahmun bear in mind that using SMS OTP is **very insecure**.
SMS can be easily hijacked and people can be victim of a SIM swap attack.
If you are curious we are building a better alternative which is not only **more secure** but it's also much **more user friendly**, and it takes care both of number verification and sim swap check to prevent these attacks. 

https://tru.id/

followbl:
@mahieyin-rahmun awesome PR - really well documented. @balazsorban44 we would be super interested in this feature in a 4.0 release, let us know if there's anything we could help with to make this happen.

mtt87:
Same here, would be happy to help make it token agnostic so that could work with our use case as well, not only with insecure 🔓  SMS ancient technology 😄

balazsorban44:
have you all seen https://github.com/nextauthjs/next-auth/discussions/1465

mtt87:
Nope! Thanks for sharing 😄

PR #2295, opened 2021-07-01
test(client): fully cover client module

0ubbe:
## Reasoning 💡

Screenshot 2021-07-16 at 21 12 40

Final tests for `<SessionProvider />` and `useSession()` 🏋🏻‍♀️

## Checklist 🧢

- [ ] ~~Documentation~~
- [x] Tests
- [ ] ~~Ready to be merged~~

## Affected issues 🎟

None


ndom91 (approved):
LGTM 👍 

I didn't follow the thread 100%, on mobile currently, but I saw your TODO, just curious why can't we clean __NEXTAUTH between test runs?

balazsorban44:
Small comment I noticed that `dist/client/__tests__` are also getting in the package published to npm. we should have a look.

don't see why those are generated at all, we only run tests, if not possible to not generate them, should we just delete after a test run?

PR #2319, opened 2021-07-06
support yuque && add `authHeaders` option

djyde:
This PR support a new provider [Yuque](https://yuque.com), which is a documentation platform from China.

According to Yuque's API authorization [definition](https://www.yuque.com/yuque/developer/api), every secured API call (such as the profile API) should pass access token in header `X-Auth-Token`, but I can't find an option that can pass the access token as a header value. So I add an `authHeaders` option in this PR.

### DIFF
--- a/types/providers.d.ts
+++ b/types/providers.d.ts
@@ -33,6 +33,7 @@ export interface OAuthConfig<P extends Record<string, unknown> = Profile>
   requestTokenUrl?: string
   authorizationUrl: string
   profileUrl: string
+  authHeaders?(info: { accessToken: string }): Record<string, string>,
   profile(profile: P, tokens: TokenSet): Awaitable<User & { id: string }>
   protection?: ProtectionType | ProtectionType[]
   clientId: string
@@ -99,6 +100,7 @@ export type OAuthProviderType =
   | "WordPress"
   | "WorkOS"
   | "Yandex"
+  | "Yuque"
   | "Zoho"
   | "Zoom"
 
--- a/www/docs/providers/yuque.md
+++ b/www/docs/providers/yuque.md
@@ -1,0 +1,38 @@
+---
+id: yuque
+title: Yuque
+---
+
+## Documentation
+
+https://www.yuque.com/yuque/developer/api
+
+## Configuration
+
+https://www.yuque.com/settings/applications
+
+## Options
+
+The **Yuque Provider** comes with a set of default options:
+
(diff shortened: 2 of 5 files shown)


types/providers.d.ts
```diff
@@ -99,6 +100,7 @@ export type OAuthProviderType =
   | "WordPress"
   | "WorkOS"
   | "Yandex"
+  | "Yuque"
```
lluia, line 103: Can you add [type tests](https://github.com/nextauthjs/next-auth/blob/788d7dc41f8767a4659136475189d6d420295855/types/tests/providers.test.ts#L259) for **Yuque** as well? 😊

types/providers.d.ts
```diff
@@ -33,6 +33,7 @@ export interface OAuthConfig<P extends Record<string, unknown> = Profile>
   requestTokenUrl?: string
   authorizationUrl: string
   profileUrl: string
+  authHeaders?(info: { accessToken: string }): Record<string, string>,
```
lluia, line 36: You get any tokens returned from the provider in the second param of the `profile` callback:
```ts
  const data = await getDataFromApi(tokens.access_token)
  return {
    id: data.sub,
    name: data.userName,
    image: data.picture
  }
}
// ...
async getDataFromApi(token) {
  return fetch('path/to/api', {
    headers: {
      'X-Auth-Token': token
    }
  })
} 
```
Users can use that to make API calls to the provider. This is [documented in our options](https://next-auth.js.org/configuration/providers#options), although maybe we should highlight it better 🤔 as it's a common use case.

www/docs/providers/yuque.md
```diff
@@ -0,0 +1,38 @@
...
+  Providers.Yuque({
+    clientId: process.env.YUQUE_CLIENT_ID,
+    clientSecret: process.env.YUQUE_CLIENT_SECRET
+  })
+]
+...
+```
+
+:::warning
+Yuque doesn't expose user's email in API
```
lluia, line 37:
```suggestion
Yuque doesn't expose the user's email in their API
```

0ubbe (changes requested):
Thanks ✨ , few amendments before we can merge 😊

types/providers.d.ts
djyde: This is used in https://github.com/djyde/next-auth/blob/69f1559ced9bc130238fd09c85c3a01578c9e6eb/src/server/lib/oauth/client.js#L222

Issue #2316, opened 2021-07-05
the new useSession should use basePath settings for redirect


Naddiseo:
### Description 🐜

Follow-up for #2236 , as I was browsing the code changes I noticed that the [redirect in `useSession`](https://github.com/nextauthjs/next-auth/blob/next/src/client/react.js#L59) doesn't use the basePath settings that are set from `NEXTAUTH_URL`. We use a slightly different url paths in our project (`/api/v1/auth/...`) so we rely on everything being parsed from from `NEXTAUTH_URL`

EDIT: I'm also providing `basePath` to the `SessionProvider` to match the `NEXTAUTH_URL`

### Is this a bug in your own project?

No

### How to reproduce ☕️

- Change `NEXTAUTH_URL` to have a different path to usual: eg: `NEXTAUTH_URL=http://localhost:3000/api/v1/auth`
- Go to an authentication required page, the redirect will redirect to an invalid/404 page

### Environment 🖥

next-auth@next

### Contributing 🙌🏽

No, I am afraid I cannot help regarding this

balazsorban44:
So I'm fairly sure that this is also the current behavior, and that this issue is a duplicate of https://github.com/nextauthjs/next-auth/issues/1713

mahieyin-rahmun:
I tested this out at my local machine. Let me know if I got something wrong, but here is what I understood:

Assuming OP's description, this is sort of his folder structure:

```bash
.
├── pages
│   ├── api
│   │   └── v1
│   │       ├── auth
│   │       │   └── [...nextauth].js
... other files
```

and, he has set `NEXTAUTH_URL` in his .env file to be:
```env
NEXTAUTH_URL=http://localhost:3000/api/v1/auth
```

Now, with the way environment variables work, they won't be available client-side. If we take a look at the `parseUrl()` [implementation](https://github.com/nextauthjs/next-auth/blob/a2e5afa16252335a0186fce304ebd466207a38c7/src/lib/parse-url.js#L8), which is what `_apiBaseUrl()` relies on:

```js
// Default values
const defaultHost = 'http://localhost:3000'
const defaultPath = '/api/auth'

if (!url) { url = `${defaultHost}${defaultPath}` }
```
This right here is the problem. The hardcoded values should instead refer to the environment variable. But, since we do not have an environment variable access client-side, initially I thought 
> "maybe we can use `window.location.pathname` as the value of `defaultPath`", 

[...]

Issue #1712, opened 2021-04-13
using a base path does not work


martinratinaud:
**Describe the bug**
whenever I use a custom NEXTAUTH_URL along with a nextjs custom basePath, it leads to this error in the client

**Steps to reproduce**

```
git clone git@github.com:nextauthjs/next-auth.git
cd next-auth
npm i 
cp .env.local.example .env.local
```
change NEXTAUTH_URL in `.env.local` to `NEXTAUTH_URL=http://localhost:3000/preprod/api/auth`

Then go to `pages/api/auth/[...nextauth].js` and comment out all providers except Credentials to keep it simple

and add a `next.config.js` file

```
module.exports = {
  basePath: '/preprod',
};
```

And finally launch `npm run dev`

**Expected behavior**
The website loads correctly on http://localhost:3000/preprod and when I click on "Sign In", I should arrive on the signing page

**Screenshots or error logs**
Instead, clicking on the "Sign In" button leads to this

```
TypeError: right-hand side of 'in' should be an object, got null

src/client/index.js (205:8) @ _callee7$

  203 | 
  204 |   // Redirect to sign in page if no valid provider specified
> 205 |   if (!(provider in providers)) {
      |        ^
  206 |     // If Provider not recognized, redirect to sign in page
  207 |     window.location = `${baseUrl}/signin?callbackUrl=${encodeURIComponent(callbackUrl)}`
  208 |     return

```
This comes from the signIn method and it seems `providers` are null

Thanks for your help

balazsorban44:
Could you try this: https://github.com/nextauthjs/next-auth/issues/689#issuecomment-699927237

balazsorban44:
Closing in favor of #1713. Feel free to comment there!

Issue #1676, opened 2021-04-08
What to do when using Zones within NextJs


divslingerx:
**Your question**

Is it possible to set the default url for all auth requests?

**What are you trying to do**

I am currently building an app that utlizes zones. [zone example](https://github.com/vercel/next.js/tree/canary/examples/with-zones).
The site is split into a "site" and an "admin" zone.
the site zone runs on port 3000 and the admin zone runs on port 4000

going to localhost:3000/admin will redirect to the admin app running at localhost:4000 behind the scenes

Nextauth api route is located in /admin/pages/api/auth/[...nextauth]

When trying to use the hooks all request are made to localhost:3000/api/auth when I believe it should be localhost:3000/admin/api/auth
using the `NEXTAUTH_URL=http://localhost:3000/admin/api/auth` environent variable does not seem to work as I expected.

logic would say that moving NextAuth into the "site" zone would solve the problem however there is no auth needed for aspect of the site app so i would prefer to keep it in the admin app if possible.

* [ ] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [x] Could not find relevant documentation
* [ ] Found the example project helpful
* [ ] Did not find the example project helpful

balazsorban44:
Could you see if https://github.com/nextauthjs/next-auth/issues/689#issuecomment-699927237 helps?

Issue #1517, opened 2021-03-14
NextAuth doesn't show basePath


thebetterjort:
My setup works perfectly locally in docker, but when I try and deploy to a server behind a nginx proxy I'm not able to set the correct url for nextauth and it shows 404 error for 

/api/auth/session
/api/auth/_log

I've set this up in my .env.local:
NEXTAUTH_URL=https://example.com/fpd/api/auth

 next.config.js
module.exports = {
  basePath: '/fpd',

PR #2320, opened 2021-07-06
feat(client): improved basepath detection

### LINKED ISSUE #1713: Improve client-side basePath detection
Currently, when the `NEXTAUTH_URL` env variable contains a basePath also defined in next.config.js, the next-auth/client still cannot pick it up. This is because when Next.js builds the production bundle, it will ignore all environment variables in client code, that is not prepended with `NEXT_PUBLIC_` and inline them otherwise. This is described in their documentation here: https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables-to-the-browser

Iain's suggestion - where you pass `basePath` to the `options` prop of `Provider` - might work, but you will have to duplicate your basePath, so we might just want to make sure that we also read in `NEXT_PUBLIC_NEXTAUTH_URL` instead.

This will ensure that all the URLs have a single source of truth and that the actual path will be inlined at build-time.

**Potential problems**

Yet another variable might be confusing to users, but I don't see a better solution immediately. Open for suggestions. 

A solution is described here: https://github.com/nextauthjs/next-auth/issues/689#issuecomment-699927237, but it requires duplication. (Eg.: You already implicitly set `basePath` through `NEXTAUTH_URL`, it shouldn't be necessary to duplicate this, especially when we provide an isomorphic `next-auth/client` module.)

**Additional context**

Related: #689, #1712, #1517, #900, #499, #1676

I will wait for this to get feedback or if anyone has a better idea, please comment below!

mahieyin-rahmun:
## Reasoning 💡

This PR attempts to improve client side path detection by leveraging a second environment variable named `NEXT_PUBLIC_NEXTAUTH_URL` to ensure client side path detections are always consistent, regardless of folder structure and URL.

Currently, there are some issues with the client side path detection. e.g. [signIn()](https://github.com/nextauthjs/next-auth/blob/next/src/client/react.js#L59) function assumes that the folder structure is like the following:

```
test-project
    └── pages
        └── api
            └── auth
                └── [...nextauth].js
```
and `NEXTAUTH_URL` to be set to `http://localhost:3000` or `https://mydomain.com`. The above breaks if the URL and/or folder structure of the user's project is different.

With these changes, users will be able to use a different folder structure if they want, such as:

```
test-project
└── pages
    └── api
       └── v1
          └── authentication
             └── [...nextauth].js
```
And set `NEXTAUTH_URL` to something like `http://localhost:3000/api/v1/authentication`.  As compensation for the flexibility, they now need to declare a second environment variable:

```
NEXTAUTH_URL=http://localhost:3000/api/v1/authentication
NEXT_PUBLIC_NEXTAUTH_URL=http://localhost:3000/api/v1/authentication
```

Luckily, this can be simplified by using [variables expansion feature of Next.js](https://nextjs.org/docs/basic-features/environment-variables)
```
NEXTAUTH_URL=http://localhost:3000/api/authentication
NEXT_PUBLIC_NEXTAUTH_URL=$NEXTAUTH_URL
```

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [ ] Ready to be merged

## Affected issues 🎟

Closes #1713, #2316. Related to #689, #1712, #1517, #900, #499, #1676

### DIFF
--- a/src/client/__tests__/parse-url.test.js
+++ b/src/client/__tests__/parse-url.test.js
@@ -1,0 +1,59 @@
+import parseUrl from "../../lib/parse-url"
+
+// https://stackoverflow.com/questions/48033841/test-process-env-with-jest
+describe("parseUrl() tests", () => {
+  const OLD_ENV = process.env
+
+  beforeEach(() => {
+    jest.resetModules() // Most important - it clears the cache
+    process.env = { ...OLD_ENV } // Make a copy
+  })
+
+  afterAll(() => {
+    process.env = OLD_ENV // Restore old environment
+  })
+
+  test("when on client side and NEXT_PUBLIC_NEXTAUTH_URL is defined, correctly returns baseUrl and basePath", () => {
+    process.env.NEXT_PUBLIC_NEXTAUTH_URL = "http://localhost:3000/api/v1/auth"
+
+    const { baseUrl, basePath } = parseUrl()
+
+    expect(baseUrl).toBe("http://localhost:3000")
+    expect(basePath).toBe("/api/v1/auth")
+  })
+
+  test("when on client side and NEXT_PUBLIC_NEXTAUTH_URL is not defined, falls back to default values", () => {
+    const { baseUrl, basePath } = parseUrl()
+
+    expect(baseUrl).toBe("http://localhost:3000")
+    expect(basePath).toBe("/api/auth")
+  })
+
+  test("when on server side, correctly parses any given URL", () => {
+    let url = "http://localhost:3000"
+    let { baseUrl, basePath } = parseUrl(url)
+
+    expect(baseUrl).toBe("http://localhost:3000")
+    expect(basePath).toBe("/api/auth")
+
+    url = "http://localhost:3000/api/v1/authentication/"
+    // this semi-colon is needed, otherwise js thinks the string
+    // above is a function
+    ;({ baseUrl, basePath } = parseUrl(url))
+
+    expect(baseUrl).toBe("http://localhost:3000")
+    expect(basePath).toBe("/api/v1/authentication")
+
+    url = "https://www.mydomain.com"
+    ;({ baseUrl, basePath } = parseUrl(url))
+
+    expect(baseUrl).toBe("https://www.mydomain.com")
+    expect(basePath).toBe("/api/auth")
+
+    url = "https://www.mydomain.com/api/v3/auth"
+    ;({ baseUrl, basePath } = parseUrl(url))
+
+    expect(baseUrl).toBe("https://www.mydomain.com")
+    expect(basePath).toBe("/api/v3/auth")
+  })
+})
--- a/src/client/react.js
+++ b/src/client/react.js
@@ -56,7 +56,7 @@ export function useSession(options = {}) {
 
   React.useEffect(() => {
     if (requiredAndNotLoading) {
-      const url = `/api/auth/signin?${new URLSearchParams({
+      const url = `${_apiBaseUrl()}/signin?${new URLSearchParams({
         error: "SessionRequired",
         callbackUrl: window.location.href,
       })}`
@@ -337,6 +337,17 @@ function _apiBaseUrl() {
     // NEXTAUTH_URL should always be set explicitly to support server side calls - log warning if not set
     if (!process.env.NEXTAUTH_URL) {
       logger.warn("NEXTAUTH_URL", "NEXTAUTH_URL environment variable not set")
+    }
+
+    if (
+      process.env.NODE_ENV === "development" &&
+      !/^http:\/\/localhost:\d+$/.test(process.env.NEXTAUTH_URL) &&
+      !process.env.NEXT_PUBLIC_NEXTAUTH_URL
+    ) {
+      logger.warn(
+        "NEXT_PUBLIC_NEXTAUTH_URL",
+        `NEXTAUTH_URL is set to "${process.env.NEXTAUTH_URL}" instead of the default "http://localhost:3000", and NEXT_PUBLIC_NEXTAUTH_URL is not set. Client side path detections will fail.`
+      )
     }
 
     // Return absolute path when called server side
--- a/src/lib/parse-url.js
+++ b/src/lib/parse-url.js
@@ -5,23 +5,36 @@
  * supporting a default value, so a simple split is sufficent.
  * @param {string} url
  */
-export default function parseUrl (url) {
+export default function parseUrl(url) {
   // Default values
-  const defaultHost = 'http://localhost:3000'
-  const defaultPath = '/api/auth'
+  const defaultHost = "http://localhost:3000"
+  const defaultPath = "/api/auth"
 
-  if (!url) { url = `${defaultHost}${defaultPath}` }
-
+  if (!url) {
+    if (process.env.NEXT_PUBLIC_NEXTAUTH_URL) {
+      /* 
+        if the user has defined NEXT_PUBLIC_NEXTAUTH_URL, which should be mandatory
+        if they are using a different path other than 'http://localhost:3000'
+        in NEXTAUTH_URL variable in their .env file, use that variable instead, 
+        otherwise, the clientside path detection will fail.
+      */
+      url = process.env.NEXT_PUBLIC_NEXTAUTH_URL
+    } else {
+      // fallback to default values, user should see a warning if NEXTAUTH_URL is different
+      // and NEXT_PUBLIC_NEXTAUTH_URL is not set
+      url = `${defaultHost}${defaultPath}`
+    }
+  }
   // Default to HTTPS if no protocol explictly specified
-  const protocol = url.startsWith('http:') ? 'http' : 'https'
+  const protocol = url.startsWith("http:") ? "http" : "https"
 
   // Normalize URLs by stripping protocol and no trailing slash
-  url = url.replace(/^https?:\/\//, '').replace(/\/$/, '')
+  url = url.replace(/^https?:\/\//, "").replace(/\/$/, "")
 
   // Simple split based on first /
-  const [_host, ..._path] = url.split('/')
+  const [_host, ..._path] = url.split("/")
   const baseUrl = _host ? `${protocol}://${_host}` : defaultHost
-  const basePath = _path.length > 0 ? `/${_path.join('/')}` : defaultPath
+  const basePath = _path.length > 0 ? `/${_path.join("/")}` : defaultPath
 
   return { baseUrl, basePath }
 }
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -20,6 +20,17 @@ if (!process.env.NEXTAUTH_URL) {
   logger.warn("NEXTAUTH_URL", "NEXTAUTH_URL environment variable not set")
 }
 
+if (
+  process.env.NODE_ENV === "development" &&
+  !/^http:\/\/localhost:\d+$/.test(process.env.NEXTAUTH_URL) &&
+  !process.env.NEXT_PUBLIC_NEXTAUTH_URL
+) {
+  logger.warn(
+    "NEXT_PUBLIC_NEXTAUTH_URL",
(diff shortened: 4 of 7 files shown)


src/lib/parse-url.js
```diff
@@ -5,23 +5,36 @@
  * supporting a default value, so a simple split is sufficent.
  * @param {string} url
  */
-export default function parseUrl (url) {
+export default function parseUrl(url) {
```
lluia, line 8: This function is also called server-side, and the logic currently works because there we pass it `NEXTAUTH_URL`. 

I think we should move that logic inside, hence whenever calling `parseUrl` it can automatically compute what the right **base URL** and **base path** are:
```ts
export default function parseBaseUrl(url) {
// ..
  if (process.env.NEXTAUTH_URL) {
    if (!process.env.NEXT_PUBLIC_NEXTAUTH_URL) {
      throw new Error("[next-auth] Trying to use a custom `NEXTAUTH_URL` without setting `NEXT_PUBLIC_NEXTAUTH_URL`, client-side redirects won't work")
```
Now that we're modifying this function, I think we should:
- rename it as it's not generic utility about URLs but is intended to compute the base URL
- `throw` in case expectations don't match so we're explicit about mis-configurations

src/server/index.js
```diff
@@ -20,6 +20,17 @@ if (!process.env.NEXTAUTH_URL) {
...
+if (
+  process.env.NODE_ENV === "development" &&
+  !/^http:\/\/localhost:\d+$/.test(process.env.NEXTAUTH_URL) &&
+  !process.env.NEXT_PUBLIC_NEXTAUTH_URL
+) {
+  logger.warn(
+    "NEXT_PUBLIC_NEXTAUTH_URL",
+    `NEXTAUTH_URL is set to "${process.env.NEXTAUTH_URL}" instead of the default "http://localhost:\${PORT}", and NEXT_PUBLIC_NEXTAUTH_URL is not set. Client side path detections will fail.`
+  )
+}
```
lluia, line 32: In my opinion, we should move this logic to `parseUrl` and the check above too to start sliming down this monolith file. 

On the other hand, we should be throwing errors in both cases rather than warnings as this misconfiguration, I understand, is critical? 🤔

www/docs/getting-started/rest-api.md
```diff
@@ -59,4 +59,17 @@ e.g.
  `NEXTAUTH_URL=https://example.com/myapp/api/authentication`
 
 `/api/auth/signin` -> `/myapp/api/authentication/signin`
+
+However, you need to do two things:
+1. You must declare a `NEXT_PUBLIC_NEXTAUTH_URL` environment variable that points to the same URL as above.
+2. Your folder structure must match the above URL. e.g.
```
lluia, line 65:
```suggestion
2. Your folder structure must match what's set on `NEXT_PUBLIC_NEXTAUTH_URL`, for instance, if it's set to `https://mywebsite.com/api/authentication`, your folder structure should be:
```

src/client/__tests__/parse-url.test.js
```diff
@@ -0,0 +1,59 @@
...
+    let url = "http://localhost:3000"
+    let { baseUrl, basePath } = parseUrl(url)
+
+    expect(baseUrl).toBe("http://localhost:3000")
+    expect(basePath).toBe("/api/auth")
+
+    url = "http://localhost:3000/api/v1/authentication/"
+    // this semi-colon is needed, otherwise js thinks the string
+    // above is a function
+    ;({ baseUrl, basePath } = parseUrl(url))
```
lluia, line 42: As I understand we don't need to re-assign `url` here, why not just do:
```ts
const url1 = "http://localhost:3000"
const { baseUrl: baseUrl1, basePath: basePath1 } = parseUrl(url1)
// ...
const url2 = "http://localhost:3000"
const { baseUrl: baseUrl2, basePath: basePath2 } = parseUrl(url2)
```
I think will make this case a bit nicer to read 👍🏽

src/client/__tests__/parse-url.test.js
```diff
@@ -0,0 +1,59 @@
...
+
+    expect(baseUrl).toBe("https://www.mydomain.com")
+    expect(basePath).toBe("/api/auth")
+
+    url = "https://www.mydomain.com/api/v3/auth"
+    ;({ baseUrl, basePath } = parseUrl(url))
+
+    expect(baseUrl).toBe("https://www.mydomain.com")
+    expect(basePath).toBe("/api/v3/auth")
+  })
```
lluia, line 58: We probably want to add test cases for the unhappy paths as well 🙏🏽

src/client/react.js
```diff
@@ -56,7 +56,7 @@ export function useSession(options = {}) {
 
   React.useEffect(() => {
     if (requiredAndNotLoading) {
-      const url = `/api/auth/signin?${new URLSearchParams({
+      const url = `${_apiBaseUrl()}/signin?${new URLSearchParams({
```
lluia, line 59: I think we could use `parseUrl()` rather than having duplicate utils?

0ubbe (changes requested):
Awesome 💯  ✨ , I think we should think this twice, whether we really want to introduce another environment variable on the setup or not 👍🏽 

I left some cleanup suggestions as well 💬

src/client/__tests__/parse-url.test.js
```diff
@@ -0,0 +1,59 @@
...
+describe("parseUrl() tests", () => {
+  const OLD_ENV = process.env
+
+  beforeEach(() => {
+    jest.resetModules() // Most important - it clears the cache
+    process.env = { ...OLD_ENV } // Make a copy
+  })
+
+  afterAll(() => {
+    process.env = OLD_ENV // Restore old environment
```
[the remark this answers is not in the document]
mahieyin-rahmun: Makes sense.

src/lib/parse-url.js
mahieyin-rahmun: > I think we should move that logic inside, hence whenever calling parseUrl it can automatically compute what the right base URL and base path are

Yes, currently, it feels like it's dangling between both client and server.  

> throw in case expectations don't match so we're explicit about mis-configurations 

Agreed.

src/server/index.js
mahieyin-rahmun: > On the other hand, we should be throwing errors in both cases rather than warnings as this misconfiguration, I understand, is critical? 

I agree on this. I could not think of a better message, so I left it as warning, but it should definitely be treated as error.

src/client/__tests__/parse-url.test.js
mahieyin-rahmun: I wanted to do this but decided against it. But it seems like it would actually be better this way.

src/client/__tests__/parse-url.test.js
mahieyin-rahmun: Yup. That would probably make more sense once we have refactored parseUrl() to throw errors.

src/client/react.js
mahieyin-rahmun: I am not sure on this one. 

As far as I understand, `_apiBaseUrl()` checks the `__NEXTAUTH` object, which in turn makes use of the `parseUrl()` function. So technically it could be refactored.

I would leave it on the core team to make decisions on this and refactor accordingly.

src/lib/parse-url.js
```diff
@@ -5,23 +5,36 @@
...
   // Normalize URLs by stripping protocol and no trailing slash
-  url = url.replace(/^https?:\/\//, '').replace(/\/$/, '')
+  url = url.replace(/^https?:\/\//, "").replace(/\/$/, "")
 
   // Simple split based on first /
-  const [_host, ..._path] = url.split('/')
+  const [_host, ..._path] = url.split("/")
   const baseUrl = _host ? `${protocol}://${_host}` : defaultHost
-  const basePath = _path.length > 0 ? `/${_path.join('/')}` : defaultPath
+  const basePath = _path.length > 0 ? `/${_path.join("/")}` : defaultPath
```
Naddiseo, line 37: Out of curiosity, is this something that could be accomplished using [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL_API) ?

Something like:
```javascript
let baseUrl = defaultHost;
let basePath = defaultPath;
try {
  url = new URL(url, url.startsWith('http') ? undefined : defaultHost);
  baseUrl = url.origin;
  basePath = url.pathname.length ? url.pathname : defaultPath;
} catch (e) {}
```

mahieyin-rahmun: Interesting PoV. I will try and test this out. Looks like this could work, with some modifications.

src/client/react.js
lluia: Good point, I didn't realise `_apiBaseUrl()` calls `parseUrl()` internally, as you mention worth of a refactor in a different PR 👍🏽

balazsorban44:
I haven't had a closer look yet, but I would definitely keep a single env variable, not introduce a new one. adding `NEXT_PUBLIC_NEXTAUTH_URL` and removing `NEXTAUTH_URL` seems reasonable, but it feels like such a drastic change since it will require changing people's pipelines etc. I know this will be a major release, but still.

Another way of inlining (making it available client-side) an env. the variable is to use the `env` option in the `next.config.js` file. This way changing the variable name wouldn't be necessary, we would only need to inform users about this.

I think only a very few of our users will actually need this change at all. Most of the time `/api/auth` is sufficient to be assumed.

I can be convinced otherwise, but I don't think we should keep both variables either way.

mahieyin-rahmun:
> I haven't had a closer look yet, but I would definitely keep a single env variable, not introduce a new one. adding `NEXT_PUBLIC_NEXTAUTH_URL` and removing `NEXTAUTH_URL` seems reasonable, but it feels like such a drastic change since it will require changing people's pipelines etc. I know this will be a major release, but still.
> 
I agree about the one variable part, and I understand about it being a big change, but I would argue that with the amount of breaking changes in **v4** already, people would have to shift around bits of their codebase anyway. An environment variable, IMHO, should not be that big of a deal. Those who don't want to, can stick with **v3**, which seems like a reasonable tradeoff. And most new users would start off with **v4** anyway.

> Another way of inlining (making it available client-side) an env variable is to use the `env` option in the `next.config.js` file [ref](https://nextjs.org/docs/api-reference/next.config.js/environment-variables). This way changing the variable name wouldn't be necessary, we would only need to inform users about this. Although I am not sure if this is something Next.js will keep around.
> 
Yes, they introduced `NEXT_PUBLIC_` prefix in lieu of this, and they recently did away with `webpack v4` in favor of `v5`, so I don't think it will stay in the long run either.

> I think only a very few of our users will actually need this change at all. Most of the time `/api/auth` is sufficient to be assumed.

IMHO, there are quite a few issues that talk about this, and although it probably is a small portion of the userbase, it would be a good addition, since it keeps everything dynamic as opposed to hardcoded or working on assumptions.

lluia:
@mahieyin-rahmun I've been speaking about this with @balazsorban44, we think adding `NEXT_PUBLIC_NEXTAUTH_URL` **v4.0.0** would be too disruptive, as consumers could easily bypass the update and end in quite a few broken pipelines 😅

We think it's something we could add as a feature in a later **v4** release, like `4.x.0` for example, and keep both environment variables around until **v5** and that would give enough time to people to migrate, and time for us to educate about it. 

Meanwhile, you can use one of the two solutions mentioned here:
https://github.com/nextauthjs/next-auth/issues/689#issuecomment-699927237

- use `env` field of `next.config.js`
- supply it as a `basePath` to `<Provider />` 

that would get you going 👍🏽

src/client/react.js
lluia: @mahieyin-rahmun could you move this change to a different PR please? 🙏🏽

lluia:
@mahieyin-rahmun, we're gonna close this 👍🏽 , but we keep an eye on it  👀  for when we add this back on `4.x.0` 🙏🏽  , super thanks for the PR 💚 💯

mahieyin-rahmun:
> @mahieyin-rahmun, we're gonna close this 👍🏽 , but we keep an eye on it  for when we add this back on `4.x.0` 🙏🏽 , super thanks for the PR

@lluia, Sure, let me know when it is time, I am willing to open another PR with the suggested refactors/modifications in this PR.

PR #2325, opened 2021-07-08
fix(oauth): correctly remove `code_verifier` cookie when used

pbr1111:
## Reasoning 💡

Once the pkce code verifier is used, the cookie is cleared but the browser blocks the deletion of this cookie, keeping it in the browser until it expires.

image

This is because when using the cookie with the useSecureCookies option enabled (https environment), the cookie is prefixed with [__Secure-](https://datatracker.ietf.org/doc/html/draft-west-cookie-prefixes-05#section-3.1). and it must use the `cookies.pkceCodeVerifier.options` options to be valid (the secure attribute must be set to true). 

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/src/server/lib/oauth/pkce-handler.js
+++ b/src/server/lib/oauth/pkce-handler.js
@@ -36,7 +36,11 @@ export async function handleCallback (req, res) {
       pkceLength: PKCE_LENGTH,
       method: PKCE_CODE_CHALLENGE_METHOD
     })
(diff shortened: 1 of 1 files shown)


balazsorban44:
Appreciate this PR, good catch! I'll try this out and merge if all is well! 🙏

PR #2342, opened 2021-07-10
feat(events): use named params for all event callbacks

balazsorban44:
Unified API for all of our user-facing methods.

NOTE: `events.error` has been removed. This method has never been called in the core, so it did actually nothing. If you want to log errors to a third-party, check out the [`logger`](https://next-auth.js.org/configuration/options#logger) option instead.

BREAKING CHANGE:

Two event signatures changed to use named params, `signOut` and `updateUser`:
```diff
// [...nextauth].js
...
events: {
- signOut(tokenOrSession),
+ signOut({ token, session }), // token if using JWT, session if DB persisted sessions.
- updateUser(user)
+ updateUser({ user })
}
```


0ubbe (approved):
Awesome 👏🏽 💯

PR #2344, opened 2021-07-10
feat(logger): simplify `logger` API

### LINKED ISSUE #1602: Expanded Logging Customisation
Additional boolean flags in options to be used in conjunction with [logger](https://next-auth.js.org/configuration/options#logger) object that allows for disables the sending of log events to the `_log` endpoint.

Serverless function invocation costs per request on Vercel. For customers that implement client and server-side exception tracking services (such as Sentry) already pay for these exceptions to be captured on said services. Having the exceptions also invoking the `_log` function is redundant in this scenario and would incur additional, unnecessary costs and double up on alerting.

Offer the ability to set `_log endpoint` or similar to `false` which disables it from being created in Vercel and disables the client and server parts of next-auth from attempting to `POST` data to the endpoint when an exception is caught. Customers are then encouraged to use the pre-existing `logger` object to capture and alert with their own logging platform.

**Potential problems**
None that I can think of.

Attempting to link Sentry to DataDog to capture `_log` events but this seems backward.

**Additional context**
I can obviously try and help with this implementation.

balazsorban44:
Similar to #2342, this aims to unify the user-facing API and provide an easier way to extend in the future.

In addition, this PR also solves the problem when the `logger.error` method sometimes did not print results, because `Error` instances are not serializable and will be printed as empty objects `"{}"`.

After this PR, we make any `Error` instances serializable as described here: https://iaincollins.medium.com/error-handling-in-javascript-a6172ccdf9af

Closes #1602
Achieved by adding a `client: true` flag when logs are coming from the frontend.

BREAKING CHANGE:

The main change is that instead of an unknown number of parameters, the log events have at most two, where the second parameter is usually an object. In the case of the `error` event, it can also be an `Error` instance (that is serializable by `JSON.stringify`). If it is an object, an `Error` instance will be available on `metadata.error`, and `message` will default to `metadata.error.message`. This is done so that an error event always provides some kind of a stack to see where the error happened

```diff
// [...nextauth.js]
import log from "some-logger-service"
...
logger: {
- error(code, ...message) {},
+ error(code, metadata) {},
- warn(code, ...message) {},
+ warn(code) {}
- debug(code, ...message) {}
+ debug(code, metadata) {}
}
```


src/client/react.js
```diff
@@ -320,25 +320,23 @@ export function SessionProvider(props) {
...
   }
 }
 
 function _apiBaseUrl() {
   if (typeof window === "undefined") {
-    // NEXTAUTH_URL should always be set explicitly to support server side calls - log warning if not set
-    if (!process.env.NEXTAUTH_URL) {
-      logger.warn("NEXTAUTH_URL", "NEXTAUTH_URL environment variable not set")
-    }
-
```
balazsorban44, line 341: We already report this server-side, I don't see the reason to do it client-side also.

src/client/react.js
```diff
@@ -320,25 +320,23 @@ export function SessionProvider(props) {
...
+    const res = await fetch(`${_apiBaseUrl()}/${path}`, options)
     const data = await res.json()
     if (!res.ok) throw data
     return Object.keys(data).length > 0 ? data : null // Return null if data empty
   } catch (error) {
-    logger.error("CLIENT_FETCH_ERROR", path, error)
+    logger.error("CLIENT_FETCH_ERROR", {
+      error,
+      path,
+      ...(req ? { header: req.headers } : {}),
```
lluia, line 332:
```suggestion
      ...(req ? { headers: req.headers } : {}),
```

types/index.d.ts
```diff
@@ -213,9 +213,17 @@ export type Theme = "auto" | "dark" | "light"
  * [Documentation](https://next-auth.js.org/configuration/options#logger)
  */
 export interface LoggerInstance {
-  warn(code: string, ...message: unknown[]): void
-  error(code: string, ...message: unknown[]): void
-  debug(code: string, ...message: unknown[]): void
+  warn(code: string): void
```
lluia, line 216: Let's add here the list of possible warning codes 👍🏽

0ubbe (approved):
Super 💪🏽 💯

PR #2349, opened 2021-07-11
docs: clarify protecting pages in _app

### LINKED ISSUE #2187: Custom _app and getServerSideProps
## Question 💬

I think the docs may be a bit misleading on how to do session checks for all pages. In [this section](https://next-auth.js.org/getting-started/client#provider) they suggest using `getServerSideProps` in `_app`, while the [NextJS docs in this section](https://nextjs.org/docs/advanced-features/custom-app#caveats) states that `getServerSideProps` does not work with custom `_app` file. From my experience the suggested approach in the NextJS Auth docs does not work.

Could I be doing it wrong, or am I correct in assuming the suggested approach in NextJS Auth docs will not work. If so they should be changed.

ndom91:
## Reasoning 💡

Small changes to clarify the documentation regarding protecting all pages in `_app.jsx`.

## Checklist 🧢

- [x] Documentation
- ~[ ] Tests~
- [x] Ready to be merged

## Affected issues 🎟

Fixes #2187


balazsorban44:
Let's do this against `next`

PR #2354, opened 2021-07-11
Include profile on signIn events

t3dotgg:
`profile` added to `event.signIn` message

## Reasoning 💡

I use the profile information to fire updates on the user object after any successful sign-in. Before this change, I had to stuff handlers to sync `profile.name` and `platform` against the `User` model in 3 separate places. This change allows me to have a complete "source of truth" within the `signIn` event :)

## Checklist 🧢

- [ ] Documentation (likely needed, will add if y'all like the changes)
- [ ] Tests (not sure if needed for this)
- [ ] Ready to be merged

I would like some thoughts from the team on this solution before cleanup + merge. I am happy to make any docs changes and add tests if this is a solution y'all are happy with :)

## Affected issues 🎟

Didn't file one, went straight to source :)

### DIFF
--- a/src/server/routes/callback.js
+++ b/src/server/routes/callback.js
@@ -139,7 +139,7 @@ export default async function callback(req, res) {
           })
         }
 
-        await dispatchEvent(events.signIn, { user, account, isNewUser })
+        await dispatchEvent(events.signIn, { user, account, profile, isNewUser })
 
         // Handle first logins on new accounts
         // e.g. option to send users to a new account landing page on initial login
@@ -286,7 +286,7 @@ export default async function callback(req, res) {
         })
       }
 
-      await dispatchEvent(events.signIn, { user, account, isNewUser })
+      await dispatchEvent(events.signIn, { user, account, profile, isNewUser })
 
       // Handle first logins on new accounts
       // e.g. option to send users to a new account landing page on initial login
@@ -409,7 +409,7 @@ export default async function callback(req, res) {
       ...cookies.sessionToken.options,
     })
 
-    await dispatchEvent(events.signIn, { user, account })
+    await dispatchEvent(events.signIn, { user, account, profile: null })
 
     return res.redirect(callbackUrl || baseUrl)
   }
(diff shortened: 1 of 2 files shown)


balazsorban44:
Thank you for the PR! I guess it's a legit change, but I'm curious if you need this change because adapters currently only save a few fields in the database?

In that case, this might be relevant https://github.com/nextauthjs/adapters/issues/151

In v4, we will try to make it easier to save anything on the User model and document on how to change your DB schema to support that.

this change is still fine I think, but hopefully your usecase can be handled more easily in the future.

TheoBr:
> Thank you for the PR! I guess it's a legit change, but I'm curious if you need this change because adapters currently only save a few fields in the database?
> 
> In that case, this might be relevant [nextauthjs/adapters#151](https://github.com/nextauthjs/adapters/issues/151)
> 
> In v4, we will try to make it easier to save anything on the User model and document on how to change your DB schema to support that.
> 
> This PR is still fine I think, but hopefully your usecase can be handled more easily in the future.

@balazsorban44 ty so much for the quick reply! That issue is also very much within my interests 👀 

Storing custom fields would be helpful, but would likely put us in a place of "stale data" pretty quick (i.e. user changes display name on Twitch)

This solution has been the quickest I've found thusfar. Resulting UX is super clean too IMO - just "re-auth" to update your name. I have a video demo here:

https://twitter.com/t3dotgg/status/1414135514169430017

TheoBr:
Oh also - I am using this change right now via `patch-package` so I'm "unblocked". Happy to migrate to a better solution once one is determined/released :)

balazsorban44:
Saw the demo, nice! I think this is a fine addition no matter the reason behind it, so if it's fine by you and you could do this against `next` instead, I'll happily merge it. 😊

TheoBr:
Tried switching base branch on this one and it was a mess (looks like the event callback pattern changed, I prefer the new one though 😄 )

New PR here: https://github.com/nextauthjs/next-auth/pull/2356

PR #2356, opened 2021-07-11
feat(events): include `profile` on `signIn` events

t3dotgg:
Redo of https://github.com/nextauthjs/next-auth/pull/2354 rebased on `next` branch

_below is copy-pasted from last PR_

## What Changed
`profile` added to `event.signIn` message

## Reasoning 💡

I use the profile information to fire updates on the user object after any successful sign-in. Before this change, I had to stuff handlers to sync `profile.name` and `platform` against the `User` model in 3 separate places. This change allows me to have a complete "source of truth" within the `signIn` event :)

## Checklist 🧢

- [x] Documentation (likely needed, will add if y'all like the changes)
- [ ] Tests (not sure if needed for this)
- [x] Ready to be merged

I would like some thoughts from the team on this solution before cleanup + merge. I am happy to make any docs changes and add tests if this is a solution y'all are happy with :)

## Affected issues 🎟

Didn't file one, went straight to source :)

### DIFF
--- a/src/server/routes/callback.js
+++ b/src/server/routes/callback.js
@@ -396,7 +396,7 @@ export default async function callback(req, res) {
       ...cookies.sessionToken.options,
     })
 
-    await events.signIn({ user, account, profile: null })
+    await events.signIn({ user, account })
 
     return res.redirect(callbackUrl || baseUrl)
   }
--- a/src/server/routes/callback.js
+++ b/src/server/routes/callback.js
@@ -132,7 +132,7 @@ export default async function callback(req, res) {
           })
         }
 
-        await events.signIn({ user, account, isNewUser })
+        await events.signIn({ user, account, profile, isNewUser })
 
         // Handle first logins on new accounts
         // e.g. option to send users to a new account landing page on initial login
@@ -274,7 +274,7 @@ export default async function callback(req, res) {
         })
       }
 
-      await events.signIn({ user, account, isNewUser })
+      await events.signIn({ user, account, profile, isNewUser })
 
       // Handle first logins on new accounts
       // e.g. option to send users to a new account landing page on initial login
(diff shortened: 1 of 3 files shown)


TheoBr:
Just updated docs as well @balazsorban44 , let me know if anything else is needed from me :)

src/server/routes/callback.js
```diff
@@ -396,7 +396,7 @@ export default async function callback(req, res) {
       ...cookies.sessionToken.options,
     })
 
-    await events.signIn({ user, account })
+    await events.signIn({ user, account, profile: null })
```
balazsorban44, line 399: Should be `userObjectReturnedFromAuthorizeHandler` (really sorry for the variable name... 😅) and so please update the documentation and TS types to reflect this. 😊

UPDATE:
Hmm. Revisited the code and actually that variable is used for `user` as well. Should we have the same value twice, or remove profile and mention in the docs that they have user? What does make the most sense in your opinion?

TheoBr: My mental model associates `profile` with "the thing the auth platform returns", so IMO nulling it out and pointing them to `user` makes more sense.

I'm also _relatively new here_ and _don't use the credentials flow at all_ so I have no strong opinions here

balazsorban44: Understandable! In that case, I think I would just drop `profile: null` and say in the docs that `user` is what they are looking for if it's a `credentials` provider. If for some reason we add `profile` later on, we could do it in a non-breaking manner then.

TheoBr: Just updated @balazsorban44 , ty for the feedback!

0ubbe (approved):

balazsorban44 (approved):
Thank you! 🙂

PR #2357, opened 2021-07-12
Use navigator.locks to prevent multiple getSession

mtt87:
## Reasoning 💡

Using [Web Locks API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API) to prevent triggering multiple `getSession` broadcast that create issues in a refresh token scenario.
https://github.com/nextauthjs/next-auth/issues/2071

Since the API is still experimental and only supported by Chrome, I've added a `navigator.locks` polyfill (less than ~1kb gzipped) that supports other browser with a fallback on `localStorage`.

The library seems to be well tested according to the specification https://github.com/aermin/web-locks

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

I have not included tests as it's still a bit unclear what's the plan with tests but If we decide that we wanna include e2e tests using Playwright then I have a test ready to be included
https://github.com/mtt87/next-auth/blob/feat/client-config/app/e2e/refresh_token.test.ts

I tested manually to login with the credentials provider and checked all the different scenario (client/server) and it seems to be working fine.

## Affected issues 🎟

It partially fixes https://github.com/nextauthjs/next-auth/issues/2071. 
I say partially because it fixes all the scenario on the client but it cannot fix server scenarios as there is no lock mechanism that can work there unless you introduce a database or persistence layer.
We could add in the documentation some info on the refresh token scenario that suggests the user to never use `getSession` on the server side but only `getToken` (of course it has some trade-offs and limitations)


balazsorban44:
my problem with this API is that it's yet another [Chrome non-standard](https://twitter.com/dmitriid/status/1399409007878619137?s=19) and also experimental, so it might be very premature of us to introduce it in this form, what do you think?

balazsorban44 (commented):
Not sure why all these new peer dependencies in this PR. 😳 Adapter's are being moved out of the core, so these are not  necessary for the core anymore. please revert the changes, also in the lock file. 🙏

mtt87:
> my problem with this API is that it's yet another [Chrome non-standard](https://twitter.com/dmitriid/status/1399409007878619137?s=19) and also experimental, so it might be very premature of us to introduce it in this form, what do you think?

I share the same feeling with you when it comes to "Chrome using non standard APIs" especially when they don't feel like they are needed or the community is moving in a totally different direction (web-components/shadow DOM etc).
In this scenario the functionality that is provided by this experimental API seems much needed for scenarios like the one we are facing (refresh token on multiple tab concurrently) and given the popularity of next.js I believe in the future there will be more people facing this issue.
Also we are not locking in anyone to use Chrome as this is a polyfill and the impact seems to be pretty minimal (less than 1kb gzipped, compared for example to 33kb polyfill for web components)

What options do we have?

1) use this `navigator.locks` polyfill library
**pros**: when other browser will eventually support this API, it will use automatically the native implementation, if not it will always rely on the `localStorage` fallback. It's available today with the `navigator.locks` package which is well tested.
**cons**: experimental API only supported by Chrome at the moment

2) build something from scratch that always uses `localStorage` so it's natively supported by all browsers
**pros**: it's not using an experimental API
**cons**: it's using a fallback even when it could use a native implementation (which is obviously better). Although we can take inspiration from the `navigator.locks` polyfill library it will basically need to be built from scratch and published.


> Not sure why all these new peer dependencies in this PR. 😳

They are just re-ordered alphabetically, it's done automatically by `npm` when installing a new library.

balazsorban44:
You might have forked an old version then?

https://github.com/nextauthjs/next-auth/blob/next/package.json#L75-L79

mtt87:
Yes apologies, I saw this and didn't press the button somehow 😆 
Screenshot 2021-07-12 at 17 13 12
Gonna fix it now

mtt87:
@balazsorban44 I've fixed it, apologies for the mistake 😄

lluia:
I see a few drawbacks using this [`navigator.locks`](https://www.npmjs.com/package/navigator.locks) polyfill as it's today. 

The library is under `1.0.0`, which means it is experimental, and any version can come with breaking changes ( which [seems to be the case](https://github.com/aermin/web-locks/pull/33/files) ) 😅 . That would require us to be extra careful when updating it. I'm also against using an experimental API that only works in Chrome 🤔

I'd say investigating the `localStorage` solution might be more bulletproof 💭

mtt87:
It feels a bit weird/inefficient to not use the `navigator.locks` API when available (Chrome web/mobile, Edge, Opera)
Screenshot 2021-07-13 at 09 27 51

Regarding the library I think nobody used it yet as it's fairly new proposal (the draft is dated 17th of June 2021) and the creator used `0.3.1` version probably because it's not a standard yet.

FYI I've also considered https://www.npmjs.com/package/broadcast-channel with the React hook to obtain a broadcast leader https://gist.github.com/tannerlinsley/a7f0db4d02f4a3780baa4842ff8d5632 but these are the considerations I've made:
- bigger library size, react hook to be added on the session provider slightly increasing the size of all pages (as it would live in `_app.js`)
- much more complex to integrate as it would need to replace the current `next-auth` broadcast implementation, handle the leader election and basically change the broadcasting of messages to delegate a leader to refresh the session.

vs `navigator.locks` polyfill:
- smaller library size
- very simple integration (2 lines of code)
- feels like a "lock mechanism" is the correct thing to do and it doesn't interfere with the current broadcasting implementation 


Anyway I can extract the `localStorage` fallback logic into a separate library (or google first if there is something already built out there) if that's easier.

balazsorban44:
I would pretty much like to wait with this addition, to be honest. We are currently focusing on shipping v4, and this is something we could introduce in a later non-breaking release. Maybe until then, even the Google draft could get some support behind it from other vendors, or the polyfill gets noticed a bit more.

The intention is very good and I see the problem, but I would like to wait with tackling it to be able to focus on more important manners. I simply don't have the free time to follow this up properly. 🙏

mtt87:
Okay no problem I understand your point. Let's wait 👍

balazsorban44:
Let's revisit this after v4 has landed in stable if there is still interest then.

I believe we are less than a month away from a stable release.

mtt87:
Sonds good, I've been actually testing v4 beta this week plus the navigation.locks but did only limited testing.

Balhut:
Any update on implementing this feature?

Misiu:
image
all browsers supports `navigator.locks`, v4 was released long ago, but the problem is still actual.
@balazsorban44 could you please take a look at this? You closed this PR, so maybe there are other ways to fix this issue?

mtt87:
If you are curious I have implemented it in this fork https://github.com/tru-ID/next-auth

Misiu:
@mtt87 thank you for such a quick reply, I'm going to try your folk.
Ideally, this should be added to the main project, so all can benefit.
Let's wait for members decision. I hope this will get solved.
I wonder how others are fixing this kind of bugs in their code.

vanenshi:
@mtt87 Thanks for the fix, we used a little different solution in our project to fix this problem. we also use the lock to prevent the refresh on the client side, since we wanted to track the refresh token with the client IP (for security), so for us it was a requirement.

asif-jalil:
@ubbe-xyz this feature is really necessary for custom provider. is there any update for this?

asif-jalil:
@mtt87 is there any alternate solution?

Issue #876, opened 2020-11-23
A network partition can make it impossible for a new user to sign up


abeluck:
**Describe the bug**

There is a transactional bug in the callback handler, specifically here:

https://github.com/nextauthjs/next-auth/blob/64084d634b0c71d37143be83e50601a0a4f773a6/src/server/lib/callback-handler.js#L194-L205

If `createUser` succeeds, but then `linkAccount` fails due to a networking issue for example, then it becomes impossible for a user to later login/sign up.

The user is created, but no linked accounts exist.

On later runs through the flow [`isSignedIn`](https://github.com/nextauthjs/next-auth/blob/64084d634b0c71d37143be83e50601a0a4f773a6/src/server/lib/callback-handler.js#L138) will be `false`, but a user with the email address will exist resulting in:

https://github.com/nextauthjs/next-auth/blob/64084d634b0c71d37143be83e50601a0a4f773a6/src/server/lib/callback-handler.js#L177-L186

**Steps to reproduce**

You can reproduce this by providing a `signIn` event that uses `setTimeout` to "sleep" for a long time. Long enough for you to stop the test database to simulate a network partition.

After the flow fails the first time, start the database again and attempt to sign up again.

**Expected behavior**

A network partition event shouldn't result in making it impossible for a user to sign up.

**Screenshots or error logs**

n/a

**Additional context**

n/a

**Feedback**
*Documentation refers to searching through [online documentation](https://next-auth.js.org), code comments and issue history. The example project refers to [next-auth-example](https://github.com/iaincollins/next-auth-example).*

* [ ] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [x] Could not find relevant documentation
* [ ] Found the example project helpful
* [ ] Did not find the example project helpful

abeluck:
I think the correct solution is for there to be a new adapter function that creates the user and links the account all at once. This will allow the underlying implementation to use a transaction.

iaincollins:
I suggest configuring an email adaptor, not least as users can lose access to accounts at third party services. Supporting email sign in prevents people from not being able to sign up in this scenario - and in the event their provider account is suspended.

I'm not particularly concerned about the scenario of handling rolling back in the unlikely event of a network problem mid way through sign up; if that is actually happening frequently enough to a be problem then I suggest investigating that issue with your service.

I'm happy to accept a patch if someone is willing to do it and provide tests for it though.

[...]

Issue #779, opened 2020-10-16
[email provider] invalidate previous login links


slaivyn:
**Summary of proposed feature**
Any call to signIn function/REST API endpoint of the email provider should invalidate any previously created login link.

**Purpose of proposed feature**
The goal is to increase the security level of this provider.

**Potential problems**
I can't see any issue with that change proposal

**Describe any alternatives you've considered**
I can't find any alternative

**Additional context**
This is a recommendation from a security auditor based on WASP guidelines.

I am a bit busy right now but later I could try to implement it myself if the feature is accepted

iaincollins:
Thanks for raising this feature!

I think this is a good suggestion and we should take it into account when working on the email provider - probably adding it at the same time as we look at rate limiting and/or adding support for short validation codes (in addition to the links).

balazsorban44:
In #2361, I am going to make it required for the user to delete the previous token used for signin, no matter if the signin was successful or not.

See `useVerificationToken`:

https://github.com/nextauthjs/next-auth/blob/79b3ab73d145be5e87a1845b0cf2106cb2ef2303/src/server/routes/callback.js#L182

If I understand the proposal here, that, should be addressed jn that PR.

slaivyn:
that sounds great!

balazsorban44:
This has been released in `4.0.0-next.22`! From that version and up, adapters are expected to only return the verification token once, and delete it from the database after usage. See the `useVerificationToken`

PR #2361, opened 2021-07-12
feat: simplify Adapter API

balazsorban44:
In this PR, I am significantly simplifying the `Adapter` API. Hopefully, this will result in a better DX (as adding new adapters will require less boilerplate) and a better UX, as we will now merge database queries into single transactions where possible for fewer database calls and thus faster responses. 🔥

If you are using an [official adapter](https://github.com/nextauthjs/adapters/tree/main/packages), you won't need to do any changes.

- [x] Session methods
- [x] User methods
- [x] Verification token methods
- [x] Docs
- [x] Tests
- [x] - remove `getAdapter`
- [x] Create an experimental release and try with an adapter

Related issues: #779, #876

BREAKING CHANGE:

`prisma-legacy` is now gone. Use `@next-auth/prisma-adapter`. Any features from the old adapter will be migrated over to the new one eventually. This is done so we can require the same default set of options from all the built-in providers, rather than allowing ambiguity on what an official adapter has to support. The `TypeORM` adapter will probably be the only one migrated as-is, but in the future we would like to break it down to lighter-weight adapters that only support single databases.

Adapters no longer have to return a `getAdapter()` method, they can return the actual adapter methods instead. All the values previously being provided through the arguments of `getAdapter` will now be available in a more digestible format directly in the concerning methods. This behavior was created so that connections could be handled more efficiently. Our review has shown that currently, the TypeORM adapter is the only one that does not handle connections out-of-the-box, so we are going to look into how we can create a wrapper/util function to make it work in the new version. For all other adapters, this will be a huge gain, as with this new API, methods are actually overrideable without creating a whole new custom adapter! 🥳

Example:

```js
function MySlightlyCustomAdapter(...args) {
  const adapter = AdapterFromSomeoneElse(...args)
  adapter.someMethodIWantToModify = (...args) => {
    // Much better implementation goes here.
  }
  return adapter
}
```

**The following method names are changing:**

```diff
- getSession
+ getSessionAndUser
```
This method now requires that you return both the user and the session as `{user, session}`. If any of these could not be retrieved, you will have to return `null` instead. (In other words, this must be a transaction.) This requires one less database call, improving the user session retrieval. Any expiry logic included in the Adapter before is now done in the core as well.

```diff
- createVerificationRequest
+ createVerificationToken
```
Better describes the functionality. This method no longer needs to call `provider.sendVerificationRequest`, we are moving this into the core. This responsibility shouldn't have fallen to the adapter in the first place.

`createVerificationToken` will now receive a `VerificationToken` object, which looks like this:
```ts
interface VerificationToken {
  identifier: string
  expires: Date
  token: string
}
```

The token provided is already hashed, so nothing has to be done, simply write it to your database. (Here we lift up the responsibility from the adapter to hash tokens)

```diff
- getVerificationRequest
+ useVerificationToken
```
Better describes the functionality. It now also has the responsibility to delete the used-up token from the database. Most ORMs should support retrieving the value while deleting it at the same time, so it will reduce the number of database calls.

``` diff
- deleteVerificationRequest
```
This method is gone. See `useVerificationToken`.

Most of the method signatures have been changed, have a look at the [TypeScript interface](https://github.com/nextauthjs/next-auth/blob/ba4ec5faa357457661ccf55a4981fcef0514dd9b/types/adapters.d.ts) to get a better picture.


types/adapters.d.ts
```diff
@@ -1,6 +1,14 @@
...
 import { EmailConfig } from "./providers"
+import { Awaitable } from "./internals/utils"
+
+export interface AdapterUser extends User {
+  emailVerified?: Date | null
+}
+
+export interface AdapterSession extends Omit<Session, "expires"> {
+  expires: Date
+}
```
balazsorban44, line 11: `Session["expires"]` is a string by default so it can easily be passed to the client, but it makes more sense to retrieve it as `Date` from DB.

See: https://github.com/vercel/next.js/discussions/11498

types/adapters.d.ts
```diff
@@ -1,6 +1,14 @@
-import { InternalOptions } from "./internals"
-import { User, Profile, Session } from "."
+import { User, Session } from "."
 import { EmailConfig } from "./providers"
+import { Awaitable } from "./internals/utils"
+
+export interface AdapterUser extends User {
+  emailVerified?: Date | null
```
balazsorban44, line 6: `emailVerified` is only used internally, so the default `User` interface never exposes it.

balazsorban44:
Use `sessionToken` instead of `id` directly to get a session from the database.

cc @iaincollins

balazsorban44:
@ndom91 @lluia As with the OAuth config improvements, I would like to go through this PR with both of you, once you have the time. :eyes:

www/docs/adapters/models.md
```diff
@@ -3,60 +3,48 @@ id: models
 title: Models
 ---
```
balazsorban44, line 5: I think this would benefit if we added a visual representation of the relations between the models.

balazsorban44:
Note to self. We should let the adapter create a user id, instead of using the first account's.

www/docs/adapters/models.md
ndom91: Added an image of the schema to the model page :+1:

ndom91 (commented):
LGTM :+1:

ndom91 (approved):
LGTM :+1:

EarthlingDavey:
I noticed one minor thing then coding the neo4j adapter.

Most of the dates are instances of JS Date. All except account.expires, which is epoch. 

Is this intentionally different, or something that should be changed?

PR #2382, opened 2021-07-16
Add optional generateVerificationToken callback to provider EmailConfig

dpeek:
Minor addition to EmailConfig type so that users can override token generation :)

### DIFF
--- a/types/providers.d.ts
+++ b/types/providers.d.ts
@@ -151,6 +151,7 @@ export interface EmailConfig extends CommonProviderOptions {
    */
   maxAge?: number
   sendVerificationRequest: SendVerificationRequest
+  generateVerificationToken?:() => Promise<string>
 }
 
 export type EmailProvider = (options: Partial<EmailConfig>) => EmailConfig


0ubbe (changes requested):
Thanks 😊 , can you also update our [documented options ](https://next-auth.js.org/configuration/providers#options-1)( just needs editing in [this file](https://github.com/nextauthjs/next-auth/blob/main/www/docs/configuration/providers.md) ) and update the [type tests](https://github.com/nextauthjs/next-auth/blob/main/types/tests/providers.test.ts#L10-L20) as well 🙏🏽  ?

balazsorban44 (commented):
Thank you! But it's being addressed in the next major release! 😊

https://github.com/nextauthjs/next-auth/blob/befd546e5e16add129be4bd9b5069b0dc43065fb/types/providers.d.ts#L156-L169

If you still really want this to be merged (next release hopefully out this summer), I'll happily do that, if you you could please copy the comment from the link above.

dpeek:
Happy to wait until next major version so you don't need to deal with another PR, I can extend types for the time being. Thank you both for your work on this great library!

dastrong:
@dpeek Hey could you share how you extended the types to include `generateVerificationToken`?

balazsorban44:
@dastrong it is, but only on the `next` channel. You can check out the version `4.0.0-next.25` and here is a WIP migration guide: https://docs-nextauthjs.vercel.app/getting-started/upgrade-v4

PR #2395, opened 2021-07-17
Migrate Providers to Typescript 🪄

0ubbe:
## Reasoning 💡

...

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## Affected issues 🎟

None


balazsorban44:
#2411 is going to cause a fair amount of merge conflicts, so I will close this since we should wait until that other PR is merged. I still would like us to introduce TS to the core, but maybe we could do it through a file that is less prone to changes in the coming days/weeks.

PR #2411, opened 2021-07-20
feat: improve OAuth provider configuration

### LINKED ISSUE #1605: Provider OAuth JWT Header Access
Provide access to the JWT header of the responding `idToken` from the OAuth provider in the `profile()` function.

It's best practice to validate the contents of an `idToken` against public keys provided by the OAuth vendor (usually found at a `.well-known` endpoint provided by the vendor. Whilst this is possible due to the async nature of the `profile()` function, we currently do have access to the JWT header in said function (the first section of the 3 sections a JWT is made up of) which contains the following payload:

```
{
  "alg": "HS256",
  "type": "JWT",
  "kid": "....",
}
```

We need access to the key ID above in order to validate the `idToken` validate the `idToken` against the `.well-known` endpoint provided by the OAuth vendor as they will rotate keys often and we need to be able to select the correct one that matches up with the `kid` we have received.

Provide another parameter on the `profile()` service that provides the header aswell as the payload and profile information. Furthermore, the ability to cache and retrieve the public keys on the Lambda would be preferable as public keys only tend to rotate every 3 months so wouldn't want to make an API call every time if it could be avoided.

**Potential problems**
None that I can tell.

I've been told that there is work currently ongoing to customise request and response methods for providers with the ability to use `async` methods such as API calls which may negate this proposal entirely but I thought it best to leave the contributors to judge this.

**Additional context**
I don't feel I have the necessary skills to contribute on code with this one but happy to help.

### LINKED ISSUE #1607: Well Known Support
For OAuth providers that have `.well-known` configuration endpoints, offer the ability to provide these and have NextAuth dynamically discover and setup against all endpoints and everything that's provided as opposed to hardcoding all the endpoints manually.

Dynamic service discovery is always advisable as it means the vendor can change their endpoints and/or configuration and a NextAuth enabled website can pick up the configuration dynamically and continue operating. Furthermore, it takes a lot of work out of manually configuring all of the endpoints which have inherent human error.

In the provider's options, provide the ability to set a `.well-known` endpoint and have the NextAuth service set up as per the configuration. Perhaps saving this configuration in the Lambda cache so it doesn't have to be requested every time.

**Potential problems**
Could conflict with other OAuth provider options. Could also lead to more time on the initial request.

Manual configuration, which is what I've done to date.

### LINKED ISSUE #1846: RFC: Improve OAuth provider configuration

balazsorban44:
> This touches on all OAuth providers, so there is a big potential for breaking by default. We have let new providers be added for contributors' specific needs, but from now on, we will require a more strict default on all new providers, so the basic behavior is predictable for everyone.
⚠ Unfortunately, we will not have the capacity to test each and every provider that has been added to the default providers, but we will do our best to test the most popular ones. (@ndom91 has worked on setting up the infrastructure for this). If you wish to make sure that the provider you are using will stay working, please reach out with your concerns and tell us how can you help us test that particular provider in the future. :pray:

That said, I will try my best to not break ANY of the currently built-in providers, or at least make the migration super easy. So hopefully, you won't have to change anything. It will most probably affect you if you defined a custom provider though.

We will monitor the default configuration much more closely, so the behavior will be more consistent across providers by default.

Closes #1846, Closes #1605, Closes #1607

BREAKING CHANGES:

Basecamp provider is removed. See the explanation [here](https://github.com/basecamp/api/blob/master/sections/authentication.md#on-authenticating-users-via-oauth)

**ALL** OAuth providers' `profile` callback is expected to only return these fields by default from now on: `id`, `name`, `email`, and `image` at most. Any of these missing values should be set to `null`.

The following new options are available:
1. `authorization` (replaces `authorizationUrl`, `authorizationParams`, `scope`)
2. `token` replaces (`accessTokenUrl`, `headers`, `params`)
3. `userinfo` (replaces `profileUrl`)

These three options map nicely to the OAuth spec's three endpoints for
1. initiating the login flow
2. retrieve OAuth tokens
3. retrieve user information

They all take the form of `EndpointHandler`:
```ts
type EndpointRequest<C, R> = (
  context: C & {
    /** `openid-client` Client */
    client: Client
    /** Provider is passed for convenience, ans also contains the `callbackUrl`. */
    provider: OAuthConfig & {
      signinUrl: string
      callbackUrl: string
    }
  }
) => Awaitable<R>

/** Gives granular control of the request to the given endpoint */
type AdvancedEndpointHandler<P extends UrlParams, C, R> = {
  /** Endpoint URL. Can contain parameters. Optionally, you can use `params`*/
  url?: string
  /** These will be prepended to the `url` */
  params?: P
  /**
   * Control the corresponding OAuth endpoint request completely.
   * Useful if your provider relies on some custom behavior
   * or it diverges from the OAuth spec.
   *
   * - ⚠ **This is an advanced option.**
   * You should **try to avoid using advanced options** unless you are very comfortable using them.
   */
  request?: EndpointRequest<C, R>
}

/** Either an URL (containing all the parameters) or an object with more granular control. */
type EndpointHandler<P extends UrlParams, C = any, R = any> =
  | string
  | AdvancedEndpointHandler<P, C, R>
```

In case of `authorization`, the `EndpointHandler` can define the `params` as [`AuthorizationParameters`](https://github.com/panva/node-openid-client/blob/51dc47d9ac619b71cd1c983b0be750a12bbae008/types/index.d.ts#L108-L143)

> Note: `authorization` does not implement `request` yet. We will have to see if there is demand for it.

From now on, instead of using the `...` spread operator when adding a new built-in provider, the user is expected to add `options` as a property at the end of the default config. This way, we can deep merge the user config with the default one. This is needed  to let the user do something like this:

```js
MyProvider({
  clientId: "",
  clientSecret: "",
  authorization: { params: {scope: ""} }
})
```
So even if the default config defines anything in `authorization`, only the user-defined parts will be overridden.


balazsorban44:
A drawback with this implementation is if the user needs to override a single/common parameter like `scope`. Since it is not top-level anymore, the would need to redefine the whole `authorization` option just to change the `scope`:
```js
authorization: {
  url: "need to duplicate",
  params: {
    // need to duplicate default params
    scope: "your scopes"
  }
}
```

Should we keep the top-level `scope` option?

balazsorban44:
Documentation is still missing, but I consider this ready for a first review.

ndom91 (approved):
So at first glance everything looks alright to me, way to go updating all those provider configs!! Dig the optional AdvancedEndpointConfig override for those who need it. 

Did you meant to leave the tests disabled for now?

If you want we can look over it together this evening on a call one more time :+1:

mod-flux:
This looks really good @balazsorban44, nice one! Also, great job on all the providers transitioning over.

bbigras:
I have a dumb question. Does the new config allow using Keycloak (OpenID Connect with  `.well-known`)?

balazsorban44:
Haven't used Keycloak myself, but if it correctly implements the spec, I assume so.

bbigras:
> Do they have like a hosted demo server like IDS4 have at https://demo.identityserver.io/ ? I would be happy to test it out.

I think you can use https://please-open.it/ for free to test.

bbigras:
Does this PR work with next-auth-example?

I got:

```
ModuleNotFoundError: Module not found: Error: Can't resolve 'next-auth/client' in '/home/bbigras/dev/next-auth-example/pages'
```

But I'm not really used to `npm link`.

balazsorban44:
It doesn't. This is an unstable branch. You have to go through the release notes to make it work. We plan to write a migration guide when it's time.

valstu:
> Note: authorization does not implement request yet. We will have to see if there is demand for it.

This is exactly what I would need :D

balazsorban44:
please open a new issue/feature request with a **detailed** description, preferably with code examples, alternatives you have considered, etc.

types/tests/providers.test.ts
```diff
@@ -136,7 +137,7 @@ Providers.IdentityServer4({
 Providers.Discord({
   clientId: "foo123",
   clientSecret: "bar123",
-  scope: "identify",
+  authorization: { params: { scope: "identify" } },
```
central419, line 140: I'm using beta.2 and this doesn't seem to work, whatever I do the scope in the URL remains as `scope=identify%20email`.

I do not want to know user's their email.

balazsorban44: Please open an issue/discussion and attach a reproduction. This is not the right place. You are commenting on a test file. 👀

PR #2432, opened 2021-07-27
docs(client): add text regarding 'logout'

ndom91:
## Reasoning 💡

Added text with the term 'logout' in order to "fix" the most searched term without results according to our algolia analytics.

image

## Checklist 🧢

- [x] Documentation
- ~[ ] Tests~
- [x] Ready to be merged


balazsorban44 (approved):
Been looking at this in the search analytics for a few weeks now. 😅 very good idea! You could try hinting this in other relevant places as well, and maybe do the same with other highly searched terms without results!

Feel free to create an issue with terms you cannot find a good place for and tag me so we can decide how to include those in the docs!

Issue #2294, opened 2021-07-01
Make `next-auth` work with other Frameworks (Vite, Vue, ...)


brillout:
Follow up of https://twitter.com/balazsorban44/status/1410641689715347461

cc @s-kris @balazsorban44

balazsorban44:
I'm very hopeful about this. We might have to do small adjustments, but I think in most cases, a tiny wrapper on top of `NextAuth` would be enough.

I already started to decouple/better namespace our client code, in v4 it will be under `next-auth/react`

See #1473

brillout:
> The ground work could be done as early as in v4 (aiming for a release this summer)

Exciting 😍.

Will I then be able to use NextAuth.js in a let's say Vite + Vue app?

I'm developing a framework ("Vike") on top of Vite + vite-plugin-ssr + RPC and I'd love to offer authentication to both my React and Vue users with NextAuth.js.

Thanks for your wonderful work, it's lovely to see the JS ecosystem maturing.

balazsorban44:
If we rewrite the backend a bit, as long as your framework can set cookies and do redirects and handle basic POST/GET requests, it should be good. We would need to separate the `server/index.js` file in a way that it is only a tiny wrapper for Next.js and the core stuff could be framework agnostic.

I liked the Sveltekit endpoint approach:

https://kit.svelte.dev/docs#routing-endpoints

brillout:
Neat & yes, the backend integration should work then.

How about Vue?

You mentioned:

> I already started to decouple/better namespace our client code, in v4 it will be under next-auth/react

Will we be able to have a `next-auth/vue`? I can see myself (and potentially other Vue users) to contribute.

balazsorban44:
Yes, that would be easy I think.

Here is the source code for the react specific frontend:

https://github.com/nextauthjs/next-auth/blob/next/src/client/react.js

We still have to work out some small things like tab syncing and invalidation of the session, but other than that it should not be hard, hopefully.

brillout:
👌.

Had a quick look over the source and yea should be fairly easy to port it to Vue's composition API; there aren't that many React specific things going on. AFACIT a common module that factors out the React/view-framework agnostic code should make porting quite easy.

Can't wait to see NextAuth.js to become the de facto standard &mdash; Vite & Vue users will love it :-).

wobsoriano:
[...]

PR #2443, opened 2021-07-28
docs(providers): add tip about async provider code

ndom91:
## Reasoning 💡

Adds tip regarding using async code in ones provider instantiation. Wasn't sure if this was the best place to put this, but it seemed appropriate enough. Im open to any other suggestions though obviously :+1: 

## Checklist 🧢

- [x] Documentation
- ~[ ] Tests~
- [x] Ready to be merged

## Affected issues 🎟

Related: https://github.com/nextauthjs/next-auth/discussions/2394

### DIFF
--- a/www/docs/configuration/options.md
+++ b/www/docs/configuration/options.md
@@ -44,6 +44,21 @@ Options are passed to NextAuth.js when initializing it in an API route.
 
 An array of authentication providers for signing in (e.g. Google, Facebook, Twitter, GitHub, Email, etc) in any order. This can be one of the built-in providers or an object with a custom provider.
 
+If you need to use an asynchronous function in your provider instantiation, you can setup your `[...nextauth].js` file like so:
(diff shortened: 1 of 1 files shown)


sastraxi:
Perhaps I'm being daft, but shouldn't the async function `return NextAuth(...`? Right now it's just instantiating it and throwing the value away.

balazsorban44:
It's a Next.js API route handler, nothing needs to be returned. It uses `res.{json,end,send}` to finish the response. `return` might be the way in other frameworks like [SvelteKit](https://kit.svelte.dev/docs#routing-endpoints) though, but we aren't there yet. (#2294)

sastraxi:
@balazsorban44 ahh right, forgot the API route handlers work just like express. Thanks for the clarification!

balazsorban44 (approved):

PR #2450, opened 2021-07-29
fix(ts): fix internal react type import

grikomsn:
This PR fixes an improper import path for the internal React client for the `next` version, from `"internals/react"` to `"./internals/react"`.

## Reasoning 💡

While `internals/react` is a valid import for the project scope (depending on the tsconfig baseUrl), on dist builds it cannot resolve `internals/react` which breaks all typings that is related to the React client functions.

This PR aims to fix this issue, which I have tested by manually changing the import path. For reproducibility, below is a patch file generated using [`patch-package`](https://github.com/ds300/patch-package):

```patch
diff --git a/node_modules/next-auth/react.d.ts b/node_modules/next-auth/react.d.ts
index 0a3b3ec..9d0f67d 100644
--- a/node_modules/next-auth/react.d.ts
+++ b/node_modules/next-auth/react.d.ts
@@ -2,7 +2,7 @@ import * as React from "react"
 import { IncomingMessage } from "http"
 import { Session } from "."
 import { ProviderType } from "./providers"
-import { SessionContextValue } from "internals/react"
+import { SessionContextValue } from "./internals/react"
 
 export interface CtxOrReq {
   req?: IncomingMessage
```

## Checklist 🧢

~- [ ] Documentation~
~- [ ] Tests~
~- [ ] Ready to be merged~

## Affected issues 🎟

So far there are no issues related to this problem.


0ubbe (approved):

balazsorban44:
Thanks!

PR #2511, opened 2021-08-10
fix(provider): Add scope string array type

Liam-Tait:
## Reasoning 💡

Documentation and code already handles `string[]` for `scope`, this pr updates the typescript types to match the current functionality and usage.

See https://next-auth.js.org/configuration/providers#options for current docs

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

Documentation is already up to date
I don't think any tests are required.
This should be ready to be merged as soon as it is reviewed

### DIFF
--- a/types/providers.d.ts
+++ b/types/providers.d.ts
@@ -27,7 +27,7 @@ export interface OAuthConfig<P extends Record<string, unknown> = Profile>
   headers?: Record<string, any>
   type: "oauth"
   version: string
-  scope: string
+  scope: string | string[]
   params: { grant_type: string }
   accessTokenUrl: string
   requestTokenUrl?: string


0ubbe (approved):
We use the Node `oauth` client for sign-in with providers, as you suggested,[ it does accept ](https://github.com/ciaranj/node-oauth/blob/a7f8a1e21c362eb4ed2039431fb9ac2ae749f26a/examples/github-example.js#L29)`string[]` for `scope` 👍🏽

balazsorban44:
The current package does support it, but the one in v4 won't by default:

https://github.com/panva/node-openid-client/blob/772306cc82b72ace9361f4a945249a5c35a8f921/types/index.d.ts#L138

The question is what should we do?

1. Merge this PR and don't apply it on the `next` branch.
2. Merge this PR and implement support for `string[]` on `next` branch.
3. Don't merge this PR, nothing to do.

PR #2516, opened 2021-08-12
Update prisma.md

RyuNIshimura:
I added the autoincrement pattern.

## Reasoning 💡

I'm using MySQL and Prisma, and I thought there would be many cases where I would use 'int' instead of 'string' for the id, so I thought it would be a good idea to document it.

I also had an error when inserting the userId into a 1:n relationship model, so I changed the id of each model from 'string' to 'int' and it worked.
(This may be a functional problem of MySQL though).

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/www/docs/adapters/prisma.md
+++ b/www/docs/adapters/prisma.md
@@ -110,7 +110,7 @@ model VerificationRequest {
 
 ```
 
-If you want to use autoincrement instead of cuid for id, rewrite it as follows.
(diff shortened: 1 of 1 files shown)


RyuNIshimura:
I found something similar in Prisma (Legacy), so I'll close.

Issue #2143, opened 2021-06-08
How to send callbackUrl in email link


lionelrudaz:
## Question 💬

In my app, I'd like to invite new user from an admin interface, then onboard them. 

In other words, I'm trying to signIn() a new user by sending a new URL by email. The user needs to land on a specific on boarding URL.

The thing is that when I specify callbackUrl as an option, it's saved in the cookies of the current user, but it won't get available in the link sent to the invited user.

Do you know a way to do that? Or to have some options in the send verification method? 

## Contributing 🙌🏽

If the solution doesn't exist, I can help creating a pull request.

ahmadigbaryia:
I have a similar problem, I'm using Email provider for authentication.
customers can login at any point of the application, and I want them to be redirected to the same point when they decided to login.

I've added a callbackUrl in the querystring, for OAuth providers it works, but for Email it doesn't.

ahallora:
You could do 

```javascript
signIn("email", {
   email: "user@email.com",
   callbackUrl: "https://the.url.you/want/your/user/to/land/on/once/validated"
}
```

lionelrudaz:
I believe that's what I did, but it wasn't working. Was there any updated in the meantime that could have fixed that?

ahallora:
Ahhh, so you want to change the link in the verification email to contain more than just the user-email and verification-token. Yeah, I want that too actually, but haven't found a way yet.

I've come some of the way by being able to override the generated HTML, by setting `sendVerificationRequest` according to this: https://next-auth.js.org/providers/email#customising-emails 

```js
Providers.Email({
  server: process.env.EMAIL_SERVER,
  from: process.env.EMAIL_FROM,
  sendVerificationRequest: customVerificationRequest,
}),
```

But I'm still unable to actually pass parameters (e.g. my own JWT) to the customVerificationRequest function. Perhaps if we could wrap `export default NextAuth({...})` and grab the `NextApiRequest` before returning NextAuth-default behaviour, we could probably intercept and add the required parameters to the sendVerificationRequest...

I've tried like this:

```js
export default async (req: NextApiRequest) => {
  console.log("nextauth query", req.query);
  return NextAuth({...});
}
```
... but it didn't work and logs out this:

```
nextauth query { nextauth: [ 'session' ] }
API resolved without sending a response for /api/auth/session, this may result in stalled requests.
```

We might come up with a way together!? 😄 🙈

balazsorban44:
@ahallora you can do:

```js
[...]

PR #2542, opened 2021-08-16
feat(callback): add consistent callback behaviour with email verification

dan-kwiat:
## Reasoning 💡

The signIn callbackUrl gets stored by NextAuth as a cookie, which works well for oauth providers. For the email provider, it works if the user clicks the email verification link from the same device that called the signIn method. However, a surprising number of users open the verification email from a different device, especially new users. E.g. They sign up from a work computer using their personal email, then their phone buzzes and they find it easier to open the email there.

This is problematic because the callbackUrl can contain important information e.g. a referral token and user preferences created during a pre-signup flow. The info gets lost if the user opens the verification link from another device.

There is potentially a simple solution, which is to put the `callbackUrl` in the query string of the verification link, without changing any of the existing cookie functionality. In fact, adding a query parameter to the link href is already possible using the `sendVerificationRequest` option on the email provider. This PR is a suggestion to check for a callback url in the query string when the user clicks the verification link, and to redirect to that url if present. That way the callback functionality remains, regardless of which device was used to open the link.

If others think this is a good idea, I'm happy to test further & add docs etc.

## Checklist 🧢

~- [ ] Documentation~
~- [ ] Tests~
~- [ ] Ready to be merged~

## Affected issues 🎟

#2143


dan-kwiat:
Update: I've changed the base to the `next` branch

dan-kwiat:
@balazsorban44 @ahallora you were discussing this issue last month - what do you think of this approach to fix the broken callback behaviour when the email verification link is opened from a different device to the one requesting sign in?

I could also add a commit to include the callback url in the default email verification link url, or leave this up to the user to implement with `sendVerificationRequest`.

dan-kwiat:
I dug deeper into how NextAuth works and found that `req.query.callbackUrl` is already considered in the [callback url handler](https://github.com/nextauthjs/next-auth/blob/next/src/server/lib/callback-url-handler.js#L16), so these changes aren't necessary :)


This means that we can fix the email provider callback functionality by simply including `callbackUrl` in the verification link. Luckily it's available in the body of the request which triggers sending the email, so we can do this:

```js
// [...nextauth].js

function emailSender(callbackUrl) {
  return async ({ identifier, url }) => {
    try {
      const emailVerificationLink = new URL(url)
      emailVerificationLink.searchParams.set("callbackUrl", callbackUrl)
      await sendVerificationEmail({
        email: identifier,
        url: emailVerificationLink.href,
      })
    } catch (error) {
      // handle error
    }
  }
}

const getOptions = (req) => ({
  providers: [
    Providers.Email({
      sendVerificationRequest: emailSender(req.body.callbackUrl),
    }),
  ],
  // ...
})

export default (req, res) => NextAuth(req, res, getOptions(req))
```

It would be much nicer if NextAuth set the query parameter internally [here](https://github.com/nextauthjs/next-auth/blob/next/src/server/lib/email/signin.js#L37). I'll close this PR now and submit a new one with that implemented.

PR #2543, opened 2021-08-16
feat(provider): add Trovo provider

IAmNotTheLeo:
## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/providers/trovo.js
+++ b/src/providers/trovo.js
@@ -1,0 +1,23 @@
+
+export default function FortyTwo(options) {
+  return {
+    id: "trovo",
+    name: "Trovo",
+    type: "oauth",
+    version: "2.0",
+    scope: "user_details_self",
+    params: {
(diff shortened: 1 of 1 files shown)


src/providers/trovo.js
```diff
@@ -0,0 +1,23 @@
+
+export default function FortyTwo(options) {
```
balazsorban44, line 2:
```suggestion
export default function Trovo(options) {
```

balazsorban44 (changes requested):
Thank you for starting this! We will need a bit more though, please see the docs: https://next-auth.js.org/configuration/providers#adding-a-new-provider

I also wonder if you know how hard it is to create a test account? In the next version, we would like to make built-in providers more reliable and test our integration with them.

IAmNotTheLeo:
Sorry, ill do that now

PR #2549, opened 2021-08-17
feat(providers): add Trovo

IAmNotTheLeo:
## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/providers/trovo.js
+++ b/src/providers/trovo.js
@@ -6,7 +6,7 @@ export default function Trovo(options) {
     version: "2.0",
     scope: "user_details_self",
     params: { grant_type: "authorization_code" },
-    accessTokenUrl: "https://${options.domain}/access_token",
+    accessTokenUrl: "https://open-api.trovo.live/openplatform/exchangetoken",
     authorizationUrl: "https://open.trovo.live/page/login.html",
     getUserInfo: "https://open-api.trovo.live/openplatform/getuserinfo",
     profile(profile) {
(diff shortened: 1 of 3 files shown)


balazsorban44 (changes requested):
Thank you, a small suggestion.

Regarding my other question, looks like we need to write an e-mail to get an app registered:

https://developer.trovo.live/docs/APIs.html#_2-register-your-application

This might take a few days once we wrote the e-mail, but I would like to wait for merging until it's done.

Unless you could provide us with a test application we could use for testing purposes? (basically to make sure that this integration will work in the future as well)

IAmNotTheLeo:
> Thank you, a small suggestion.
> 
> Regarding my other question, looks like we need to write an e-mail to get an app registered:
> 
> https://developer.trovo.live/docs/APIs.html#_2-register-your-application
> 
> This might take a few days once we wrote the e-mail, but I would like to wait for merging until it's done.
> 
> Unless you could provide us with a test application we could use for testing purposes? (basically to make sure that this integration will work in the future as well)

I don't mind to wait until it's merged, don't have a  testing application I’m afraid :(, just want other developers to have the authentication for the streaming Platform Trovo

balazsorban44:
This should be reopened against the `beta` branch.

@ndom91, @lluia this might need extra attention, see https://github.com/nextauthjs/next-auth/pull/2549#pullrequestreview-732011703

PR #2554, opened 2021-08-18
docs(provider): remove en-us from Azure urls

aaronpowell:
MS Docs has a lot of local language translations, so it's best to remove locale information from the URLs so that when someone follows them, they land on the right language version of the content.


balazsorban44 (approved):
Makes sense, thank you!

PR #2577, opened 2021-08-21
Changed default Github scope to read-only

nathanclevenger:
When using the default settings of the Github provider, with the "user" scope, it grants read/write access to profile info only.  By changing to "user:email" it will only request read-only access https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps

## Reasoning 💡

When using the next-auth defaults for the GitHub provider, it requests read/write access to user profile info, which is more access than should be provided.  This pull requests reduces the scope of the request to be read-only

## Checklist 🧢

- ~[ ] Documentation~
- ~[ ] Tests~
- [ X] Ready to be merged

PR #2579, opened 2021-08-21
Changed default Github scope to read-only

nathanclevenger:
When using the default settings of the Github provider, with the "user" scope, it grants read/write access to profile info only. By changing to "read:user" and "user:email" it will only request read-only access https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps

## Reasoning 💡

When using the next-auth defaults for the GitHub provider, it requests read/write access to user profile info, which is more access than should be provided. This pull requests reduces the scope of the request to be read-only

## Checklist 🧢

- ~[ ] Documentation~
- ~[ ] Tests~
- [ X] Ready to be merged

### DIFF
--- a/src/providers/github.js
+++ b/src/providers/github.js
@@ -3,7 +3,7 @@ export default function GitHub(options) {
     id: "github",
     name: "GitHub",
     type: "oauth",
-    authorization: "https://github.com/login/oauth/authorize?scope=user",
+    authorization: "https://github.com/login/oauth/authorize?scope=read:user+user:email",
     token: "https://github.com/login/oauth/access_token",
     userinfo: "https://api.github.com/user",
     profile(profile) {


ndom91:
Just for documentation, this is a redo of #2577

balazsorban44:
Thank you! Just wondering (don't have a pc to check right now), but wouldn't `read:user` include the email as well?

nathanclevenger:
> Thank you! Just wondering (don't have a pc to check right now), but wouldn't `read:user` include the email as well?

When I tested that, I didn't get an email with just `read:user` ... I only got email with `user:email` ... the documentation does explicitly say you get `user:email` scope with `user` but it doesn't say email comes with `read:user`

balazsorban44:
Hmm. actually when logging in with GitHub here:

https://next-auth-example.vercel.app/

it does give email though. 🤔 (only `read:user` scope)

nathanclevenger:
> Hmm. actually when logging in with GitHub here:
> 
> https://next-auth-example.vercel.app/
> 
> it does give email though. 🤔 (only `read:user` scope)

That is very odd ... because when I log in with GitHub on that sample, it does not provide email for my account:

```
{
  "user": {
    "name": "nathanclevenger",
    "email": null,
    "image": "https://avatars.githubusercontent.com/u/4130910?v=4"
  },
  "expires": "2021-09-20T20:07:24.024Z"
}
```

nathanclevenger:
Maybe if the full `user` scope had been previously authorized, then email would be available?  When I did this on my account for my testing, I had revoked all previous tokens ...

balazsorban44:
That's not it! I also revoked access, reauthenticated (with only `read:user` scope), and got my email also! I remember now it was something special with GitHub in this case.

For the email to come through, you have to add public email here: https://github.com/settings/profile

Can you maybe verify that the `read:email` scope also works if the email was not public?

nathanclevenger:
Ahh, that makes sense about your email address being part of your public profile.  My email address is private not in my public profile ... but yes, if I add the `user:email` scope then I do get my email address even though it's private

balazsorban44:
Thank you! In that case, I am happy to keep `user:email` in the default scopes as well!

PR #2552, opened 2021-08-17
refactor: rewrite core to TypeScript

balazsorban44:
- [x] re-enable tests
- [x] fix bundling (split between client and server, follow Next.js)
- [x] fix CSS import
- [ ] ~fix dev app imports (check out https://github.com/vercel/next.js/pull/22867)~ will do in subsequent PR

PR #2598, opened 2021-08-24
feat: github codespace setup

ndom91:
## Reasoning 💡

This would make it super easy to checkout any PR, issue, etc. in a preconfigured dev env in the browser in the comfortable VS Code interface!

So I added a `.devcontainer` setup for enabling people to spin up a dev environment with codespaces in no time. Next to where you can set secrets for github actions, you can also set env vars for codespaces. There I set the auth0 id / secret / domain for testing. 

What works so far is spinning up the node container, installing the `next-auth` dependencies, as well as the `/app` dependencies and starting the dev server. You can then visit the example app homepage via the generated preview URL :tada: 

Things that still do not work:
- [ ] `copy:css` for some reason doesn't ocpy `dist/css/index.css` to where it should be, although `copy:app` **does** work :shrug: 
- [ ] programmatically getting the codespace preview URL to correctly be able to set `NEXTAUTH_URL`

Whenever you start a codespace, any ports you forward from the container get a dynamic preview URL like this:
image

You can set env vars in the `.devcontainer.json` file in a few ways - but I'm not sure how we'd get that dynamic URL ahead of time, or if there is any programmatic way to get it..

You can open up any PR (like this one :wink:) in a codespace here:

image

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -25,6 +25,6 @@
   // Use 'postCreateCommand' to run commands after the container is created.
   "postCreateCommand": "npm run dev:setup && npm run dev",
 
-  // Comment out connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
+  // Comment out to connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
   "remoteUser": "node"
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,20 +3,27 @@
   "build": {
     "dockerfile": "Dockerfile",
     // Update 'VARIANT' to pick a Node version: 12, 14, 16
-    "args": { "VARIANT": "16" }
+    "args": { "VARIANT": "16" },
+    "context": ".."
   },
 
   // Set *default* container specific settings.json values on container create.
   "settings": {},
 
   // Add the IDs of extensions you want installed when the container is created.
-  "extensions": ["dbaeumer.vscode-eslint"],
+  "extensions": ["dbaeumer.vscode-eslint", "esbenp.prettier-vscode"],
 
   // Use 'forwardPorts' to make a list of ports inside the container available locally.
   "forwardPorts": [3000],
+  "portsAttributes": {
+    "3000": {
+      "label": "next-auth-demo-app",
+      "onAutoForward": "notify"
+    }
+  },
 
   // Use 'postCreateCommand' to run commands after the container is created.
-  "postCreateCommand": "cd app && npm install && npm run dev",
+  "postCreateCommand": "npm install && cd app && npm install && npm run dev",
 
   // Comment out connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
   "remoteUser": "node"
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -4,7 +4,7 @@
     "dockerfile": "Dockerfile",
     // Update 'VARIANT' to pick a Node version: 12, 14, 16
     "args": { "VARIANT": "16" },
-    "context": ".."
+    "context": "../app"
   },
 
   // Set *default* container specific settings.json values on container create.
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -23,7 +23,7 @@
   },
 
   // Use 'postCreateCommand' to run commands after the container is created.
-  "postCreateCommand": "npm install && cd app && npm install && npm run dev",
+  "postCreateCommand": "npm run dev:setup && npm run dev",
 
   // Comment out connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
   "remoteUser": "node"
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,23 @@
+{
+  "name": "Node.js",
+  "build": {
+    "dockerfile": "Dockerfile",
+    // Update 'VARIANT' to pick a Node version: 12, 14, 16
+    "args": { "VARIANT": "16" }
+  },
+
+  // Set *default* container specific settings.json values on container create.
+  "settings": {},
+
+  // Add the IDs of extensions you want installed when the container is created.
+  "extensions": ["dbaeumer.vscode-eslint"],
+
+  // Use 'forwardPorts' to make a list of ports inside the container available locally.
+  "forwardPorts": [3000],
+
+  // Use 'postCreateCommand' to run commands after the container is created.
+  "postCreateCommand": "cd app && npm install && npm run dev",
+
+  // Comment out connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
+  "remoteUser": "node"
+}
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,14 @@
+# [Choice] Node.js version: 16, 14, 12
+ARG VARIANT=16
+FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:${VARIANT}
+
+# [Optional] Uncomment this section to install additional OS packages.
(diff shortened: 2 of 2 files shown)


balazsorban44:
Interesting! After https://github.com/nextauthjs/next-auth/pull/2552 lands though, the dev app set up might change drastically. I am hoping that https://github.com/vercel/next.js/pull/22867 will simplify things.

Also, could the `.devcontainer` folder go under `config` or `.github`? Would like to reserve the root for only necessary things.

ndom91:
@balazsorban44 unfortunately the `.devcontainer` directory needs to be in the root as far as I can tell. Regarding the typescript changes, although a lot may change under the hood, in general it'll still be abstracted into a few npm scripts (build, run, etc.), right?

We obviously don't need to go forward with this, but I was experimenting with the devcontainers / codespaces for work and thought it'd be a nice addition here for you and others to easily spin up env's out of PR's, check things out, etc.

balazsorban44:
I am not against it, happy to discuss it!

belgattitude:
@balazsorban44 

> After #2552 lands though, the dev app set up might change drastically.

OMG ! You did this magic :star2: :1st_place_medal: 

>  I am hoping that vercel/next.js#22867 will simplify things.

Yes indeed and actually even if experimental I feel it's safe to go for it in the next-auth examples. I converted a [repo](https://github.com/belgattitude/nextjs-monorepo-example/pull/286) lately without problem.

<details>
<summary>For NextJS 11.1+, simply enable experimental flag</summary>
      
```js
const nextConfig = {
  experimental: {
    externalDir: true,
  },
};
export default nextConfig;
```
    
</details>


If there's ome risk to enable experimental, this one works with all next versions since NextJs 10

<details>
 <summary>For NextJS < 11, simply add a webpack override</summary>

```js
const nextConfig = {
  webpack: (config, { defaultLoaders }) => {
    // Will allow transpilation of shared packages through tsonfig paths
    // @link https://github.com/vercel/next.js/pull/13542
    const resolvedBaseUrl = path.resolve(config.context, "../../");
    config.module.rules = [
      ...config.module.rules,
      {
        test: /\.(tsx|ts|js|jsx|json)$/,
        include: [resolvedBaseUrl],
        use: defaultLoaders.babel,
        exclude: (excludePath) => {
          return /node_modules/.test(excludePath);
        },
      },
    ];
    return config;
  },
};
``` 
</details>

@ndom91 

> unfortunately the .devcontainer directory needs to be in the root as far as I can tell.

I think so too, but maybe not a bad idea... Are you trying to get multiple apps / example running ?

> I'm not sure how we'd get that dynamic URL ahead of time, or if there is any programmatic way to get it..

Just to be sure, you mean to set NEXTAUTH_URL ?

balazsorban44:
#2631 will fix the dev app.

Dev containers might already have an env variable we could alias in NEXTAUTH_URL?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,30 @@
...
+  "forwardPorts": [3000],
+  "portsAttributes": {
+    "3000": {
+      "label": "next-auth-demo-app",
+      "onAutoForward": "notify"
+    }
+  },
+
+  // Use 'postCreateCommand' to run commands after the container is created.
+  "postCreateCommand": "npm install && cd app && npm install && npm run dev",
```
balazsorban44, line 26:
```suggestion
  "postCreateCommand": "npm run dev:setup && npm run dev",
```

After merging #2631, this should be enough.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,30 @@
...
+
+  // Comment out connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
```
balazsorban44, line 28:
```suggestion
  // Comment out to connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
```

balazsorban44 (changes requested):

kriswuollett:
Interesting! I use [devcontainers myself locally](https://github.com/kriswuollett/code-in-container) based on the "Docker-from-Docker" configuration. If meant to be for development-level activities, I have some suggestions that may be helpful, and some general questions since I'm new to NodeJS development:

* How does this setup pull in  the `next-auth` code dependencies? In [`devcontainer.json`](https://github.com/nextauthjs/next-auth/pull/2598/files#diff-24ad71c8613ddcf6fd23818cb3bb477a1fb6d83af4550b0bad43099813088686) I just see `"context": "../app"`, so wouldn't `../next-auth/...` files not be available in the container, let alone the `process.cwd()` lines in [`app/next.config.js`](https://github.com/nextauthjs/next-auth/blob/e8a58a01b6b8c456268771f9b382fa300ccab8ac/app/next.config.js#L9) working? Is there something I'm missing since this is being tested in [Codespaces](https://github.com/features/codespaces) and not just Visual Studio Code locally with Docker Desktop?
* Perhaps the `Dockerfile` should have Next.js start up in [debug mode](https://nextjs.org/docs/advanced-features/debugging#step-1-start-nextjs-in-debug-mode) and a launch config for attaching a debugger added? For codespaces specifically, I don't know if the debugging port would need to be exported, but it would be needed for some locally running devcontainer configurations. For an issue I'm having currently, using the debugging has been very useful.
* Codespaces [seems inexpensive](https://docs.github.com/en/billing/managing-billing-for-github-codespaces/about-billing-for-codespaces), but would we expect project or developers to set up paid account to use it (for primary targeting a codespaces configuration)? I don't see any mention of being free for open-source projects. Are there any of the [codespaces limitations](https://code.visualstudio.com/docs/remote/codespaces#_known-limitations-and-adaptations) this project may hit?
* Maybe the [`[...nextauth].js`](https://github.com/nextauthjs/next-auth/blob/e8a58a01b6b8c456268771f9b382fa300ccab8ac/app/pages/api/auth/%5B...nextauth%5D.js#L34) file in the demo app could dynamically add providers only if the required environment variables were set?
* Given the above with the providers, and not necessarily in an initial PR, maybe it could be updated to be based on [Docker-from-Docker compose](https://github.com/microsoft/vscode-dev-containers/tree/main/containers/docker-from-docker-compose) and bundle something like [Keycloak](https://www.keycloak.org/) running too? That would allow for the base configuration to have a decently complex setup out of the box without having to customize environment variables in `devcontainer.json`.

ndom91:
@kriswuollett hey thanks for the all the feedback! So this was just an experimental thing, definitely not 100% thought through yet, at least on my part haha. 

But I can hopefully answer some of your questions.

We're not running it in a 'docker-in-docker' type setup, so it clones the repo and runs the container containing node from `.devcontainers/Dockerfile` and then executes the `postCreateCommand` string of npm scripts. Our `../app` directory was the example application whose `package.json` linked up one level to the `next-auth` source code. I'm not 100% sure how it is now that its been refactored to typescript, I haven't had a chance to look into it thoroughly in the past few days. But thats the general gist of it.

Github codespaces currently is free (See blue note box here: https://docs.github.com/en/codespaces/codespaces-reference/understanding-billing-for-codespaces - "**Note: Codespaces is free to use for all organizations on a GitHub Team..**"). However after reading your message and looking into it, it looks like the initial free period is expiring Sept. 10th, so in just about a week :(

Regarding dynamic providers.. I was talking to Balazs recently about how best we could include some OAuth configs here for testing, without leaking or potentialy exposing our internal testing OAuth keys. Maybe if we conditionally add the configs to `[...nextauth].js`, something along these lines..

```js

const nextAuthConfig = {
  providers: [],
  ..
}

process.env.GITHUB_SECRET && nextAuthConfig.providers.push(GithubProvider({
  ..  
}))

export default NextAuth(nextAuthConfig)
```

We could allow users to add their own OAuth keys, and it would automatically add / enable that provider if the appropriate provider key / secret is available. Is that kind of what you were picturing? I think thats  apretty good way of avoiding having to insert our own keys and potentially losing them, and making it easy for the users / contirbutors to setup and use theirs. 

I'm not familiar with Keycloak, but bundling another Auth provider in there to test could be another option, instead of the oauth configs. I'm open to whatever makes it easy for users / contirbutors to get up to speed :+1:

PR #2631, opened 2021-08-31
refactor(dev): use Next.js `externalDir` to fix dev app

balazsorban44:
This fixes the developer application after the TS refactor #2552 

We rely on the experimental `externalDir` feature of Next.js found here: https://github.com/vercel/next.js/pull/22867

This basically means we can import `next-auth` source code - without needing to compile it first - into the dev server even if it lives outside of Next.js's root folder, and HMR will "just work".

The only exception is the CSS which we use in Server-Side rendered pages that won't hot-reload anyway. They generate HTML through API routes, so no `getServerSideProps` is involved. This means editing the CSS won't trigger a manual page reload, but that has been the case before this PR as well.


0ubbe (approved):
Very neat 💯 💪🏽

ndom91 (approved):
Yeah awesome solution!

PR #2683, opened 2021-09-06
fix(ts): forward credentials fields to authorize

### LINKED ISSUE #2677: TypeScript complains on `CredentialsConfig`
### Discussed in https://github.com/nextauthjs/next-auth/discussions/2672

<div type='discussions-op-text'>

<sup>Originally posted by **Rhym** September  6, 2021</sup>
Good morning! I am updating my next auth in preparation for the 4.0 release and my previously working Credentials provider has an issue with the `authorize` callback.

I receive a 
```
"Types of parameters 'credentials' and 'credentials' are incompatible. Type 'Record<never, string>' is missing the following properties from type '{ username: string; password: string; }': username, password". "The expected type comes from property 'authorize' which is declared here on type 'Partial<CredentialsConfig<{}>>'"
```

I see that the `CredentialsConfig` takes a generic, but I'm not actually sure where to set the config. Any help would be appreciated, thank you. Example of my code: 

```typescript
providers: [
  // https://next-auth.js.org/configuration/providers
  CredentialsProvider({
    id: "login",
    name: "Login",
    credentials: {
      username: {},
      password: {},
    },

    authorize: async (credentials: { username: string; password: string }) => {
      const DATA = { username: credentials.username, password: credentials.password };
      // Login logic...
    },
  }),
],
```
</div>

balazsorban44:
Fixes #2677

### DIFF
--- a/src/providers/credentials.ts
+++ b/src/providers/credentials.ts
@@ -26,13 +26,20 @@ export type CredentialsProvider = <C extends Record<string, CredentialInput>>(
 
 export type CredentialsProviderType = "Credentials"
 
-export default function Credentials(
-  options: Partial<CredentialsConfig>
-): CredentialsConfig {
+type UserCredentialsConfig<C extends Record<string, CredentialInput>> = Partial<
+  Omit<CredentialsConfig<C>, "options">
+> &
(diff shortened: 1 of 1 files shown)


ndom91 (approved):
LGTM :+1:

PR #2698, opened 2021-09-08
Allow user options id for signinUrl and callbackUrl

Thisen:
## Reasoning 💡

This will change the way providers are parsed and use the user `options.id` from their provider config in the `signinUrl` and `callbackUrl`.

This reason for this suggestion is to better align with v3 behavior, where it used the users chosen provider ID for these urls. In the current beta, it uses the `defaultOptions.id`

## Checklist 🧢

- [X] Documentation (N/A - Couldn't find any related documentation)
- [X] Tests (N/A - Couldn't find any related tests)
- [X] Ready to be merged

## Affected issues 🎟

Haven't found any.


balazsorban44 (approved):
Good catch, thanks!

Issue #2708, opened 2021-09-09
CognitoProvider fails in Next-auth v.4 with required client_secret parameter


javigonz:
### Description 🐜

Trying to upgrade Next-auth `v3.14.7` to Next-auth `v4.0.0-beta.2`.

We´ve got an error message: `client_secret_basic client authentication method requires a client_secret`.

Into `[...nextauth].js` with version 3, we haven´t got any `clientSecret` setted, and the authorization / login with cognito works fine. But with the same configuration into version 4 the error is showed.

We don´t want to have a client secret parameter into the provider.

The problem may be that you are using as dependency `"openid-client": "^4.7.4"`, and it set by default the auth method to `client_secret_basic` which it sent as header the `client_secret` option that we don´t have.
But related with their documentation (https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1), `"The client MAY omit the parameter if the client secret is an empty string"`, and this is seems that not happens because we are passing an empty string as `client_secret` parameter.

### Is this a bug in your own project?

No

### How to reproduce ☕️

**1 Reproduce with Next-auth `v3.14.7`:**
   

>  - Launch an application with cognito Provider setting in `/api/auth/[...nextauth].js` without any `clientSecret` parameter, like this:

```js
import NextAuth from 'next-auth';
import Providers from 'next-auth/providers';

export default NextAuth({
 Providers.Cognito({
      clientId: process.env.COGNITO_CLIENT_ID,
      domain: process.env.COGNITO_DOMAIN'
    })
  ],
  callbacks: {...}
});
```
Authorization and login should works fine.

**2 Reproduce with Next-auth `v4.0.0-beta.2`:**
   

>  - Launch an application with new CognitoProvider setting in `/api/auth/[...nextauth].js` with `clientSecret`setting to an empty string, like this:

```js
import NextAuth from 'next-auth';
import CognitoProvider from 'next-auth/providers/cognito';

export default NextAuth({
  providers: [
    CognitoProvider({
      clientId: process.env.COGNITO_CLIENT_ID,
      clientSecret: '',
      domain: process.env.COGNITO_DOMAIN,
      issuer: `https://${process.env.COGNITO_DOMAIN}/`
    })
  ],
  callbacks: {...}
});
```
An error message is showed in the console:
```
[next-auth][error][OAUTH_CALLBACK_ERROR]
https://next-auth.js.org/errors#oauth_callback_error client_secret_basic client authentication method requires a client_secret {
[...]

Issue #2524, opened 2021-08-13
Manually Test Providers (v4)


lluia:
### Description 📓

Before the release of **v4** we want to manually test most (if not all) the current **OAuth providers** to make sure the changes that happen on **v4** to their configuration didn't break any of them 🤞🏽

Here's the table where we keep track of this testing:

| Provider     | Account created | Manually tested |
| ------------ | :-------------- | --------------: |
| `42`         |                 |                 |
| `Apple`      |                 |                 |
| `Atlassian`  | ✅              |                 |
| `Auth0`      | ✅              |                 |
| `Azure B2C`  |                 |                 |
| `Azure`      |                 |                 |
| `Basecamp`   | ✅              |                 |
| `BattleNet`  |                 |                 |
| `Box`        |                 |                 |
| `Bungie`     |                 |                 |
| `Cognito`    |                 |                 |
| `Coinbase`   |                 |                 |
| `Discord`    | ✅              |                 |
| `Dropbox`    | ✅              |                 |
| `Eve Online` |                 |                 |
| `Facebook`   | ✅              |    ✅   @balazsorban44              |
| `FaceIT`     | ✅              |                 |
| `FourSquare` | ✅              |                 |
| `Freshbooks` | ✅              |                 |
| `FusionAuth` |                 |                 |
| `Github`     | ✅              |      ✅      @ndom91         |
| `Gitlab`     | ✅              |  ✅      @ndom91             |
| `Google`     | ✅              |                 |
| `IS4`        | ✅              |                 |
| `Instagram`  | ✅              |                 |
| `Kakao`      | ✅              |                 |
| `Line`       | ✅              |                 |
| `LinkedIn`   | ✅              |                 |
| `MailChimp`  | ✅              |                 |
| `MailRu`     |                 |                 |
| `Medium`     |                 |                 |
| `Naver`      |                 |                 |
| `Netlify`    |                 |                 |
| `Okta`       |                 |                 |
| `OneLogin`   |                 |                 |
| `Osso`       |                 |                 |
| `Reddit`     |                 |                 |
| `Salesforce` |                 |                 |
| `Slack`      |                 |                 |
| `Spotify`    |                 |                 |
| `Strave`     |                 |                 |
| `Twtich`     | ✅              |                 |
| `Twitter`    | ✅              |                 |
| `VK`         |                 |                 |
| `Wordpress`  |                 |                 |
| `WorkOS`     |                 |                 |
| `Yandex`     |                 |                 |
[...]

PR #2717, opened 2021-09-10
fix: expose `openid-client` client and jwks options

balazsorban44:
In #2411, we introduced `openid-client` as the new underlying OAuth library. It came with a lot of configuration options, which we unintentionally obscured as it seemed to be unnecessary. In the issues below it was proven wrong, so for advanced usecases, we want to let the user configure everything according to their likings.

There is still an option we don't allow to configure, which is https://github.com/panva/node-openid-client/blob/main/docs/README.md#new-issuermetadata 

If a need arise, we can expose that as `issuer` as well. That property is already reserved for a url though, so similar to `authorization`, `token` and `userinfo` it will be able to take the form of either an url or an object.

Relevant issues: #2524, #2708


javigonz:
@balazsorban44 That´s works really fine. Now we could pass more options to `client` and be more precise in their configuration:

```js
import NextAuth from 'next-auth';
import CognitoProvider from 'next-auth/providers/cognito';

export default NextAuth({
  providers: [
    CognitoProvider({
      clientId: process.env.COGNITO_CLIENT_ID,
      domain: process.env.COGNITO_DOMAIN,
      client: {
        token_endpoint_auth_method: 'none'
      },
      issuer: `https://${process.env.COGNITO_DOMAIN}/`
    })
  ],
  callbacks: {...}
});
```

Thanks so much **Balázs Orbán** for your fast and excellent work, this library is amazing.

balazsorban44:
good to hear! I very much wanted to check this myself, but registering to AWS required my credit card, which I wasn't going to give them. 🤷‍♂️

Do you know if AWS have some kind of a solution for us to be able to register to Cognito without a credit card? we need the absolute bare minimum setup, just to be able to check the login flow.

balazsorban44:
I think I found a way. 😄

Utilizing [our OpenCollective page] (https://opencollective.com/nextauth), I could request a virtual bank card:

https://blog.opencollective.com/virtual-card-support

I'll test this out later today.

javigonz:
This looks great, let me know if you need some help from userPool or whatever.

Fronix:
@balazsorban44 
I don't think there is a way to register to Cognito without a credit card. However, you could sign up for [revolut](https://www.revolut.com/) and create a virtual credit card with no money on it. That would ensure no money can be charged and you don't give them your real cc info.

balazsorban44:
Cognito has been resolved, without this. :eyes: See #2829 If not necessary, then I don't want to expose additional options. I'll close this for now. Feel free to comment if there is an objection, but please check #2829 first.

danieltott:
I had just landed on this PR as a solution that has taken me 2 days to figure out - using a custom OAuth Provider - the only way to get the token request through was to change the `Client` call in `node_modules/next-auth/server/lib/oauth/client.js`

```
  const client = new issuer.Client({
    client_id: provider.clientId,
    client_secret: provider.clientSecret,
    redirect_uris: [provider.callbackUrl],
    token_endpoint_auth_method: 'client_secret_post'
  });
```

I would love it if this PR got back in...

balazsorban44:
@danieltott, no problem, if there is actual interest! :sweat_smile: Could you please check if `next-auth@0.0.0-pr.2717.06130bef` works for you? I did release it in experimental, but there was no feedback on it.

It exposes new options, see this PR for details.

danieltott:
@balazsorban44 YES! Just tested it, worked perfectly 👍  🥳

balazsorban44:
Let's get it in then! 😬 No ETA on new beta release yet, I want to do less of those then `next` releases.. you can use the above version until then. Thanks for checking it out!

Issue #2701, opened 2021-09-08
TypeScript error for the Credentials provider


dragoncorf:
### Description 🐜

The changes to Providers in [...nextauth].ts in V4 appered a TypeScript error for the Credentials provider.

``` TypeScript
Type 'CredentialsConfig<{ username: { label: string; type: string; }; password: { label: string; type: string; }; }>' is not assignable to type 'Provider'.
  Type 'CredentialsConfig<{ username: { label: string; type: string; }; password: { label: string; type: string; }; }>' is not assignable to type 'CredentialsConfig<{}>'.
    Types of property 'authorize' are incompatible.
      Type '(credentials: Record<"username" | "password", string>, req: NextApiRequest) => Awaitable<Omit<User, "id"> | { id?: string | undefined; } | null>' is not assignable to type '(credentials: Record<never, string>, req: NextApiRequest) => Awaitable<Omit<User, "id"> | { id?: string | undefined; } | null>'.
        Types of parameters 'credentials' and 'credentials' are incompatible.
          Type 'Record<never, string>' is missing the following properties from type 'Record<"username" | "password", string>': username, password
```

### Is this a bug in your own project?

No

### How to reproduce ☕️

``` TypeScript
import bcrypt from 'bcryptjs';
import prisma from 'lib/prisma';
import NextAuth from 'next-auth';
import CredentialsProvider from 'next-auth/providers/credentials';

export default NextAuth({
  providers: [
    CredentialsProvider({
      name: 'Credentials',
      credentials: {
        username: { label: 'Username', type: 'text' },
        password: { label: 'Password', type: 'password' }
      },
      async authorize(credentials: any, req) {
        const user = await prisma.user.findUnique({
          where: {
            username: credentials.username
          }
        });

        if (!user) {
          return null;
        }

        const valid = await bcrypt.compare(credentials.password, user.password);

        if (!valid) {
          console.log(`Credentials not valid`);
          return null;
        }

        if (user) {
          return { ...user, email: user.username };
        }
        return null;
      }
    })
  ]
});
```

### Environment 🖥

  System:
    OS: Windows 10 10.0.19042
    CPU: (8) x64 Intel(R) Core(TM) i5-1035G1 CPU @ 1.00GHz 
    Memory: 3.44 GB / 11.81 GB
  Binaries:
    Node: 14.15.4 - D:\Node\node.EXE
    npm: 6.14.10 - D:\Node\npm.CMD
  Browsers:
[...]

PR #2763, opened 2021-09-15
Delete default type for generic param in CredentialsConfig

kjugi:
## Reasoning 💡

By setting the default type for generic like `{}` we are blocking the inference of type in credentials as the whole typing is intended to do (I think so).

After that change we are able to do the following code in TS with `strict` mode enabled:

```ts
import CredentialsProvider from 'next-auth/providers/credentials';

CredentialsProvider({
  credentials: {
    email: {
      label: 'email',
      type: 'text',
    },
    password: {
      label: 'password',
      type: 'password',
    },
  },
  async authorize() {
    try {
      await fetch('/testing');

      return {
        id: '',
      }
    } catch (err) {
      return null;
    }
  },
}),
```

I'm aware that marking it as `any` is never the best solution as I'm doing in the `Provider` type. But it can get any object shape from us in that case I guess.

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## Affected issues 🎟

Don't know if resolves the following issues so they are just marked as affected.

#2701
#2677

### DIFF
--- a/src/providers/credentials.ts
+++ b/src/providers/credentials.ts
@@ -10,7 +10,7 @@ export interface CredentialInput {
 }
 
 export interface CredentialsConfig<
-  C extends Record<string, CredentialInput> = {}
+  C extends Record<string, CredentialInput>
 > extends CommonProviderOptions {
   type: "credentials"
   credentials: C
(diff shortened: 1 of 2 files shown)


balazsorban44:
I believe this has been fixed in #2802 by @Thisen

PR #2788, opened 2021-09-18
feat: built-in page theme updates

ndom91:
## Reasoning 💡

Add some very minimal customisation to the built-in signin page so people might not immediately need to replace it. This way they can customise some things with their brand color and add their company / project logo just above the card with the signin buttons. We explicitly **do not** want to go overboard styling this page. This is not an authentication component library or next.js app template!

Example:
```js
export default NextAuth({
  providers: [...],
  jwt: {...},
  theme: {
    colorScheme: 'auto',
    brandColor: '#67b246',
    logo: 'https://company.com/assets/logo.png'
  }
})
```

Would generate this:

Signin:
image
image

Signout:
image
image

Error:
image
image

Verify Request:
image
image

What do yall think?

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/app/pages/api/auth/[...nextauth].ts
+++ b/app/pages/api/auth/[...nextauth].ts
@@ -139,8 +139,8 @@ export default NextAuth({
   },
   debug: true,
   theme: {
-    color: "auto",
-    logo: "https://crm.newtelco.de/images/logo_newtelco.png",
-    brandColor: "#67B246"
+    colorScheme: "auto",
+    logo: "https://next-auth.js.org/img/logo/logo-sm.png",
+    brandColor: "#1786fb"
   },
 })
--- a/app/pages/api/auth/[...nextauth].ts
+++ b/app/pages/api/auth/[...nextauth].ts
@@ -138,5 +138,9 @@ export default NextAuth({
     secret: process.env.SECRET,
   },
   debug: true,
-  theme: "auto",
+  theme: {
+    color: "auto",
+    logo: "https://crm.newtelco.de/images/logo_newtelco.png",
+    brandColor: "#67B246"
+  },
 })
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -96,7 +96,11 @@ async function NextAuthHandler(
   const options: InternalOptions<any> = {
     debug: false,
     pages: {},
-    theme: "auto",
+    theme: {
+      color: "auto",
+      logo: 'https://crm.newtelco.de/images/logo_newtelco.png',
+      brandColor: '#67b246'
+    },
     // Custom options override defaults
     ...userOptions,
     // These computed settings can have values in userOptions but we override them
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -97,9 +97,9 @@ async function NextAuthHandler(
     debug: false,
     pages: {},
     theme: {
-      color: "auto",
-      logo: 'https://crm.newtelco.de/images/logo_newtelco.png',
-      brandColor: '#67b246'
+      colorScheme: "auto",
+      logo: '',
+      brandColor: ''
     },
     // Custom options override defaults
     ...userOptions,
--- a/src/server/types.ts
+++ b/src/server/types.ts
@@ -202,7 +202,7 @@ export interface NextAuthOptions {
  * [Pages](https://next-auth.js.org/configuration/pages)
  */
 export type Theme = {
-  color: "auto" | "dark" | "light"
+  colorScheme: "auto" | "dark" | "light"
   logo?: string
   brandColor?: string
 }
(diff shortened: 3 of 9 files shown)


balazsorban44:
I really like the idea! The other pages would benefit from this as well, I think. See: https://github.com/nextauthjs/next-auth/tree/beta/src/server/pages

I don't think we need individual configuration for each page, just stretch the color and logo across all of them.

ndom91:
Regarding the Credentials placeholder, it's empty in fact. If you don't specify a placeholder value in the config, which we dont in the example dev app

https://github.com/nextauthjs/next-auth/blob/e8a58a01b6b8c456268771f9b382fa300ccab8ac/src/server/pages/signin.js#L97-L99

https://github.com/nextauthjs/next-auth/blob/e8a58a01b6b8c456268771f9b382fa300ccab8ac/app/pages/api/auth/%5B...nextauth%5D.js#L58-L62

image

ndom91:
@balazsorban44 alright so I themed the rest of the pages. I didn't find a good use for the brandColor on other pages.. 

Do you like the border of the card being in the brandColor?

image
image

Doesn't look so bad with this green, but I can imagine it wouldn't be so nice with other colors.

app/pages/api/auth/[...nextauth].ts
```diff
@@ -138,5 +138,9 @@ export default NextAuth({
     secret: process.env.SECRET,
   },
   debug: true,
-  theme: "auto",
+  theme: {
+    color: "auto",
+    logo: "https://crm.newtelco.de/images/logo_newtelco.png",
+    brandColor: "#67B246"
```
Comment (balazsorban44) [inline app/pages/api/auth/[...nextauth].ts:144]: I'm fine to leave it like this, but maybe adding a NextAuth.js logo/color would make more sense? 😜

src/server/index.ts
```diff
@@ -96,7 +96,11 @@ async function NextAuthHandler(
   const options: InternalOptions<any> = {
     debug: false,
     pages: {},
-    theme: "auto",
+    theme: {
+      color: "auto",
+      logo: 'https://crm.newtelco.de/images/logo_newtelco.png',
+      brandColor: '#67b246'
```
balazsorban44, line 102:
```suggestion
```

We should have no default preferences here, I think. Even adding NextAuth.js as default might be confusing and more often than not people would probably be annoyed by it anyway.

I think the current unbranded version is a good default.

src/server/types.ts
```diff
@@ -201,7 +201,11 @@ export interface NextAuthOptions {
  * [Documentation](https://next-auth.js.org/configuration/options#theme) |
  * [Pages](https://next-auth.js.org/configuration/pages)
  */
-export type Theme = "auto" | "dark" | "light"
+export type Theme = {
+  color: "auto" | "dark" | "light"
+  logo?: string
+  brandColor?: string
+}
```
balazsorban44, line 208:
```suggestion
export type Theme = {
  colorScheme: "auto" | "dark" | "light"
  logo?: string
  brandColor?: string
}
```

I might want to call this `colorScheme` (after the CSS query "prefers-color-scheme"), or `mode` instead, but not really important to me, I'm fine with `color`.

Do we want to support `"auto" | "dark" | "light"` as well for backward compatibility?

We kind of stopped breaking things since the first beta, but it seems like a reasonably simple thing to upgrade and if we wouldn't do it now, then we would have to wait until v5 to deprecate. But I'm fine with anything we agree on.

balazsorban44 (changes requested):

balazsorban44:
Docs will have to be upgraded:

https://github.com/nextauthjs/docs/blob/main/docs/configuration/options.md#theme

https://github.com/nextauthjs/docs/blob/main/docs/configuration/pages.md#theming

src/server/types.ts
ndom91: Yeah good call on this! I didn't like it, but was strugglign to come up with a better name. `colorScheme` is great!

src/server/index.ts
ndom91: Oops my bad! Didn't mean to keep those test values of mine in there :facepalm:

app/pages/api/auth/[...nextauth].ts
Comment (ndom91) [reply app/pages/api/auth/[...nextauth].ts]: Oops my bad! Didn't mean to keep those test values of mine in there :facepalm:

ndom91:
Theme docs PR: https://github.com/nextauthjs/docs/pull/35

balazsorban44 (approved):
LGTM!

leon-do:
colorScheme works great but I"m unable to get logo.

Is there an example repo with logo?

alsherif-khalaf:
> ```js
> heme: {
>     colorScheme: 'auto',
>     brandColor: '#67b246',
>     logo: 'https://company.com/assets/logo.png'
>   }
> ```

yes the same problem with me , no logo

PR #2813, opened 2021-09-22
fix(provider): remove extra brace from `jwks_uri`

jrasanen:
## Reasoning 💡

The Azure B2C adapter in 4.0.0-beta has an extra brace in the template literal, breaking the `jwks_uri` by default if it is not overridden using `options`.

## Checklist 🧢

~- [ ] Documentation~
- [X] Tests
- [X] Ready to be merged


balazsorban44 (approved):
Good catch!

PR #2829, opened 2021-09-23
feat(providers): refactor Cognito provider

balazsorban44:
Fixed `cognito` provider config to support v4.

Also logs `error_description` if present. (partially related to #1820)

#2524

Thanks for @s-kris providing the AWS infrastructure to test this!

Proof after signing in:

~WIP: I would like to see if we can get the `image` by default as well.~


ndom91 (approved):
Ah finally a straightforward openid-connect following provider 😅 

LGTM 👍

PR #2842, opened 2021-09-26
fix(providers): make string endpoint handlers overrideable

balazsorban44:
Fixes the issue outlined in #2807 and also improves some internal TypeScript types.

In short, if we defined a provider like this:

```js
{
  ...
  authorization: "https://example.com/auth/token?scope=email"
}
```

and the user tried overriding like this:

```js
YourProvider({
  authorization: {params: {scope: "not-email"}}
})
```

This would still result in `scope` being set to `email`. See the linked discussion for more info.

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -11,22 +11,38 @@ import type {
   Awaitable,
 } from ".."
 
-import type { Provider } from "../providers"
+import type {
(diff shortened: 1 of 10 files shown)


ndom91 (approved):
LGTM 👍

PR #2848, opened 2021-09-27
feat(providers): refactor Slack provider

balazsorban44:
Fixed `slack` provider config to support v4.

#2524

Proof after signing in:

One downside is that Slack required the `redirect_uri` or `callbackUrl` to start with `https`.
I used `ngrok` to bypass this problem, also setting NEXTAUTH_URL to the same URL what `ngrok` gave me.


ndom91 (approved):
LGTM :+1:

PR #2485, opened 2021-08-05
add Keycloak provider

bbigras:
## Reasoning 💡

Add a Keycloak provider.

I tested with a self-hosted Keycloak.

I based this PR on the auth0 provider.

I was first wondering if it was possible in https://github.com/nextauthjs/next-auth/pull/2411#issuecomment-892768889.

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## Affected issues 🎟

Related: #1261

PR #2851, opened 2021-09-28
fix(providers): convert Keycloak to TS, add `picture` by default

balazsorban44:
Slight iteration on #2485 to add a profile picture by default, so it is more consistent with other Provider defaults:

To achieve this, I opened up the Keycloak Admin Console, and under Users -> User -> Attributes, I added a `picture` attribute pointing to our logo, which is hen mapped in the `profile` callback.


ndom91 (approved):
LGTM :+1:

PR #2855, opened 2021-09-28
feat: rename `callbacks.signIn` to `callbacks.shouldSignIn`

balazsorban44:
Due to constant misuse of this callback (people use this to mutate the `user` object, and create stuff in the database, etc. This should be achieved either in the `jwt` callback or in the appropriate `events` that we expose instead. The mutation is probably the worse of these two, because that is not something we can guarantee to work in the future.)

I propose that we change the name to better describe the intended functionality.

In this PR, I introduced it in a non-breaking manner since I did not want to make new breaking changes in the `beta` releases, but it is up for debate whether we should just rip off the plaster...


ndom91 (approved):
Nice! Good API design move, `mergeCallbacks` for backward compat, and actionable typescript note. :+1: :+1: PR haha

balazsorban44:
After discussing, since we don't have a better alternative for some use cases, we should do it at a later point.

ivan-kleshnin:
> Due to constant misuse of this callback (people use this to mutate the user object, and create stuff in the database, etc. This should be achieved either in the jwt callback or in the appropriate events that we expose instead. The mutation is probably the worse of these two, because that is not something we can guarantee to work in the future.)

I can reply to this as a newcomer. Documentation is really lacking, should be a top priority rather than new features IMO.

1. To check if the user can login we have to query a database.
Refetching the same user data in `jwt` callback sounds like an extra query.

Fetching the data in `signIn` and mutating the `user` object to expose e.g. `user.role` downstream
is the only working approach I've found.

2. If there's a DB read, people will add a DB write unless it's explicitly recommended otherwise AND an alternative is proposed.

3. Events exists in the context of the Adapter. Docs do not explain how to register a user without an adapter to begin with.

Issue #2612, opened 2021-08-27
Credentials provider: request headers undefined in authorize callback on new Nodes


svkoskin:
### Description 🐜

When using Credentials provider, the authorize callback should receive the incoming message object as the second parameter. However, on certain Node versions the incoming message has `undefined` as `headers`.

This happens on Node versions 14.15.2, >=15.1 and >=16.0 and is related to this change: https://github.com/nodejs/node/pull/35281 There is an issue on Node but it is closed (https://github.com/nodejs/node/issues/36550)

 A developer seems to be advising against using the spread operator on `headers` (https://github.com/nodejs/node/issues/36550#issuecomment-747601297) which is exactly what NextAuth is doing before calling `authorize` callback (https://github.com/nextauthjs/next-auth/blob/0fae0c7a8eb224964e38da409ecb9ba0497e8c50/src/server/routes/callback.js#L339).

### Is this a bug in your own project?

No

### How to reproduce ☕️

```
import Providers from `next-auth/providers`
...
providers: [
  Providers.Credentials({
    name: 'Credentials',
    credentials: {
      username: { label: "Username", type: "text", placeholder: "jsmith" },
      password: { label: "Password", type: "password" }
    },
    async authorize(credentials, req) {
      // The following statement should output a structure with the HTTP request headers.
      // On Node 14.15.2, >=15.1 and >=16.0 it outputs undefined and none of the headers can be accessed.
      console.log(req.headers);
      }
    }
  })
]
```

### Environment 🖥

```
  System:
    OS: Linux 4.19 Ubuntu 20.04.2 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i5-6500 CPU @ 3.20GHz
    Memory: 16.37 GB / 25.02 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 16.2.0 - ~/.nvm/versions/node/v16.2.0/bin/node
    npm: 7.13.0 - ~/.nvm/versions/node/v16.2.0/bin/npm
  npmPackages:
    next: 11.1.0 => 11.1.0
    next-auth: 3.28.0 => 3.28.0
    react: 17.0.2 => 17.0.2

```

### Contributing 🙌🏽

No, I am afraid I cannot help regarding this

ilijaNL:
[...]

Issue #2433, opened 2021-07-27
REST API bearer tokens without JWT?


jaredpalmer:
### Question 💬

It appears that it is currently not possible for to use next-auth to secure a REST API that uses cors + bearer tokens, unless you use jwt. However, JWT is not feasible for my use case as I must have revokable sessions and (api keys) and I do not want to maintain a blocklist. 

It seems that `getSession({ req })` could easily extract the session token and fetch/load the session if it the request has `headers.authorization` set but no cookie.

I could do this on my own for now, but callbacks wouldn't fire. 

Please advise. 

### How to reproduce ☕️

try to make an authenticated request with an HTTP client, without JWT.

```ts
export default (req, res, next) => {
      const session = await getSession({ req });
      // this doesn't work unless the token is passed via cookie. 
      // it _should_ work if it is passed in authorization header though
      if (session) {
        req.user = session.user as Express.User;
        next();
      } else {
        next(getUnauthorized());
      }
    });
```

### Contributing 🙌🏽

Yes, I am willing to help answer this question in a PR

balazsorban44:
Could you use the `getToken` method instead? It will extract the session from Authorization headers as well, for convenience:

https://github.com/nextauthjs/next-auth/blob/75ca097ff7585016264f4c6848e58da74c77aaf8/src/lib/jwt.js#L126

jaredpalmer:
No. The `getToken` only works with JWTs. It will not load the session object because I am not using a JWT.

What you actually want on the sever is this:

```tsx
const getServerSession = ({ req, options }) => {   ​
 ​const token = getTokenFromCookie(req) || getTokenFromAuthHeader(req)
  ​if (options.jwt) {
    return options.callbacks.session(decode(token), token)
  }
   // lookup session and user directly in the database
  const session = adapter.getSessionByToken(token)
  const user = adapter.getUser(session.userId)
  return options.callbacks.session(session, user)
}
```

balazsorban44:
I agree, I plan to add something like `getServerSession` when v4 is stable. :+1:

balazsorban44:
@jaredpalmer this is under development now: https://github.com/nextauthjs/next-auth/issues/1535#issuecomment-931785322

jaredpalmer:
Bump

PR #2857, opened 2021-09-29
refactor: decouple Next.js from core

balazsorban44:
Using a similar API to SvelteKit endpoints (https://kit.svelte.dev/docs#routing-endpoints), we can create pure functions without the need of mutating `req` and `res` deeply nested in the core of NextAuth.js.

This will open up new possibilities like expanding to other frameworks #2294 and implement methods like `getServerSession` #1535

Thanks to the TypeScript refactoring of the core I did earlier in #2552, I have so much confidence doing this, it might actually make it into v4 after all...

pages:
- [x] signin
- [x] signout
- [x] error
- [x] verify-request
- [x] new-user

routes:
- [x] session
- [x] providers
- [x] csrf
- [x] _log
- [x] signin
- [x] signout
- [x] callback

Also, related: #2612, #2433


balazsorban44:
Done, but I will do some cleanup and testing.

ndom91 (approved):
LGTM :+1:

Issue #2735, opened 2021-09-12
Type error when using Azure B2C


renebrummel:
### Question 💬

Hi all,

I'm trying to set up next-auth for Azure AD B2C. When I sign in I get this error. Can anyone give me  a clue? I'm on version "next-auth": "^4.0.0-next.26"

```
https://next-auth.js.org/errors#signin_oauth_error Cannot read property 'url' of undefined {
  error: {
    message: "Cannot read property 'url' of undefined",
    stack: "TypeError: Cannot read property 'url' of undefined\n" +
      '    at openidClient (X:\\Sync\\Red and Bundle\\Development\\Source Code\\Red.CustomerPortal.Web\\.next\\server\\pages\\api\\auth\\[...nextauth].js:2094:69)\n' +
      '    at getAuthorizationUrl (X:\\Sync\\Red and Bundle\\Development\\Source Code\\Red.CustomerPortal.Web\\.next\\server\\pages\\api\\auth\\[...nextauth].js:1799:51)\n' +
      '    at Object.signin (X:\\Sync\\Red and Bundle\\Development\\Source Code\\Red.CustomerPortal.Web\\.next\\server\\pages\\api\\auth\\[...nextauth].js:3332:68)\n' +
      '    at NextAuthHandler (X:\\Sync\\Red and Bundle\\Development\\Source Code\\Red.CustomerPortal.Web\\.next\\server\\pages\\api\\auth\\[...nextauth].js:1045:31)\n' +
      '    at runMicrotasks (<anonymous>)\n' +
      '    at processTicksAndRejections (internal/process/task_queues.js:94:5)\n' +
      '    at async X:\\Sync\\Red and Bundle\\Development\\Source Code\\Red.CustomerPortal.Web\\.next\\server\\pages\\api\\auth\\[...nextauth].js:1094:32\n' +
      '    at async Object.apiResolver (X:\\Sync\\Red and Bundle\\Development\\Source Code\\Red.CustomerPortal.Web\\node_modules\\next\\dist\\server\\api-utils.js:101:9)\n' +
      '    at async DevServer.handleApiRequest (X:\\Sync\\Red and Bundle\\Development\\Source Code\\Red.CustomerPortal.Web\\node_modules\\next\\dist\\server\\next-server.js:770:9)\n' +
[...]

PR #2860, opened 2021-09-29
[POC/Review] - Issue 2735/type error in azure b2c

BenjaminWFox:
## Reasoning 💡

This is a POC Example PR for Issue #2735

It shows updates that were required in NextAuth in order to successfully log in to Azure AD B2C, as well as my Provider setup in the nextauth config file.

### Additionally

This required using Custom Policies in order to create a UserInfo endpoint for the B2C flow.

This appears to be related to the usage (which is new in v4? unsure.) of the `openid-client` library, when it creates an `Issuer`.

In order to create the UserInfo endpoint is Azure B2C, I followed this documentation:

1. [Create Custom User Flows](https://docs.microsoft.com/en-us/azure/active-directory-b2c/tutorial-create-user-flows)
2. [Implement UserInfo Endpoint](https://docs.microsoft.com/en-us/azure/active-directory-b2c/userinfo-endpoint)

### Important note

The Custom Policy documentation worked well and was relatively straight forward (if complex 😬) _EXCEPT_ for one issue in the `SignUpOrSignin.xml` file where the line:

`<OutputClaim ClaimTypeReferenceId="email" />`

should change to

`<OutputClaim ClaimTypeReferenceId="signInNames.emailAddress" PartnerClaimType="email" />`

(reference: https://stackoverflow.com/q/59669666/1763258)

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## Affected issues 🎟

Addresses, but does not necessarily fully fix or close, #2735 - type error in azure b2c


src/providers/azure-ad-b2c.js
```diff
@@ -32,10 +36,24 @@ export default function AzureADB2C(options) {
...
+      }
+
       return {
-        id: profile.oid,
+        id,
         name,
-        email: profile.emails[0],
+        email,
         image: null,
       }
```
BenjaminWFox, line 58: The way the properties are returned with a Custom Policy is rather different than the way the properties were returned with an out-of-the-box User Flow.

balazsorban44:
it looks like B2C supports OIDC 

 https://docs.microsoft.com/en-us/azure/active-directory-b2c/openid-connect

which would make the configuration much easier, and the userinfo endpoint config wouldn't be necessary, since you get an id_token.

I would like to avoid any customization if possible. we only need an id, name, email and picture.

src/providers/azure-ad-b2c.js
balazsorban44: We likely want to go with the most basic flow for the built-in provider though, and rather educate people better on how to achieve custom behaviour in https://next-auth.js.org/providers/azure-ad-b2c

balazsorban44:
Closing in favour of #2862, thanks for the help!

PR #2875, opened 2021-10-01
feat(providers): refactor Apple provider

balazsorban44:
Thanks for helping out with testing this, @sergeymishin

Fixed `apple` provider config to support v4.

#2524

Proof (🤣...) after signing in:
image

**NOTES:** 
- no name
- email only in certain situations
- no image
- needs `https` for local development
- We removed the option to generate `clientSecret` on the fly. From now on, you will have to do it in userland like so:
- Docs need to be updated https://github.com/nextauthjs/docs/blob/main/docs/providers/apple.md

```ts
export default async function auth(req, res) {
    ...
    providers: [
      AppleProvider({
        clientSecret: await generateClientSecret({
        })
      })
    ]
  })
}
```


piraka9011:
Bumping this since we need this PR for our project please.

balazsorban44:
> Bumping this since we need this PR for our project please.

https://github.com/nextauthjs/next-auth/pull/2875#issuecomment-937812600

app/pages/api/auth/[...nextauth].ts
```diff
@@ -23,6 +23,19 @@ import CognitoProvider from "next-auth/providers/cognito"
...
+// import { generateClientSecret } from "next-auth/providers/apple"
+// const appleSecret = await generateClientSecret({
+//   clientId: process.env.APPLE_ID,
+//   keyId: process.env.APPLE_KEY_ID,
+//   teamId: process.env.APPLE_TEAM_ID,
+//   privateKey: process.env.APPLE_PRIVATE_KEY,
+// })
+// console.log(
+//   `APPLE_SECRET=${appleSecret} # Created: ${new Date().toISOString()} Update if more than 6 months old.`
+// )
```
Comment (lluia) [inline app/pages/api/auth/[...nextauth].ts:38]: I think we could move this to the [`README`](https://github.com/nextauthjs/next-auth/blob/main/app/README.md) within the example app folder.

0ubbe (approved):
So far changes look good to me 🌟 , I understand we still need to fix the issue of no name/email consistently returned and update the docs 👍🏽

balazsorban44:
@piraka9011, @mmmoussa I would appreciate it if you could test out the `experimental` version listed here: https://github.com/nextauthjs/next-auth/pull/2875#issuecomment-980467597

It is on par with `beta.7` in any other way. We will release v4 next week, but I don't have access to an Apple developer account, and that makes testing hard.

Could you report back what the `profile` callback returns? When I tried together with @sergeymishin, I got no visible info back. :confused:


There is also the `generateClientSecret` method, could you test that as well? :pray:

piraka9011:
Glad to see this!
FYI, Apple requires redirect urls to be served over https and does not allow redirect urls containing localhost.
We had to do the following steps when testing locally:

1. From the root directory of the project, make a subdirectory called certificates:

  ```sh
  mkdir certificates; cd certificates
  ```

2. Create an SSL certificate:

  ```sh
  openssl req -x509 -out localhost.crt -keyout localhost.key -newkey rsa:2048 -nodes -sha256 -subj '/CN=localhost' -extensions EXT -config <( printf "[dn]\nCN=localhost\n[req]\ndistinguished_name = dn\n[EXT]\nsubjectAltName=DNS:localhost\nkeyUsage=digitalSignature\nextendedKeyUsage=serverAuth")
  ```

3. In your `.env` file, set the `NEXTAUTH_URL` environment variable to your app's domain (ex. `mysite.io`)

  ```sh
  NEXTAUTH_URL=https://mysite.io:3000
  ```

4. Add `127.0.0.1 mysite.io` to your hosts file (`sudo vim /private/etc/hosts`). This allows you to use `mysite.io` as an alias to localhost. 

5. Create a node server using https with your SSL certificate.

```js
const { createServer } = require("https")
const { parse } = require("url")
const next = require("next")
const fs = require("fs")

const dev = process.env.NODE_ENV !== "production"
const app = next({ dev })
const handle = app.getRequestHandler()

const httpsOptions = {
  key: fs.readFileSync("./certificates/localhost.key"),
  cert: fs.readFileSync("./certificates/localhost.crt"),
}

app.prepare().then(() => {
  createServer(httpsOptions, (req, res) => {
    const parsedUrl = parse(req.url, true)
    handle(req, res, parsedUrl)
  }).listen(3000, (err) => {
    if (err) throw err
    console.log("> Ready on https://localhost:3000")
  })
```

Now you can visit `https://mysite.io:3000` and try logging in w/ Apple credentials.

balazsorban44:
For testing only, I would just use https://ngrok.com

No hassle, no custom next server.

balazsorban44:
I've merged this. if any issues arise we can open a new PR.

PR #2876, opened 2021-10-01
feat(signin): Support passing any argument to credentials input

zakaryan2004:
## Reasoning 💡

Spreading the object into the input tag allows developers to specify any attribute for the input tag used in the builtin sign-in page, such as 'autocomplete', 'autofocus', etc.

## Checklist 🧢

- [x] Documentation (https://github.com/nextauthjs/docs/pull/55)
- [x] Tests
- [x] Ready to be merged

Tested manually, by passing the attributes previously supported (e.g. type, value), new ones (e.g. autocomplete), and nonexisting ones (e.g. bla), as well as not passing anything to test the defaults.

Documentation changes will be pushed to the docs repo.

## Affected issues 🎟

No behavior change.
Removed the hardcoded attributes which just set the defaults of the 'input' tag and won't cause any behavior change in case they are absence from the object.
I left the placeholder attribute, as removing it would change behavior, but maybe this one can be removed as well.

### DIFF
--- a/src/server/pages/signin.tsx
+++ b/src/server/pages/signin.tsx
@@ -106,11 +106,10 @@ export default function Signin({
                       <input
                         name={credential}
                         id={`input-${credential}-for-${provider.id}-provider`}
-                        type={provider.credentials[credential].type || "text"}
-                        value={provider.credentials[credential].value || ""}
                         placeholder={
                           provider.credentials[credential].placeholder || "Password"
                         }
+                        {...provider.credentials[credential]}
                       />
                     </div>
                   )


balazsorban44 (approved):
Thanks! Sounds good!

zakaryan2004:
After using the new version on one of my other projects (which I didn't test this change with), I found out that, even though all major browsers have input's type set to "text" as default, CSS doesn't consider this the default, and the current stylesheet won't style an input field which doesn't have the "type" attribute set. I am really really sorry I let something simple like this pass by. Should I create a new PR to fix this? I suppose it's not a good idea to force push the branch, even though it's beta.

balazsorban44:
yes, please create a fix PR, no worries! thanks for checking back!

PR #2881, opened 2021-10-02
fix(signin): Set default input type to text to fix CSS

zakaryan2004:
Partially reverts c5bd99d92a4728d8ace6f41e07da02a3a94f0e15, as CSS doesn't consider type="text" the default for the input tag, so the styles break if no type is specified.

Signed-off-by: Gegham Zakaryan <zakaryan.2004@outlook.com>

## Reasoning 💡

Partially reverts c5bd99d92a4728d8ace6f41e07da02a3a94f0e15.

Even though all major browsers have HTML input tag's type set to "text" by default, CSS doesn't consider this the default, and the current stylesheet won't style an input field which doesn't have the "type" attribute set.

## Checklist 🧢

- [ ] ~Documentation~
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/server/pages/signin.tsx
+++ b/src/server/pages/signin.tsx
@@ -106,6 +106,7 @@ export default function Signin({
                       <input
                         name={credential}
                         id={`input-${credential}-for-${provider.id}-provider`}
+                        type={provider.credentials[credential].type || "text"}
                         placeholder={
                           provider.credentials[credential].placeholder || "Password"
                         }


zakaryan2004:
I don't know what else can break, but let's leave it unmerged for a day, maybe I'll find another problem, lol

balazsorban44 (approved):
Thanks!

balazsorban44:
> I don't know what else can break, but let's leave it unmerged for a day, maybe I'll find another problem, lol

sorry, I merged. 😅 I only saw this after...

zakaryan2004:
> > I don't know what else can break, but let's leave it unmerged for a day, maybe I'll find another problem, lol
> 
> sorry, I merged. 😅 I only saw this after...

No problem, I doubt anything would even break anyway

PR #2910, opened 2021-10-07
fix(providers): resize default AzureAD profile picture

balazsorban44:
See https://github.com/nextauthjs/next-auth/issues/2524#issuecomment-933764240 for the explanation.

the docs should be updated as well: https://github.com/nextauthjs/docs/issues/62

Relevant issue: #2919


ndom91 (approved):
LGTM :+1:

PR #2916, opened 2021-10-07
fix: conditionally render theme logo on builtin pages

ndom91:
## Reasoning 💡

Theme logo was rendering the img tag with undefined src if no `theme.logo` was defined, which is obviously unwanted behavior.

I rendered a span in its place when no logo is defined to have something with that class to retain the space so the first provider isn't smushed up against the top of the screen. 

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

## Affected issues 🎟

Fixes #2914

### DIFF
--- a/src/css/index.css
+++ b/src/css/index.css
@@ -152,13 +152,12 @@ a.site {
   position: absolute;
   width: 100%;
   height: 100%;
-  display: table;
+  display: grid;
+  place-items: center;
   margin: 0;
   padding: 0;
 
   > div {
-    display: table-cell;
-    vertical-align: middle;
     text-align: center;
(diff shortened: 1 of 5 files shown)


balazsorban44:
Why do we need an empty span instead of `null` when no logo?

ndom91:
> I think I would prefer no span, actually. That's how it looked before as well.

Yeah so I checked and previously we were using a `table-cell` div and centering it in the page actually. 

image

But yeah doesn't matter to me, just let me know what ya wanna do :+1:

balazsorban44 (approved):

PR #2932, opened 2021-10-10
feat(providers): add Notion provider

### LINKED ISSUE #2905: [V4] Cognito Authentication Errors
### Description 🐜

Hi I'm using version "next-auth": "^4.0.0-beta.4"

```
    CognitoProvider({
      clientId: `${process.env.COGNITO_CLIENT_ID}`,
      clientSecret: `${process.env.COGNITO_CLIENT_SECRET}`,
      issuer: `${process.env.COGNITO_ISSUER}`,
    }),

```

I'm getting below error.
```
[next-auth][error][SIGNIN_OAUTH_ERROR]
https://next-auth.js.org/errors#signin_oauth_error Cannot read property 'params' of undefined {
  error: {
    message: "Cannot read property 'params' of undefined",
    stack: "TypeError: Cannot read property 'params' of undefined\n" +
      '    at getAuthorizationUrl (D:\\Work\\NodeJS\\BDiLabs\\next+mui\\frontend\\node_modules\\next-auth\\server\\lib\\oauth\\authorization-url.js:25:48)\n' +
      '    at Object.signin (D:\\Work\\NodeJS\\BDiLabs\\next+mui\\frontend\\node_modules\\next-auth\\server\\routes\\signin.js:30:68)\n' +    
      '    at NextAuthHandler (D:\\Work\\NodeJS\\BDiLabs\\next+mui\\frontend\\node_modules\\next-auth\\server\\index.js:202:31)\n' +
      '    at runMicrotasks (<anonymous>)\n' +
      '    at processTicksAndRejections (internal/process/task_queues.js:93:5)\n' +
      '    at async D:\\Work\\NodeJS\\BDiLabs\\next+mui\\frontend\\node_modules\\next-auth\\server\\index.js:251:32\n' +
      '    at async Object.apiResolver (D:\\Work\\NodeJS\\BDiLabs\\next+mui\\frontend\\node_modules\\next\\dist\\server\\api-utils.js:101:9)\n' +
      '    at async DevServer.handleApiRequest (D:\\Work\\NodeJS\\BDiLabs\\next+mui\\frontend\\node_modules\\next\\dist\\server\\next-server.js:770:9)\n' +
      '    at async Object.fn (D:\\Work\\NodeJS\\BDiLabs\\next+mui\\frontend\\node_modules\\next\\dist\\server\\next-server.js:661:37)\n' +   
      '    at async Router.execute (D:\\Work\\NodeJS\\BDiLabs\\next+mui\\frontend\\node_modules\\next\\dist\\server\\router.js:205:32)',      
    name: 'TypeError'
  },

  message: "Cannot read property 'params' of undefined"
}
````

No

### How to reproduce ☕️

I'm guessing its coming from here : https://github.com/nextauthjs/next-auth/blob/beta/src/server/lib/oauth/authorization-url.js#L21

```
    const params = { ...provider.authorization.params,
      ...req.query
    };
```
When i make it optional provider.authorization?.params  it works, but then another issue comes up, session is missing data it used to name like 'name'. This data was there in v3

```
  profile: {
    id: 'c5ceee18-999-43de-8ea5-d847e6c48b12',
    name: undefined,
    email: 'niranjith.perumal@moo.com',
    image: undefined
  },
``


### Screenshots / Logs 📽

I can share cognito credentials if it will help 

### Environment 🖥

  System:
    OS: Windows 10 10.0.19043
    CPU: (8) x64 Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
    Memory: 4.36 GB / 15.85 GB
  Binaries:
    Node: 14.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.5 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 7.24.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 94.0.4606.71
    Edge: Spartan (44.19041.1023.0), Chromium (93.0.961.44)
    Internet Explorer: 11.0.19041.1202
  npmPackages:
    next: 11.1.3-canary.7 => 11.1.3-canary.7
    next-auth: ^4.0.0-beta.4 => 4.0.0-beta.4
    react: ^17.0.2 => 17.0.2

### Contributing 🙌🏽

balazsorban44:
Currently trying to figure out why the `owner` is not correctly returned.

https://developers.notion.com/docs/authorization#exchanging-the-grant-for-an-access-token

Fixes #2905

Proof after login:

image

_The image does not show because I got rate-limited with all my failed attempts. 😅_

### DIFF
--- a/app/pages/api/auth/[...nextauth].ts
+++ b/app/pages/api/auth/[...nextauth].ts
@@ -23,6 +23,7 @@ import CognitoProvider from "next-auth/providers/cognito"
 import SlackProvider from "next-auth/providers/slack"
 import Okta from "next-auth/providers/okta"
 import AzureB2C from "next-auth/providers/azure-ad-b2c"
+import Notion from "next-auth/providers/notion"
 
 // import { PrismaAdapter } from "@next-auth/prisma-adapter"
 // import { PrismaClient } from "@prisma/client"
@@ -167,6 +168,10 @@ export default NextAuth({
       clientSecret: process.env.AZURE_B2C_SECRET,
       tenantId: process.env.AZURE_B2C_TENANT_ID,
       primaryUserFlow: process.env.AZURE_B2C_PRIMARY_USER_FLOW,
+    }),
+    Notion({
+      clientId: process.env.NOTION_ID,
+      clientSecret: process.env.NOTION_SECRET,
     }),
   ],
   jwt: {
--- a/src/providers/notion.ts
+++ b/src/providers/notion.ts
@@ -1,0 +1,58 @@
+import { OAuthConfig, OAuthUserConfig } from "."
+
+export interface NotionResponse {
+  /**
+   * https://developers.notion.com/reference/user#all-users
+   * https://developers.notion.com/reference/user#people
+   */
+  owner: {
+    id: string
+    name: string
+    avatar_url: string
+    email?: string
+    person: any
+  }
+  workspace_name: string
+  workspace_id: string
+  workspace_icon: string
+}
+
+export type NotionProfile = NotionResponse["owner"] &
+  Omit<NotionResponse, "owner">
+
+export default function Notion<P extends Record<string, any> = NotionProfile>(
+  options: OAuthUserConfig<P>
+): OAuthConfig<P> {
+  return {
+    id: "notion",
+    name: "Notion",
+    type: "oauth",
+    authorization: "https://api.notion.com/v1/oauth/authorize?owner=user",
+    token: "https://api.notion.com/v1/oauth/token",
+    userinfo: {
+      request(params) {
+        // https://developers.notion.com/docs/authorization#exchanging-the-grant-for-an-access-token
+        const tokens = params.tokens as unknown as NotionResponse
+        return {
+          id: tokens.owner.id,
+          name: tokens.owner.name,
+          email: tokens.owner.email,
+          avatar_url: tokens.owner.avatar_url,
+          workspace_id: tokens.workspace_id,
+          workspace_name: tokens.workspace_name,
+          workspace_icon: tokens.workspace_icon,
+        }
+      },
+    },
+    profile(profile) {
+      return {
+        id: profile.id,
+        name: profile.name,
+        email: profile.email,
+        image: profile.avatar_url,
+      }
+    },
+    checks: ["state"],
+    options,
+  }
+}
--- a/src/providers/notion.ts
+++ b/src/providers/notion.ts
@@ -1,27 +1,56 @@
 import { OAuthConfig, OAuthUserConfig } from "."
 
-export interface NotionResponse {
+interface NotionWorkspaceOwner {
+  type: "workspace"
+  workspace: true
+}
+
+interface NotionUserUser {
+  object: "user"
+  id: string
+  name: string
+  avatar_url: string
+  type: "person"
+  person: { email: string }
+}
+
+interface NotionBotUser {
+  object: "user"
+  id: string
+  name: string
+  avatar_url: string
+  type: "bot"
+  bot: any
+}
+
+interface NotionUserOwner {
+  type: "user"
+  user: NotionUserUser | NotionBotUser
+}
+
+export interface NotionResponseFields {
+  workspace_name: string
+  workspace_id: string
+  workspace_icon: string
+  bot_id: string
+}
+
+export interface NotionResponse extends NotionResponseFields {
   /**
    * https://developers.notion.com/reference/user#all-users
    * https://developers.notion.com/reference/user#people
    */
-  owner: {
-    id: string
-    name: string
-    avatar_url: string
-    email?: string
-    person: any
-  }
-  workspace_name: string
-  workspace_id: string
-  workspace_icon: string
+  owner: NotionWorkspaceOwner | NotionUserOwner
 }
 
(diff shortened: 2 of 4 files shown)


balazsorban44:
Thank you @justjake! Unfortunately adding a `Notion-Version: 2021-08-16` header to the token request had no effect. Whatever I do, `owner` is always returned as `{ type: 'workspace', workspace: true }`. I created a new workspace from scratch, with a new e-mail and a new integration, where the first thing I did was to change to OAuth. :confused:

justjake:
Maybe what’s happening here is that when you create an integration, it gets an initial token as a workspace token in its “home” workspace. When you make repeated OAuth requests for the same (user, workspace) pair you’ll receive the same token response until the integration is de-authorized on the Notion side of things.  

We could test this theory two ways:

- De-authorize your integration from your existing Notion workspace. Then on your next pass through the OAuth flow you’ll receive a new token, which should be a owner=user token. https://www.notion.so/Add-and-manage-integrations-with-the-API-910ac902372042bc9da38d48171269cd
- Create a new workspace with your existing Notion account, and go through the OAuth flow targeting the new workspace. https://www.notion.so/Create-join-switch-workspaces-3b9be78982a940a7a27ce712ca6bdcf5

I  also have more example code here which is working fine: https://github.com/justjake/passport-notion/blob/main/lib/passport-notion/strategy.ts

In any case, I apologize for the confusing API experience. I’ll pass this thread along to the API & Platform team as feedback. Thank you for your time.

justjake:
> By the way, is it possible to set Notion-Version as part of the URL, a query parameter, or a body? The header is the only part of a request that is cumbersome to set even with openid-client (the underlying OAuth library) 😅

I don’t think so. `headers` is a Provider option in v3 and working fine, is this removed in v4?

balazsorban44:
the top level header option is removed, because it did not make sense there. I might add it back later on in another form. 99% of the providers don't need it (read: only notion would require it so far out of the 30+ providers I've checked, but I have a feeling it's not my problem here)

I'll check your suggestions, but both the workspace/integration/account has been created from scratch, just an FYI for this purpose, so everything started off "clean".

I'll report back with my findings thanks for the help!

justjake:
Nice work. Before you merge please make sure that the provider produces a unique Account per workspace you auth with - because integrations are granted permission on a per workspace basis in Notion, an iteration needs a separate access_token per workspace.

justjake:
Since OAuth tokens are issued per-workspace, I think it's better to treat a Notion OAuth token as an Account, rather than treating the Notion User as an Account. 

In Notion, it's quite common to be in multiple workspaces. My workspace switcher for my personal Notion account has 6 workspaces. 

<details>
<summary>screenshot</summary>

image

</details>

If the provider uses the Notion User's `id` for the Account ID (one Account per Notion User), I won't be able to connect my next-auth app to more than one workspace at a time -- I'd have to re-authenticate each time I want to act in a different workspace. That would be quite frustrating, especially for power users or consultants who work in many more workspaces. 

Instead, it's better to use the OAuth token's `bot_id` field as the provider Account ID. That way, next-auth apps that use the Notion provider can easily implement multiple workspace support by querying for Accounts from the Notion provider, and it's easy to add a relational linkage between a NextAuth Account and the related resources fetched from the Notion API.

For example in the app I'm working on, i have my own kind of workspace switcher to select Notion workspaces, and if I add support for Github or otherwise those accounts will show up in the list easily.

image

balazsorban44:
Probably it never came up. Usually Slack integrations target a single workspace, maybe? 🤷‍♂️

If it worked in v3 it should in v4, as we didn't explicitly remove functionality, if anything at all, we significantly simplified the public APIs (https://github.com/nextauthjs/next-auth/pull/2361, https://github.com/nextauthjs/adapters/pull/171), I just feel like this extra layer of complexity will confuse users (as it confuses me as a first-time user of Notion).

As @ndom91 pointed out to me, this is only a real problrm for when you want to use an Adapter. When the session/user is stateless (persisted in a JWT formatted session cookie), you kind of have to re-authenticate between different workspaces anyway, as you would when signing in with eg. GitHub instead of Google.

jordn:
Hi. Trying to follow along here but thought i'd just add to the conversation. I don't think I have opinions about the IDs/workspaceID.
 
I had the issue marked above (using the v3 next-auth and @justjake's example)... first time through owner was just a bot (i think) `owner: { type: 'workspace', workspace: true }`. I deactivated the integration on notion and tried again and i did get the user ID etc.

So... that's all a bit weird. Not sure I understand why that was the case or whether it will always be the case. (good chance i'm doing something wrong here too).

And other oddity is that on re-login with the same workspace I select the pages given access each time. I wouldn't really expect that.

feliche93:
@jordn or @balazsorban44 were you able to create a working provider integration? I tried doing it myself and get an `[next-auth][error][GET_AUTHORIZATION_URL_ERROR] ` although the authorization url perfectly works. Could I take your example from the PR or could you post any working version? Thanks 🙏

balazsorban44:
This seemed to be a bit more complicated than expected because of Notion workspaces.

There is a good amount of useful code in this PR, but the confusion around the different token owners would probably be confusing for users of other providers, that do not have those concepts. We might revisit this after v4 has been released to v4, or we could add a "simplified" version, declaring the constraints in the documentation.

PR #2956, opened 2021-10-14
Fix | JWT Options | Use plain object instead for string for `signingKey`

LeoBorai:
Currently as pointed out in the docs for JWT options:
https://next-auth.js.org/configuration/options#json-web-token-options

An object must be provided to the `signingKey` JWT options object.
But actually a `string` is required by the types definition.

In order to make it fit the documentation I'm replacing the `string`
type with the same structure provided in the docs.

### DIFF
--- a/src/lib/jwt.js
+++ b/src/lib/jwt.js
@@ -31,7 +31,7 @@ export async function encode({
 } = {}) {
   // Signing Key
   const _signingKey = signingKey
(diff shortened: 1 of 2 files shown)


balazsorban44:
This would be a breaking change. I believe we should align the docs with the implementation instead. See https://github.com/nextauthjs/docs/issues/5

PR #2980, opened 2021-10-18
feat(react): preserve history on client-side navigation

### LINKED ISSUE #2676: Navigating to custom sign in page using `signIn()` replaces history. Is there a way to configure that or is this a bug?
### Question 💬

When invoking the front end `signIn()` method, I get redirected to my custom sign in page. If I then hit the back button, it goes back to a blank page with `about:blank` in the URL. 

**Is there a way to configure history or back behaviour when invoking the front end `singIn()` method or is this a bug?**

I'm following the [pages documentation v3](https://next-auth.js.org/v3/configuration/pages).

### How to reproduce ☕️

`[...nextauth].ts`
```typescript
...
pages: {
    signIn: '/auth/signin',
},
...
```

`Nav` component
```tsx
...
<button onClick={() => signIn()}>
    Log in
</button>
...
```

My sign in page is basically a combo of email and providers (not email).

`/pages/auth/signin.tsx`
```typescript
import { FunctionComponent } from 'react';
import { GetServerSideProps } from 'next';
import {
    signIn,
    getSession,
    getProviders,
    getCsrfToken,
    ClientSafeProvider,
} from 'next-auth/client';

interface SigninProps {
    providers: ClientSafeProvider[];
    csrfToken: string;
}

const Signin: FunctionComponent<SigninProps> = ({ providers, csrfToken }) => {
    return (<>...</>);
};

export default Signin;

export const getServerSideProps: GetServerSideProps = async (ctx) => {
    const {req, res} = ctx;
    const session = await getSession();

    if (session && res && session.access_token){
        res.writeHead(302, {
            location: '/'
        });
        res.end();
        return;
    }

    const providers = await getProviders();

    let csrfToken: string = null;
    if (Object.values(providers).find((p) => p.type === 'email')) {
        csrfToken = await getCsrfToken(ctx);
    }

    const providersWithoutEmail = Object.values(providers).filter(
        (p) => p.type !== 'email'
    );

    return {
        props: {
            providers: providersWithoutEmail,
            csrfToken,
        },
    };
};
```

### Contributing 🙌🏽

balazsorban44:
Fixes #2676

Using `window.location.href` instead of `window.location.replace` as pointed out by @ndom91 in https://github.com/nextauthjs/next-auth/issues/2676#issuecomment-913875136 will preserve the history.


ndom91 (approved):
LGTM! :+1:

PR #3011, opened 2021-10-23
feat(providers): add Pipedrive provider

estevanjantsk:
## Reasoning 💡

Added Pipedrive as a provider, because the lib does not have yet.
I tested for both version v3 and v4 and it's working.

## Checklist 🧢

- [x] Documentation [[PR](https://github.com/nextauthjs/docs/pull/83)]
- [ ] Tests
- [ ] Ready to be merged


balazsorban44 (changes requested):
Thank you for this PR!

I left some comments. I do not wish to "backport" this to v3, so the v3 related configuration is unnecessary. Let us focus on getting this working in v4 for now. Could you address my concerns and let me know if it works for you?

By the way, are you interested in adding this as TypeScript? You can see any of the `.ts` files in this directory for inspiration: https://github.com/nextauthjs/next-auth/tree/beta/src/providers

src/providers/pipedrive.ts
```diff
@@ -0,0 +1,63 @@
...
+  company_domain?:        string;
+  company_country?:       string;
+  company_industry?:      string;
+  language?:              Language;
+}
+
+interface Language {
+  language_code?: string;
+  country_code?:  string;
+}
```
balazsorban44, line 40: Could you have this as a single interface? Also, if you have it configured, could you run Prettier on the file for consistent formatting with the rest of the codebase? :pray:

balazsorban44 (changes requested):

src/providers/pipedrive.ts
estevanjantsk: I've fixed the interface. however, I don't have Prettier configured, could you run for me, please?

lluia: @estevanjantsk I run `prettier` for you on the file 👍🏽 , for next time, if you already did `npm install` on the cloned project just run `npm run lint:fix` and it'll format the file for you 💅🏽

I think this provider is ready to go then @balazsorban44  🤞🏽 ?

balazsorban44 (approved):
Thanks! I will take your word that you have tested, but as a proof, could you just link to an image or the JSON returned by the `profile` callback that we can refer to later on if someone has problems with this provider in the future?

Something like here: https://github.com/nextauthjs/next-auth/pull/2848#issue-1008276523 or just the JSON output.

Thank you!

PR #3045, opened 2021-10-27
Added support for zoom in beta

dijs:
## Reasoning 💡

- `scope` needs to be an empty string
- profile `image` needs to read from `pic_url`


balazsorban44 (changes requested):

balazsorban44 (approved):
Thanks! I just ran Prettier on the file for consistency with the rest of the code.

Issue #3037, opened 2021-10-27
Support Next.js 12 Middleware


ramiel:
### Description 📓

After watching the recent Nextjs conf, I'm wondering if there's a way to move authentication api (here it's the `/session` endpoint) to run on the edge instead that on server. Looking at this very basic example 

https://youtu.be/PWi-49ODKAs?t=3203

maybe there's a way to take advantage of this in NextAuth too, even if in an opt-in way.

### How to reproduce ☕️

Nothing to reproduce

### Contributing 🙌🏽

Yes, I am willing to help implement this feature in a PR

balazsorban44:
FYI there is a working example here https://github.com/nextauthjs/next-auth-example/pull/40

ramiel:
That's great, a good starting point.    
Is RSC  "Rendered server components" or similar?

ramiel:
Trying this by myself I get a `TypeError: getCurves is not a function` error on jose.

balazsorban44:
React Server Components: https://reactjs.org/blog/2020/12/21/data-fetching-with-react-server-components.html

You will need `4.0.0-beta.6` at least, `jose` is working fine for the use case we have in `next-auth`

ramiel:
Sure,that one 😀. Version beta 6 has problems for me, so I cannot upgrade. I'll have another look later.

erlinghaugstad:
> FYI there is a working example here [nextauthjs/next-auth-example#40](https://github.com/nextauthjs/next-auth-example/pull/40)
> 
> it's only working with jwt sessions at the moment and there is no way to pass the session down as props to the page yet. So the client will have to make an additional API request and the authenticated page will experience a loading state, unlike if you use `getServerSideProps`. In my opinion, it means there is little gain using middleware here at the moment.

@balazsorban44 One thing that you can gain by using middlewares as is, is to for instance limit access to an entire subtree of paths, i.e adding a middleware to `/client/[id]/_middleware.ts` and check in the middleware if the user actually should have access to the client in question, i.e authorize. I can live with the double requests for now(not live yet) 😄 

Looking forward to a fully fledged integration with this, it does seem quite powerful 💪🏻

batuhanbilginn:
import { getToken } from "next-auth/jwt"
import { NextResponse } from "next/server"

/** @param {import("next/server").NextRequest} req */
export async function middleware(req) {
  if (req.nextUrl.pathname === "/middleware-protected") {
[...]

PR #3062, opened 2021-10-29
refactor: use universal modules in `next-auth/jwt`

panva:
This updates the `src/jwt/index.ts` module to only use universal modules for uuid and hkdf (in addition to already universal jose) to get around bundle issues and sizes towards #3037


balazsorban44 (approved):
Thank you!

PR #3066, opened 2021-10-29
docs(readme): opencollective domain

nicovak:
I had 502 using `opencollective.org` for some time, also, the correct domain looks like `opencollective.com`


balazsorban44 (changes requested):
Thanks, could you maybe rephrase the sentence to remove the "recently" part as well? 🙏

nicovak:
I ran prettier also to format the file

balazsorban44 (approved):

PR #3100, opened 2021-11-03
fix: correctly transpile all client-side submodules

balazsorban44:
Some modules weren't correctly transpiled for the client-side code.


ndom91 (approved):
Good catch!

PR #3142, opened 2021-11-07
refactor: decouple CSRF-state

balazsorban44:
Instead of calculating the state value from the CSRF token, we create a unique value for every flow, save it in a cookie, and delete it after usage, much like with the PKCE flow.

This PR introduces a new cookie called `next-auth.state`.

The default expiry is 15 minutes.


ndom91 (approved):
LGTM :+1: 

rfc6749 rubbing off on ya? Haha

balazsorban44:
Yeah... This will somewhat increase security as well. :muscle:

PR #3143, opened 2021-11-07
feat: make missing secret an error

balazsorban44:
If no `secret` specified in `NextAuthOptions`, `next-auth` generates one based on the user configuration. If the user does not use an OAuth provider, the generated secret will not contain enough entropy. Someone could guess the user's config, thus gaining access to the JWT encryption method. We already make this clear in the docs by saying:

> The default behaviour is volatile, and it is strongly recommended you explicitly specify a value to avoid invalidating end user sessions when configuration changes are deployed.

This PR makes this an error in production.

The corresponding documentation is also made more clear, see https://github.com/nextauthjs/docs/pull/103


0ubbe (dismissed):
Looks good 👏🏽 ✅

balazsorban44:
I moved the user config assertion to the core, and refactored the code so that in case of a configuration error, we either render the default error page or redirect to the user's custom one. The error is still logged through `logger`, but the client won't expose the cause of the actual error, it will show something like this:

image

src/core/lib/assert.ts
```diff
@@ -0,0 +1,71 @@
...
+  })
+
+  if (hasCredentials) {
+    if (options.session?.strategy === "database" || options.adapter) {
+      return new UnsupportedStrategy(
+        "Signin in with credentials only supported if JWT strategy is enabled"
+      )
+    }
+    if (
+      options.providers.every((p) => p.type !== "credentials" || p.authorize)
```
balazsorban44, line 60: use `.some`

src/core/lib/assert.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  let hasCredentials, hasEmail
+
+  options.providers.forEach(({ type }) => {
+    if (type === "credentials") hasCredentials = true
+    else if (type === "email") hasEmail = true
+  })
+
+  if (hasCredentials) {
+    if (options.session?.strategy === "database" || options.adapter) {
```
balazsorban44, line 54: `options.adapter` would be OK, if there are other provider types

0ubbe (approved):
Pair review, looking great 💪🏽

PR #3144, opened 2021-11-07
feat: rename session strategy

balazsorban44:
This PR makes it hopefully more clear how the user can use the different session strategies.

1. No adapter, `strategy: "jwt"`: This is the default. The session is saved in a cookie and never persisted anywhere.
2. With Adapter, `strategy: "database"`: If an Adapter is defined, this will be the implicit setting. No user config is needed.
3. With Adapter, `strategy: "jwt"`: The user can explicitly instruct `next-auth` to use JWT even if a database is available. This can result in faster lookups in compromise of lowered security. Read more about: https://next-auth.js.org/faq#json-web-tokens

Docs: https://github.com/nextauthjs/docs/pull/105

### DIFF
--- a/src/core/init.ts
+++ b/src/core/init.ts
@@ -85,7 +85,8 @@ export async function init({
     providers,
     // Session options
     session: {
-      jwt: !userOptions.adapter, // If no adapter specified, force use of JSON Web Tokens (stateless)
+      // If no adapter specified, force use of JSON Web Tokens (stateless)
+      strategy: userOptions.adapter ? "database" : "jwt",
       maxAge,
       updateAge: 24 * 60 * 60,
       ...userOptions.session,
--- a/src/core/lib/callback-handler.ts
+++ b/src/core/lib/callback-handler.ts
@@ -36,7 +36,7 @@ export default async function callbackHandler(params: {
     adapter,
     jwt,
     events,
-    session: { jwt: useJwtSession },
+    session: { strategy: sessionStrategy },
   } = options
 
   // If no adapter is configured then we don't have a database and cannot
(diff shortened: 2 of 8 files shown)


0ubbe (approved):
💯

zigang93:
still get CALLBACK_CREDENTIALS_JWT_ERROR ,
even I already set the strategy to jwt
```ts
session: {
    strategy: 'jwt',
  },
```

davecarlson:
Same as above, unable to get a csrf token:

[next-auth][error][CALLBACK_CREDENTIALS_JWT_ERROR] 
https://next-auth.js.org/errors#callback_credentials_jwt_error Signin in with credentials only supported if JWT strategy is enabled UnsupportedStrategy [UnsupportedStrategyError]: Signin in with credentials only supported if JWT strategy is enabled

aokigit:
Having the exact same problem as the two people above me. Tried both the old and new strategy way and always receiving the `CALLBACK_CREDENTIALS_JWT_ERROR` error.

balazsorban44:
Addressed already.

#3216 reported this, the fix has landed, release coming :soon:. Use the version here ttps://github.com/nextauthjs/next-auth/pull/3217#issuecomment-970646869 or `patch-package` if you cannot wait.

rabin-nyaundi:
I'm having the same issue with next-auth v4. 
```
[next-auth][error][CALLBACK_CREDENTIALS_JWT_ERROR] 
https://next-auth.js.org/errors#callback_credentials_jwt_error Signin in with credentials only supported if JWT strategy is enabled UnsupportedStrategy [UnsupportedStrategyError]: Signin in with credentials only supported if JWT strategy is enabled

```

renchris:
Is there a way to use next-auth for using jwt tokens in session storage but not using it nor including a provider for user login authentication?

PR #3168, opened 2021-11-10
lazy load nodemailer

anguskeatinge:
lazy load nodemailer


balazsorban44:
You can test it yourself, but this won't actually be tree-shaken. From my experience, when built with `next build`, `nodemailer` was still part of the bundle, and I could not even start a server with `next dev`. This is because (if my understanding is correct), Next.js requires dynamic imports in development even if it's in conditionals.

anguskeatinge:
Makes sense - thanks for the detailed explanation!

PR #3189, opened 2021-11-13
fix(providers): set 42 default scope

estarossa0:
## Reasoning 💡
Specifying the scope for the 42 provider as `public`, since it get defaulted to `openid` which is a invalid scope for 42 api.
Default has been set to `public` not an empty string which is mandatory for all 42 api apps.

## Checklist 🧢

~- [ ] Documentation~
~- [ ] Tests~
- [x] Ready to be merged

## Affected issues 🎟
NONE

### DIFF
--- a/src/providers/42.js
+++ b/src/providers/42.js
@@ -4,7 +4,10 @@ export default function FortyTwo(options) {
     id: "42-school",
     name: "42 School",
     type: "oauth",
-    authorization: "https://api.intra.42.fr/oauth/authorize",
+    authorization: {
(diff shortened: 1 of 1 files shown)


alaadotsol (approved):

balazsorban44:
Thank you! So you can confirm that this provider works in v4?

estarossa0:
Signing proof:
- `useSession` output:
image
- `signIn` callback output:
image

balazsorban44:
Thank you so much! 🙏

balazsorban44 (approved):

PR #1872, opened 2021-04-28
feat(provider): add 42 School provider

mohouyizme:
## Reasoning 💡

Add 42 School provider.

### What is 42?

42 is an IT school with a network of more than 29 campuses around the globe, sharing the same learning platform and unique pedagogy with more than 8,000 students, for more information: https://www.codam.nl/en/the-42-network, https://www.42.fr/en/

### Why this provider?

42 School provider will make it easy for students to build useful Next.js apps for their campuses and the network, I already used it as a custom provider and I would like to share it with the community.

## Checklist 🧢

- [x] Documentation
- [ ] Tests N/A (manually)
- [x] Ready to be merged

PR #3211, opened 2021-11-16
refactor(providers): convert 42 to TypeScript

estarossa0:
### Move 42 provider to typescript
* Add default OAuth types
* Profile type
### Changing 42 provider file name to 42-school
In https://github.com/nextauthjs/next-auth/pull/1872 @balazsorban44 asked to change id of provider and file name(https://github.com/nextauthjs/next-auth/pull/1872#pullrequestreview-646614765), but only id was changed, and that created a conflict between `AUTOGENERATED` types and id,
image
and when trying to use  "[#Starts OAuth sign-in flow when clicked](https://next-auth.js.org/getting-started/client#starts-oauth-sign-in-flow-when-clicked)" `signIn` function suggest `42` in parameter but only work with `42-school`. 

## Checklist 🧢

~- [ ] Documentation~
~- [ ] Tests~
- [x] Ready to be merged


src/providers/42-school.ts
```diff
@@ -1,5 +1,15 @@
-/** @type {import(".").OAuthProvider} */
-export default function FortyTwo(options) {
+import { OAuthConfig, OAuthUserConfig } from "./oauth"
+
+export interface FortyTwoProfile {
+  id: string
+  fullname: string
+  email: string
+  image: string
```
balazsorban44, line 7: Do you know about any extra fields here?

balazsorban44 (changes requested):
Found a potential improvement for the `FortyTwoProfile` interface

src/providers/42-school.ts
estarossa0:
```js
profile: {
    id: 62377,
    email: 'arraji@student.1337.ma',
    login: 'arraji',
    first_name: 'Abdelbari',
    last_name: 'Rraji',
    usual_full_name: 'Abdelbari Rraji',
    usual_first_name: null,
    url: 'https://api.intra.42.fr/v2/users/arraji',
    phone: 'hidden',
    displayname: 'Abdelbari Rraji',
    image_url: 'https://cdn.intra.42.fr/users/arraji.jpg',
    'staff?': false,
    correction_point: 8,
    pool_month: 'august',
    pool_year: '2019',
    location: 'e1r5p15',
    wallet: 195,
    anonymize_date: '2022-11-16T00:00:00.000+01:00',
    created_at: '2019-08-05T10:11:36.151Z',
    updated_at: '2021-11-16T06:51:13.996Z',
    alumni: false,
    'is_launched?': false,
    groups: [],
    cursus_users: [ [Object], [Object] ],
    projects_users: [
      [Object], [Object], [Object], [Object], [Object],
      [Object], [Object], [Object], [Object], [Object],
      [Object], [Object], [Object], [Object], [Object],
      [Object], [Object], [Object], [Object], [Object],
      [Object], [Object], [Object], [Object], [Object],
      [Object], [Object], [Object], [Object], [Object],
      [Object], [Object], [Object], [Object], [Object],
      [Object], [Object], [Object], [Object], [Object],
      [Object], [Object], [Object], [Object], [Object],
      [Object], [Object], [Object], [Object], [Object],
      [Object], [Object], [Object], [Object], [Object],
      [Object], [Object], [Object], [Object], [Object],
      [Object], [Object], [Object], [Object], [Object],
      [Object], [Object], [Object]
    ],
    languages_users: [ [Object], [Object] ],
    achievements: [
      [Object], [Object], [Object],
      [Object], [Object], [Object],
      [Object], [Object], [Object],
      [Object], [Object], [Object],
      [Object], [Object], [Object],
      [Object], [Object], [Object],
      [Object], [Object], [Object],
      [Object]
    ],
    titles: [],
    titles_users: [],
    partnerships: [],
    patroned: [],
    patroning: [],
    expertises_users: [ [Object], [Object], [Object], [Object], [Object], [Object] ],
    roles: [],
    campus: [ [Object] ],
    campus_users: [ [Object] ],
    user: null
  }
```
I can try to create an interface from the response I get but I can't guarantee the exact value of some field (since it's not documented and I don't have the privilege to edit them to know what they return) so it will be set as `any`

balazsorban44: sounds good!

estarossa0:
I typed most of the response, except some properties that I don't have an idea of its value, I also tried to keep readable and easy to edit in case I typed something wrong or there was any change from the 42 api side.

(self note: I think I wrote enough types and interfaces for the rest of 2021 🤓)

balazsorban44 (approved):
Awesome, thank you! I'll merge this now, and clean up a bit to match other similar user interfaces, but all in all, this looks good.

balazsorban44:
I'm not sure why the filename has changed, I'll change it back to only 42 as well.

estarossa0:
> I'm not sure why the filename has changed, I'll change it back to only 42 as well.

As I mentioned in https://github.com/nextauthjs/next-auth/pull/3211#issue-1054665231 cause the file name is different from the provider `id` it caused an issue with the `OAuthProviderType` type generated in `next-auth > providers > oauth-types.d.ts` file. When using `signIn` function with a string argument typescript suggest `42` 
image
but it doesn't work unless the argument was the provider `id` which is `42-school`.
That's was why I changed the file name just to fix the generated type, and cause you asked for a file name change in https://github.com/nextauthjs/next-auth/pull/1872#pullrequestreview-646614765 first time provider was added. I didn't know that it'll cause a problem with something internal. Let me know what you think and the right way fix for it. Thank you 😄

balazsorban44:
So true! My bad, I've rushed this... :facepalm: I'll open a new PR to fix this.

PR #3221, opened 2021-11-16
refactor(providers): cleanup 42

balazsorban44:
Continuation of #3211

- renamed file back to `42`, so it doesn't change the submodule import at `next-auth/providers/42`
- using `Array` or `any[]` array types instead of tuples
- rename interfaces to be in the singular form
- export all interfaces


0ubbe (approved):
💅🏽 👍🏽

PR #3222, opened 2021-11-17
fix: assert `action` when `req.query` isn't available

balazsorban44:
`getServerSession` does not pass `req.query` to `NextAuthHandler`, and thus the newly introduced assertion that checks the existence of `req.query.nextauth`, but ignored the fact that methods relying directly on core might not have that passed down.

I also cleaned up `method` and used `GET` as a fallback value.


src/core/index.ts
```diff
@@ -13,7 +13,7 @@ import type { ErrorType } from "./pages/error"
 export interface IncomingRequest {
   /** @default "http://localhost:3000" */
   host: string
-  method: string
+  method?: string
```
lluia, line 16: I think we could type this more strictly?
```suggestion
  method?: 'GET' | 'POST' | 'PUT' | 'DELETE' | 'PATCH' | 'HEAD' | 'OPTIONS' | 'TRACE'
```

0ubbe (approved):
Nice 🙌🏽 , a small nitpick 😊 💬

src/core/index.ts
balazsorban44: I don't see an immediate value in it. The core handler only handles `"GET" | "POST"` anyway, otherwise, it will fall back to generate a 400 bad request response at the bottom.

I checked `Request["method"]` for reference, if that defines this type as `string`:

image

If we find a real use case/confusion, we can address this in a different PR.

PR #3268, opened 2021-11-23
pass session and user into signin callback for oauth provider

anguskeatinge:


anguskeatinge:
https://github.com/nextauthjs/next-auth/discussions/3264

balazsorban44:
This will need to go through a proper feature request issue first, explaining the use case and the problem what this would solve. From the link above, I am not sure how this will solve anything. Let's continue the discussion there.

PR #3332, opened 2021-12-02
fix: remove `nextauth` from authorization params

### LINKED ISSUE #3322: Extra query param nextauth
### Question 💬

Our custom provider is very strict what params can be passed and oauth flow fails because there is this NEXTAUTH query param. Is there anything I can do to not send it to the oauth server? I know it's required for next-auth to work correctly so i cant just omit it anywhere in the backend.

Exact error:
VARIABLES IN FORM NOT IN PROCEDURE: NEXTAUTH

### Config, [...nextauth].js
```
export default NextAuth({
  providers: [{
    type: 'oauth',
    id: 'oauth2',
    name: 'duoIdentifier',
    client: {
      client_id: '<client_id>',
      client_secret: '<client_secret>'
    },
    authorization: {
      url: '<auth_url>',
      params: {
        scope: 'duoIdentifier',
      }
    },
    accessTokenUrl: '<access_token_url>'
  }],
})

```

### The problem

This line cause an error on our end `&nextauth=signin%2Coauth2&` if I remove it. Everything works.

`http://*****/oauth2/auth?client_id=<client_id>&scope=duoIdentifier&response_type=code&redirect_uri=<redirect_uri>&nextauth=signin%2Coauth2&state=<state>`

This will work just fine

`http://*****/oauth2/auth?client_id=<client_id>&scope=duoIdentifier&response_type=code&redirect_uri=<redirect_uri>&state=<state>`

### How to reproduce ☕️

Custom provider so can't reproduce

### Contributing 🙌🏽

dnikomon:
## Reasoning 💡

Query param only needed internally and can cause errors if oauth provider only accepts certain query params 
## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## Affected issues 🎟
Fixes #3322

### DIFF
--- a/src/next/index.ts
+++ b/src/next/index.ts
@@ -18,17 +18,18 @@ async function NextAuthNextHandler(
   res: NextApiResponse,
   options: NextAuthOptions
 ) {
+  const { nextauth, ...query } = req.query
   const handler = await NextAuthHandler({
     req: {
       host: (process.env.NEXTAUTH_URL ?? process.env.VERCEL_URL) as string,
       body: req.body,
-      query: req.query,
+      query,
       cookies: req.cookies,
       headers: req.headers,
       method: req.method,
-      action: req.query.nextauth?.[0] as NextAuthAction,
-      providerId: req.query.nextauth?.[1],
-      error: (req.query.error as string | undefined) ?? req.query.nextauth?.[1],
+      action: nextauth?.[0] as NextAuthAction,
+      providerId: nextauth?.[1],
+      error: (req.query.error as string | undefined) ?? nextauth?.[1],
     },
     options,
   })


src/core/routes/signin.ts
```diff
@@ -7,7 +7,7 @@ import { Account, User } from "../.."
 /** Handle requests to /api/auth/signin */
 export default async function signin(params: {
   options: InternalOptions<"oauth" | "email">
-  query: IncomingRequest["query"]
+  query: IncomingRequest["query"] & { nextauth: string[] }
```
balazsorban44, line 10: I don't think this is necessary

balazsorban44 (changes requested):
Thank you for looking into this!

I think we could get rid of the `nextauth` query param, even before it is sent to the `NextAuthHandler` here: https://github.com/nextauthjs/next-auth/blob/1f4b7d8089937e356225dcab463d042e9780ec5a/src/next/index.ts#L21-L34

src/core/routes/signin.ts
dnikomon: Typescript complains for me but might be local development issue only. Would not be the first time.

balazsorban44: See my recommendation. If you do the necessary change much higher up, this file won't even have to change.

dnikomon: Fixed

dnikomon:
> Thank you for looking into this!
> 
> I think we could get rid of the `nextauth` query param, even before it is sent to the `NextAuthHandler` here:
> 
> https://github.com/nextauthjs/next-auth/blob/1f4b7d8089937e356225dcab463d042e9780ec5a/src/next/index.ts#L21-L34

Thank you ✋🏼

balazsorban44 (approved):

PR #3346, opened 2021-12-03
fix(ts): improve types for encode/decode functions

DmitryScaletta:
## Reasoning 💡

Improve types to be able to use synchronous functions that encodes/decodes JWT.

```ts
import NextAuth from 'next-auth';
import type { JWT } from 'next-auth/jwt';
import { sign, verify } from 'jsonwebtoken';

export default NextAuth({
  // ...
  jwt: {
    encode: ({ token, secret }) => verify(token, secret),
    decode: ({ token, secret }) => sign(token, secret) as JWT,
  },
});
```

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/src/jwt/types.ts
+++ b/src/jwt/types.ts
@@ -1,4 +1,4 @@
-import { decode, encode } from "."
+type MaybePromise<T> = T | Promise<T>
 
 export interface DefaultJWT extends Record<string, unknown> {
   name?: string | null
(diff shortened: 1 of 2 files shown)


src/jwt/types.ts
```diff
@@ -1,4 +1,4 @@
-import { decode, encode } from "."
+type MaybePromise<T> = T | Promise<T>
```
balazsorban44, line 1: We have a type for this already, called `Awaitable`.

DmitryScaletta: fixed

balazsorban44 (approved):

PR #3381, opened 2021-12-06
fix(providers): forward missing data to token request of Instagram

andersonba:
## Reasoning 💡

In the Instagram provider, the token request was missing `client_id` and `client_secret` in the body ([reference](https://developers.facebook.com/docs/instagram-basic-display-api/getting-started#step-5--exchange-the-code-for-a-token)), resulting in a bad request error as you see below:

image

image

I tried to fix using `params` object as suggested in the documentation, but the error persists.

```javascript
    token: {
      url: "https://api.instagram.com/oauth/access_token",
      params: {
        client_id: options.clientId,
        client_secret: options.clientSecret,
      },
    }
```

Is this request GET or POST? When I use fetch + POST without these two data, Instagram API returns an error informing the lack of these data instead of just Bad Request.

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/providers/instagram.js
+++ b/src/providers/instagram.js
@@ -32,7 +32,30 @@ export default function Instagram(options) {
     type: "oauth",
     authorization:
       "https://api.instagram.com/oauth/authorize?scope=user_profile",
-    token: "https://api.instagram.com/oauth/access_token",
+    token: {
+      url: "https://api.instagram.com/oauth/access_token",
+      async request(context) {
+        const {
+          provider,
+          params: { code },
+        } = context
+        const body = new URLSearchParams([
+          ["grant_type", "authorization_code"],
+          ["code", code],
+          ["client_id", provider.clientId],
+          ["client_secret", provider.clientSecret],
+          ["redirect_uri", provider.callbackUrl],
+        ])
+        const response = await (
+          await fetch(provider.token.url, {
+            method: "POST",
+            body,
+          })
+        ).json()
+        const { access_token } = response
+        return { tokens: { access_token } }
+      },
+    },
     userinfo:
       "https://graph.instagram.com/me?fields=id,username,account_type,name",
     async profile(profile) {


src/providers/instagram.js
```diff
@@ -32,7 +32,30 @@ export default function Instagram(options) {
...
+        const response = await (
+          await fetch(provider.token.url, {
+            method: "POST",
+            body,
+          })
+        ).json()
+        const { access_token } = response
+        return { tokens: { access_token } }
+      },
+    },
```
balazsorban44, line 58: This is most likely unnecessary/easier to fix through the `client` option. https://next-auth.js.org/configuration/providers/oauth#client-option

balazsorban44 (changes requested):
Thank you. Please have a look at my comment.

Would you mind converting this provider into TypeScript as well maybe? There are multiple examples here: https://github.com/nextauthjs/next-auth/tree/main/src/providers

stryt2:
Hi, I can confirm this problem still exists in NextAuth 4.3.2. I can also confirm that using a custom provider provided by @andersonba works as expected.

I am happy to submit in a PR to fix this but I did not understand @balazsorban44's comment about using the `client` option. 

@andersonba is suggesting that the POST request for "exchanging the code for token" is lacking `client_id` and `client_secret`. I'm unsure how the `client` option could be helpful in this case since the client would already have these information (confirmed after `console.log(context)` when provide custom `async request(context)` in the `token` option, see below for output).

 ```
 {
  provider: {
    id: 'my-instagram',
    name: 'Instagram',
    type: 'oauth',
    authorization: {
      url: 'https://api.instagram.com/oauth/authorize',
      params: [Object]
    },
    token: {
      url: 'https://api.instagram.com/oauth/access_token',
      request: [AsyncFunction: request]
    },
    userinfo: { url: 'https://graph.instagram.com/me', params: [Object] },
    profile: [AsyncFunction: profile],
    idToken: false,
    checks: [ 'state' ],
    clientId: 'XXXXXXXXXXXXX',
    clientSecret: 'XXXXXXXXXXXXXXXXXXXXXXXX',
    signinUrl: 'https://localhost:3000/api/auth/signin/my-instagram',
    callbackUrl: 'https://localhost:3000/api/auth/callback/my-instagram'
  },
  params: {
    code: 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX',
    state: 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'
  },
  checks: { state: 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX' },
  client: Client {
    authorization_signed_response_alg: 'RS256',
    client_id: 'XXXXXXXXXXXXX',
    client_secret: 'XXXXXXXXXXXXXXXXXXXXXXXX',
    grant_types: [
      'authorization_code'
    ],
    id_token_signed_response_alg: 'RS256',
    redirect_uris: [
      'https://localhost:3000/api/auth/callback/my-instagram'
    ],
    response_types: [
      'code'
    ],
    token_endpoint_auth_method: 'client_secret_basic'
  }
}
```

PR #3409, opened 2021-12-09
Fix Apple Provider first and last name

ryanmalesic:
Re-adding apple firstName and lastName. The name is not sent in the response but the fix was removed in a recent commit. This commit re-adds the logic to add the first and last name to the profile.

## Reasoning 💡

The name for a user is important.

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## Affected issues 🎟
Fixes the changes made in #2875

### DIFF
--- a/src/core/lib/oauth/callback.ts
+++ b/src/core/lib/oauth/callback.ts
@@ -137,6 +137,8 @@ export default async function oAuthCallback(params: {
       })
     }
 
+    profile.user = JSON.parse(body?.user ?? query?.user ?? null)
+
     const profileResult = await getProfile({
       profile,
       provider,
--- a/src/core/lib/oauth/callback.ts
+++ b/src/core/lib/oauth/callback.ts
@@ -137,6 +137,9 @@ export default async function oAuthCallback(params: {
       })
     }
 
+    // Contains the email, firstName, and lastName of Apple.
+    // Needed beacuse apple does not send the name in the token (profile callback).
+    // profile: {user: {email: 'johndoe@example.com', name: {firstName: 'John', lastName: 'Doe'}}}
     profile.user = JSON.parse(body?.user ?? query?.user ?? null)
 
     const profileResult = await getProfile({
--- a/src/providers/apple.ts
+++ b/src/providers/apple.ts
@@ -111,8 +111,9 @@ export default function Apple<P extends Record<string, any> = AppleProfile>(
     profile(profile) {
       return {
         id: profile.sub,
-        name: profile.name,
-        email: profile.email,
+        firstName: profile.user?.name?.firstName ?? null,
+        lastName: profile.user?.name?.lastName ?? null,
+        email: profile?.email ?? null,
         image: null,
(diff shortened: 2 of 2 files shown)


src/core/lib/oauth/callback.ts
```diff
@@ -137,6 +137,11 @@ export default async function oAuthCallback(params: {
       })
     }
 
+    // Contains the email, firstName, and lastName of Apple.
+    // Needed beacuse apple does not send the name in the token (profile callback).
+    // profile: {user: {email: 'johndoe@example.com', name: {firstName: 'John', lastName: 'Doe'}}}
+    profile.user = JSON.parse(body?.user ?? query?.user ?? null)
```
balazsorban44, line 143: I really object this change, especially if it's Apple only.

We tested this provider a few days ago, and this wasn't necessary.

balazsorban44 (commented):
Could you please show me the input argument of your profile callback both with and without your change in the core?

I really want to avoid putting provider specific code into the core.

src/core/lib/oauth/callback.ts
ryanmalesic: {
  iss: 'https://appleid.apple.com',
  aud: 'io.ilyeo.Ilyeo.sid',
  exp: 1639202356,
  iat: 1639115956,
  sub: '001195.bb514806f96e4d848b097a33c7f57085.0504',
  at_hash: '2soySvxdgQRZauIl1eZa9w',
  email: 'ryanmalesic@icloud.com',
  email_verified: 'true',
  auth_time: 1639115954,
  nonce_supported: true
}

This is the contents of profile without the changes. Name is not included. Name also isn't stored in the database without the changes I made.

Documentation isn't clear but 
> The modified name is only shared with your app and not with Apple, and hence isn’t included in the ID token.
https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple#3383768

FernandVEYRIER:
Hello!

I would like to open this thread again because indeed the name for Apple provider does not work. The information is only contained in the body, Apple does not provide it elsewhere.

Inside the body:
```
{
  "code": "someValue",
  "user": "{\"name\":{\"firstName\":\"Fernand\",\"lastName\":\"VEYRIER\"},\"email\":\"myEmail@email.com\"}"
}
```

But later on when `tokens.claims()` is called, the value of the profile is the following:
```
{
  "iss": "https://appleid.apple.com",
  "aud": "...",
  "exp": 1646363341,
  "iat": 1646276941,
  "sub": "...",
  "at_hash": "vhu7PI4b6vTNlJvGgTE-PQ",
  "email": "myEmail@email.com",
  "email_verified": "true",
  "auth_time": 1646276829,
  "nonce_supported": true
}
```

The user data is totally omitted and cannot be retrieved anymore. Would love to have an update on this, thanks!

rogaha:
+1

PR #3413, opened 2021-12-09
[discussion] Add support for anonymous sessions

### LINKED ISSUE #568: Anonymous sessions
How can anonymous sessions be used? It is necessary that before the user is authorized, it is possible to collect the data associated with him. For example, the items in the cart. Without sessions, I will not be able to add products on the server side, first I will need to store them on the client side and send them to the server side only after authorization.

mbarton:
## Reasoning 💡

Some sites, especially e-commerce, use anonymous tokens to support actions for users that are not logged in. These are then exchanged when the user logs in. See #568 for discussion.

At the moment, NextAuth has no notion of an anonymous session. As part of a poc for adopting the library, I've had a look at the smallest possible change to add it. Whilst we can obtain these tokens outside of the library, it simplifies the design of the poc quite a bit.

I have modified the `session` API route to invoke the `jwt` callback if no session cookie is already present. In this case, `token` will be empty and `user`, `profile` etc undefined. The callback can then modify the token, in our case adding an anonymous access token.

I've added an `anonymousSession` setting under `jwt` to opt in, as this changes behaviour.

This PR is not ready to merge but I'd like to raise it early and start a discussion about whether this feature is right for NextAuth generally and if this is the right way to implement it. I'm more than happy to take on any additional work to refine the implementation :)

I've already got a few questions:

- Is it OK to generate a token on the `GET session` API route?
  - I'm slightly worried about CSRF implications with the current design funnelling it through the existing JWT callback
  - In our case the worst an attacker could do is generate a new anonymous token but that's specific to our callback
  - Perhaps there should be an additional `anonSession` API route that the client calls, passing a CSRF token?
- Should the anon token callback be separate for clarity?
  - We could potentially avoid the setting entirely and only set the anonymous session if the callback is defined

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## Affected issues 🎟

Fixes #568

### DIFF
--- a/src/core/routes/session.ts
+++ b/src/core/routes/session.ts
@@ -37,14 +37,21 @@ export default async function session(
 
   const sessionToken = sessionStore.value
 
-  if (!sessionToken) return response
-
   if (sessionStrategy === "jwt") {
     try {
-      const decodedToken = await jwt.decode({
-        ...jwt,
-        token: sessionToken,
-      })
+      let decodedToken;
+
+      if(!jwt.anonymousSessions && !sessionToken) {
+        return response
+      } else if(jwt.anonymousSessions && !sessionToken) {
+        // TODO MRB: should we verify a CSRF token here or maybe a separate callback?
+        decodedToken = {}
+      } else {
+        decodedToken = await jwt.decode({
+          ...jwt,
+          token: sessionToken,
+        })
+      }
 
       const newExpires = fromDate(sessionMaxAge)
 
@@ -59,7 +66,6 @@ export default async function session(
         expires: newExpires.toISOString(),
       }
 
-      // @ts-expect-error
       const token = await callbacks.jwt({ token: decodedToken })
       // @ts-expect-error
       const newSession = await callbacks.session({ session, token })
@@ -89,6 +95,8 @@ export default async function session(
       response.cookies?.push(...sessionStore.clean())
     }
   } else {
+    if (!sessionToken) return response
+
     try {
       const { getSessionAndUser, deleteSession, updateSession } =
         adapter as Adapter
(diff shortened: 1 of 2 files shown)


balazsorban44:
Please open a Feature Request first.

https://github.com/nextauthjs/next-auth/issues/new?assignees=&labels=enhancement&template=feature_request.yaml

balazsorban44:
I believe you could also easily mock a fake session with the [Credentials Provider](https://next-auth.js.org/providers/credentials)

mbarton:
Thanks @balazsorban44, apologies for the noise. I've switched to using a dummy credentials provider as you suggested and it's working great

balazsorban44:
We could add an example to our tutorials docs!

mbarton:
I've opened a docs PR here: https://github.com/nextauthjs/docs/pull/177

dmusicb:
@balazsorban44 If I've already added authentication using providers w/ a database backend, is it not possible to also provide this as an option? I have users that want to try out the application w/ a guest account before linking their emails and real information through one of the providers. At the same time I don't necessarily want to give up on storing the sessions in a database for users that did sign in through a provider.

I'm thinking is it possible to:

1. When new users come to the site, they see the login page and can choose from the OAuth providers or the Anonymous Login.
2. If they choose the Anonymous login the request that is sent contains a custom header that says this is a user that is trying to create a guest account. Otherwise if they choose OAuth it's missing this header.
3. In the auth backend initialize two NextAuth classes, one for JWT session w/ just the anonymous CredentialsProvider and another with the backend session that contains the OAuth Providers.
4. When the request to sign in comes, we can use the appropriate NextAuth class. And when it's through the JWT one, we will continue to have something in the cookies that tells us this is the guest user so on future requests we always know to use the JWT NextAuth.

Would something like this be possible? I'm having trouble trying to get it set up as I can't seem to pull anything off the request that tells me which provider they are trying to log in to to switch between the different NextAuth classes when using the signIn(provider) call on the client.

PR #3454, opened 2021-12-16
add NEXT_PUBLIC_ variable support

RobbyUitbeijerse:
Hey there!

Our team was opting to move away from the 'old' way of handling variables in next.js (away from the [runtime configuration](https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration) options and the env object in next.config.js.

The issue we were facing with next-auth is that it didn't seem to support the NEXT_PUBLIC_ variables, while I checked the source I also noticed that it wasn't using the NEXT_PUBLIC_VERCEL_URL variable compared to the VERCEL_URL. What I did was simply create a new variable and assign the urls by first checking for public variables before falling back on the process variables.

We found this issue: https://github.com/nextauthjs/next-auth/issues/1713 

Let us know what you think


balazsorban44:
Hi, thanks for the PR, but I think this is slightly more complicated. First of all, if we introduce a new env variable, it might be a bit confusing.

I also think that we should try to eliminate the need for it in the first place, where possible. From client-side code (where NEXT_PUBLIC is relevant), we don't actually need the full URL, relative URL `fetch` calls should just work, so in theory, there is no need for that variable.

That would mean if you use a different path than `/api/auth`, you might have to duplicate it in `NEXTAUTH_URL` and the `basePath` option of `SessionProvider`. :thinking:

On the other hand, I'm not even sure if there is any point in supporting other paths than `/api/auth`. I have a feeling nobody/very few actually overrides that.

I'm just not sure what the correct/least painful solution is.

PR #3474, opened 2021-12-19
feat(oauth): Improve support for non-web clients such as react native

stevesouth:
## Reasoning 💡

I have been trying to understand how to extend next auth so that it can be used for react native apps. I think the main change required is to ensure key oauth endpoints optionally support not redirecting and instead return the key information back to the client.

With this change, posting redirect:false to the signon endpoint will now return the redirect url, instead of redirecting to it. This will allow non-web apps, such as those built with react native, to access the codeChallenge, redirectUrl and state, and handle the provider's sign on flow. The app can then call the redirectUrl itself once the provider flow is complete, including the code returned from the provider and the state.

Feedback welcome. I will try and put together an example repo with next auth and react native expo (which also needs a minor change to accept a custom codeChallenge https://github.com/expo/expo/pull/15535). 

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -176,7 +176,16 @@ export async function NextAuthHandler<
             options,
           })
           if (signin.cookies) cookies.push(...signin.cookies)
-          return { ...signin, cookies }
+
+          if (req.body?.redirect === false) {
+            return {
+              headers: [{ key: "Content-Type", value: "application/json" }],
+              body: { redirect: signin.redirect } as any,
+              cookies,
+            }
+          } else { 
+            return { ...signin, cookies }
+          }
         }
 
         return { redirect: `${options.url}/signin?csrf=true`, cookies }


balazsorban44:
I appreciate this PR, and I understand the confusion, since expanding to other frameworks is currently just a plan (but we intend to work on it the coming year #2294)

Although, I'm not sure why this PR is needed :thinking:

The whole point of separating `next-auth/core` from `next-auth/next` was to allow adding a wrapper of your liking.

Remember, you don't have to import from `next-auth` or `next-auth/next`, you can import the core directly and create your own wrapper. (`next-auth/core`).

Here is the one used for Next.js:
https://github.com/nextauthjs/next-auth/blob/main/src/next/index.ts

Am I missing something?

stevesouth:
There is a good chance i'm the one missing something, just trying to get my head around the code here, and in expo, and the oath2 spec. 

I think the fact that 'core' does http redirects adds to the confusion a little. Perhaps if core was more 'headless' and the redirect code was elsewhere it would help, i'm not sure.

In any case, I think i've hit an oauth hurdle at the last step. I've put an example repo here and written up what I was trying and where I got to. Even if this doesn't end up working it might save others some time!

https://github.com/stevesouth/next-auth-react-native-expo

stevesouth:
Ah, now I also see json: true is already an option which seems to do exactly the same thing. I think its just missing from the docs?

balazsorban44:
Core does not do redirects. This file does https://github.com/nextauthjs/next-auth/blob/main/src/next/index.ts

stevesouth:
I opened it as draft for this kind of feedback. To be clear:

1) It turned out the change I made already existed. I just didn't spot the json: true option as its not documented.

2) #2294 is useful to capture the general change to support different frameworks. However I think it would be good to have a specific thread to capture support for react native as its non trivial, especially as the implicit flow is now deprecated.

In the meantime I'll keep documenting any successes or failures here. https://github.com/stevesouth/next-auth-react-native-expo

Klin-Coders:
any update?

SushritPasupuleti:
@stevesouth have you made any progress with attempt 3?

PR #3509, opened 2021-12-23
update strava auth and token urls

krautwigundrueben:
there must be another issue still as it's still throwing a callback error, please support

built-in strava oauth throws callback error, they changed oauth and token urls

## Reasoning 💡

changed oauth and token urls according to https://developers.strava.com/docs/authentication/#requestingaccess

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

no, there's still a bug I haven't had time to dig into

### DIFF
--- a/src/providers/strava.js
+++ b/src/providers/strava.js
@@ -4,8 +4,8 @@ export default function Strava(options) {
     id: "strava",
     name: "Strava",
     type: "oauth",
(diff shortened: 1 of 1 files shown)


krautwigundrueben:
seems I got something wrong... still doesn't work like other providers though...

PR #3502, opened 2021-12-23
fix(providers): use `idToken` by default in Slack provider

joematune:
## Reasoning 💡

Add `idToken: true` by default to Slack provider config to solve oauth callback error.

<details open>
  <summary>Next-Auth console error</summary>

```
[next-auth][error][OAUTH_CALLBACK_ERROR] 
https://next-auth.js.org/errors#oauth_callback_error id_token detected in the response, you must use client.callback() instead of client.oauthCallback() {
  error: {
    message: 'id_token detected in the response, you must use client.callback() instead of client.oauthCallback()',
    stack: 'RPError: id_token detected in the response, you must use client.callback() instead of client.oauthCallback()\n' +
      '   ...,
    name: 'RPError'
  },
  providerId: 'slack',
  message: 'id_token detected in the response, you must use client.callback() instead of client.oauthCallback()'
}
```
</details>

## Checklist 🧢

- [x] Documentation
- [x] Tests

## Affected issues 🎟

Fixes the same problem as here: #3383 but for Slack provider.

PR #3514, opened 2021-12-25
feat: better out-of-the-box `id_token` detection

balazsorban44:
Many of our providers can (and do) return `id_token`s and we should make a better job at detecting it.

Good indicators are:

1. well-known endpoint with `openid` in the url
2. `openid` scope request

In these cases, if the user did not define a top-level `idToken` property, we should check these cases.

Fixes #3502

### DIFF
--- a/src/core/lib/providers.ts
+++ b/src/core/lib/providers.ts
@@ -45,17 +45,17 @@ function normalizeProvider(provider?: Provider) {
       typeof value === "string"
     ) {
       const url = new URL(value)
-      ;(acc as any)[key] = {
+      acc[key] = {
         url: `${url.origin}${url.pathname}`,
         params: Object.fromEntries(url.searchParams ?? []),
       }
     } else {
-      acc[key as keyof InternalProvider] = value
+      acc[key] = value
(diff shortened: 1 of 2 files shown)


src/core/lib/providers.ts
```diff
@@ -56,13 +57,18 @@ function normalizeProvider(provider?: Provider) {
...
-    !provider.checks
-  ) {
-    ;(normalizedProvider as InternalProvider<"oauth">).checks = ["state"]
+  if (provider.type === "oauth" && !provider.version?.startsWith("1.")) {
+    // If provider has as an "openid-configuration" well-known endpoint
+    // or an "openid" scope request, it will also likely be able to receive an `id_token`
+    ;(normalized as any).idToken = Boolean(
+      (normalized as any).idToken ??
+        (normalized as any).wellKnown?.includes("openid-configuration") ??
+        (normalized as any).token?.params?.scope?.includes("openid")
```
panva, line 66:
```suggestion
        (normalized as any).authorization?.params?.scope?.includes("openid")
```

PR #3537, opened 2021-12-29
Setting the http options befor discover the .wellKnown endpoint 

yfridelance:
Set custom.setHttpOptionsDefaults before Issuer.discover(.wellKnown). This allow discover the .wellKnown endpoint behind a proxy

Proxy settings was not setted on discovering the .wellKnown Endpoint

The http settings are set by the with this change before the discover

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

Yes

### DIFF
--- a/src/core/lib/oauth/client.ts
+++ b/src/core/lib/oauth/client.ts
@@ -12,7 +12,9 @@ export async function openidClient(
   options: InternalOptions<"oauth">
 ): Promise<Client> {
   const provider = options.provider
-
+  
+  if (provider.httpOptions) custom.setHttpOptionsDefaults(provider.httpOptions)
+  
   let issuer: Issuer
   if (provider.wellKnown) {
     issuer = await Issuer.discover(provider.wellKnown)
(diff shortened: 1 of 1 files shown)


src/core/lib/oauth/client.ts
```diff
@@ -44,7 +46,7 @@ export async function openidClient(
   // and https://github.com/nextauthjs/next-auth/issues/3067
   client[custom.clock_tolerance] = 10
 
-  if (provider.httpOptions) custom.setHttpOptionsDefaults(provider.httpOptions)
+  //if (provider.httpOptions) custom.setHttpOptionsDefaults(provider.httpOptions)
```
balazsorban44, line 49: could we just remove this line?

balazsorban44 (changes requested):
Thank you.

Have you tested if this actually makes a difference? Any idea how I could reproduce it? 🙏

balazsorban44 (approved):

PR #3538, opened 2021-12-29
unused code, maybe a possible fix.

7Hero:
## Reasoning 💡
On production this function doesn't return the token when used inside _middleware, no problem with the code above or below, so this could be the only problem.

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged


balazsorban44:
Please open a bug report before a PR. That code is necessary to handle cookie chunking, a new feature introduced in v4.

Issue #2156, opened 2021-06-09
3.26.0 ERROR post install npx husky install


RosayGaspard:
## Description 🐜

It seems the last build isn't working. I've spent an half hour trying to undersand why Vercel was suddenly unable to deploy my project... until I saw an update of next-auth had been released.

## How to reproduce ☕️

Run npm i next-auth 

## Screenshots / Logs 📽

```
❯ npm i next-auth

> next-auth@3.26.0 postinstall /Users/rosaygaspard/Documents/dev/Web/ReactJS/nextjs/ac-eup/node_modules/next-auth
> npx husky install

npx: installed 1 in 1.241s
.git can't be found (see https://typicode.github.io/husky/#/?id=custom-directory)
npm WARN ac-eup@1.0.0 No repository field.

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! next-auth@3.26.0 postinstall: `npx husky install`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the next-auth@3.26.0 postinstall script.
```
## Environment 🖥

System:
    OS: macOS 11.3.1
    CPU: (4) x64 Intel(R) Core(TM) i5-4278U CPU @ 2.60GHz
    Memory: 128.24 MB / 8.00 GB
    Shell: 5.8 - /usr/local/bin/zsh
  Binaries:
    Node: 14.11.0 - /usr/local/bin/node
    npm: 6.14.8 - /usr/local/bin/npm
  Browsers:
    Chrome: 91.0.4472.101
    Firefox: 88.0.1
    Safari: 14.1

Also happened while deploying to Vercel

ivanmaeder:
Seeing the same thing locally:

```
$ npm install --save next-auth

> npx husky install

npx: installed 1 in 5.989s
.git can't be found (see https://typicode.github.io/husky/#/?id=custom-directory)
...
...
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! next-auth@3.26.0 postinstall: `npx husky install`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the next-auth@3.26.0 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
...
```

Tried the previous version and it worked fine:

```
+ next-auth@3.25.0
added 80 packages from 111 contributors and audited 811 packages in 16.43s
found 0 vulnerabilities
```

balazsorban44:
[...]

PR #3548, opened 2022-01-01
fix: add eslintIgnore in package.json

ThangHuuVu:
## Reasoning 💡

The `lint` command is returning (38) errors: 
```
  0:0  error  Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: app/components/access-denied.js.
The file must be included in at least one of the projects provided
...
  0:0  error  Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: src/client/__tests__/client-provider.test.js.
The file must be included in at least one of the projects provided
...
```
I fix this by adding the `eslintIgnore` object in package.json with the value get from `tsconfig.json` `exclude`. 

## Checklist 🧢

- ~[ ] Documentation~
- ~[ ] Tests~
- [x] Ready to be merged

### DIFF
--- a/app/pages/api/auth/[...nextauth].ts
+++ b/app/pages/api/auth/[...nextauth].ts
@@ -1,5 +1,5 @@
 import NextAuth, { NextAuthOptions } from "next-auth"
-import EmailProvider from "next-auth/providers/email"
+// import EmailProvider from "next-auth/providers/email"
 import GitHubProvider from "next-auth/providers/github"
 import Auth0Provider from "next-auth/providers/auth0"
 import KeycloakProvider from "next-auth/providers/keycloak"
--- a/package.json
+++ b/package.json
@@ -175,12 +175,9 @@
     ]
   },
   "eslintIgnore": [
-    "./*.js",
     "./*.d.ts",
-    "app",
     "**/tests",
-    "**/__tests__",
-    "config"
+    "**/__tests__"
   ],
   "release": {
     "branches": [
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
   "eslintConfig": {
     "parser": "@typescript-eslint/parser",
     "parserOptions": {
-      "project": "./tsconfig.json"
+      "project": "./tsconfig.eslint.json"
     },
     "extends": ["standard-with-typescript", "prettier"],
     "ignorePatterns": [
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -174,6 +174,14 @@
       }
     ]
   },
+  "eslintIgnore": [
+    "./*.js",
+    "./*.d.ts",
+    "app",
+    "**/tests",
+    "**/__tests__",
+    "config"
```
balazsorban44, line 183:
```suggestion
    "./*.d.ts",
    "**/tests",
    "**/__tests__",
```

Ideally, I would like ESLint to report in most of these directories.

balazsorban44 (changes requested):

package.json
ThangHuuVu: Since we want to run the lint command on `app`, `src` and `config`, but at the same time want `tsc` to compile files in `src` only, I introduced a separate `tsconfig.eslint.json` file for `eslint`, extending from the base `tsconfig.json` file. Do you think this makes good sense?

balazsorban44: I'll have to check. I generally don't run the lint command, just let VSC warn me and I fix the warnings immediately.

ThangHuuVu: hmm, how do you think about adding lint-staged and husky to lint before committing?

balazsorban44: Considering the low amount of external PRs, I never really thought this would be necessary. We tried adding `husky` once, but it had issues and got reverted. (#2156, https://github.com/nextauthjs/next-auth/pull/2130#issuecomment-858145520)

Since it's mainly me who creates code, it seemed unnecessary until this point. If we can do it in a reliable way, I guess it would be alright.

ThangHuuVu: Got it. After checking, please let me know if there's anything that should be changed in this PR.

app/pages/api/auth/[...nextauth].ts
```diff
@@ -1,5 +1,5 @@
 import NextAuth, { NextAuthOptions } from "next-auth"
-import EmailProvider from "next-auth/providers/email"
+// import EmailProvider from "next-auth/providers/email"
```
Comment (balazsorban44) [inline app/pages/api/auth/[...nextauth].ts:2]: Could this be commented out by default? I often use the dev app to check something, without an adapter

balazsorban44 (approved):

PR #3581, opened 2022-01-06
Added patreon provider - tested and working

Tetedeiench:
## Reasoning 💡
As per https://github.com/nextauthjs/next-auth/discussions/3519 , the patreon provider was missing. I added one. Make sure you review thoroughly what I'm doing, despite the changes being simple. I'm updating the docs right after this PR, as they have moved to a separate repo.

## Checklist 🧢
Added Patreon provider

- [ x ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## Affected issues 🎟
https://github.com/nextauthjs/next-auth/discussions/3519


app/.env.local.example
```diff
@@ -1,45 +0,0 @@
-# Rename file to .env.local (or .env) and populate values
```
balazsorban44, line 1: Why was this file removed?

package-lock.json
```diff
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "next-auth",
```
balazsorban44, line 8: What happened here? Could we revert this change?

src/providers/patreon.js
```diff
@@ -0,0 +1,27 @@
...
+export default function Patreon(options) {
+    return {
+        id: 'patreon',
+        name: 'Patreon',
+        type: 'oauth',
+        version: '2.0',
+        authorization: { 
+          url : 'https://www.patreon.com/oauth2/authorize?response_type=code',
+          params: { scope: 'identity identity[email] identity.memberships', grant_type: 'authorization_code' }
+        },
```
balazsorban44, line 11:
```suggestion
        authorization: { 
          url : 'https://www.patreon.com/oauth2/authorize',
          params: { scope: 'identity identity[email] identity.memberships' }
        },
```

src/providers/patreon.js
```diff
@@ -0,0 +1,27 @@
+/** @type {import(".").OAuthProvider} */
+export default function Patreon(options) {
+    return {
+        id: 'patreon',
+        name: 'Patreon',
+        type: 'oauth',
+        version: '2.0',
+        authorization: { 
+          url : 'https://www.patreon.com/oauth2/authorize?response_type=code',
```
balazsorban44, line 9:
```suggestion
          url : 'https://www.patreon.com/oauth2/authorize',
```

I believe this should be unnecessary

src/providers/patreon.js
```diff
@@ -0,0 +1,27 @@
...
+          params: { scope: 'identity identity[email] identity.memberships', grant_type: 'authorization_code' }
```
balazsorban44, line 10:
```suggestion
          params: { scope: 'identity identity[email] identity.memberships' }
```

I believe `grant_type` is set by default.

Also, what's the reason for the `identity.memberships` scope?

Do we need this to get either of these `id`, `name`, `image`, `email`?

If not, it should be best if we set as few scopes as possible by default.

src/providers/patreon.js
```diff
@@ -0,0 +1,27 @@
...
+        },
+        token : {
+          url : 'https://www.patreon.com/api/oauth2/token',
+        },
+        userinfo : {
+          url: 'https://www.patreon.com/api/oauth2/api/current_user'
+        },
```
balazsorban44, line 17:
```suggestion
        token: "https://www.patreon.com/api/oauth2/token",
        userinfo: "https://www.patreon.com/api/oauth2/api/current_user",
```

Just some simplification.

src/providers/patreon.js
```diff
@@ -0,0 +1,27 @@
...
+        userinfo : {
+          url: 'https://www.patreon.com/api/oauth2/api/current_user'
+        },      
+        profile: (profile) => {
+          return {
+            id: profile.data.id,
+            name: profile.data.attributes.full_name,
+            email: profile.data.attributes.email
+          }
+        },
```
balazsorban44, line 24:
```suggestion
        profile(profile) {
          return {
            id: profile.data.id,
            name: profile.data.attributes.full_name,
            email: profile.data.attributes.email,
            image: null,
          }
        },
```

So we don't receive a profile image/picture? Could you check? If so, let's set `image: null`.

balazsorban44 (changes requested):
Thank you for this PR! I added some comments but generally looks good to me.

I was wondering if you were interested in converting this to TypeScript? It does not require much, see an example here: https://github.com/nextauthjs/next-auth/blob/main/src/providers/auth0.ts

Basically, an added interface to hint the user what they can receive through the `profile` callback.

Could you share a screenshot of sorts that shows a logged-in user as some kind of proof we can refer to later on? :pray:

Thank you once again for your contribution! :green_heart:

app/.env.local.example
Tetedeiench: Me being a n00b. I just renamed it, and didn't copy it.

package-lock.json
Tetedeiench: We can, I just ran npm install, sorry.

src/providers/patreon.js
Tetedeiench: Indeed, I'll remove the unnecessary scope

Tetedeiench:
Hi,

Thanks for your review ! Sorry for having so many mistakes.

I switched the provider to Typescript. I did my best to restore the files to their original states when necessary.

As for the proof, here's a gif of the full workflow running locally (with the address to show the URLs) using the latest version, switched to typescript : 
patreon login proof

Let me know if there's still things that need some work ! And again, sorry for the poor quality.

balazsorban44 (approved):

balazsorban44:
A stellar contribution, thank you! :green_heart:

PR #3601, opened 2022-01-10
feat(react): getSession

### LINKED ISSUE #596: getSession call does not trigger session object update until window loses focus
I'm currently trying to update some page components when the session changes after an `onClick`-triggered `Promise` is fulfilled. Here the function `DeleteGroup` returns a `Promise`. I print the value of the session before the `Promise` is returned, and then print it again after it is fulfilled by calling `getSession()`.

```jsx
<Button
  onClick={() => {
    console.log(session);
    DeleteGroup(value.id).then(async () => console.log(await getSession()));
  }}
>
```

According to the console, everything I've logged looks right. The initial `console.log(session)` output shows `groups` as an `Array` of length 2, while the second `console.log(await getSession())` output shows it as length 1.

Screen Shot 2020-08-20 at 6 02 35 PM

However, until the window loses focus, my UI continues to show 2 groups despite the console output clearly showing the result of the `getSession` call having only 1 group. My UI is immediately synced once the window loses focus, so I do not believe there is anything wrong with the way I've written my UI components. If that is the case, why wouldn't my `getSession()` call immediately update the UI accordingly?

Apologies if this is a repeat; I find session updating to be the trickiest part of Next-Auth.

### LINKED ISSUE #2208: useSession/Provider does not update when a new session is created until page reload.
## Description 🐜

When a page is not reloaded after login, the useSession/Provider is never updated, even after calling getSession.

## How to reproduce ☕️

1. Here is a [reproduction ](https://github.com/PaulPCIO/next-auth-session) of the issue.
2. I have set up a modal based login using a security code, for this example, I have [hardcoded ](https://github.com/PaulPCIO/next-auth-session/blob/434b417d24e3ae1d4b1190db2a97fb6da169c296/pages/api/auth/%5B...nextauth%5D.js#L14) the security code to `TEST`
3. When you log in successfully, the [callpack API](https://github.com/PaulPCIO/next-auth-session/blob/434b417d24e3ae1d4b1190db2a97fb6da169c296/components/login-view.js#L95-L97) triggers an additional [getSession](https://github.com/PaulPCIO/next-auth-session/blob/434b417d24e3ae1d4b1190db2a97fb6da169c296/components/login-view.js#L99).
4. At this stage the cookies are created and the session should be there, but the provider does not update until you refresh the page.

## Screenshots / Logs 📽

example

## Environment 🖥

```
  System:
    OS: Linux 5.4 Ubuntu 20.04.2 LTS (Focal Fossa)
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 11.23 GB / 24.97 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Yarn: 1.22.10 - ~/.nvm/versions/node/v14.16.1/bin/yarn
    npm: 6.14.12 - ~/.nvm/versions/node/v14.16.1/bin/npm
```

inshatan:
## Reasoning 💡

The changes cover following issues:
- https://github.com/nextauthjs/next-auth/issues/596
- https://github.com/nextauthjs/next-auth/issues/2208

When retrieving session on the client with the method `getSession` it is reasonable to update the data stored in the context

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## Affected issues 🎟

Fixes #596
Fixes #2208

### DIFF
--- a/src/client/_utils.ts
+++ b/src/client/_utils.ts
@@ -15,6 +15,8 @@ export interface NextAuthClientConfig {
    * trigger session updates from places like `signIn` or `signOut`
    */
   _getSession: (...args: any[]) => any
+  /** Updates session data */
+  _setSession: (session: Session | null | undefined) => void
 }
 
 export interface CtxOrReq {
--- a/src/react/index.tsx
+++ b/src/react/index.tsx
@@ -60,6 +60,7 @@ const __NEXTAUTH: NextAuthClientConfig = {
   _lastSync: 0,
   _session: undefined,
   _getSession: () => {},
+  _setSession: () => {},
 }
 
 const broadcast = BroadcastChannel()
@@ -121,7 +122,7 @@ export type GetSessionParams = CtxOrReq & {
   broadcast?: boolean
 }
 
-export async function getSession(params?: GetSessionParams) {
+async function fetchSession(params?: GetSessionParams) {
   const session = await fetchData<Session>(
     "session",
     __NEXTAUTH,
@@ -131,6 +132,12 @@ export async function getSession(params?: GetSessionParams) {
   if (params?.broadcast ?? true) {
     broadcast.post({ event: "session", data: { trigger: "getSession" } })
   }
+  return session
+}
+
+export async function getSession(params?: GetSessionParams) {
+  const session = await fetchSession(params)
+  __NEXTAUTH._setSession(session);
   return session
 }
 
@@ -323,11 +330,10 @@ export function SessionProvider(props: SessionProviderProps) {
         // We should always update if we don't have a client session yet
         // or if there are events from other tabs/windows
         if (storageEvent || __NEXTAUTH._session === undefined) {
-          __NEXTAUTH._lastSync = now()
-          __NEXTAUTH._session = await getSession({
+          const data = await fetchSession({
             broadcast: !storageEvent,
           })
-          setSession(__NEXTAUTH._session)
+          __NEXTAUTH._setSession(data)
           return
         }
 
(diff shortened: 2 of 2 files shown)


balazsorban44:
`getSession` is intended to be an isomorphic method that works both client and server-side, and likely won't have the capability to update the state, as it is often used outside of React lifecycle methods even on client-side. For that, please rely on `useSession`, and well, on the server-side this wouldn't make sense anyway.

inshatan:
> h client and server-sid

Thank you for the quick reply)

The proposed changes will remain `getSession` isomorphic.

About the issues - currently, you should reload the page to update the session. It is inappropriate in a lot of cases.

There is a workaround: https://github.com/nextauthjs/next-auth/issues/596#issuecomment-943453568. But this workaround will not work when a user is unauthenticated (`__NEXTAUTH.session = null`). 

Please provide some mechanism to trigger session reload/update programmatically. I can help you with implementation if it is necessary.

inshatan:
> `getSession` is intended to be an isomorphic method that works both client and server-side, and likely won't have the capability to update the state, as it is often used outside of React lifecycle methods even on client-side. For that, please rely on `useSession`, and well, on the server-side this wouldn't make sense anyway.

Thank you for the quick reply)

The proposed changes will remain `getSession` isomorphic.

About the issues - currently, you should reload the page to update the session. It is inappropriate in a lot of cases.

There is a workaround: #596 (comment). But this workaround will not work when a user is unauthenticated (`__NEXTAUTH.session = null`).

Please provide some mechanism to trigger session reload/update programmatically. I can help you with implementation if it is necessary.

PR #3648, opened 2022-01-16
feat(core): allow user updates

### LINKED ISSUE #2269: make updating session easier
### Discussed in https://github.com/nextauthjs/next-auth/discussions/2267

<div type='discussions-op-text'>

<sup>Originally posted by **phileasmah** June 28, 2021</sup>
## Summary 💭

Include a ```setState``` to mutate the ```session``` state when using the ```useSession()``` hook. 

## Description 📓
In my app, I'm trying to build a settings page for profiles, like changing the display name etc, but I found that after changing the display name (name in database has been changed), the ```session``` state in ```useSession()``` stayed the same.

So, I was trying to find a way to mutate the session object of the current user and stumbled upon ```getSession()``` which (correct me if I'm wrong) is suppose to get a new session and replace it with the current one but according to this [`issue`](https://github.com/nextauthjs/next-auth/issues/596) it doe not work as intended. 

I was looking at the source code and thought wouldn't it be easier to just allow us to mutate the session state directly? (or are there security concerns I'm not aware of? In which case, it would be interesting to know). Looking at the code, it should be pretty easy to implement.

An advantage when compared to ```getSession()``` would obviously be that 1 less call needs to be made to the database, but again, I'm not sure if I am glossing over any security concerns etc 
</div>

There are probably other related issues, which could be resolved with this. I'll try to gather them below, but if anyone reading this knows any, please comment below, and I would also like to hear your thoughts on this!

balazsorban44:
Fixes #2269

TODO

- Explain caveats (DB only, `user` must exist in `session`)
- Docs (new warnings/errors)
- Future improvements (update session?)

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -222,6 +222,19 @@ export async function NextAuthHandler<
           }
         }
         return {}
+      case "session":
+        // Verified CSRF Token required for session updates
+        if (options.csrfTokenVerified) {
+          const session = await routes.session({
+            method: "POST",
+            user: req.body?.user,
+            options,
+            sessionStore,
+          })
+          if (session.cookies) cookies.push(...session.cookies)
+          return { ...session, cookies } as any
+        }
+        return { cookies, status: 401 }
       default:
     }
   }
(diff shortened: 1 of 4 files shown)


shadiramadan:
@balazsorban44 Why was this abandoned?  Is there an alternative mechanism to tackle this?

PR #3656, opened 2022-01-17
Fix: (Azure AD) Check for valid profile picture response before converting to base64

davidchalifoux:
## Reasoning 💡
Currently, the `profile()` function for Azure AD takes a response from the Microsoft Graph API and converts it to base64 for profile pictures. Unfortunately, this has caused bugs for my project because it doesn't make sure that the response from Microsoft is valid before converting. Users without a photo set on their account will not recieve a valid response, but an invalid "image" will still be returned.

This PR adds a small check to the profile photo response to make sure the HTTP-status is 200-299. If the status is outside of that range, the profile image is left as an empty string.

## Checklist 🧢

- [ ] ~Documentation~
- [ ] ~Tests~
- [x] Ready to be merged

## Affected issues 🎟
None

### DIFF
--- a/src/providers/azure-ad.ts
+++ b/src/providers/azure-ad.ts
@@ -28,7 +28,7 @@ export default function AzureAD<P extends Record<string, any> = AzureADProfile>(
     wellKnown: `https://login.microsoftonline.com/${tenant}/v2.0/.well-known/openid-configuration`,
     authorization: {
       params: {
-        scope: "User.Read",
+        scope: "openid profile email",
       },
     },
     async profile(profile, tokens) {
--- a/src/providers/azure-ad.ts
+++ b/src/providers/azure-ad.ts
@@ -36,7 +36,7 @@ export default function AzureAD<P extends Record<string, any> = AzureADProfile>(
         id: profile.sub,
         name: profile.name,
         email: profile.email,
-        image: "",
+        image: null,
       };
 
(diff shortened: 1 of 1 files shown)


src/providers/azure-ad.ts
```diff
@@ -32,6 +32,13 @@ export default function AzureAD<P extends Record<string, any> = AzureADProfile>(
       },
     },
     async profile(profile, tokens) {
+      const profileObject = {
+        id: profile.sub,
+        name: profile.name,
+        email: profile.email,
+        image: "",
```
balazsorban44, line 39:
```suggestion
        image: null,
```

balazsorban44 (changes requested):
Thanks! Added a small suggestion.

src/providers/azure-ad.ts
davidchalifoux: Looks like that angered TS. Shouldn't `""` be fine since `!!""` returns false?

balazsorban44: It will potentially save it to the database as an empty string, also when used in the front-end, people will use it as `src={user.image}` which will break.

Not sure why `null` wouldn't be accepted by TS though. :thinking:

davidchalifoux: ea5647d Fixes it.
Unfortunately I'm not too familiar with TS so I just used separate return statements.

balazsorban44 (approved):

PR #3657, opened 2022-01-18
feat(middleware): introduce `withAuth` Next.js method

### LINKED ISSUE #3037: Support Next.js 12 Middleware

balazsorban44:
This is currently an idea :bulb: but it fixes #3037

To test it out, see https://github.com/nextauthjs/next-auth/pull/3657#issuecomment-1014990129

Useful resources:
- https://nextjs.org/docs/middleware#execution-order

### Authentication only

You allow any logged-in user to access a page

```js
// pages/_middleware.js

export { default } from "next-auth/middleware"
```

### Authorization

Certain conditions must be met for the user to continue, for example, to restrict an admin dashboard, you can do:

```js
// pages/admin/_middleware.js

export default withAuth({
  callbacks: {
    authorized: ({ token }) => token?.user.isAdmin
  }
})
```

See the source code for the proposed API.

Suggestion: We could detect if `NEXTAUTH_SECRET` is present, sparing the user to fill an extra configuration field. (NOTE: This could be done in `NextAuth` in `pages/api/auth/[...nextauth]` as well)

Open for feedback.

- [ ] Docs https://github.com/nextauthjs/docs/pull/218
- [ ] Update [Live Demo](https://next-auth-example.vercel.app/)

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,0 +1,2 @@
+export { default } from "./next/middleware"
+export * from "./next/middleware"
--- a/src/next/middleware.ts
+++ b/src/next/middleware.ts
@@ -1,0 +1,97 @@
+import type { NextRequest } from "next/server"
+import type { Awaitable, NextAuthOptions } from ".."
+import type { JWT } from "../jwt"
+
+import { NextResponse } from "next/server"
+import { getToken } from "../jwt"
+
+export interface NextAuthMiddlewareOptions {
+  /**
+   * The secret used to create the session.
+   * @note Must match as `secret` in `NextAuth`.
+   * @default process.env.NEXTAUTH_SECRET
+   *
+   * ---
+   * [Documentation](https://next-auth.js.org/configuration/options#secret)
+   */
+  secret?: string
+  /**
+   * Where to redirect the user in case of an error if they weren't logged in.
+   * Similar to `pages` in `NextAuth`.
+   *
+   * ---
+   * [Documentation](https://next-auth.js.org/configuration/pages)
+   */
+  pages?: NextAuthOptions["pages"]
+  /**
+   * Callback that receives the user's JWT payload
+   * and returns `true` to allow the user to continue.
+   *
+   * If it returns `false`, the user is redirected to the sign-in page instead
+   *
+   * The default is to let the user continue if they have a valid JWT (basic authentication).
+   *
+   * How to restrict a page and all of it's subpages for admins-only:
+   * @example
+   *
+   * ```js
+   * // `pages/admin/_middleware.js`
+   * import { withAuth } from "next-auth/next/middleware"
+   *
+   * export default withAuth({
+   *   authorized: ({ token }) => token?.user.isAdmin
+   * })
+   * ```
+   */
+  authorized: (options: {
+    token: JWT | null
+    req: NextRequest
+  }) => Awaitable<boolean>
+}
+
+/**
+ * Middleware that checks if the user is authenticated/authorized.
+ * If if they aren't, they will be redirected to the login page.
+ * Otherwise, continue.
+ *
+ * @example
+ *
+ * ```js
+ * // `pages/_middleware.js`
+ * export { withAuth as default } from "next-auth/next/middleware"
+ * ```
+ *
+ * ---
+ * [Documentation](https://next-auth.js.org/getting-started/middleware)
+ */
+export function withAuth(options?: NextAuthMiddlewareOptions) {
+  const secret = options?.secret ?? process.env.NEXTAUTH_SECRET
+  if (!secret) {
+    const code = "NO_SECRET"
+    console.error(
+      `[next-auth][error][${code}]`,
+      `\nhttps://next-auth.js.org/errors#${code.toLowerCase()}`
+    )
+    return NextResponse.redirect(
+      options?.pages?.error ?? "/api/auth/error?error=Configuration"
+    )
+  }
+
+  return async function middleware(req: NextRequest) {
+    const token = await getToken({
+      req: {
+        headers: req.headers as any,
+        cookies: req.cookies,
+      },
+      secret,
+    })
+
+    const isAuthorized = options?.authorized
+      ? options?.authorized?.({ token, req })
+      : !!token
+
+    if (isAuthorized) return
+
+    return NextResponse.redirect(options?.pages?.signIn ?? "/api/auth/signin")
+  }
+}
--- a/src/next/middleware.ts
+++ b/src/next/middleware.ts
@@ -1,4 +1,4 @@
-import type { NextRequest } from "next/server"
+import type { NextFetchEvent, NextRequest } from "next/server"
 import type { Awaitable, NextAuthOptions } from ".."
 import type { JWT } from "../jwt"
 
@@ -64,7 +64,35 @@ export interface NextAuthMiddlewareOptions {
  * ---
  * [Documentation](https://next-auth.js.org/getting-started/middleware)
  */
-export function withAuth(options?: NextAuthMiddlewareOptions) {
+export async function withAuth(
+  ...args: [NextAuthMiddlewareOptions] | [NextRequest, NextFetchEvent]
+) {
+  if (args.length === 2) {
+    const secret = process.env.NEXTAUTH_SECRET
+
+    if (!secret) {
+      const code = "NO_SECRET"
+      console.error(
+        `[next-auth][error][${code}]`,
+        `\nhttps://next-auth.js.org/errors#${code.toLowerCase()}`
+      )
+      return NextResponse.redirect("/api/auth/error?error=Configuration")
+    }
+
+    const req = args[0]
+    const token = await getToken({
+      req: {
+        headers: req.headers as any,
+        cookies: req.cookies,
+      },
+      secret,
+    })
+    if (token) return
+
+    return NextResponse.redirect("/api/auth/signin")
+  }
+
+  const options = args[0]
   const secret = options?.secret ?? process.env.NEXTAUTH_SECRET
   return async function middleware(req: NextRequest) {
     const pages = options?.pages ?? {}
--- a/src/next/middleware.ts
+++ b/src/next/middleware.ts
@@ -14,15 +14,6 @@ type AuthorizedCallback = (params: {
 
 export interface NextAuthMiddlewareOptions {
   /**
-   * The secret used to create the session.
-   * @note Must match as `secret` in `NextAuth`.
-   * @default process.env.NEXTAUTH_SECRET
-   *
-   * ---
-   * [Documentation](https://next-auth.js.org/configuration/options#secret)
-   */
-  secret?: NextAuthOptions["secret"]
-  /**
    * Where to redirect the user in case of an error if they weren't logged in.
    * Similar to `pages` in `NextAuth`.
    *
@@ -46,10 +37,12 @@ export interface NextAuthMiddlewareOptions {
      *
      * ```js
      * // `pages/admin/_middleware.js`
-     * import { withAuth } from "next-auth/next/middleware"
+     * import { withAuth } from "next-auth/middleware"
      *
      * export default withAuth({
-     *   authorized: ({ token }) => token?.user.isAdmin
+     *   callbacks: {
+     *     authorized: ({ token }) => token?.user.isAdmin
+     *   }
      * })
      * ```
      *
@@ -76,7 +69,7 @@ async function handleMiddleware(
     return
   }
 
-  const secret = options?.secret ?? process.env.NEXTAUTH_SECRET
+  const secret = process.env.NEXTAUTH_SECRET
   if (!secret) {
     console.error(
       `[next-auth][error][NO_SECRET]`,
--- a/src/next/middleware.ts
+++ b/src/next/middleware.ts
@@ -122,3 +122,5 @@ export async function withAuth(
     )
   }
 }
+
+export default withAuth
--- a/src/next/middleware.ts
+++ b/src/next/middleware.ts
@@ -136,7 +136,7 @@ export function withAuth(...args: WithAuthArgs) {
     const options = args[1] as NextAuthMiddlewareOptions | undefined
     return async (...args: Parameters<NextMiddleware>) =>
       await handleMiddleware(args[0], options, async (token) => {
-        ;(args[0] as any).token = token
+        ;(args[0] as any).nextauth = { token }
         return await middleware(...args)
       })
   }
--- a/src/next/middleware.ts
+++ b/src/next/middleware.ts
@@ -78,10 +78,12 @@ export async function withAuth(
       )
       return NextResponse.redirect("/api/auth/error?error=Configuration")
     }
+    const req = args[0]
+    if (await getToken({ req: req as any, secret })) return
 
-    if (await getToken({ req: args[0] as any, secret })) return
-
-    return NextResponse.redirect("/api/auth/signin")
+    return NextResponse.redirect(
+      `/api/auth/signin?${new URLSearchParams({ callbackUrl: req.url })}`
+    )
   }
 
   const options = args[0]
@@ -114,6 +116,9 @@ export async function withAuth(
 
     if (isAuthorized) return
 
-    return NextResponse.redirect(pages.signIn ?? "/api/auth/signin")
+    const redirectUrl = pages.signIn ?? "/api/auth/signin"
+    return NextResponse.redirect(
+      `${redirectUrl}?${new URLSearchParams({ callbackUrl: req.url })}`
+    )
   }
 }
--- a/src/next/middleware.ts
+++ b/src/next/middleware.ts
@@ -69,8 +69,7 @@ async function handleMiddleware(
     return
   }
 
-  const secret = process.env.NEXTAUTH_SECRET
-  if (!secret) {
+  if (!process.env.NEXTAUTH_SECRET) {
     console.error(
       `[next-auth][error][NO_SECRET]`,
       `\nhttps://next-auth.js.org/errors#no_secret`
@@ -81,11 +80,11 @@ async function handleMiddleware(
     }
   }
 
-  const token = await getToken({ req: req as any, secret })
+  const token = await getToken({ req: req as any })
 
-  const authorized: AuthorizedCallback =
-    options?.callbacks?.authorized ?? (({ token }) => !!token)
-  const isAuthorized = await authorized({ req, token })
+  const isAuthorized =
+    (await options?.callbacks?.authorized?.({ req, token })) ?? !!token
+
   // the user is authorized, let the middleware handle the rest
   if (isAuthorized) return await onSuccess?.(token)
 
--- a/src/next/middleware.ts
+++ b/src/next/middleware.ts
@@ -79,15 +79,7 @@ export async function withAuth(
       return NextResponse.redirect("/api/auth/error?error=Configuration")
     }
 
-    const req = args[0]
-    const token = await getToken({
-      req: {
-        headers: req.headers as any,
-        cookies: req.cookies,
-      },
-      secret,
-    })
-    if (token) return
+    if (await getToken({ req: args[0] as any, secret })) return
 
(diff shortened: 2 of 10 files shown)


pdevito3:
Love this idea and is what i'm used to doing on the backend side if I was making a BFF!

FWIW, something like the below is what I might use in a .NET app. The api is really easy and powerful. Might provide some inspiration here:

```c#
          services.AddAuthentication(options =>
          {
              options.DefaultScheme = "cookie";
              options.DefaultChallengeScheme = "oidc";
              options.DefaultSignOutScheme = "oidc";
          })
          .AddCookie("cookie", options =>
          {
              options.Cookie.Name = "__Host-bff";
              options.Cookie.SameSite = SameSiteMode.Strict;
          })
          .AddOpenIdConnect("oidc", options =>
          {
              options.Authority = "https://localhost:3385";
              options.ClientId = "react.bff";
              options.ClientSecret = "secret";
              options.ResponseType = "code";
              options.ResponseMode = "query";
              options.usePkce = true;

              options.GetClaimsFromUserInfoEndpoint = true;
              options.MapInboundClaims = false;
              options.SaveTokens = true;

              options.Scope.Clear();
              options.Scope.Add("openid");
              options.Scope.Add("profile");
              options.Scope.Add("role");
              options.Scope.Add("email");
              options.Scope.Add("recipe_management_access");
              options.Scope.Add("offline_access");

              options.TokenValidationParameters = new()
              {
                  NameClaimType = "name",
                  RoleClaimType = "role"
              };
          });
```

src/next/middleware.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (!secret) {
+      const code = "NO_SECRET"
+      console.error(
+        `[next-auth][error][${code}]`,
+        `\nhttps://next-auth.js.org/errors#${code.toLowerCase()}`
+      )
+      return NextResponse.redirect("/api/auth/error?error=Configuration")
+    }
+    const req = args[0]
+    if (await getToken({ req: req as any, secret })) return
```
Code-Hex, line 82: Is this OK because the type of parameter is `req: NextApiRequest | Pick<NextApiRequest, "cookies" | "headers">`?
```suggestion
    if (await getToken({ req, secret })) return
```

zpg6:
Thanks for addressing this feature! Some feedback:

- How would you combine this with other middleware in the same scope? My use case involves some unrelated items, like rate limiting, alongside the `withAuth` method for most of my API. 

- I'll often times be decoding the tokens again in the handler functions to get the userId as a part of the database query. Would it be more efficient (and would it be secure) to have the middleware store the token claims it already decoded in the request?

src/next/middleware.ts
balazsorban44: The issue is with `getToken` types under the hood. Since this is internal code, I did not bother fixing it for now.

balazsorban44:
@zpg6 I haven't seen much middleware chaining yet, but It could work similarly to https://next-auth.js.org/configuration/initialization#advanced-initialization I guess.

zpg6:
1. Thumbs up on your suggested wrapper, and good point... still to be seen how the nextjs community handles chaining more generally. Having a lightweight escape hatch like this until then would be plenty.  

2. Thanks for the insights on performance of decoding twice. Just one of those things that feels redundant, but I suppose storing in the request does add extra complexity without any real code savings.


Planning to test our your pr soon, will follow up with any DX feedback.

src/next/middleware.ts
nyedidikeke: @balazsorban44 Could this break on `tsc`?

always-maap:
great work.
in the case of external authorization, what is the best practice in next auth middleware? for example, google handle [IAM roles](https://cloud.google.com/iam/docs/understanding-roles) like so:
- accesscontextmanager.accessPolicies.create
- accesscontextmanager.accessPolicies.update
- ...

each user(admin) has different roles indicated like above. 
calling user role list in the token callback and attach to the token, then checking in _middleware? is it possible?

```javascript
// pages/accessPolicies/_middleware.js
import { withAuth } from "next-auth/next/middleware"

export default withAuth({
  authorized: ({ token }) => token?.roles.includes('accesscontextmanager.accessPolicies')
})
```

balazsorban44:
@always-maap `next-auth` has its own session, so it remains agnostic about the Provider you are using. In your case, as you say, using the [`jwt` callback](https://next-auth.js.org/configuration/callbacks#jwt-callback) could be used to save the user's roles at sign in, and then you can just check that as you showed in your code above.

Rees1993:
Is the NextRequest req variable available inside the authorized function? I have a scenario where specific paths are accessible when the user isn't logged in and I need to check the current pathname.

balazsorban44:
@Rees1993 literally the comment above yours :D

https://github.com/nextauthjs/next-auth/pull/3657#issuecomment-1015445553

There is also another possibility. You can include a middleware only for a subset of pages, if your URL structure allows it. (eg.: `pages/admin/_middleware.js` will only call the middleware for pages under /admin, like /admin/a /admin/b, /admin/a/b etc. A page like /home wouldn't even invoke the Middleware. )

Useful resource: https://nextjs.org/docs/middleware#execution-order

Rees1993:
@balazsorban44 Well I think I might need to go to the opticians!

Unfortunately these pages are all at the base level (pages/contact.js, pages/login.js etc.) so I don't think it would work. Thanks for that link though as I forgot about that functionality.

Otherwise, I'm all for this middleware and look forward to testing it!

balazsorban44:
Update, I did two changes:

1. export `withAuth` as default
2. change import of middleware from `next-auth/middleware`, instead of `next-auth/next/middleware`

So this will let you do:


```js
export { default } from "next-auth/middleware"
```

For an even shorter syntax. The reasoning is that we might not see Middleware support in other frameworks very soon, and if we do, we might release that under the `@next-auth` scope anyway instead. `next-auth` has been for Next.js, always, so it just makes sense to use it that way as well for now.

balazsorban44:
It looks like this has been very well received by the community ( :green_heart: ), so I'm going to wrap things up by adding the chaining support discussed at https://github.com/nextauthjs/next-auth/pull/3657#issuecomment-1015019410

I'll add documentation, and update our official example page, and this will be good to go!

If you want to test this out already, check out https://github.com/nextauthjs/next-auth/pull/3657#issuecomment-1014990129

which contains an updated experimental release available at npm.

NOTE: Since Next.js Middleware is in public beta still, we might have to accommodate for it if we release this to `latest` before Middleware goes stable.

To avoid confusion, I'll probably release this with a warning and reserve the right to change the API if we need to accommodate for anything. (Hopefully, we won't have to :crossed_fingers: )

erlinghaugstad:
This is perfect for our use cases; Flexible and powerful, good job 🥳

zpg6:
Great work 🚀

adambrgmn:
Great work! Really been looking forward to this!
I've been trying it out on a small personal project. But I'm having issues :/ 

I'm planning on restricting all pages to signed in users. So I'm putting the handler in `pages/_middleware.ts`.

The first option: 
```ts
export {default} from 'next-auth/middleware';
```
This gives me an infinite redirect loop.

The second:
```ts
import withAuth from 'next-auth/middleware';
export default withAuth({ authorized: ({ token }) => token != null });
```

Gives me an error saying that I'm not exporting a middleware (`Error: The Middleware "pages/_middleware" must export a `middleware` or a `default` function`). `export const middleware = withAuth(...)` doesn't work either.

Is this something you've encountered before?

I'm going to try it out a bit more during the afternoon and will get back to you if I solve it!

adambrgmn:
Okey, I came a bit further pretty fast and found an issue.
When I do the `export {default} from '...'` version the infinite redirect happens because the middleware (if placed in the root) because the middleware will run for _all_ pages and api routes. This might be good to emphasize in the documentation.

The reason I get the error that a middleware isn't exported properly is because `withAuth(options)` actually gives back a promise, since it is defined as `async function withAuth`. The version where we pass options, instead of export directly, must resolve sync.

src/next/middleware.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    return NextResponse.redirect(
+      `/api/auth/signin?${new URLSearchParams({ callbackUrl: req.url })}`
+    )
```
adambrgmn, line 86: Since the function can't use await in the upper scope this needs to be changed to something like this:

```suggestion
    return getToken({ req: req as any, secret }).then((token) => {
      if (token) return;
      return NextResponse.redirect(
        `/api/auth/signin?${new URLSearchParams({ callbackUrl: req.url })}`
      );
    });
```

balazsorban44: Very good catch! I'll address it together with letting the first argument be a middleware to support chaining.

src/next/middleware.ts
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * ```js
+ * // `pages/_middleware.js`
+ * export { withAuth as default } from "next-auth/next/middleware"
+ * ```
+ *
+ * ---
+ * [Documentation](https://next-auth.js.org/getting-started/middleware)
+ */
+export async function withAuth(
```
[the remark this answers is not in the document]
balazsorban44: Thanks again, I've fixed these! See the files for some usage examples.

balazsorban44:
I've cleaned up the code, and added some different use cases in this file:

https://github.com/nextauthjs/next-auth/blob/2976116e6b8d94cd1b3d336355548c9c11ed8115/app/pages/middleware-protected/_middleware.js

I updated the experimental release at https://github.com/nextauthjs/next-auth/pull/3657#issuecomment-1014990129, so please try them out, and tell me if I missed any cases! (I think I did not).

I tested all of them, so they should be good to go.

I'll try adding the documentation this weekend probably, and I intend to ship this to `latest` probably next week. :tada:

Cheers! :beers:

src/next/middleware.ts
```diff
@@ -0,0 +1,162 @@
...
+function initConfig(
+  req: NextRequest,
+  optionsOrEv?: NextAuthMiddlewareOptions | NextFetchEvent
+) {
+  const options = optionsOrEv as NextAuthMiddlewareOptions // This is relatively safe as we don't share properties with NextFetchEvent
+  const signInPage = options?.pages?.signIn ?? "/api/auth/signin"
+  const errorPage = options?.pages?.error ?? "/api/auth/error"
+
+  // Avoid infinite redirect loop
+  if ([signInPage, errorPage].includes(req.nextUrl.pathname)) return
```
adambrgmn, line 71: Maybe this goes outside the scope of this middleware since I can run my own. While this solves the infinite redirect issue it does still prevents sign in 😬 Since when the providers does their callback they try to go to `/api/auth/callback/{provider}` –  but that page is also covered by the middleware, if the middleware lives in `./pages/_middleware.js`, and is not among the ignored paths. So when the callback comes the user is not signed in and gets redirected back to sign in 😛

Maybe all `/api/auth` routes should be ignored by this middleware? `req.nextUrl.pathname.startsWith('/api/auth')`.

adambrgmn (commented):
Great work! Just one more edge case mentioned in a comment below :) 
I guess it is not really the most popular way to use the middleware, using it in on all pages. But nonetheless it might be worth taking into account :) do as you please!

src/next/middleware.ts
GeeWee: Do note that you _can_ specify custom `auth` paths (ours is at `api/next-auth/` for example), so you can't just hard-code it in.

adambrgmn: Ah that is true! I did not think about that. Maybe `signInPage` and `errorPage` should be consolidated into `authBase` with default `/api/auth`?

balazsorban44: We can use NEXTAUTH_URL here, no problem.

src/next/middleware.ts
```diff
@@ -0,0 +1,162 @@
...
+ * export { default } from "next-auth/middleware"
+ * ```
+ *
+ * ---
+ * [Documentation](https://next-auth.js.org/getting-started/middleware)
+ */
+export function withAuth(...args: WithAuthArgs) {
+  if (args[0] instanceof NextRequest) {
+    const config = initConfig(args[0], args[1])
+    if (!config || config.redirect) return config?.redirect
```
lluia, line 112: I see two flows here, either we want to **(1)** redirect because there was an error in the configuration or **(2)** or do nothing to prevent the infinite loop.

I suggest to re-write this bit to be more explicit:
```ts
// if there was an error reading the middleware config redirect to the error page
if (config.errorRedirect)  return config.errorRedirect
// prevent an infinite loop when coming from an OAuth flow like `/api/auth/callback/{provider}`
if (!config) return
```
so it's easier to grasp at first glance what's happening here without having to look at the implementation of `initConfig()`

src/next/middleware.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    const token = await getToken({ req: req as any, secret })
+
+    if (await authorized({ req, token })) return
+
+    return NextResponse.redirect(
+      `${signInPage}?${new URLSearchParams({ callbackUrl: req.url })}`
+    )
+  }
+}
```
lluia, line 160: I wonder if abstracting the request handling flow into a **helper function** would help in this case 🤔 :
```ts
async function handleRequest(config, onSuccess = () => {}) {
  // if there was an error reading the middleware config redirect to the error page
  if (config.errorRedirect)  return config.errorRedirect
  
  // prevent an infinite loop when coming from an OAuth flow like `/api/auth/callback  /{provider}`
  if (!config) return
  
  const { req, secret, signInPage, authorized } = config
  
  const token = await getToken({ req: req as any, secret })
  const isAuthorized = await authorized({ req, token })
  
  // the user is authorized, let the middleware handle the rest
  if (isAuthorized) return onSuccess()
  
  // the user is not logged in, re-direct to the sign-in page
  return NextResponse.redirect(
    `${config.signInPage}?${new URLSearchParams({ callbackUrl: req.url })}`
  )
}
```
It seems that for the **3 cases**:
- `args[0] instanceof NextRequest` 
- `typeof args[0] === "function"`
- `args[0] === options`

The flow is the same apart from what happens on success, right? 🤔

0ubbe (commented):
Superb feature 💯 💚 , leaving some thoughts to improve the readability of the code ☕️

bbigras:
Are the [nextauth callbacks](https://next-auth.js.org/configuration/callbacks) called with the middleware auth? It doesn't seem so, right now.

I'm using the `jwt()` callback to refresh the openid token.

src/middleware.ts
```diff
@@ -0,0 +1,2 @@
+export { default } from "./next/middleware"
+export * from "./next/middleware"
```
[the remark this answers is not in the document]
balazsorban44: From personal experience, not always.

src/next/middleware.ts
```diff
@@ -0,0 +1,162 @@
...
+      options,
+      secret,
+      signInPage,
+      authorized: options?.authorized ?? (({ token }) => token),
+    }
+
+  console.error(
+    `[next-auth][error][NO_SECRET]`,
+    `\nhttps://next-auth.js.org/errors#no_secret`
+  )
```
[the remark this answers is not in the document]
balazsorban44: We could add it as a feature later on maybe, I'm currently not sure how fetching in Middleware works.

src/next/middleware.ts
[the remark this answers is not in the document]
balazsorban44: This is an internal function and does not affect the end-user, I'm happy to improve internal typing in a follow-up PR! :+1:

balazsorban44:
small update, this hasn't been forgotten, but I've been traveling for the last few days. I plan to push this out this week! thank you for everyone's patience! those who could not wait hopefully used the experimental verison from higher up and tested it that everything works well. 🙏

0ubbe (approved):
Super 💚 ✨

src/next/middleware.ts
balazsorban44: You were right! I did refactor everything according to this.

balazsorban44:
Moved `authorized` to `callbacks.authorized` to resemble the [`callbacks.signIn` method](https://next-auth.js.org/configuration/callbacks#sign-in-callback) as they have the same purpose. This aligns our APIs a bit better.

src/next/middleware.ts
```diff
@@ -0,0 +1,141 @@
...
+      `[next-auth][error][NO_SECRET]`,
+      `\nhttps://next-auth.js.org/errors#no_secret`
+    )
+
+    return {
+      redirect: NextResponse.redirect(`${errorPage}?error=Configuration`),
+    }
+  }
+
+  const token = await getToken({ req: req as any })
```
sergeyshaykhullin, line 83: @balazsorban44 Does this middleware will work with `strategy: 'database'`?

I have similar middleware, but with `getSession` request, because 'getToken' was null if i remember correct...

wrod7: "Only supports the "jwt" [session strategy](https://next-auth.js.org/configuration/options#session). We need to wait until databases at the Edge become mature enough to ensure a fast experience. (If you know of an Edge-compatible database, we would like if you proposed a new [Adapter](https://next-auth.js.org/tutorials/creating-a-database-adapter))" 

[caveats](https://next-auth.js.org/configuration/nextjs#caveats)

mind sharing your sessions middleware function?  👀

sergeyshaykhullin: Yes, it is pretty simple

```ts
export async function middleware(request: NextRequest) {

  const cookie = request.headers.get('cookie')

  const session = cookie
    ? await getSession({ req: { headers: { cookie } } as any })
    : undefined

  if (!session) {
    return NextResponse.redirect(new URL('/api/auth/signin', request.url))
  }
}
```

zanzlender:
I tried to do this but for some reason it doesn't work... I always get redirected to Next-auth 's default login page... However looking at the middleware example I made this implementation which works for me.

```
// pages/_middleware.ts
import type { NextFetchEvent, NextRequest } from 'next/server'
import { getToken } from 'next-auth/jwt';

export function middleware(req: NextRequest, ev: NextFetchEvent) {
  let token = await getToken({
        req: request,
        secret: process.env.JWT_SECRET
      });

  if(token) return NextResponse.next();
  
  return NextResponse.redirect("login");
}
```

However like some mentioned in this and other issues, if you are using TypeScript, the req in getToken is compatible with NextApiRequest, but not NextRequest... I don't know if there is some hidden logic there, but for me it works just fine if you don't define the type and just use NextRequest.
Because of that, if you want type safety, I just made these little changes in @next-auth/jwt/index.d.ts. 

I imported this
`import type { NextRequest } from "next/server";`

And changed the acceptable types in GetTokenParams --> req
```
export interface GetTokenParams<R extends boolean = false> {
    /** The request containing the JWT either in the cookies or in the `Authorization` header. */
    req: NextApiRequest | NextRequest | Pick<NextApiRequest, "cookies" | "headers">;   <-- added NextRequest
    ....
}
```

Correct me if this is wrong and if this could break the app, but for now it works for me, and I had no problems what so ever.

talDoFlemis:
On build time it complains about the req when I deploy on Vercel, one way that I've found that worked was this

```
import { NextResponse } from "next/server"
import { NextApiRequest } from "next"

export async function middleware(req: NextApiRequest) {
  const url = req.url
    req,
    secret: process.env.NEXTAUTH_SECRET,
    secureCookie: process.env.NODE_ENV === "production",
  })

  if (url?.includes("/admin") && !token) {
    const parsedUrl = new URL(url)
    parsedUrl.pathname = "/stop"
    return NextResponse.redirect(parsedUrl.href)
  }
}
```
In that way, we still can use the token for the middleware autentication.

balazsorban44:
Hi, please note that commenting on a merged PR is not going to be looked at. Please open a bug report with your reproduction if having issues. :pray:

PR #3710, opened 2022-01-23
fix(providers): use battlenet openid

jcyamacho:
## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged


src/providers/battlenet.ts
```diff
@@ -0,0 +1,44 @@
...
+    id: "battlenet",
+    name: "Battle.net",
+    type: "oauth",
+    idToken: true,
+    wellKnown: `${base}/.well-known/openid-configuration`,
+    profile(profile) {
+      return {
+        id: profile.sub,
+        name: profile.battle_tag
+      }
```
balazsorban44, line 40: So there is no E-mail or profile picture?

balazsorban44 (changes requested):
Thanks, a small comment.

If there is anything special to document, could you open a PR for https://github.com/nextauthjs/docs/blob/main/docs/providers/battlenet.md

if not, that's fine.

src/providers/battlenet.ts
jcyamacho: Unfortunately no.. :( 
https://us.forums.blizzard.com/en/blizzard/t/get-email-on-oauth/11364/2

jcyamacho:
> Thanks, a small comment.
> 
> If there is anything special to document, could you open a PR for https://github.com/nextauthjs/docs/blob/main/docs/providers/battlenet.md
> 
> if not, that's fine.

done: https://github.com/nextauthjs/docs/pull/213 thanks!!

src/providers/battlenet.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const base = region.toUpperCase() === "CN"
+    ? "https://www.battlenet.com.cn/oauth"
+    : `https://${region}.battle.net/oauth`
+
+  return {
+    id: "battlenet",
+    name: "Battle.net",
+    type: "oauth",
+    wellKnown: `${base}/.well-known/openid-configuration`,
```
balazsorban44, line 34:
```suggestion
  options: OAuthUserConfig<P>
): OAuthConfig<P> {

  return {
    id: "battlenet",
    name: "Battle.net",
    type: "oauth",
    wellKnown: `${options.issuer}/.well-known/openid-configuration`,
```

Let's not introduce a new concept (`region`), to keep providers uniform :+1:

src/providers/battlenet.ts
```diff
@@ -0,0 +1,43 @@
...
+    profile(profile) {
+      return {
+        id: profile.sub,
+        name: profile.battle_tag
```
balazsorban44, line 38:
```suggestion
        id: profile.sub,
        name: profile.battle_tag,
        email: null,
        image: null,
```

Let's keep the `null` values, as this would change the behaviour of adapters (some does not work with `undefined`)

src/providers/battlenet.ts
```diff
@@ -0,0 +1,43 @@
+import { OAuthConfig, OAuthUserConfig } from "."
```
balazsorban44, line 1:
```suggestion
import type { OAuthConfig, OAuthUserConfig } from "."
```

balazsorban44 (changes requested):

src/providers/battlenet.ts
jcyamacho: looks good, but maybe is better to throw an error instead of having China issuer (or any other region issuer) as a default, wdyt?

jcyamacho: hmm...
image
I will set one of the regions of global partition as a default

packages/docs/docs/providers/battlenet.md
```diff
@@ -15,21 +15,100 @@ https://develop.battle.net/access/clients
 
 The **Battle.net Provider** comes with a set of default options:
 
-- [Battle.net Provider options](https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/providers/battlenet.js)
+- [Battle.net Provider options](https://github.com/nextauthjs/next-auth/blob/main/src/providers/battlenet.ts)
```
balazsorban44, line 18:
```suggestion
- [Battle.net Provider options](https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/providers/battlenet.ts)
```

URL is this. (for now :sweat_smile:)

packages/docs/docs/providers/battlenet.md
```diff
@@ -15,21 +15,100 @@ https://develop.battle.net/access/clients
...
 
 You can override any of the options to suit your own use case.
 
+### Issuer
+
+Battle.net oprrate with account partitions, an issuer should be selected according to the region: https://develop.battle.net/documentation/guides/regionality-and-apis
```
balazsorban44, line 24:
```suggestion
Battle.net operates with account partitions, an issuer should be selected according to the region: https://develop.battle.net/documentation/guides/regionality-and-apis
```

packages/docs/docs/providers/battlenet.md
```diff
@@ -15,21 +15,100 @@ https://develop.battle.net/access/clients
...
+      }
+      return token
+    },
+    async session ({ session, token }) {
+      session.accessToken = token.accessToken
+      return session
+    }
+  }
+})
+```
```
balazsorban44, line 114: This whole section is a duplicate of what's documented already in https://next-auth.js.org/configuration/callbacks and also https://next-auth.js.org/getting-started/example#using-nextauthjs-callbacks

We should just link to those sections to keep provider docs consistent.

What you could mention is the need for the extra `wow.profile` scope I assume, and link to some Battle.net documentation about it.

balazsorban44, line 114: This whole section is a duplicate of what's documented already in https://next-auth.js.org/configuration/callbacks and also https://next-auth.js.org/getting-started/example#using-nextauthjs-callbacks

We should just link to those sections to keep the provider docs consistent and simple.

What you could mention is the need for the extra `wow.profile` scope I assume, and add a link to some Battle.net documentation about it that explains why do you need that scope.

packages/next-auth/src/providers/battlenet.ts
```diff
@@ -0,0 +1,36 @@
...
+    id: "battlenet",
+    name: "Battle.net",
+    type: "oauth",
+    wellKnown: `${options.issuer}/.well-known/openid-configuration`,
+    profile(profile) {
+      return {
+        id: profile.sub,
+        name: profile.battle_tag,
+        email: null,
+        image: null,
```
balazsorban44, line 31: Do you know if there are any endpoints this could be retrieved from?

balazsorban44 (changes requested):
Thank you for the PR! I'm finally able to look at them again, and I had some comments. Could you address them, please? :pray:

As proof that everything is working, could you share a small screenshot/snippet of code of the output about the user when you are logged in?

packages/next-auth/src/providers/battlenet.ts
jcyamacho: nop :(

jcyamacho:
@balazsorban44  looks like the changes are done here: https://github.com/nextauthjs/next-auth/commit/75cb175259fdb8bc5e5745f72ccf143209387fdb, https://github.com/nextauthjs/next-auth/commit/163149b4243b857f73a2cee9afa62a958d57bb51

PR #3450, opened 2021-12-15
feat(react): add `disableRefetchOnRefocus` option to SessionProvider

filipditrich:
Adding new option `disableRefetchOnRefocus` to SessionProvider component which disables session refetching in `visibilitychange` event callback.

## Reasoning 💡

Automatic refetching on tab switching is usefull but it should be controllable by user, since it may be an expensive operation.

## Checklist 🧢

- [x] Documentation
- [ ] ~Tests~
- [x] Ready to be merged

## Affected issues 🎟

Fixes #3405

PR #3730, opened 2022-01-25
feat(react): add refetchOnWindowFocus option to SessionProvider

### LINKED ISSUE #3405: Session revalidation on refocus causes unwanted component rerenders
### Description 🐜

With the removal of `clientMaxAge` and `keepAlive` in favor of `refetchInterval`, every time a tab is switched, the session gets revalidated, thus updating a session state in `SessionProvider` which then causes to rerender all components using `useSession`.

Is it possible to disable this "revalidation" upon window refocusing?
Thanks in advance.

No

### How to reproduce ☕️

```tsx
<SessionProvider
session={pageProps.session}
// re-validate stale session every hour
refetchInterval={60 * 60}
>
  <Component /> {/* <-- on tab switch this component gets rerendered, it "resets" its states */}
</SessionProvider>
```

### Screenshots / Logs 📽

Screenshot 2021-12-08 at 7 21 49 PM

### Environment 🖥

 System:
    OS: macOS 12.0.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 8.40 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.18.1 - ~/.nvm/versions/node/v14.18.1/bin/node
    Yarn: 1.22.17 - /opt/homebrew/bin/yarn
    npm: 6.14.15 - ~/.nvm/versions/node/v14.18.1/bin/npm
  Browsers:
    Firefox Developer Edition: 96.0
    Safari: 15.1
  npmPackages:
    next: 11.1.2 => 11.1.2 
    next-auth: 4.0.5 => 4.0.5 
    react: 17.0.2 => 17.0.2 

### Contributing 🙌🏽

timakin:
The original feature request (#3450) looks left behind for about a month.

But because you can't control whether to refetch session when you focus on a browser, some issues still occur (like form resetting, duplicated API calls, etc).

So I re-open this PR instead of @filipditrich (almost all of this feature depend on his effort) and handle @balazsorban44 's reviews.

Following descriptions are copies from [here](https://github.com/nextauthjs/next-auth/pull/3450).

----

Adding new option refetchOnWindowFocus to SessionProvider component which disables session refetching in visibilitychange event callback.

## Reasoning 💡

Automatic refetching on tab switching is usefull but it should be controllable by user, since it may be an expensive operation.

## Checklist 🧢

- [x] Documentation
- [ ] ~Tests~
- [x] Ready to be merged

## Affected issues 🎟

Fixes #3405

### DIFF
--- a/src/react/index.tsx
+++ b/src/react/index.tsx
@@ -379,16 +379,18 @@ export function SessionProvider(props: SessionProviderProps) {
   }, [])
 
   React.useEffect(() => {
+    const { refetchOnWindowFocus = true } = props
     // Listen for when the page is visible, if the user switches tabs
-    // and makes our tab visible again, re-fetch the session.
+    // and makes our tab visible again, re-fetch the session, but only if
+    // this feature is not disabled.
     const visibilityHandler = () => {
-      if (document.visibilityState === "visible")
+      if (refetchOnWindowFocus && document.visibilityState === "visible")
         __NEXTAUTH._getSession({ event: "visibilitychange" })
     }
     document.addEventListener("visibilitychange", visibilityHandler, false)
     return () =>
       document.removeEventListener("visibilitychange", visibilityHandler, false)
-  }, [])
+  }, [props.refetchOnWindowFocus])
 
   React.useEffect(() => {
     const { refetchInterval } = props
--- a/src/react/types.ts
+++ b/src/react/types.ts
@@ -70,4 +70,9 @@ export interface SessionProviderProps {
    * If set to `0` (default), the session is not polled.
    */
   refetchInterval?: number
+  /**
+   * `SessionProvider` automatically refetches the session when the user switches between windows.
+   * This option activates this behaviour if set to `true` (default).
+   */
+  refetchOnWindowFocus?: boolean
 }


balazsorban44 (approved):
Thank you! :pray: :green_heart:

hboylan:
Thank you for implementing this feature. I need this as well since we already have a refresh token flow in our app.

Any idea when this can be published to NPM?

timakin:
@hboylan 
You mean you'd re-fetch the session regardless of this `refetchOnWindowFocus` option when you refresh an access token, right?
This package will automatically try to fetch a new access token every `refetchInterval` as before.
Isn't it enough?

hboylan:
@timakin no, I will use the refresh token to fetch a new access token and retry the service request when my service responds with 401 unauthorized. Meaning the access token is expired.

Therefore, I would like to use this new SessionProvider prop to prevent next-auth from fetching a new session and hitting the OAuth service unnecessarily

This is exactly the feature I've been looking trying to reproduce. 😄

timakin:
@hboylan 
Thanks for a detail. I see.
However, it looks it is not the issue of this option, but the another one of preventing useless API calls if the session was expired.
So, it's better to report as a new issue to block session re-fetch when it's already unauthorized. 👍

hboylan:
@timakin Basically, I would like to set `refetchOnWindowFocus={false}` so to prevent `GET /api/auth/session` from firing every time the tab gains focus. This is causing several issues w/ my application. Instead, I will rely on `refetchInterval` or simply refresh the session if my other service responds with 401.

> Any idea when this can be published to NPM?

Code-Hex:
@balazsorban44 Hi 👋 
I want to use this feature as soon as possible. When will the feature be available?

Issue #2414, opened 2021-07-21
Recommended approach to share session information between applications on subdomains


raphaelpc:
### Question 💬

Hello! First of all, thanks for all the great work on this amazing library!

I'm trying to implement a solution in which i have the same authentication on subdomains as well as on main domain.
What is the recommended approach to share session information between all applications?

_This is a continuation of a discussion that started on some previous issues, like #405, #794, #1668.
All those issues are now closed without a full example or definitive answer, so i decided to make a summary of what was previously discussed, in hope that i can get some help on deciding what is the best approach to do that integration on all domains and subdomains of my application and help document that approach for others in the same need._

### How to reproduce ☕️

The application i'm working on is, actually, an ecosystem of applications, and not all of them are made with Next.js (or even with React). Many are legacy applications written in JSP.

The main one isn't an Next.js app and will live in "www.my-domain.com".
It will have sub-domains, as in "app1.my-domain.com", "app2.my-domain.com", etc.
As explained [in the FAQ](https://next-auth.js.org/faq#can-i-use-nextauthjs-with-a-website-that-does-not-use-nextjs):
> If you are using a different framework for you website, you can create a website that handles sign in with Next.js and then access those sessions on a website that does not use Next.js as long as the websites are on the same domain.
>
> If you use NextAuth.js on a website with a different subdomain then the rest of your website (e.g. auth.example.com vs www.example.com) you will need to set a custom cookie domain policy for the Session Token cookie.

So, since the main application isn't a React / Next.js app, I will have to implement an auth application so i can use NextAuth: "auth.my-domain.com". Notice that it's the ONLY application that includes NextAuth - from my understanding, it makes no sence to add NextAuth to all my other Next.js applications.

I created a diagram to help illustrate the ecosystem:

![image](https://user-images.githubusercontent.com/30241943/126547656-6939cc44-8df8-4f9f-8bd6-518f104f9089.png)

To set a custom cookie domain policy for the Session Token cookie, i did in my [...nextauth] API route what was originally posted by @Xodarap in https://github.com/nextauthjs/next-auth/issues/405#issuecomment-737593528:

```
const useSecureCookies = process.env.NEXTAUTH_URL.startsWith('https://')
const cookiePrefix = useSecureCookies ? '__Secure-' : ''
const hostName = Url(process.env.NEXTAUTH_URL).hostname
const options = {
  cookies: {
    sessionToken: 
    {
      name: `${cookiePrefix}next-auth.session-token`,
[...]

PR #3735, opened 2022-01-25
feat(react): custom fetch options

timosaikkonen:
Introducing the ability to customize options passed to [fetchData](https://github.com/nextauthjs/next-auth/blob/4824f8c02ad1b0799f4c15eea9c86e93fc5c72fc/src/client/_utils.ts#L32) by passing `fetchOptions` as a prop to `SessionProvider`

## Reasoning 💡

Primarily to facilitate setting `"credentials": "include"` in a case where a single next-auth server is serving multiple Next sites.

## Checklist 🧢

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

## Affected issues 🎟

https://github.com/nextauthjs/next-auth/issues/2414

### DIFF
--- a/packages/next-auth/src/client/__tests__/client-provider.test.js
+++ b/packages/next-auth/src/client/__tests__/client-provider.test.js
@@ -155,6 +155,33 @@ test("allows to customize the URL for session fetching", async () => {
   })
 })
 
+test("allows to customize headers for session fetching", async () => {
+  server.use(
+    rest.get(`/api/auth/session`, (req, res, ctx) =>
+      res(ctx.status(200), ctx.json(mockSession))
+    )
+  )
+
+  const fetchOptions = {
+    headers: {
+      'X-Foo-Bar': 'hello'
+    }
+  }
+
+  render(<ProviderFlow session={mockSession} fetchOptions={fetchOptions} />)
+
+  // there's an existing session so it should not try to fetch a new one
+  expect(fetchSpy).not.toHaveBeenCalled()
+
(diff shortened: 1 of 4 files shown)


balazsorban44:
I'm not sure of the immediate security implications here, as I haven't used that setting before. :thinking:

timosaikkonen:
@balazsorban44 Anything I can do to get this to land? I'm dealing with some schedule pressure here :(

adamelmore:
@balazsorban44 I'd also love to see this PR land. Any thoughts?

timosaikkonen:
@balazsorban44 Gentle nudge :) Anything I can do to help this land?

timosaikkonen:
@balazsorban44 Apologies for pestering you about this but I'm still looking to get this to land. As always, let me know if there's anything more I can do to move this forward. Happy to jump on a call to run through the changes if that's easier for you.

ranmocy:
This is still a relevant requests. I think Next Auth JS should consider support this.
By default people won't use it. And for people knows what they are doing, this is almost the only way to do it without forking the package.
The primary use case we have is to support a single page application calling the API server on a different subdomain (e.x. `api.example.com` vs `app.example.com`). Access to the CORS settings for those auth requests are needed.
It would be great Next Auth JS could support it, since it's designed as a self-hosting solution, providing the ability for customization seems reasonable here.

dimamarkus:
Bump. This still seams to be one of the cleanest solution to a real problem. https://github.com/nextauthjs/next-auth/issues/2414

PR #3740, opened 2022-01-25
feat: added types for sign in errors

### LINKED ISSUE #3665: Type the error message
### Question 💬

It seems that the error messages are not typed. If we could have typed errors, it could further improve customizations.

> According to what I found here: https://github.com/nextauthjs/next-auth/blob/82447f8e3ebc7004cf91121725b4f5970d2276d8/src/core/pages/signin.tsx#L42

An example I can think of is:

```typescript
export type ErrorType = 
  'SignIn' | 
  'OAuthSignin'
```

### How to reproduce ☕️

N/A

### Contributing 🙌🏽

norbitrial:
## Reasoning 💡

  This PR is intended to add typing for `<SigninPage />` component's `error` property as the following:

Screenshot 2022-01-25 at 21 53 02

After this change it will accept only the giving `string` values. Originally it accepted any `string` value.

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## Affected issues 🎟

Fixes #3665

### DIFF
--- a/src/core/pages/signin.tsx
+++ b/src/core/pages/signin.tsx
@@ -1,12 +1,25 @@
 import { Theme } from "../.."
 import { InternalProvider } from "../../lib/types"
 
+export type SignInErrorTypes =
+  | "Signin"
+  | "OAuthSignin"
+  | "OAuthCallback"
+  | "OAuthCreateAccount"
+  | "EmailCreateAccount"
+  | "Callback"
+  | "OAuthAccountNotLinked"
+  | "EmailSignin"
+  | "CredentialsSignin"
+  | "SessionRequired"
+  | "default"
+
 export interface SignInServerPageParams {
   csrfToken: string
   providers: InternalProvider[]
   callbackUrl: string
   email: string
-  error: string
+  error: SignInErrorTypes
   theme: Theme
 }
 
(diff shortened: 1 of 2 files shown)


src/core/pages/signin.tsx
```diff
@@ -59,16 +72,17 @@ export default function SigninPage(props: SignInServerPageParams) {
 
   return (
     <div className="signin">
-      
-      { theme.brandColor && <style
```
norbitrial, line 63: Not 100% sure if this is caused by a wrong Prettier config on my end or it's still a valid change.

Let me know if I missed something here, thanks! 👀

src/core/pages/signin.tsx
```diff
@@ -1,12 +1,25 @@
 import { Theme } from "../.."
 import { InternalProvider } from "../../lib/types"
 
+export type SignInErrorTypes =
```
balazsorban44, line 4: could we add a JSDoc comment on top of this that takes the user to the https://next-auth.js.org/configuration/pages#sign-in-page page, or better yet, each value should have a comment on them. See the link for reference.

balazsorban44 (changes requested):
Thanks! Could we re-export this from `next-auth`?

Could you do the same for the error page as well?

https://github.com/nextauthjs/next-auth/blob/714d80a4f55a1ccd556dc9e1510a54d8ccd4d8b1/src/core/pages/error.tsx#L7

https://next-auth.js.org/configuration/pages#error-page

src/core/pages/signin.tsx
norbitrial: Thanks for the review! 👍 Just added the `ErrorType` for the error page, also linked in both cases the documentations.

balazsorban44 (approved):

Issue #340, opened 2020-06-26
WebAuthn support


shouze:
*Please stick to one distinct feature request per issue where possible and raise additional feature quests as separate issues. Try to avoid adding feature requests to existing issues in the comments of issues raised by other users.*

**Summary of proposed feature**
Since Safari 14 will support [WebAuthn](https://webauthn.io/) through [Touch/FaceID](https://www.theverge.com/2020/6/24/21301509/apple-safari-14-browser-face-touch-id-logins-webauthn-fido2) this is a true game changer (like passwordless) in authentication domain. It's supported on Android too (since at least Android 7).

**Purpose of proposed feature**
Seamless mobile authentication with better pricacy control

**Detail about proposed feature**
[more than 60% of mobile devices have a biometric hardware part](https://www.statista.com/statistics/804269/global-smartphone-fingerprint-sensor-penetration-rate/).

**Potential problems**
Not sure yet, I've not familiar enough with WebAuthn & FIDO2.
Limitations are more on desktop platforms, should not be a good idea at the moment to expose this feature as very few computers get biometric hardware so it defaults to security keys that very few people owns. This probably imply device detection.

**Describe any alternatives you've considered**
Looks like there's [a good repository to start about WebAuthn](https://github.com/herrjemand/awesome-webauthn). And some [js code available here](https://github.com/web-auth/webauthn-helper).

**Additional context**
![Screenshot_20200626-103547](https://user-images.githubusercontent.com/54712/85839520-4e518280-b79b-11ea-9bc7-a429edce2d8d.png)
![Screenshot_20200626-103556](https://user-images.githubusercontent.com/54712/85839526-4eea1900-b79b-11ea-908b-8011a73e4a93.png)
![Screenshot_20200626-103926](https://user-images.githubusercontent.com/54712/85839528-4f82af80-b79b-11ea-8350-fa4642a9bca9.png)

*Please indicate if you are willing and able to help implement the proposed feature.*
Yes for sure!

balazsorban44:
@shouze I am reopening this issue, as we have some more capacity now. Just opened an RFC https://github.com/nextauthjs/next-auth/discussions/1465 and I was wondering if that would fit under there? I am not totally familiar with how WebAuthn works just yet, so if you figured out more about it since this post, it would help a lot to determine how could we proceed with this feature request.

gamedevsam:
[...]

PR #3750, opened 2022-01-26
feat(providers): Added database session strategy for Credentials

nixuuu:
New provider type: dbCredentials

That allows to use custom credentials authentication with `{ ...NextAuthOptions, session: { strategy: "database" } }` - only when users are stored in same database table.

Example of use:
```typescript
CredentialsProvider({
  name: "Credentials",
  credentials: {
  },
    if (credentials.password === "pw") {
      return {
        id: '1', // must match id from database
      }
    }
    return null
  },
}),
```
And that's all, now we can use our users from database along with oauth providers and session strategy `database`

image

## Reasoning 💡
Current Credentials provider does not allow to login using session strategy `database` (rather than `jwt`). It makes Credentials useless with this method - even 2FA token is impossible to implement with this strategy.

## Checklist 🧢

- [ ] ~Documentation~
- [ ] ~Tests~
- [x] Ready to be merged

Fixes #3749

### DIFF
--- a/src/core/lib/callback-handler.ts
+++ b/src/core/lib/callback-handler.ts
@@ -29,7 +29,7 @@ export default async function callbackHandler(params: {
   // Input validation
   if (!account?.providerAccountId || !account.type)
     throw new Error("Missing or invalid provider account")
-  if (!["email", "oauth"].includes(account.type))
+  if (!["email", "oauth", "dbCredentials"].includes(account.type))
     throw new Error("Provider not supported")
 
   const {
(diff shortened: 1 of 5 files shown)


src/core/lib/callback-handler.ts
```diff
@@ -216,6 +216,17 @@ export default async function callbackHandler(params: {
 
       return { session, user, isNewUser: true }
     }
+  } else if (account.type === "dbCredentials") {
```
KATT, line 219: Does this need to be a separate `account.type` than `credentials`? The OAuth flows can seamlessly work with both `jwt` or `database`

KATT (commented):
Awesome! I just opened #3749 😊

I don't have enough knowledge of the codebase to properly review :)

src/core/lib/callback-handler.ts
nixuuu: I didn't want it to disturb the current `Credential` flow, because credentials in `jwt` strategy doesn't need database at all. It's not problem for me to change this if any reviewer say that - like I say, it's just to not destroy whole internet 😄

nixuuu: @KATT I removed recently added type `dbCredentials` to use existing one `credentials`

balazsorban44:
There are similar/better ideas at

#340 and #1465

PR #3782, opened 2022-02-02
feat(providers): add BoxyHQ SAML Jackson provider

deepakprabhakara:
https://github.com/boxyhq/jackson is an open-source SAML SSO service that implements SAML login as an OAuth 2.0 flow. Adding it as a provider here for the convenience of our users.

## Reasoning 💡

This simplifies the usage for our customers and we hope other NextAuth users will find our SAML integration useful for their enterprise apps.

## Checklist 🧢

- [x] Documentation (https://github.com/nextauthjs/docs/pull/217)
- ~[ ] Tests~
- [x] Ready to be merged

## Affected issues 🎟

n/a


src/providers/saml-jackson.ts
```diff
@@ -0,0 +1,51 @@
...
+      params: {
+        scope: "",
+        response_type: "code",
+        provider: "saml",
+      },
+    },
+    token: {
+      url: `${options.issuer}/api/oauth/token`,
+      params: { grant_type: "authorization_code" },
+    },
```
balazsorban44, line 33:
```suggestion
    token: `${options.issuer}/api/oauth/token`,
```

I believe that parameter is added by default, no need for you to specify. But please check.

src/providers/saml-jackson.ts
```diff
@@ -0,0 +1,51 @@
...
+    userinfo: `${options.issuer}/api/oauth/userinfo`,
+    profile(profile) {
+      return {
+        id: profile.id || "",
+        firstName: profile.firstName || "",
+        lastName: profile.lastName || "",
+        email: profile.email || "",
+        name: `${profile.firstName || ""} ${profile.lastName || ""}`.trim(),
+        email_verified: true,
+        image: null,
```
balazsorban44, line 43:
```suggestion
        id: profile.id,
        email: profile.email ?? null,
        name: [profile.firstName, profile.lastname].filter(Boolean).join(" "),
        image: null,
```

The initial profile is the same for all providers on purpose.

What's the reasoning for `|| ""` on multiple parameters? I would assume at least an `id` is always returned. :thinking: Is there a chance of that not being returned? That would complicate flows where the user wants to use a [database adapter](https://next-auth.js.org/adapters/overview).

What about email? Is it also not guaranteed to return?

src/providers/saml-jackson.ts
```diff
@@ -0,0 +1,51 @@
...
+      }
+    },
+    options: {
+      clientId: options.clientId || "dummy",
+      clientSecret: options.clientSecret || "dummy",
+    },
```
balazsorban44, line 49:
```suggestion
    options
```

This should be kept as-is, so we can merge the user's overrides on top of the defaults from above.

If "dummy" is a special client, I recommend adding this information to the docs instead (https://github.com/nextauthjs/docs/pull/217), to mention that users can experiment with this provider using the `"dummy"` client id and secret values.

balazsorban44 (changes requested):
Thanks, looking good, but I added some comments. Could you also provide a screenshot of a successful login for future reference?

deepakprabhakara:
Thanks @balazsorban44, I'll address the suggestions shortly.

src/providers/saml-jackson.ts
deepakprabhakara: id and email are guaranteed to return.

deepakprabhakara:
Screenshot-BoxyHQ-example

deepakprabhakara:
Thanks @balazsorban44, for very useful suggestions. I wasn't aware of a few things you mentioned there.

deepakprabhakara:
@balazsorban44 Adjusted the code for the monorepo changes but unable to test due to some issues with running the dev app, will look into it soon. https://github.com/nextauthjs/docs/pull/217 can be closed

packages/docs/docs/providers/saml-jackson.md
```diff
@@ -0,0 +1,58 @@
+---
+id: saml-jackson
+title: BoxyHQ SAML Jackson
+---
+
+## Documentation
+
+BoxyHQ SAML Jackson is an open source service that handles the SAML login flow as an OAuth 2.0 flow, abstracting away all the complexities of the SAML protocol.
+
+You can deploy SAML Jackson as a separate service or embed it into your app using our NPM library. [Please check our documentation for more details](https://boxyhq.com/docs/jackson/deploy)
```
balazsorban44, line 10:
```suggestion
You can deploy SAML Jackson as a separate service or embed it into your app using our NPM library. [Check out the documentation for more details](https://boxyhq.com/docs/jackson/deploy)
```

packages/docs/docs/providers/saml-jackson.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Configuration
+
+SAML login requires a configuration for every tenant of yours. One common method is to use the domain for an email address to figure out which tenant they belong to. You can also use a unique tenant ID (string) from your backend for this, typically some kind of account or organization ID.
+
+Check our [documentation](https://boxyhq.com/docs/jackson/getting-started#2-saml-config-api) for more details.
```
balazsorban44, line 16:
```suggestion
Check out the [documentation](https://boxyhq.com/docs/jackson/getting-started#2-saml-config-api) for more details.
```

packages/docs/docs/providers/saml-jackson.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Options
+
+The **SAML Jackson Provider** comes with a set of default options:
+
+- [BoxyHQ Provider options](https://github.com/nextauthjs/next-auth/blob/main/src/providers/boxyhq.ts)
```
balazsorban44, line 22:
```suggestion
- [BoxyHQ Provider options](https://github.com/nextauthjs/next-auth/tree/main/packages/next-auth/src/providers/boxyhq.ts)
```

Let's keep the filename/id the same everywhere. You can decide on which one to use. `boxyhq` or `saml-jackson` or something else?

packages/docs/docs/providers/saml-jackson.md
```diff
@@ -0,0 +1,58 @@
...
+
+  // Map your users's email to a tenant and product
+  const tenant = email.split("@")[1];
+  const product= 'my_awesome_product';
+...
+  <Button
+    onClick={async (event) => {
+      event.preventDefault();
+
+      signIn("saml", {}, { tenant: tenant, product: product });
```
balazsorban44, line 55:
```suggestion
      signIn("saml-jackson", {}, { tenant, product });
```

The first argument must match the provider implementation's `id`. You can choose another name, but let's keep it consistent.

packages/docs/docs/providers/saml-jackson.md
```diff
@@ -0,0 +1,58 @@
...
+## Example
+
+```ts
+import SAMLJacksonProvider from "next-auth/providers/saml-jackson"
+...
+providers: [
+  SAMLJacksonProvider({
+    issuer: "http://localhost:5000",
+    clientId: "dummy", // The dummy here is necessary since we'll pass tenant and product custom attributes in the client code
+    clientSecret: "dummy", // The dummy here is necessary since we'll pass tenant and product custom attributes in the client code
```
balazsorban44, line 35:
```suggestion
    issuer: "http://localhost:5000",
    clientId: "dummy", // The dummy here is necessary since we'll pass tenant and product custom attributes in the client code
    clientSecret: "dummy", // The dummy here is necessary since we'll pass tenant and product custom attributes in the client code
```

This is an unusual flow, so I would like to have more context here.

If the `clientId` and `clientSecret` values aren't important, we should just omit them, or set them by default in the provider code.

packages/docs/docs/providers/saml-jackson.md
```diff
@@ -0,0 +1,58 @@
...
+  })
+}
+...
+```
+
+On the client side you'll need to pass additional parameters `tenant` and `product` to the `signIn` function. This will allow SAML Jackson to figure out the right SAML configuration and take your user to the right SAML Identity Provider to sign them in.
+
+```tsx
+import { getCsrfToken, signIn } from "next-auth/react";
```
balazsorban44, line 44:
```suggestion
import { signIn } from "next-auth/react";
```

packages/docs/docs/providers/saml-jackson.md
```diff
@@ -0,0 +1,58 @@
...
+...
+
+  // Map your users's email to a tenant and product
+  const tenant = email.split("@")[1];
+  const product= 'my_awesome_product';
```
balazsorban44, line 49:
```suggestion
  const product = 'my_awesome_product';
```

Could you explain where this value comes from? Is it static for the app's lifetime? Is it coming from a dropdown/select?

packages/docs/docs/providers/saml-jackson.md
```diff
@@ -0,0 +1,58 @@
...
+
+```ts
+import SAMLJacksonProvider from "next-auth/providers/saml-jackson"
+...
+providers: [
+  SAMLJacksonProvider({
+    issuer: "http://localhost:5000",
+    clientId: "dummy", // The dummy here is necessary since we'll pass tenant and product custom attributes in the client code
+    clientSecret: "dummy", // The dummy here is necessary since we'll pass tenant and product custom attributes in the client code
+  })
```
balazsorban44, line 36:
```suggestion
  SAMLJacksonProvider({ issuer: process.env.SAML_JACKSON_ISSUER })
```

This is an unusual flow, so I would like to have more context here. Why are client id and client secret "dummy" values?

If the `clientId` and `clientSecret` values aren't important somehow (very curious), we should just omit them, or set them by default in the provider code.

From the localhost issuer, I will also assume that you don't have a hosted service/it's something the user will always need to set up themselves? Do you have something similar to https://demo.identityserver.io/ that the user could play with? It would be a nice mention on this docs page.

Is this referring to the same as here: https://github.com/nextauthjs/next-auth/pull/3782/files#diff-e4ec1563a89a51e2b61e12a89fc66bb226e55e3f27173919bdb9b9f75a713d85R10

packages/next-auth/src/providers/saml-jackson.ts
```diff
@@ -0,0 +1,37 @@
+import type { OAuthConfig, OAuthUserConfig } from "."
+
+export interface SAMLJacksonProfile {
+  id: string;
+  email: string;
+  firstName: string | null | undefined;
+  lastName: string | null | undefined;
```
balazsorban44, line 7:
```suggestion
  email?: string;
  firstName?: string | null;
  lastName?: string | null;
```

Let's prefer optional parameters.

packages/next-auth/src/providers/saml-jackson.ts
```diff
@@ -0,0 +1,37 @@
...
+      },
+    },
+    token: `${options.issuer}/api/oauth/token`,
+    userinfo: `${options.issuer}/api/oauth/userinfo`,
+    profile(profile) {
+      return {
+        id: profile.id,
+        email: profile.email || "",
+        name: [profile.firstName, profile.lastName].filter(Boolean).join(" "),
+        image: null,
```
balazsorban44, line 32:
```suggestion
        email: profile.email || "",
        name: [profile.firstName, profile.lastName].filter(Boolean).join(" "),
        image: null,
```

In which case will the user not receive an email here?

Is there any API to get a profile picture?

balazsorban44 (changes requested):
Thanks for the PR! I've read through it and had some comments. Could you address those? Also, could you provide a screenshot/code snippet of a logged-in user as proof that this setup works as intended?

I asked in one of the comments, but I'll ask here. Do you have something similar to https://demo.identityserver.io/ so I could verify this myself without needing to do this: https://boxyhq.com/docs/jackson/deploy#with-docker (I can if there is no online demo).

Small note, the above link makes the scroll "stuck" on your docs page, you might want to check it out.

As the monorepo is fresh, I'm still investigating if we can automate this, but for now, could you also add the provider (title) somewhere here?:

https://github.com/nextauthjs/next-auth/blame/main/.github/ISSUE_TEMPLATE/bug_1_provider.yml#L33

packages/next-auth/src/providers/saml-jackson.ts
deepakprabhakara: Unfortunately SAML does not provide a way to get a profile picture

packages/docs/docs/providers/saml-jackson.md
deepakprabhakara: Makes sense, I'll make this consistent.

packages/docs/docs/providers/saml-jackson.md
deepakprabhakara: `tenant` and `product` are decided by the user, typically unique values that identify their customer. SAML (unlike OAuth 2.0) is tenanted per customer (and product).

packages/docs/docs/providers/saml-jackson.md
deepakprabhakara: `clientId` and `clientSecret` aren't as critical here (unlike OAuth 2.0) because SAML is tenanted (unique config per customer, that is also configured by the customer using their SAML Identity Provider). When `tenant` and `product` are used then `clientId` and `clientSecret` become merely artifacts for the OAuth 2.0 flow. More details explained [here](https://boxyhq.com/docs/jackson/security#client-secret).

We just released a mock SAML service that can be used for testing - https://mocksaml.com, I'll add it to the docs. Please also try our demo here to get an understanding of the flow - https://saml-demo.boxyhq.com/

deepakprabhakara: I had originally internalised the `dummy` logic here - https://github.com/nextauthjs/next-auth/pull/3782/commits/8955ffd887a63fc40b670ed4251488e2c608e177#diff-76f06144da0418fa936f7f4af37fcf4b1c21448f94716360feef15e4c2be4013L46

deepakprabhakara: @balazsorban44 I have setup a demo that requires no configuration from your side. Please run the app from this branch - https://github.com/nextauthjs/next-auth/pull/3782/commits/afbdc5b4aee33ccc4ef1f5292ca44217a5fe94c3

deepakprabhakara:
@balazsorban44 Thanks again for the review, I have made the necessary changes and also added an example for you to test.

packages/next-auth/src/providers/boxyhq-saml.ts
```diff
@@ -0,0 +1,37 @@
+import type { OAuthConfig, OAuthUserConfig } from "."
+
+export interface SAMLJacksonProfile {
+  id: string;
+  email?: string;
```
balazsorban44, line 5: You mentioned `email` is guaranteed at https://github.com/nextauthjs/next-auth/pull/3782#discussion_r799064357
```suggestion
  email: string
```

packages/next-auth/src/providers/boxyhq-saml.ts
```diff
@@ -0,0 +1,37 @@
...
+      params: {
+        provider: "saml",
+      },
+    },
+    token: `${options.issuer}/api/oauth/token`,
+    userinfo: `${options.issuer}/api/oauth/userinfo`,
+    profile(profile) {
+      return {
+        id: profile.id,
+        email: profile.email || "",
```
balazsorban44, line 30:
```suggestion
        email: profile.email,
```

You mentioned this being guaranteed: https://github.com/nextauthjs/next-auth/pull/3782#discussion_r799064357

packages/next-auth/src/providers/boxyhq-saml.ts
```diff
@@ -0,0 +1,37 @@
+import type { OAuthConfig, OAuthUserConfig } from "."
+
+export interface SAMLJacksonProfile {
+  id: string;
```
balazsorban44, line 4:
```suggestion
  id: string
```

packages/next-auth/src/providers/boxyhq-saml.ts
```diff
@@ -0,0 +1,37 @@
...
+  email?: string;
+  firstName?: string | null;
+  lastName?: string | null;
```
balazsorban44, line 7:
```suggestion
  firstName?: string | null
  lastName?: string | null
```

Are these really optional, or guaranteed to return at least `null`, when missing on your side? Should we drop either the `?` or `null`?

balazsorban44 (changes requested):
Thanks. Besides the minor concerns and formatting preferences, is there a way to use consistent naming in all places?

I am seeing BoxyHQ SAML Jackson and all its permutations in different places:

- BoxyHQ SAML Jackson
- `"boxyhq-saml"`
- `"saml"`
- `SAMLJacksonProfile`
- `BoxyHQSAMLProvider`
- SAML Jackson

We should try to stick to a similar name, derived from the same brand naming in all places like docs, TypeScript, function names, ids, etc.

Could we unify these? :pray:

balazsorban44:
I have to add, the demo provider is wonderful, I wish most providers would have a public example like that :pray: :sparkles:

docs/docs/providers/boxyhq-saml.md
```diff
@@ -0,0 +1,58 @@
...
+## Example
+
+```ts
+import BoxyHQSAMLProvider from "next-auth/providers/boxyhq-saml"
+...
+providers: [
+  BoxyHQSAMLProvider({
+    issuer: "http://localhost:5000",
+    clientId: "dummy", // The dummy here is necessary since we'll pass tenant and product custom attributes in the client code
+    clientSecret: "dummy", // The dummy here is necessary since we'll pass tenant and product custom attributes in the client code
```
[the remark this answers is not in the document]
deepakprabhakara: @balazsorban44 Please see here - https://github.com/nextauthjs/next-auth/pull/3782#discussion_r817225917. User can opt for a proper clientSecret or the convenience of tenant and product.

packages/next-auth/src/providers/boxyhq-saml.ts
deepakprabhakara: We can drop the null.

docs/docs/providers/boxyhq-saml.md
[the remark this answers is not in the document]
deepakprabhakara: Also this configuration can be set during deployment to change the default `dummy` value - https://boxyhq.com/docs/jackson/deploy/env-variables#client_secret_verifier

deepakprabhakara:
Adding a screenshot of our hosted demo: Screenshot 2022-03-04 at 18 26 07

deepakprabhakara:
> I have to add, the demo provider is wonderful, I wish most providers would have a public example like that 🙏 ✨

Thanks @balazsorban44 🙏, we built the Mock SAML tool specifically for this purpose (and to ease testing as well).

balazsorban44 (approved):
Thank you!

deepakprabhakara:
Many thanks @balazsorban44!

PR #3845, opened 2022-02-06
chore: separate build commands for core and app

ThangHuuVu:
## Reasoning 💡

Separate build command for `next-auth` and `next-auth-app`.

## Checklist 🧢

- ~[ ] Documentation~
- [x] Move `emitDeclarationOnly` to `next-auth` only
- [x] Tests
- [x] Ready to be merged


balazsorban44:
I don't remember the context here. Could you help? I think we should keep the `build` command for CI/CD purposes, but we only need to build the core app and the adapter packages right now, not the docs (it has its own pipeline at Vercel) or the dev app (never published)

ThangHuuVu:
@balazsorban44 You mentioned that it's not very common to build `app`, so I thought this should add some improvements to the pipelines.
> I think we should keep the build command for CI/CD purposes, but we only need to build the core app and the adapter packages right now (later on other framework packages like @next-auth/svelte etc), not the docs (it has its own pipeline at Vercel) or the dev app (never published)

I just updated the build command accordingly 👍

balazsorban44 (approved):
Nice! Thank you!

PR #3872, opened 2022-02-07
docs: add Gatsby example, move Next.js example

balazsorban44:
## Reasoning 💡

This was asked on Twitter by @queen-raae: https://twitter.com/raae/status/1489915309255708673

The proposal aligns well with our intent to explore support for other frameworks #2294, and because of the similarities of the Gatsby API with Next.js ones, this is a good candidate for our first official non-Next.js example.

> NOTE: Even though this is official, I would consider it subject to change, as there might be incompatible cases that I have not considered yet while creating this example. We can address those in the future, but I believe we already cover the most basic usecases.

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged


ndom91 (approved):
LGTM :+1:

Issue #3650, opened 2022-01-17
Move all NextAuth.js projects into a monorepo


balazsorban44:
## History

When NextAuth.js was first created, it had a single Adapter, and no docs. Later on, a documentation page was added, and until v4, it lived under a `www` folder in this repository.

As the need for more adapters appeared, we decided to separate the project into multiple repositories:

Currently we have:

Core: https://github.com/nextauthjs/next-auth
Adapters: https://github.com/nextauthjs/adapters
Docs: https://github.com/nextauthjs/docs
JS Example: https://github.com/nextauthjs/next-auth-example
TS Example: https://github.com/nextauthjs/next-auth-typescript-example

I propose that we move Core, Adapters and Docs into the same repository again, but this time we do it properly as a monorepo.

Adapters is a good starting point, as it is already a monorepo, using Lerna under the hood.

For the new monorepo, [Turborepo](https://turborepo.org/) looks like a very good candidate. It currently does not support versioning, and I propose we have a look at [Changesets](https://github.com/changesets/changesets) as an alternative to `lerna`. I've had difficulties with it on the Adapters repo (mainly because of lack of experience, I admit), and I've heard good things about Changesets.

As for the examples, ideally, we could also move those into the monorepo, but they are currently available as [repository templates](https://docs.github.com/en/repositories/creating-and-managing-repositories/creating-a-repository-from-a-template#about-repository-templates). We could add a cron job that copies our example packages from the monorepo to those repos, so we don't have to maintain them separately.

A reason for this proposal is that although the initial intention was a better separation of bug reports for easier maintenance, most of the time people don't check where they post a bug report, and it has been difficult to keep up with that (I mostly had to manually transfer issues to the correct place, which is tedious).

Further, as we plan to expand to other frameworks (#2294), it might be easier to keep everything under the same repository.

In the long term, the core package should be available as `@next-auth/core` probably, but for now let's keep it as `next-auth`.

With this transition to a monorepo, we could start accepting PRs for other frameworks.

If you have any suggestions, let us know!

### How to reproduce ☕️

-

### Contributing 🙌🏽

Yes, I am willing to help implement this feature in a PR

ThangHuuVu:
To replace `lerna`, it is a question about which features of `lerna` we'd like to keep:
- Installer: candidates are [yarn workspaces](https://yarnpkg.com/) (v1), [pnpm](https://pnpm.js.org)
[...]

PR #3873, opened 2022-02-07
feat: update Firebase adapter to use new API

### LINKED ISSUE #3827: Support for the modular Firebase 9

chanceaclark:
Ports and refactors `adapter-firebase` to use new Adapter API. Ported from this PR: https://github.com/nextauthjs/adapters/pull/183

Breaking Changes:
- Requires Firebase@9+
- Adapter now consumes `FirebaseOptions`
- Additional `emulator` options to connect to Firestore emulator

## Reasoning 💡

Uses new firebase modular + adapter api.

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## Affected issues 🎟

Closes #3827

cc. @balazsorban44


balazsorban44:
I really appreciate it! Once we finish the migration to a monorepo setup #3650, I'll have a look!

My main blocker was to figure out if we could use [Firebase Authentication](https://firebase.google.com/docs/auth)'s [User management](https://firebase.google.com/docs/auth/web/manage-users) somehow. Do you have any ideas?

chanceaclark:
>I really appreciate it! Once we finish the migration to a monorepo setup https://github.com/nextauthjs/next-auth/issues/3650, I'll have a look!

For sure, no worries!

>Do you have any ideas? It would certainly be cleaner for Firebase users, but we could also just use Firestore for now and re-visit this later on.

I was thinking about this for while and tried to take a stab at it myself adding it to this adapter. But I gave up after a little because it was like fitting a square peg in a round hole. Firebase Auth essentially acts similarly like this library – providers in next-auth are similar in nature "Sign-in providers". Potentially, we could make something work in y'alls Provider API but it might not be an all-in-one solution. For example I was thinking of something like `FirebaseEmailProvider`, `FirebaseGoogleProvider`, etc... I'm not too familiar with this library, but what do you think about that?

>I also know people wanted a security rule suggestion, so if you have any, could you add it to the docs at the same time?

Do you know what kind of suggestion people are looking for? Like rules around the `next-auth` created collections (`users`, `accounts`, `sessions` and `verificationTokens`)? Or how to protect resources you create after a user has been authenticated? Also, is there a issue around this that I can get more context?

lukef7fywmrp (approved):

alibidjandy:
HI
excuse me , just wanna regenerate my new job's web app with next.js so I'm going to implement next-auth package for authentication as their authentication on legend old web app implemented with firebase I have to use it on the new  but actually I'm so blind with next-auth and firebase 
could you plz Introduce me an implemented repo somewhere could help me to figure out the logic , in next-auth document couldn't found anything adequate 
https://next-auth.js.org/adapters/firebase#options

chanceaclark:
> HI excuse me , just wanna regenerate my new job's web app with next.js so I'm going to implement next-auth package for authentication as their authentication on legend old web app implemented with firebase I have to use it on the new but actually I'm so blind with next-auth and firebase could you plz Introduce me an implemented repo somewhere could help me to figure out the logic , in next-auth document couldn't found anything adequate https://next-auth.js.org/adapters/firebase#options

Hey @alibidjandy, I would probably bring this question into a GitHub issue or discussion to bring more visibility to it. PR's should be focused more on the code submitted.

But to answer your question from my perspective, I don't have an example for you that I know off the top of my head. Here's a quick [GitHub search](https://github.com/search?l=&q=next-auth+filename%3Apackage.json&type=code) in which you can potentially find some examples.

thiras:
Hello. Is there any ETA for the merge?

chanceaclark:
Hey @thiras, I think we are still kinda waiting for https://github.com/nextauthjs/next-auth/discussions/3952 to be resolved. Not sure what the status is on that, maybe @balazsorban44 or @ndom91 can enlighten us?

balazsorban44:
The monorepo migration has been finished, currently making sure the release pipeline is ready. I'll look into PRs like this when I'm confident about releases. :+1:

Thank you for your patience. In the meantime, please test this adapter by copy-pasting it into your projects and see if it works.

chanceaclark:
You got it @balazsorban44. I'll go ahead and rebase and test it. Also, it seems like one of the CI checks needs approval be a maintainer before running, would you mind checking that out?

DiegoGonzalezCruz:
Hi guys, I just found this and I'm loving it! Sending y'all good energies for a successfull PR!

MarinaMijatovic:
Can this adapter work with firebase realtime database?

ndom91:
> Can this adapter work with firebase realtime database?

This adapter is specifically designed to work with `Firestore` only. I'm curious what use-case you would need a realtime database for when storing user metadata (email, username, etc.)

MarinaMijatovic:
> > Can this adapter work with firebase realtime database?
> 
> This adapter is specifically designed to work with `Firestore` only. I'm curious what use-case you would need a realtime database for when storing user metadata (email, username, etc.)

Sign in should be done by email provider, and the email provider requires adapter to store verification token. The project itself requires users metadata, and tracking current flow of their progress on their project. I'm not sure whether realtime database is the best solution, still researching.

Sal-man:
Hi guys, any ETA on the merge?

ndom91:
> > > Can this adapter work with firebase realtime database?
> > 
> > This adapter is specifically designed to work with `Firestore` only. I'm curious what use-case you would need a realtime database for when storing user metadata (email, username, etc.)
> 
> Sign in should be done by email provider, and the email provider requires adapter to store verification token. The project itself requires users metadata, and tracking current flow of their progress on their project. I'm not sure whether realtime database is the best solution, still researching.

Normal firestore -like database should be fine for this 👍

Sanjay-B:
ETA?

DiegoGonzalezCruz:
I checked the code and is pretty much the configuration of firebase as an adapter.

abrevinci:
Any updates?

ndom91:
There still seem to be some merge conflicts. Once we get those cleaned up I'd be happy to take a closer look and test this 👍

chanceaclark:
> There still seem to be some merge conflicts. Once we get those cleaned up I'd be happy to take a closer look and test this 👍

@ndom91 resolved ✅

ndom91:
@chanceaclark so I wasn't able to get it to run in a dev project, but the tests did all pass.

Next to not being able to get it to run, I had a few small notes:

- tsconfig shuold be extending `tsconfig/adapters.json` not `../base.json`.
- can you fix the names of your npm scripts to remove the `:wip` part?

Then in terms of not getting it to run, I had my editor telling me the new imports didn't work. i.e. `import { getFirestore } from "firebase/firestore"` wasn't exported from "firebase/firestore" and in the console it said `WEBPACK_IMPORTED....getFirestore is not a function`.

I assume this might just be a problem with my local env, however, as that is obviously also how firestore now recommends using it. I cleaned my `node_modules` and reinstalled, making sure I had the newest package.json with the 9.x version checked out, but still no luck :shrug:

I got the same warning in my editor with the `initializeApp` import, however that _seemed_ to work, but then threw an error in the console like this: `FirebaseError: Firebase: Firebase App named '[DEFAULT]' already exists (app/duplicate-app)`.

So in short, I'm new to firebase myself so I'm assuming I might just have done something wrong here. The tests are all passing and if someone else can confirm it works for them I'm happy tentatively merging this, maybe under a "beta" tag or something until we have more confirmation?

Sanjay-B:
@ndom91 I desperately need this for a app that I am working on. While we wait to get this approved, what is a good method to get this into a dev environment? I tried to move the package manually and got errors. Wondering how you got it to work.

ndom91:
> @ndom91 I desperately need this for a app that I am working on. While we wait to get this approved, what is a good method to get this into a dev environment? I tried to move the package manually and got errors. Wondering how you got it to work.

Unfortunately i did not get it to work, as stated in my comment above. You could try something like [`patch-package`](https://npm.im/patch-package)

chanceaclark:
>Next to not being able to get it to run, I had a few small notes:
>
> - tsconfig shuold be extending tsconfig/adapters.json not ../base.json.
> - can you fix the names of your npm scripts to remove the :wip part?

@ndom91 You got it.

>Then in terms of not getting it to run, I had my editor telling me the new imports didn't work. i.e. import { getFirestore } from "firebase/firestore" wasn't exported from "firebase/firestore" and in the console it said WEBPACK_IMPORTED....getFirestore is not a function.

I've had this happen before too. I assume you might be using vscode, if so, I've had to do one/all-of a few of these things:
- Restart your editor (have you tried turning it on and off again 😅 )
- Open the command palette and run `ESLint: Restart ESLint Server`
- Open the command palette and run `TypeScript: Restart TS server`

If those don't work, then I'm not really sure. Maybe there's some turborepo cache that you're pulling from 🤷 

---

So I tested it in my app by copy+pasta the files directly and imported them in my `[...nextauth].ts` file. Here's a screenshot of how I have it setup to work.
Screen Shot 2022-03-22 at 09 25 09

Sanjay-B:
> So I tested it in my app by copy+pasta the files directly and imported them in my [...nextauth].ts file. Here's a screenshot of how I have it setup to work. @chanceaclark 
I did exactly that but, I moved it to a js (not typescript project) and was not able to run it. Do you happen to know a solution to this or it is just something that I have to deal with?

Sanjay-B:
> What sort of errors were you getting @Sanjay-B?

image

The error seems to be ``error - TypeError: Cannot read property 'getProvider' of undefined``. The error seems ambiguous and originates from the module. You can repro this error by converting the ``index.ts`` and ``converter.js`` without a ``tsconfig`` file through ``tsc index.ts`` and ``tsc converter.ts`` via terminal.

DiegoGonzalezCruz:
Hi guys, does this adapter authenticates with Firebase? I was able to write some callbacks to JWT and signIn, but users are unauthenticated. Ergo, I can't set firebase rules or use Storage.
Any lights about it?

mjamado:
Did this issue die?

Sanjay-B:
Alright so, I got this working. 

Here are some steps for some of y'all who want to use / test this:

Download or Clone this: https://github.com/chanceaclark/next-auth/tree/feat/firebase-adapter

Inside ``./packages``, you'll see a bunch of adapters. 
Ref: https://github.com/chanceaclark/next-auth/tree/feat/firebase-adapter/packages

Take ``adapter-firebase`` and place it anywhere but ``node_modules`` in your project. For me, I have a directory called ``temp_modules``. If you place it in ``node_modules``, ``nom install`` or ``npm ci`` will clean or override the package with whatever is stored in npm. Obviously, this not what you want.
Ref: https://github.com/chanceaclark/next-auth/tree/feat/firebase-adapter/packages/adapter-firebase

If you're using TypeScript, it's drag and drop. If you're using JavaScript (like me), it's a bit different.

First, install TypeScript. ``npm install typescript``. Flags are case-by-case so, that's up to you. 
Make sure you have access to ``tsc``. If you don't, it's not in PATH or you need to restart your shell / terminal. 

Inside the ``adapter-firebase`` directory, delete everything but the ``src`` directory. Personally, I usually get rid of the ``src`` directory (not the contents; just the folder) so that there are no directories and just the sole files that I need. However, it really doesn't matter so, do whatever you feel is right. In this case, ``index.ts`` and ``converter.ts`` are still in the ``src`` directory.

Next, create a ``tsconfig.json`` file inside your ``adapter-firebase`` directory. 

Add the following contents:
```json
{
  "compilerOptions": {
    "target": "es5",
    "module": "commonjs",
    "sourceMap": true
  }
}
```

Now, open up terminal or a shell, ``cd`` into wherever ``adapter-firbase`` is and run the following:
``tsc src/index.ts``
``tsc src/converter.ts``

This transpiles TypeScript code to JavaScript. From there on out, we're able to use said methods and whatnot you'd typically do. One major thing to note, there's a setting in ``next.config.json`` that breaks this. It's experimental but, for some reason it's turned on by default. We want to disable it. This is what mine looks like:
```json
/** @type {import('next').NextConfig} */
const nextConfig = {
  reactStrictMode: true,
  experimental: {
    esmExternals: false
  }
}

module.exports = nextConfig
```
``esmExternals`` causes it to break. Disabling it fixes this issue.

So far, no issues. Here are some pictures if you're confused or you're a visual learner:

Screen Shot 2022-04-06 at 2 07 30 PM

Screen Shot 2022-04-06 at 2 08 32 PM

Best of luck out there!

abrevinci:
@Sanjay-B your solution worked for me.

chanceaclark:
> Did this issue die?

Sorry, should've communicated better... I've had a pretty hectic couple of weeks at my day job plus been taking some holiday. I can take a better look at this sometime in the next 2 weeks 😬

mjamado:
> > Did this issue die?
> 
> Sorry, should've communicated better... I've had a pretty hectic couple of weeks at my day job plus been taking some holiday. I can take a better look at this sometime in the next 2 weeks grimacing

That's ok, just making sure it wasn't dead in the water. I had to decide if I keep on waiting on a merge, copy it @Sanjay-B style, or abandon this path altogether. I guess I'll wait, then.

DiegoGonzalezCruz:
Hi for everyone wondering about firebase rules and this adapter, I wrote this gist as an example.
[https://gist.github.com/DiegoGonzalezCruz/0f75b1548b6122da9fd2e78b257c9d40](url)

atanaskanchev:
> Hi for everyone wondering about firebase rules and this adapter, I wrote this gist as an example.
> 
> [https://gist.github.com/DiegoGonzalezCruz/0f75b1548b6122da9fd2e78b257c9d40](url)

Hi Diego, the gist appears private

DiegoGonzalezCruz:
> 

fixed it

nibori:
Hi guys, this feature would be very useful! Any updates for the coming merge?

ndom91:
> > 
> 
> fixed it

Hey @DiegoGonzalezCruz do you mind making your gist public so that others can check it out?

chanceaclark:
Hey y'all, I'm back on this for a quick minute (onto another holiday next week, but I'll be back for good after as far as the eye can see).

So I ended up changing the api a bit... for a few reasons.
1. You should now just be able to pass in the firebase config and the adapter should do the rest for you. *The big issue is that after building the files, the app/firestore instance wasn't binding to the adapter.* I didn't have enough time to fully dig into this, and I'm no expert in build tools, so this was the best solution I could come up with.
2. You can now pass in `emulator` settings if you need to emulate firestore. I couldn't pass the app/firestore instance into the adapter anymore so I enable this as a config to attach to your local emulated firestore instance.

If you have any better alternatives let me know. I will go ahead and update the PR description with this info.

Just a note...

Looking at the overall intention of this adapter, it's not really a "Firebase" adapter, but rather a "Firestore" one. The name of it may be confusing to some people, thinking it'll work with the Firebase Auth.

lowfront:
@chanceaclark @balazsorban44 Hello, I confirmed that to protect firebase data, sign-in is required as a user with data rights registered in firebase authentication. So I made an firestore adapter example that works by either signing in to the firebase before the firebase adapter is run, or signing in when it is run, [Example](https://github.com/lowfront/next-auth-firebase-adapter-suggestion)

What do you think of an approach that allows you to optionally include authorized user sign-in information as shown in the example?

Sanjay-B:
> @chanceaclark @balazsorban44 Hello, I confirmed that to protect firebase data, sign-in is required as a user with data rights registered in firebase authentication. So I made an firestore adapter example that works by either signing in to the firebase before the firebase adapter is run, or signing in when it is run, [Example](https://github.com/lowfront/next-auth-firebase-adapter-suggestion)
> 
> What do you think of an approach that allows you to optionally include authorized user sign-in information as shown in the example?
> 
> Firebase sign-in provider includes email/password, phone, OAuth, etc. Other providers except Email/Password require user input every time they sign-in. So I think Email/Password provider is the best way to use it for server.

I'm a bit confused by this. The purpose of the adapter iirc is meant to allow sign-ins to be stored via firebase. This means that your auth tokens, etc are tied to your provider which are then stored into firestore as a session. For example, I am using GoogleProvider where I have a full oauth-based sign in. User signs in with a valid google email and nextauth handles the rest for me via the adapter (in this case firebase).

lowfront:
> > @chanceaclark @balazsorban44 Hello, I confirmed that to protect firebase data, sign-in is required as a user with data rights registered in firebase authentication. So I made an firestore adapter example that works by either signing in to the firebase before the firebase adapter is run, or signing in when it is run, [Example](https://github.com/lowfront/next-auth-firebase-adapter-suggestion)
> > What do you think of an approach that allows you to optionally include authorized user sign-in information as shown in the example?
> > Firebase sign-in provider includes email/password, phone, OAuth, etc. Other providers except Email/Password require user input every time they sign-in. So I think Email/Password provider is the best way to use it for server.
> 
> I'm a bit confused by this. The purpose of the adapter iirc is meant to allow sign-ins to be stored via firebase. This means that your auth tokens, etc are tied to your provider which are then stored into firestore as a session. For example, I am using GoogleProvider where I have a full oauth-based sign in. User signs in with a valid google email and nextauth handles the rest for me via the adapter (in this case firebase). All I have to do is track the session if it's a protected route. Otherwise, it's the same treatment as any other public-facing page.

@Sanjay-B That's right, if the firebase is used only on the server side, the session data stored in next-auth is protected without any problems.

However, firebase can also be used on the frontend. Suppose this firebase project is the same project as the one saved in next-auth and there are no rules (for convenience, I think this is how it can be configured). Anyone other than the author can access session, account, and user data stored in next-auth, if they only know the key used to store it in the firebase in next-auth.

So my example is a configuration for firebase to be used without security risk in both backend and frontend.
If you block the reading of key subdata stored in next-auth through the rule in firebase, you can protect it, but the rule also applies to the server side, making it impossible to read from the adapter.
What I know is that the rule can give you permission is Firebase authentication or any string token. Among them, I thought Firebase authentication was more suitable.

What do you think about it?

ndom91:
I think a lot of the confusion here comes from what @chanceaclark alluded to, the difference of "Firebase (auth)" adapter vs. "Firestore" adapter. 

This is specifically only supposed to be an adapter which uses Firebase's **Firestore** as a data-layer to do the same thing the Prisma, MongoDB, etc. adapters do for next-auth, which is save user data to tables.

I think we should focus on that and only that for now with this adapter, and potentially rename it to "Firestore Adapter", as has been suggested as well, to avoid further confusion.

If there is enough demand, and there seems to be, maybe someone can then begin a separate adapter to leverage Firebase Auth together with next-auth. I'm not sure what the requirements there would be though as it would be rather different than the typical DB based next-auth adapters. However, an adapter is nothing more than a set of about 10-15 predefined methods (like `createUser`, `updateUser`, `linkAccount`, etc.), so if Firebase Auth exposes an API so that those methods could all be implemented it seems possible :+1:

chanceaclark:
>This is specifically only supposed to be an adapter which uses Firebase's Firestore as a data-layer to do the same thing the Prisma, MongoDB, etc. adapters do for next-auth, which is save user data to tables.

I think this is crucial context. A tl'dr from my understanding that the `next-auth` library addresses is that `adapters` = `databases`, whereas `providers` = auth providers (e.g. Firebase Auth, Google, Github, etc...). It's weird with Firebase Auth though since you can have more providers through them.

>I think we should focus on that and only that for now with this adapter, and potentially rename it to "Firestore Adapter", as has been suggested as well, to avoid further confusion.

We can rename this in another PR if we want. It would be a breaking change. Or we can throw a `@deprecated` tag on `FirebaseAdapter` and re-export this as `FirestoreAdapter` 🤔  e.g...
```ts
const adapter = (...) = {...}
// ...
/**
 * @deprecated Adapter name will be deprecated, use `FirestoreAdapter` export instead.
 */
export const FirebaseAdapater = adpater;

export const FirestoreAdapter = adapter;
```

>If there is enough demand, and there seems to be, maybe someone can then begin a separate adapter to leverage Firebase Auth together with next-auth. I'm not sure how hard that would be though as it would be rather different than the typical DB based next-auth adapters and I have very little experience with Firebase Auth myself.
>
>However, an adapter is nothing more than an implementation of a set of about 10-15 methods (like createUser, updateUser, linkAccount, etc.), so if the Firebase Auth SDK exposes an API so that those methods could all be implemented it seems possible 👍

I think based off our these replies, and my understanding of the repo, a Firebase Auth implementation would be better suited for being a provider, not adapter. I mentioned something about [this above](https://github.com/nextauthjs/next-auth/pull/3873#issuecomment-1032879199). I don't have enough experience with Firebase Auth, but looking at the APIs I'm not sure they have the appropriate methods to build off of the adapter implementation (I could be wrong though).

Anyways, I think we should pull this into another discussion and not here so we can reserve this PR for code comments.

ndom91:
> >This is specifically only supposed to be an adapter which uses Firebase's Firestore as a data-layer to do the same thing the Prisma, MongoDB, etc. adapters do for next-auth, which is save user data to tables.
> 
> I think this is crucial context. A tl'dr from my understanding that the `next-auth` library addresses is that `adapters` = `databases`, whereas `providers` = auth providers (e.g. Firebase Auth, Google, Github, etc...). It's weird with Firebase Auth though since you can have more providers through it.
> 
> >I think we should focus on that and only that for now with this adapter, and potentially rename it to "Firestore Adapter", as has been suggested as well, to avoid further confusion.
> 
> We can rename this in another PR if we want. It would be a breaking change. Or we can throw a `@deprecated` tag on `FirebaseAdapter` and re-export this as `FirestoreAdapter` 🤔  e.g...
> ```ts
> const adapter = (...) = {...}
> // ...
> /**
>  * @deprecated Adapter name will be deprecated, use `FirestoreAdapter` export instead.
>  */
> export const FirebaseAdapater = adpater;
> 
> export const FirestoreAdapter = adapter;
> ```
> 
> >If there is enough demand, and there seems to be, maybe someone can then begin a separate adapter to leverage Firebase Auth together with next-auth. I'm not sure how hard that would be though as it would be rather different than the typical DB based next-auth adapters and I have very little experience with Firebase Auth myself.
> >
> >However, an adapter is nothing more than an implementation of a set of about 10-15 methods (like createUser, updateUser, linkAccount, etc.), so if the Firebase Auth SDK exposes an API so that those methods could all be implemented it seems possible 👍
> 
> I think based off our these replies, and my understanding of the repo, a Firebase Auth implementation would be better suited for being a provider, not adapter. I mentioned something about [this above](https://github.com/nextauthjs/next-auth/pull/3873#issuecomment-1032879199). I don't have enough experience with Firebase Auth, but looking at the APIs I'm not sure they have the appropriate methods to build off of the adapter implementation (I could be wrong though).
> 
> Anyways, I think we should pull this into another discussion and not here so we can reserve this PR for code comments.

Hey thanks for the thorough reply!

Yeah so a rename would definitely be a breaking change, your right. But what do you think, worth it maybe in the long run to avoid more confusion down the road? 

Classically adapters have been strictly databases, but technically they could be anything that can fulfill those required methods, right. I'm not very experienced with firebase generally and firebase auth more specifically, but last time balazs and I looked into it, as far as I remember there were a few api methods missing from firebase auth as well in order to fulfill all required methods of a nextauth adapter

Sanjay-B:
> > > @chanceaclark @balazsorban44 Hello, I confirmed that to protect firebase data, sign-in is required as a user with data rights registered in firebase authentication. So I made an firestore adapter example that works by either signing in to the firebase before the firebase adapter is run, or signing in when it is run, [Example](https://github.com/lowfront/next-auth-firebase-adapter-suggestion)
> 
> > > What do you think of an approach that allows you to optionally include authorized user sign-in information as shown in the example?
> 
> > > Firebase sign-in provider includes email/password, phone, OAuth, etc. Other providers except Email/Password require user input every time they sign-in. So I think Email/Password provider is the best way to use it for server.
> 
> > 
> 
> > I'm a bit confused by this. The purpose of the adapter iirc is meant to allow sign-ins to be stored via firebase. This means that your auth tokens, etc are tied to your provider which are then stored into firestore as a session. For example, I am using GoogleProvider where I have a full oauth-based sign in. User signs in with a valid google email and nextauth handles the rest for me via the adapter (in this case firebase). All I have to do is track the session if it's a protected route. Otherwise, it's the same treatment as any other public-facing page.
> 
> 
> 
> @Sanjay-B That's right, if the firebase is used only on the server side, the session data stored in next-auth is protected without any problems.
> 
> 
> 
> However, firebase can also be used on the frontend. Suppose this firebase project is the same project as the one saved in next-auth and there are no rules (for convenience, I think this is how it can be configured). Anyone other than the author can access session, account, and user data stored in next-auth, if they only know the key used to store it in the firebase in next-auth.
> 
> 
> 
> So my example is a configuration for firebase to be used without security risk in both backend and frontend. If you block the reading of key subdata stored in next-auth through the rule in firebase, you can protect it, but the rule also applies to the server side, making it impossible to read from the adapter. What I know is that the rule can give you permission is Firebase authentication or any string token. Among them, I thought Firebase authentication was more suitable. The email and password that I put in as environment variables are to pass the rule of firebase unlike Google OAuth.
> 
> 
> 
> What do you think about it?

This is an interesting use-case. I think that it's possible and valid to have a username-password based login but, personally, I'm more inclined to move towards GoogleProvider when using Firebase-Adapter. Both services are maintained by Google so, both have consistencies that make development and overall integration easier. 

That being said, this is all subjective and how you go about your solution is purely up to you.

Sanjay-B:
> > ... 
> > I think based off our these replies, and my understanding of the repo, a Firebase Auth implementation would be better suited for being a provider, not adapter. I mentioned something about [this above](https://github.com/nextauthjs/next-auth/pull/3873#issuecomment-1032879199). I don't have enough experience with Firebase Auth, but looking at the APIs I'm not sure they have the appropriate methods to build off of the adapter implementation (I could be wrong though).
> > 
> > Anyways, I think we should pull this into another discussion and not here so we can reserve this PR for code comments.
> 
> Hey thanks for the thorough reply!
> 
> Yeah so a rename would definitely be a breaking change, your right. But what do you think, worth it maybe in the long run to avoid more confusion down the road? It would obviously have to be planned thoroughly. 
> 
> Classically adapters have been strictly databases, but technically they could be anything that can fulfill those required methods, right. I'm not very experienced with firebase generally and firebase auth more specifically, but last time balazs and I looked into it, as far as I remember there were a few api methods missing from firebase auth as well in order to fulfill all required methods of a nextauth adapter. I think you're right there as well 👍 

I think that this should stay just a firebase-adapter. First & foremost, the firebase node module (a dep for this) is maintained by Google and provides a very nice API along with docs. Secondly, the purpose of this seems to be an adapter strictly to integrate a session with a database flavor. Moving outside that purpose trying to essentially fulfill what the firebase does seems like a pointless effort. Especially when, it adds time to this, is a breaking change and may take time to document before it's considered useful. My philosophy is really, "Hey don't break stuff unless you absolutely need to" and personally, there's really no need outside of a few edge cases that warrant a breaking change.

lowfront:
@Sanjay-B  Firebase can be used at the front-end and it is more common. [link](https://stackoverflow.com/a/54389420) explains the reason well. Therefore, I think it is necessary to know firebase auth and rule. I have some experience and can explain Firebase auth. The provider for Firebases auth is a provider for firebase services only to access data in the database (realtime database or firestore) of Firebase. No new provider can be added in addition to the limited provider provided by firebase auth. This is known as a disadvantage of firebase service. 

> I think based off our these replies, and my understanding of the repo, a Firebase Auth implementation would be better suited for being a provider, not adapter. I mentioned something about https://github.com/nextauthjs/next-auth/pull/3873#issuecomment-1032879199. I don't have enough experience with Firebase Auth, but looking at the APIs I'm not sure they have the appropriate methods to build off of the adapter implementation (I could be wrong though).

@chanceaclark  So the connection between next-auth's provider and firebase auth is impossible as far as I know. Because the provider list of next-auth and firebase-auth are different and the purpose is different. The provider in next-auth provides a way to log in to the service I created, and the provider in firebase auth provides a way to log in to the firebase service to obtain permission to store data in the firebase database. In other words, the provider in firebase auth gives you permission to store the login information generated by the next-auth provider in the firebase

If the OAuth provider supported by Firebase auth is enough, it would be more convenient to build it with Firebase service alone. But I chose next-auth because I wanted more providers in my service. What I am curious about is this. How do you set the rule of firebase and use it? Considering that there is no authentication procedure through Firebase auth in the code you wrote, it seems that you are using a rule that allows reading and writing to all data, is that right? Firebase service does not recommend it. When Firebase service creates a new database, it creates all data disclosure rules, which are time-limited when in test mode, but all data is private when in production mode.

I decided to make an adapter with the intention of using firebase to store the data of next-auth. And I wanted to read and write data at the front-end using the method of firebase SDK for convenience, instead of rewriting all CRUD APIs in nextjs. But I'm in trouble here. That is, if the data received through next-auth's provider and the data used by the front end are stored in the same database and allowed to read and write to all data, the person who finds the firebaseConfig in the front end code will be able to find all session tokens and all user information stored in the session.

So I used firebase-auth to create an example to protect the data stored in next-auth. Of course, the way I wrote it may not be the best way. I'm still thinking about finding a natural way to use it on the frontend. However, considering that firebase services can also be commonly used at the front-end, I think that even if don't handle Firebase-auth, at least this adapter should give a reminder to use it only on the server for data protection.

Please let me know if there is anything wrong with me. Or let me know if there's anything you don't understand. Or if this seems to be a hasty optimization, please let me know your opinion. I left an opinion here because I thought this was indispensable in the case of firebase use, but if you think differently, I will propose a new PR for firebase-adapter including firebase-auth later.

gianmarcodelfreo:
Is there any news on the release date of this adapter?

atanaskanchev:
> Is this PR still active?

It has been more than 3 months, I'm starting to lose hope 😕

chanceaclark:
Hey y'all,

There's a lot discussed in this PR and I don't want this to be lingering on. **The intention for this PR was to update the implementation to the newest version of Firebase libraries and to refactor to the newest adapter API.**

### `FirebaseAdapter` -> `FirestoreAdapter`
I added a commit to this PR, essentially adding a `@deprecated` tag to the `FirebaseAdapter` export and added a new `FirestoreAdapter` export. Along with that, I posted https://github.com/nextauthjs/next-auth/discussions/4592 to discuess that there. I'll go with what @Sanjay-B and @ndom91 recommend and can remove that commit if we need to discuss it more.

### Security Rules
I would love to tackle this, but it seems a bit out of scope for the intention of this PR. I believe there are already some discussions already posted to talk about this, so let's continue that there.

### Firebase Auth
Let's pull this into a discussion since `next-auth` and Firebase Auth are pretty similar in nature. Trying to make Firebase Auth work in `next-auth` is like using an extra tool for the same job, at least how I think about it.

ndom91:
Hey y'all, just a quick status update - we've been discussing this internally lately. We really want to move forward with it as soon as possible. Just don't have a ton of time to work on it ourselves at the moment. 

But don't despair, we've not forgotten it and want to move forward with it too!

Sanjay-B:
If it helps any, I've been using for a while now and it works.

adrianvanlan:
> If it helps any, I've been using for a while now and it works. That being said, it's great to see movement on this.

Can you provide a working example? I've tried it and I don't know what I'm doing wrong

fotoflo:
> > Hi for everyone wondering about firebase rules and this adapter, I wrote this gist as an example.
> > [https://gist.github.com/DiegoGonzalezCruz/0f75b1548b6122da9fd2e78b257c9d40](url)
> 
> Hi Diego, the gist appears private

im still getting a 404

packages/adapter-firebase/src/index.ts
```diff
@@ -1,283 +1,250 @@
...
+    port?: number;
+  };
 }
 
-export type FirebaseSession = Session & {
-  id: string
-  expires: Date
-}
+/**
+ * @deprecated Use FirestoreAdapter export insteader
```
ndom91, line 32:
```suggestion
 * @deprecated Use FirestoreAdapter export instead
```

chanceaclark: 🤦

chanceaclark:
@ndom91 went ahead and updated the jest config to fix that previous broken workflow

ndom91:
Unfortunately I can't get the dev setup up and running to test this with a dummy firebase account. I don't think its got to do with the firebase adapter though. Seems to be an unrelated problem on my end. 

@Sanjay-B said he's been using this PR for a while now already, what's your experience with it? Any successes actually using it @chanceaclark ?

chanceaclark:
Been super successful using it! Also, just in case, I tested it out by pullings y'alls [example repo](https://github.com/nextauthjs/next-auth/tree/main/apps/example-nextjs), `yarn link`'ed my local repo, installed firestore locally, and ran the emulator.

Screen Shot 2022-06-23 at 16 05 01

Also, here's a diff of what I did in the example repo (although it doesn't take into account the `yarn link`): https://gist.github.com/chanceaclark/2b8cb408429cab20a787b11cdbb1e191

ndom91:
Okay awesome, after Balazs's recent PR updating a lot of the dev dependencies, I was able to get it to run locally as well! Connected and worked right away with the local firebase emulator. 

Got close to it working with an actual test firebase app, however, I also kept getting errors about "missing permissions" to write to the DB. I'm assuming this is to do with the security rules that have been mentioned here before? I don't know though, not familiar with firebase at all really :/

chanceaclark:
>Maybe the Firebase setup instructions could use a bit of improvement? Or maybe they are just basic Firebase settings that any experienced Firebase user would know, that I just happened to miss. What do you think?

Good point! I think for the most part it's having a lot of Firebase knowledge, but I  do see that there might be some gaps in the docs. Let me touch those up just a bit and rebase.

chanceaclark:
@ndom91 I don't have access to see why that Vercel deployment is failing. Mind clueing me in?

chanceaclark:
Ah looks like it's an issue introduced in https://github.com/nextauthjs/next-auth/pull/4757 should be no cause for concern. Everything should be good to go on my end. Maybe review it once more and let me know if anything else needs to be looked at.

kamami:
When will this be available and documented? :)

ndom91:
We fixed the docs failed deploy, should be good to go now. 

It looks good to me, will try to get the teams support to merge this as soon as possible!

balazsorban44:
Thanks for the PR, and sorry for the delay! I'm thinking renaming to `FirestoreAdapter` is not a good idea, since the package is still called `@next-auth/firebase-adapter`. Won't that be confusing? :thinking:

Do we see a future with a different Firebase-based adapter? Should we either:
1. release `@next-auth/firestore-adapter`
2. add this in a submodule `@next-auth/firebase-adapter/firestore`
3. Let the export name be `FirebaseAdapter`

chanceaclark:
> Thanks for the PR, and sorry for the delay! I'm thinking renaming to `FirestoreAdapter` is not a good idea, since the package is still called `@next-auth/firebase-adapter`. Won't that be confusing? 🤔

I do agree with you that it is a little confusing, but I will say it's already confusing as is since this adapter is purely built for Firestore. We've seen the many comments of _when Firebase Auth will be added to this adapter_ and we probably want to make it clear what this adapter actually supports: Firestore.
 
> Do we see a future with a different Firebase-based adapter? Should we either:
> 
> 1. release `@next-auth/firestore-adapter`
> 2. add this in a submodule `@next-auth/firebase-adapter/firestore`
> 3. Let the export name be `FirebaseAdapter`

Unless I'm missing some context y'all might have, the only thing I potentially see is exporting a Firebase Realtime DB adapter. However, I've never worked with it so I don't know if it's even feasible/wanted. In regards to those options:

1. I think that would be a lot of work and it's up to y'all if you want to go down that path of deprecating `@next-auth/firebase-adapter` – IMO I would leave it as is.
2. Submodules is a fine approach, but you could also just export additional modules along with it if you are worried about namespacing them e.g. `import { FirestoreAdapter, RealtimeAdapter } from '@next-auth/firebase-adapter` 🤷 
3. I assume this is just revert the last commit just leave it as `FirebaseAdapter`.

My recommendation is the 2nd option I proposed, **but at the end of the day it's your call. If you don't want to deal with the renaming quite yet, I'll rip it out, just give me the word.** We can continue the discussion in #4592.

> Apart from that, I had some code nitpicks, which I'll hold to myself for now, and can do in a follow-up PR, you have done great work here and I don't want to unnecessarily prolong the merge of this. 😅

I appreciate it! I also don't mind tackling them if need be, just let me know.

cdriesler:
Since we're right at the finish line I want to jump in - 

@chanceaclark do you have a sponsorship link somewhere, GitHub or otherwise? I've been eagerly following this PR for a while now. It'll solve a pile of problems for me, and I'd like to buy you a coffee-or-two as thanks for the months of work.

chanceaclark:
I appreciate the offer @cdriesler! However, I don't have anything setup for sponsorship currently and I'm doing this for the fun of it 😄 Maybe in the future, but for right now, just giving back to the community 🙇‍♂️

packages/adapter-firebase/src/index.ts
```diff
@@ -1,283 +1,250 @@
-import type firebase from "firebase"
-import { createHash, randomBytes } from "crypto"
-import { Adapter } from "next-auth/adapters"
+import { FirebaseOptions, initializeApp } from 'firebase/app';
```
balazsorban44, line 1: let's use separate imports for types (`import type`) where possible

packages/adapter-firebase/src/index.ts
```diff
@@ -1,283 +1,250 @@
...
+  connectFirestoreEmulator
+} from 'firebase/firestore';
+import { Account } from 'next-auth';
+import { Adapter, AdapterSession, AdapterUser, VerificationToken } from 'next-auth/adapters';
+
+import { getConverter } from './converter';
+
+type IndexableObject = Record<string, unknown>;
+
+interface FirestoreAdapterOptions {
```
balazsorban44, line 24:
```suggestion
export interface FirestoreAdapterOptions {
```

packages/adapter-firebase/src/index.ts
```diff
@@ -1,283 +1,250 @@
...
 }
 
-export type FirebaseSession = Session & {
-  id: string
-  expires: Date
-}
+/**
+ * @deprecated Use FirestoreAdapter export instead
+ */
+export const FirebaseAdapter = ({ emulator, ...firebaseOptions }: FirebaseOptions & FirestoreAdapterOptions): Adapter => {
```
balazsorban44, line 34: Since we are doing a breaking change, let's just drop it.

```suggestion
/**
 * @deprecated Use FirestoreAdapter export instead
 */
export function FirestoreAdapter({ emulator, ...firebaseOptions }: FirebaseOptions & FirestoreAdapterOptions): Adapter {
```

packages/adapter-firebase/src/index.ts
```diff
@@ -1,283 +1,250 @@
...
+
+        return verificationToken;
+      }
+
+      return null;
+    },
+  };
+};
+
+export const FirestoreAdapter = FirebaseAdapter;
```
balazsorban44, line 250:
```suggestion
```

packages/adapter-firebase/src/index.ts
```diff
@@ -1,283 +1,250 @@
...
-          await client
-            .collection("verificationRequests")
-            .doc(verificationRequest.id)
-            .delete()
-        },
+    async createUser(newUser) {
+      const userRef = await addDoc(Users, newUser);
+      const userSnapshot = await getDoc(userRef);
+
+      if (userSnapshot.exists() && Users.converter) {
```
balazsorban44, line 54: I see these checks from here and down. In which case would `.converter` not be defined?

packages/adapter-firebase/tests/index.test.ts
```diff
@@ -1,67 +1,77 @@
...
-).firestore()
-firestore.useEmulator("localhost", 8080)
+const app = initializeApp({ projectId: "next-auth-test" });
+const firestore = getFirestore(app);
+
+connectFirestoreEmulator(firestore, 'localhost', 8080);
+
+type IndexableObject = Record<string, unknown>;
+
+const Users = collection(firestore, 'users').withConverter(getConverter<AdapterUser>());
```
balazsorban44, line 17: Does it make sense to test `getConverter` before being used in tests themselves? We potentially test the adapter against itself which makes tests less reliable.

packages/adapter-firebase/src/index.ts
chanceaclark: Technically, yeah but TypeScript doesn't know that 😂

balazsorban44: :shushing_face:

adrianvanlan:
When will this version be available in npm?

dellwatson:
i just created new project and now have problem with firebaseAdapter:
` name: 'GetUserByAccountError` . it seems working now without next-auth, i suppose must be related to this?

DanielBailey-web:
> My main blocker on [my original PR](https://github.com/nextauthjs/adapters/pull/183) was to figure out how could we integrate with [Firebase Authentication](https://firebase.google.com/docs/auth)'s [User management](https://firebase.google.com/docs/auth/web/manage-users).

Is this part of a new issue?

PR #3874, opened 2022-02-07
chore: add v4 migration script for adapters

ndom91:
## Reasoning 💡

Initial attempt at some DB migration scripts in the `v4 upgrade` docs. For migrating adapters created under v3 to the new v4 schema.

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged


packages/docs/docs/getting-started/upgrade-to-v4.md
```diff
@@ -439,6 +439,106 @@ See the [`session` option docs](https://next-auth.js.org/configuration/options#s
...
+DROP COLUMN "provider_type"
+DROP COLUMN "compound_id"
+DROP COLUMN "created_at"
+DROP COLUMN "updated_at"
+
+ADD COLUMN "token_type" varchar(191) NULL
+ADD COLUMN "scope" varchar(191) NOT NULL
+ADD COLUMN "id_token" varchar(191) NOT NULL
+ADD COLUMN "oauth_token_secret" varchar(191) NOT NULL
+ADD COLUMN "oauth_token" varchar(191) NOT NULL
```
balazsorban44, line 467: So these are only needed if they are using Twitter OAuth 1.0. We should comment and link to the docs. Otherwise these can be dropped.

packages/docs/docs/getting-started/upgrade-to-v4.md
```diff
@@ -439,6 +439,106 @@ See the [`session` option docs](https://next-auth.js.org/configuration/options#s
...
+ALTER TABLE accounts
+CHANGE "access_token_expires" "expires_at" int
+CHANGE "user_id" "userId" varchar(191)
+ADD CONSTRAINT fk_user_id FOREIGN KEY (user_id) REFERENCES users(id)
+RENAME COLUMN "provider_id" "provider"
+RENAME COLUMN "provider_account_id" "providerAccountId"
+DROP COLUMN "provider_type"
+DROP COLUMN "compound_id"
+DROP COLUMN "created_at"
+DROP COLUMN "updated_at"
```
balazsorban44, line 461: Let's comment that these have never been necessary for NextAuth.js to work, hence they are dropped, but can be kept if they want to.

packages/docs/docs/getting-started/upgrade-to-v4.md
```diff
@@ -439,6 +439,106 @@ See the [`session` option docs](https://next-auth.js.org/configuration/options#s
...
+ADD COLUMN "scope" varchar(191) NOT NULL
+ADD COLUMN "id_token" varchar(191) NOT NULL
+ADD COLUMN "oauth_token_secret" varchar(191) NOT NULL
+ADD COLUMN "oauth_token" varchar(191) NOT NULL
+ADD COLUMN "session_state" varchar(191) NOT NULL
+
+/* USER */
+ALTER TABLE users
+DROP COLUMN "created_at"
+DROP COLUMN "updated_at"
```
balazsorban44, line 473: Same as https://github.com/nextauthjs/next-auth/pull/3874/files#r800927573

packages/docs/docs/getting-started/upgrade-to-v4.md
```diff
@@ -439,6 +439,106 @@ See the [`session` option docs](https://next-auth.js.org/configuration/options#s
...
+RENAME COLUMN "email_verified" "emailVerified"
+
+/* SESSION */
+ALTER TABLE sessions
+RENAME COLUMN "session_token" "sessionToken"
+CHANGE "user_id" "userId" varchar(191)
+ADD CONSTRAINT fk_user_id FOREIGN KEY (user_id) REFERENCES users(id)
+DROP COLUMN "access_token"
+DROP COLUMN "created_at"
+DROP COLUMN "updated_at"
```
balazsorban44, line 483: https://github.com/nextauthjs/next-auth/pull/3874/files#r800927573

packages/docs/docs/getting-started/upgrade-to-v4.md
```diff
@@ -439,6 +439,106 @@ See the [`session` option docs](https://next-auth.js.org/configuration/options#s
...
+
+/* VERIFICATION REQUESTS */
+ALTER TABLE verification_requests
+DROP COLUMN id
+DROP COLUMN "created_at"
+DROP COLUMN "updated_at"
```
balazsorban44, line 489: https://github.com/nextauthjs/next-auth/pull/3874/files#r800927573

balazsorban44 (changes requested):
Only commented on the MySQL one, but should be the same for Postgres

TheoBr:
Looks good overall, really appreciate the separate examples for MySQL and Postgres 

Two concerns:
- “Rename” isn’t an option for a lot of sharded dbs [like Planetscale](https://github.com/planetscale/beta/discussions/122)
- by making multiple new fields non nullable, you're likely breaking any existing entries in those tables

ndom91:
Thanks for the feedback everyone! I've addressed the comments so far. 

Regarding `RENAME` not working with sharded DBs like planetscale, I think that basically rules out using an automated migration like this, right. In cases like those, if the user wants to upgrade from v3 -> v4, they'll have to manually adjust the DB if possible or potentially just create a new one and move the data over. I know that sucks, but I don't see another option if `RENAME` isn't available. 

What do yall think?

TheoBr:
```
  type               String   @map("providerType")
  refresh_token      String?  @map("refreshToken") @db.Text
  provider           String   @map("providerId")
```
^ is the fix we're using at Ping for now

ndom91:
Yeah just remap the names in Prisma. Makes sense. For those using prisma this seems like a nice workaround, will keep it in mind!

TheoBr:
My other concern was w/ the non-nullable bindings to `expires_at` and `session_state`. Neither seem to be used, this screenshot is from a minimal production app on nextauthjs v4

image

ndom91:
@TheoBr yeah I changed those to no longer be non-nullable, sticking to the previous implementation. New v4 setups can keep the non-nullable variants going forward, but for migrations you're right - that would probably break some current implementations :+1: 

Or do you mean to allow NULL on `created_at` + `updated_at` ?

balazsorban44 (approved):
LGTM :heavy_check_mark:

Issue #2509, opened 2021-08-09
NextAuth not working behind corporate proxy


raphaelpc:
### Description 🐜

NextAuth is not working behind corporate proxy.

I studied the current NextAuth implementation and found that this error happens because NextAuth makes use of the "node-auth" library (npm package "oauth") in it's "oAuthClient" (src/server/lib/oauth/client.js), which requires manually setting up an agent to be used by the "https" library.

I created a PR with the solution of the problem:
https://github.com/nextauthjs/next-auth/pull/2493

Related discussion: https://github.com/nextauthjs/next-auth/discussions/676
Related StackOverflow: https://stackoverflow.com/questions/32130471/node-js-https-not-working-behind-corporate-proxy

If the PR is not going to be accepted, i believe it is important do register the issue here for other people facing the same problems that i'm facing right now (which are quite time consuming to debug).

### Is this a bug in your own project?

No

### How to reproduce ☕️

Just run any application with NextAuth behind a corporate proxy and try to log in with a Provider (like Google or GitHub).

### Screenshots / Logs 📽

```
[next-auth][error][oauth_get_access_token_error] 
https://next-auth.js.org/errors#oauth_get_access_token_error Error: connect ETIMEDOUT 216.58.202.141:443
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1146:16) {
  errno: -4039,
  code: 'ETIMEDOUT',
  syscall: 'connect',
  address: '216.58.202.141',
  port: 443
} undefined undefined
[next-auth][error][oauth_get_access_token_error]
https://next-auth.js.org/errors#oauth_get_access_token_error Error: connect ETIMEDOUT 216.58.202.141:443
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1146:16) {
  errno: -4039,
  code: 'ETIMEDOUT',
  syscall: 'connect',
  address: '216.58.202.141',
  port: 443
} google 4/0AX4XfWiDMdHUjXWR1xqLs8bGa1g5CEJOdBiypyNPA3b4vzJHeqxOGFownSkn5ABA885Asw
[next-auth][error][oauth_callback_error]
https://next-auth.js.org/errors#oauth_callback_error Error: connect ETIMEDOUT 216.58.202.141:443
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1146:16) {
  errno: -4039,
  code: 'ETIMEDOUT',
  syscall: 'connect',
[...]

PR #3900, opened 2022-02-09
chore(docs): add tutorial for avoiding corporate email scanning HEAD reqs

ndom91:
## Reasoning 💡

Saw someone else had this problem, and the "tutorial" / fix was relatively quick, so I decided to just write it up for the future.

See:
1. https://github.com/nextauthjs/next-auth/issues/2509
2. https://github.com/nextauthjs/next-auth/issues/1840

**EDIT**: There may need to be some extra checks before just returning a 200 all willy-nilly haha. What do yuo think Balazs?

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged


balazsorban44:
The write-up was very good, I actually am very clear about this problem now!

Although I'm not sure in the case of `HEAD` how this would be an issue, as we only run NextAuth.js logic when the request is `POST` or `GET`.

https://github.com/nextauthjs/next-auth/blob/68e412b063e824685529bddc1288ec88d41afe4b/packages/next-auth/src/core/index.ts#L234

balazsorban44 (commented):
The write-up was very good, I actually am very clear about this problem now!

Although I'm not sure in the case of `HEAD` how this would be an issue, as we only run NextAuth.js logic when the request is `POST` or `GET`.

https://github.com/nextauthjs/next-auth/blob/68e412b063e824685529bddc1288ec88d41afe4b/packages/next-auth/src/core/index.ts#L234

I have to assume that some of those IT checks actually send a `GET` request instead, which could invalidate the token indeed.

I would rather add some kind of user-agent detection to the tutorial. Do we know SafeLink or any of those if they send some detectable info?

ndom91:
Did some more research here, and it looks like opencollective were also not able to find any other distinguishing factors, other than that they were using "old" versions of chrome (See: https://github.com/opencollective/opencollective-frontend/blob/2852c113685ffc8273419839c6012f11a7751fbb/lib/robots-detector.js).

But you're right, it does seem to be `GET` requests in most cases, not `HEAD`.

balazsorban44:
Tha is for the thorough research! Some of these are nice tricks, but reusing the one-time token doesn't sound right to me. 😅

ndom91:
I think this one can just be merged. Any of the other work-arounds / potential changes we'd have to make are a little bigger, plus theres no "best-practice" out there in the community for dealing with this kind of thing yet so it'd require some more discussion about which route to even take, etc.. 

So I'd say having a basic tutorial like this is a decent middle-ground for now. What do you think? @balazsorban44

rnbrady:
@balazsorban44 wrote:
> I would rather add some kind of user-agent detection to the tutorial. Do we know SafeLink or any of those if they send some detectable info?

@ndom91 wrote:
> What I can definitely picture is that they have a distinguishable user-agent or something like that for their bot. I just don't know how we'd get it 🤔

I have a user who can't log in with the email provider so I ran some debug on Vercel. As soon as we send the verification email, we get two requests from their corporate firewall or proxy:

HEAD request with from user agent `Barracuda Sentinel (EE)`

GET request from `Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 6.1; WOW64; Trident/4.0; SLCC2; .NET CLR 2.0.50727; .NET CLR 3.5.30729; .NET CLR 3.0.30729)`

Thought that might be useful to you guys.

ndom91:
@rnbrady thanks for taking a look. This soundsn exactly like the Office 365 issue that sparked this whole issue/PR initially.

Were you able to get it to work, or the issue is still active for you? If so, I'd recommend opening a fresh issue and maybe pointing to this PR / existing issue(s).

Thanks!

rnbrady:
@ndom91 yes got it working as follows:

```javascript
const authHandler = NextAuth(authOptions);

export default function auth(req: NextApiRequest, res: NextApiResponse) {
  // Workaround for known email scanners that send GET or HEAD requests which have
  // the effect of cancelling the one time token. We have seen:

  // HEAD request with user-agent: Barracude Sentinel (EE)
  // GET request with user-agent: Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 6.1; WOW64; Trident/4.0; SLCC2; .NET CLR 2.0.50727; .NET CLR 3.5.30729; .NET CLR 3.0.30729)
  if (
    req.method === 'HEAD' ||
    /Mozilla.+MSIE.+Windows NT.+WOW64.+Trident.+SLCC2.+NET CLR/.test(
      req.headers['user-agent'],
    ) ||
    /lua-resty-http.+ngx_lua/.test(req.headers['user-agent'])
  ) {
    return res.status(200).send('Please visit the link from a browser.');
  }

  // GET request with double URL encoded param callbackUrl=https%253A%252F%252F
  if (req.url.includes('callbackUrl=https%253A%252F%252F'))
    return res
      .status(400)
      .send('Your proxy has mangled the callbackUrl parameter in the URL.');

  return authHandler(req, res);
}
```

Andrew781123:
Is this solution still valid? I came across the same issue but this solution doesn't solve it. I checked the logs and realise that there is no`HEAD` request, instead it is a`GET` request that triggered the magic link.

PR #3931, opened 2022-02-12
chore(docs): add corporate proxy tutorial

ndom91:
## Reasoning 💡

Added a quick tutorial for using NextAuth.js behind an HTTP(S) proxy, as detailed by @raphaelpc in #2509.

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged


balazsorban44:
I think if we actually are going to document this, we are as good as adding support for this instead.

ndom91:
@balazsorban44 yeah, up to you. I think the tutorial is easy enough to follow on ones own. Its also a relatively niche requirement, right?

But being niche also doesn't mean its not core, I feel like users in corporate networks / big enterprise-y apps will often need this. Maybe it just hasn't come up because people run into this issue and then switch auth libraries? Or other devs have just fixed it themselves before posting an issue here? Beats me.

If we wanted to add native support, it would require basically adding that http agent library as a dependency and adding a boolean toggle for `useProxy` or something like that, right? Oh and a few options to configure it I guess. Anything else?

ndom91:
@balazsorban44 what do you think about this? While adding support wouldn't be a major thing, it is adding another dependency.

Maybe we just add this tutorial for now and note down the proxy support for the next minor version bump or something like that?

herdma:
Thanks, this helped me a lot! I think there might be another patch needed to make the AzureAD provider work? I am running into the following error, do you have any idea what else might need to be changed?

```
[next-auth][error][OAUTH_PARSE_PROFILE_ERROR]
https://next-auth.js.org/errors#oauth_parse_profile_error request to https://graph.microsoft.com/v1.0/me/photos/48x48/$value failed, reason: getaddrinfo ENOTFOUND graph.microsoft.com {
  error: {
    message: 'request to https://graph.microsoft.com/v1.0/me/photos/48x48/$value failed, reason: getaddrinfo ENOTFOUND graph.microsoft.com',
    stack: 'FetchError: request to https://graph.microsoft.com/v1.0/me/photos/48x48/$value failed, reason: getaddrinfo ENOTFOUND graph.microsoft.com\n' +
      '    at ClientRequest.<anonymous> (/app/node_modules/next/dist/compiled/node-fetch/index.js:1:64142)\n' +
      '    at ClientRequest.emit (node:events:526:28)\n' +
      '    at TLSSocket.socketErrorListener (node:_http_client:442:9)\n' +
      '    at TLSSocket.emit (node:events:526:28)\n' +
      '    at emitErrorNT (node:internal/streams/destroy:157:8)\n' +
      '    at emitErrorCloseNT (node:internal/streams/destroy:122:3)\n' +
      '    at processTicksAndRejections (node:internal/process/task_queues:83:21)',
    name: 'FetchError'
  },
  OAuthProfile: {
    .. all good ..
  },
  message: 'request to https://graph.microsoft.com/v1.0/me/photos/48x48/$value failed, reason: getaddrinfo ENOTFOUND graph.microsoft.com'
}
```

ndom91:
@herdma ah yes, for providers where an extra request needs to be made for the profile image (i.e. the AzureAD one), the http proxy would also need applied there. Great catch! I'll make sure to make a note of that here.

PR #3964, opened 2022-02-13
feat(providers): add Shopify provider

prabureddy:
## Reasoning 💡
Added Shopify Provider

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged


docs/docs/providers/shopify.md
```diff
@@ -0,0 +1,50 @@
+---
+id: shopify
+title: Shopify
+---
+
+## Documentation for Shopify apps
+https://shopify.dev/apps
+
+
```
balazsorban44, line 9:
```suggestion
## Documentation for Shopify apps

https://shopify.dev/apps

```

apps/dev/pages/api/auth/[...nextauth].ts
```diff
@@ -91,6 +92,13 @@ export const authOptions: NextAuthOptions = {
       clientId: process.env.GITHUB_ID,
       clientSecret: process.env.GITHUB_SECRET,
     }),
+    ShopifyProvider({
+      clientId: process.env.SHOPIFY_ID,
+      clientSecret: process.env.SHOPIFY_SECRET,
+      shop: "unique-master", // Don't include .myshopify.com
+      scope: "read_products,write_products,read_orders,write_orders",
+      apiVersion: "2022-01",
```
Comment (balazsorban44) [inline apps/dev/pages/api/auth/[...nextauth].ts:100]:
```suggestion
```

These should not be necessary, scope is also not a top-level option anymore

docs/docs/providers/shopify.md
```diff
@@ -0,0 +1,50 @@
+---
```
balazsorban44, line 1: Let's make sure that this documentation page has the same structure as others, with the following headings:

- Documentation
- Configuration
- Options
- Example

docs/docs/providers/shopify.md
```diff
@@ -0,0 +1,50 @@
...
+```js
+import ShopifyProvider from "next-auth/providers/shopify";
+...
+providers: [
+  ShopifyProvider({
+    clientId: process.env.SHOPIFY_ID,
+    clientSecret: process.env.SHOPIFY_SECRET,
+    shop: "unique-master", // Don't include .myshopify.com
+    scope: "read_products,write_products,read_orders,write_orders",
+    apiVersion: "2022-01",
```
balazsorban44, line 29:
```suggestion
```

docs/docs/providers/shopify.md
```diff
@@ -0,0 +1,50 @@
...
+:::
+
+:::tip
+Create your shopify partner account https://www.shopify.com/shopifypartnersvap and create an app from there
+
+Under "URLs" in the App Setup, configure the following for "OAuth 2.0":
+
+- Redirect URL
+  - http://localhost:3000/api/auth/callback/shopify
+:::
```
balazsorban44, line 50: Let's use `tip` with moderation. Also, keep the casings consistent.

OAuth, Shopify,

docs/docusaurus.config.js
```diff
@@ -1,73 +1,73 @@
 module.exports = {
```
balazsorban44, line 1: Unnecessary file change, could you please revert?

docs/scripts/generate-providers.js
```diff
@@ -1,13 +1,13 @@
-const path = require("path")
```
balazsorban44, line 1: Unnecessary file change, could you please revert?

docs/sidebars.js
```diff
@@ -1,78 +1,78 @@
 module.exports = {
```
balazsorban44, line 1: Unnecessary file change, could you please revert?

docs/src/components/ProviderMarquee.js
```diff
@@ -1,29 +1,30 @@
-import React from "react"
```
balazsorban44, line 1: I see you added a Shopify logo here, which is fine, but all the `'` characters need to be changed back to `"`

docs/src/components/ProviderMarqueeStyle.js
```diff
@@ -1,4 +1,4 @@
-import styled from "styled-components"
```
balazsorban44, line 1: Unnecessary file change, could you please revert?

docs/src/pages/index.js
```diff
@@ -1,18 +1,18 @@
-import React, { useEffect } from "react"
```
balazsorban44, line 1: Unnecessary file change, could you please revert?

docs/src/pages/seo.js
```diff
@@ -1,6 +1,6 @@
-import React from "react"
```
balazsorban44, line 1: Unnecessary file change, could you please revert?

packages/next-auth/src/providers/shopify.js
```diff
@@ -0,0 +1,79 @@
+export default function Shopify(options) {
+  const {
+    shop = "",
+    scope = "",
+    clientId = "",
+    clientSecret = "",
+    apiVersion = "",
+  } = options
```
balazsorban44, line 8:
```suggestion
  const { apiVersion = "2022-01" } = options
```

Let's set a default value for the apiVersion. `scope` is misplaced, and the others are required values, so no need to default them anything.

packages/next-auth/src/providers/shopify.js
```diff
@@ -0,0 +1,79 @@
...
+        if (res.ok) {
+          const result = await res.json()
+          return { tokens: result }
+        }
+
+        throw new Error(
+          "Something went wrong while trying to get the access_token"
+        )
+      },
+    },
```
balazsorban44, line 49:
```suggestion
    client: {
      token_endpoint_auth_method: "client_secret_post",
    },
    token: `https://${shop}.myshopify.com/admin/oauth/access_token`,
```

This is a standard, so no need to reach for `request`.

packages/next-auth/src/providers/shopify.js
```diff
@@ -0,0 +1,79 @@
...
+          }
+        )
+        if (res.ok) {
+          const result = await res.json()
+          return result
+        }
+
+        throw new Error("Something went wrong while trying to access your shop")
+      },
+    },
```
balazsorban44, line 70:
```suggestion
    userinfo: `https://${shop}.myshopify.com/admin/api/${apiVersion}/shop.json`,
```

Potentially this should be enough, although it would need verification due to this I found: https://community.shopify.com/c/shopify-apis-and-sdks/x-shopify-access-token-vs-authorization-bearer/m-p/564674

Looks like Shopify is claiming OAuth 2.0 compliance but they are skipping this step...

https://datatracker.ietf.org/doc/html/rfc6749#section-7.1

Or they might allow it now, could you just verify, please?

packages/next-auth/src/providers/shopify.js
```diff
@@ -0,0 +1,79 @@
...
+  } = options
+
+  return {
+    id: "shopify",
+    name: "Shopify",
+    type: "oauth",
+    authorization: {
+      url: `https://${shop}.myshopify.com/admin/oauth/authorize`,
+      params: {
+        scope,
```
balazsorban44, line 17: Let's set some sensible default scopes that are only enough to show basic profile information. If the developer wants to set additional scopes, they can easily override these defaults.

packages/next-auth/src/providers/shopify.js
```diff
@@ -0,0 +1,79 @@
...
+        throw new Error("Something went wrong while trying to access your shop")
+      },
+    },
+    profile: (profile) => {
+      return {
+        id: profile.shop.id,
+        name: profile.shop.name,
+      }
+    },
+    ...options,
```
balazsorban44, line 77:
```suggestion
    options,
```

packages/next-auth/src/providers/shopify.js
```diff
@@ -0,0 +1,79 @@
...
+          return result
+        }
+
+        throw new Error("Something went wrong while trying to access your shop")
+      },
+    },
+    profile: (profile) => {
+      return {
+        id: profile.shop.id,
+        name: profile.shop.name,
```
balazsorban44, line 74:
```suggestion
        id: profile.shop.id,
        name: profile.shop.name,
        email: null,
        image: null
```

If e-mail and a profile image isn't available, we have to set these values to `null`.

balazsorban44 (changes requested):
Thank you! I've some comments. Generally, a lot of files have been touched which shouldn't have been. You might have a Prettier setting that conflicts with this project's preference. We prefer double-quotes. I added a rule to it in our `package.json` to avoid problems like this in the future.

In addition, I would like you to make sure the documentation page is similar to other provider documentation pages in structure. See https://next-auth.js.org/providers/auth0 for an example.

The provider implementation had a bit too many customizations, I tried to give hints on how to simplify it and make it more similar to other providers.

I would like to ask you how you feel about TypeScript? We prefer Providers to be written in TypeScript, but it's fine if you are not comfortable, we can merge it as JS and we could convert ourselves (cc @ThangHuuVu, @lluia)

If you are interested, here is an example: https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/providers/auth0.ts

docs/scripts/generate-providers.js
prabureddy: I executed `yarn lint:fix` for linting and it modified some of the files.

packages/next-auth/src/providers/shopify.js
prabureddy: Shopify is ignoring this step [https://datatracker.ietf.org/doc/html/rfc6749#section-7.1](https://datatracker.ietf.org/doc/html/rfc6749#section-7.1).

Shopify is using `x-shopify-access-token` to fetch user-app related data and not an `authorization` header. So, we have to implement the userinfo -> request method from scratch.

packages/next-auth/src/providers/shopify.js
prabureddy: We can add default scopes related to the profile but if the Shopify app wants to access the shop orders and their customers then again the shop admin need to reinstall the app with new scopes and it breaks the user experience. This is the reason all Shopify apps are suggested to add all the required scopes at once while installation.

apps/dev/pages/api/auth/[...nextauth].ts
Comment (prabureddy) [reply apps/dev/pages/api/auth/[...nextauth].ts:100]: For every Shopify apps scope is a must, each app has different features and according to that the app request Shopify with the required scopes.

The scope should be received from the app. Is there any way to receive scope from the user if the scope is also not a top-level option anymore?

prabureddy:
@balazsorban44 Thanks for the review and the guidelines. I've made the requested changes. Please review it.

docs/src/pages/index.js
```diff
@@ -288,4 +288,4 @@ export default NextAuth({
 })
 `.trim()
 
-export default Home
+export default Home
```
lluia, line 291: Let's revert this change as it's from **Prettier** and I believe unintentional.

docs/docs/providers/shopify.md
```diff
@@ -0,0 +1,45 @@
...
+title: Shopify
+---
+
+## Documentation
+
+https://shopify.dev/apps/auth/oauth/getting-started
+
+## Configuration
+
+This is your **Redirect URL** <ins>http://localhost:3000/api/auth/callback/shopify</ins>. Add this callback URL in your app settings.
```
lluia, line 12: I would re-word this. 

`http://localhost:3000/api/auth/callback` works if the user is trying the application locally, however in other environments like production the domain is going to be different. 

For instance, when using **Vercel** it would be: `https://my-app.vercel.app/api/auth/callback`.

```suggestion
Add `/api/auth/callback/shopify in your app settings as the **callback URL**. For instance, if you're trying the sign in locally it'll be: `http://localhost:3000/api/auth/callback`.
```

docs/docs/providers/shopify.md
```diff
@@ -0,0 +1,45 @@
...
+
+```js
+import ShopifyProvider from "next-auth/providers/shopify";
+...
+providers: [
+  ShopifyProvider({
+    clientId: process.env.SHOPIFY_ID,
+    clientSecret: process.env.SHOPIFY_SECRET,
+    shop: "unique-master", // Don't include .myshopify.com
+    scope: "read_products,write_products,read_orders,write_orders",
```
lluia, line 38:
```suggestion
    authorization: {
      // if needed, you can ask for extra scopes here...
      params: { scope: "read_products" },
    },
```

packages/next-auth/src/providers/shopify.js
```diff
@@ -0,0 +1,49 @@
+export default function Shopify(options) {
+  const { shop, scope, apiVersion = "2022-01" } = options
```
lluia, line 2: As @balazsorban44 mentioned, `scope` [is no longer a top-level option](https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/providers/oauth.ts#L92-L148). Given this is a **JS** file you don't see any complaints, but in **TS** you wouldn't be allowed to read it like this ☕️

Instead, users of this provider can supply extra scopes like this:
```ts
  ShopifyProvider({
    clientId: process.env.SHOPIFY_ID,
    clientSecret: process.env.SHOPIFY_SECRET,
    authorization: { params: { scope: 'my-scopes' } }   
```
There's no need for you to handle this case in the code, as the options are going to be deeply merged on our side.

0ubbe (commented):
Thanks for the contribution! 💚  Some small suggestions 💬

prabureddy:
Changes modified @lluia @balazsorban44

packages/next-auth/src/providers/shopify.js
balazsorban44: Thanks. I took this up with the Shopify Engineering team. Let's hope for a good answer.

jamie-kieran-martin: Any update on this? Keen to get this working, happy to assist wherever required.

In playing around with this, I've noticed a few things.

# Access Mode

Shopify has two different access modes in their auth flow: `online` and `offline` (defaults to `offline`).

Adding `grant_options[]` (as shown below) to the params in the authorization endpoint turns the `online` mode on.

```js
{
...
authorization: {
 url: `https://${shop}.myshopify.com/admin/oauth/authorize`,
 params: {
  "grant_options[]": "per-user",
  scope: "read_products",
 },
},
...
}
```

### `offline` response

```js
{
  access_token: 'shpat_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
  scope: '...'
}
```

### `online` response

```js
{
  access_token: 'shpat_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
  scope: '...',
  expires_at: 1645954815,
  associated_user_scope: '...',
  session: '...',
  account_number: 0,
  associated_user: {
    id: 12345,
    first_name: '...',
    last_name: '...',
    email: '...',
    account_owner: false,
    locale: 'en-AU',
    collaborator: false,
    email_verified: true
  }
}
```

`online` mode breaks adapters as the data returned no longer conforms with the NextAuth [Models](https://next-auth.js.org/adapters/models). (Testing done with Prisma adapter)

As far as I can tell this is the only way to get the actual logged-in user. Shopify's User API seems to be locked behind Shopify Plus [[reference](https://shopify.dev/api/admin-rest/2022-01/resources/user#top)]. But the `associated_user.id` value is **not** global across different Shopify instances... so a user has multiple IDs across one email.

# `userinfo`

If using the offline mode token, this results in a staff user being able to bypass scopes if they have access to the configured App in Shopify. The current setup in the `userinfo.request` callback using the `https://${shop}.myshopify.com/admin/api/${apiVersion}/shop.json` endpoint returns information on the Shopify shop. This means all staff on the Shopify instance authenticate to the same user.

# `store_name` aka `shop`

All endpoints in the Shopify Admin API follow this pattern:
https://{store_name}.myshopify.com/admin/api/{apiVersion}/{resource}.json

Requests made to Shopify must be made via the Shopify instance (as in `abc.myshopify.com`, where `abc` is the `store_name`). This must be stored somewhere for Public Apps. As far as I can tell this can never be changed [reference](https://help.shopify.com/en/manual/your-account/manage-account). 

In testing, I was using this temporary configuration for the profile callback to store the store_name as the users' name:

```js
profile: (profile) => {
 return {
  id: profile.shop.id,
  name: profile.shop.myshopify_domain,
  email: profile.shop.email,
  image: null,
 };
},
```

Note the `myshopify_domain`. This is the `store_name` and what I was using to make requests to the Shopify Admin API with. It might be wise to store the shop somewhere semantic.

Hope this helps!

An aside from this, the shop query param gets lost in the auth flow after the initial redirect from Shopify to your app. Following the advice in #3776, I pull the `shop` variable from the query params and add it to the cookie session so I can use it between the requests of the auth flow. Not great, but it works. This implementation also assumes that the user comes from Shopify, and doesn't navigate directly resulting in the shop variable being undefined :\

prabureddy:
This feature is ready to be merged. Need approval from admins!

ndom91:
@prabureddy looks like there is now a conflict in `docs/providers.json`. Can you take a look at that? Thanks!

prabureddy:
@ndom91 conflicts resolved.... Thanks!

Please let me know if anything up there!

ndom91:
I just tested this locally, and everything looks good, except for one small thing.

When I tried to connect to my test app with a test store, it came back with an error that the scope: `openid` was invalid. So I tried it with an empty scope, no luck. But putting any sort of valid scope in seemed to work. So for my testing I used `read_customers`. 

i.e.

```
    id: "shopify",
    name: "Shopify",
    type: "oauth",
    authorization: {
      url: `https://${shop}.myshopify.com/admin/oauth/authorize`,
      params: {
        scope: "read_customers",
      },
    },
...
```

prabureddy:
Yes, we need to add some required scopes to install the app in a store. It is mandatory.

ndom91:
> Yes, we need to add some required scopes to install the app in a store. It is mandatory.

Right, so the provider can't be used with a defined `scope`, right? Let us add a basic one already, or at least mention it in the setup docs then :+1:

packages/next-auth/src/providers/shopify.js
```diff
@@ -0,0 +1,46 @@
+export default function Shopify(options) {
+  const { shop, apiVersion = "2022-01" } = options
+
+  return {
+    id: "shopify",
+    name: "Shopify",
+    type: "oauth",
+    authorization: {
+      url: `https://${shop}.myshopify.com/admin/oauth/authorize`,
```
balazsorban44, line 9: I much preferred if instead of adding a new provider-specific config option `shop`, we required `issuer` to be set to the URL `https://${shop}.myshopify.com` to align with other providers.

ndom91: Are you recommending renaming `shop` to `issuer` here basically?

ThangHuuVu:
Hi, @prabureddy 👋 are you still interested in shipping this? 👀

prabureddy:
@balazsorban44  Can we merge this PR?

mgavin-origen:
I'm very interested, is there anything new?

PR #3966, opened 2022-02-13
fix: custom sign in errors without redirection

### LINKED ISSUE #1564: Custom SignIn errors without redirection
Custom SignIn errors without redirection

SignIn without redirection is possible, but we couldn't retrieve and show custom errors without redirection

Login page:

`const handleLogin = async (values) => {
        const response = await signIn('credentials', {
            identifier: values.email,
            password: values.password,
            redirect: false
        })
        if (response.ok) router.push('/')
        else setError({ error: response.error })`

[...nextauth].js:

`authorize: async (credentials) => {
            try {
                const { data } = await api.post('auth', {
                    identifier: credentials.identifier,
                    password: credentials.password
                })
                return data
            } catch (error) {
                const errorMessage = error.message;
                console.error('login error:', errorMessage)
                throw Error(errorMessage)
            }`

When an error is thrown, it should be available through the signIn response if redirect is set to false

**Potential problems**
None

Using it as is but it breaks UX

**Additional context**
None

ThangHuuVu:
## Reasoning 💡

For `credentials` provider:
- If `provider.authrize()` return no user, redirect to signin instead of `/error` to skip an unnecessary redirect.
- If  any errors are thrown during the process, return a status code `401`.
## Checklist 🧢

- ~[ ] Documentation~
- [ ] Tests
- [ ] Ready to be merged

## Affected issues 🎟

- Fixes #1564


balazsorban44 (commented):
Thanks! I'm trying to remember if the following only exists because of historical reasons or it has some use case:

https://github.com/nextauthjs/next-auth/blob/614d9cc4339b51b450b9c6c6f2ed1b2621929f39/packages/next-auth/src/core/index.ts#L142-L157

I mean, in theory, all these could be solved similarly to what this PR proposes :thinking:, right?

ThangHuuVu:
> I'm trying to remember if the following only exists because of historical reasons or it has some use case:

So here's a relevant PR: https://github.com/nextauthjs/next-auth/pull/3576

>  in theory, all these could be solved similarly to what this PR proposes 🤔, right?

IMO it's totally feasible to remove that block of code in [next-auth/packages/next-auth/src/core/index.ts](https://github.com/nextauthjs/next-auth/blob/614d9cc4339b51b450b9c6c6f2ed1b2621929f39/packages/next-auth/src/core/index.ts#L142-L157), and handle redirect to `/signin` in various places (mostly in callback.ts). I can search for all the errors except for `Signin`, do you know where it's used?

balazsorban44 (approved):

balazsorban44:
Sorry, this was still a draft. `Signin` might not even be something we have. See:

https://next-auth.js.org/configuration/pages#sign-in-page

We might want to strongly type all the available errors to avoid confusion.

ThangHuuVu:
> We might want to strongly type all the available errors to avoid confusion.
@balazsorban44 how about we merge this PR first for the issue and address this problem in a follow-up PR?

Issue #912, opened 2020-12-05
Regarding stale issues


balazsorban44:
Hi there! 👋 

If you are reading this, you probably met our little non-human 🤖 friend who told you to come here.

In an effort to reduce the number of open issues, we use [stalebot](https://github.com/probot/stale) to close old/inactive issues.

**If you think your issue is still relevant and/or unresolved, we encourage you to comment on the issue, so it will be marked as not stale again.**

We hope that this will help us maintain the project better.

If you have any thoughts on this decision, feel free to comment on this issue.

jaredpalmer:
Bump

4mohdisa:
I get error: Try signing in with a different account.
when i try to login with google auth i am using next-auth

PR #3969, opened 2022-02-13
chore: improve maintenance

balazsorban44:
I cleaned up the following:

- [x] stale action: Does not link to #912 anymore. If we want to make that info more prominent, maybe the `CONTRIBUTING.md` file is more prominent, we should link to that instead. This PR does not address that.
- [x] PR labeler action: Cleaned it up and locked the version on the action to prevent future issues
- [x] Issue labeler action: New bot that auto triages adapter bug reports. based on the dropdown choice of the user, it will add the adapter specific issue label
- [x] Bug report issue template: This is our default template, should mostly be used for core-specific reports and later on (#2294) other frameworks.
- [x] Bug report (Adapter): Since moving to a monorepo (#3952), we will need to manage adapters better. This new template asks adapter questions like a version of that adapter and provides a dropdown for the user of our available adapters (and another category for custom adapters). Used by the Issue labeler action to auto-label these with the specific adapter type to be able to scan the issues more easily.
- [x] Bug report (Provider): Similar to the above one, but we don't auto-label these individually since all of them are in the core.
- [x] default label for bug reports/feature requests is now "triage". A maintainer will need to mark something as a "bug" or "accepted" before it is considered for further investigation.
- [x] issues with the label "triage" won't be auto-closed
- [x] issues with the label "accepted" won't be auto-closed

Check the labels page for the new and existing labels: https://github.com/nextauthjs/next-auth/labels

What we could do in addition is to create PR templates specific for new providers and adapters with precise checkpoints (testing, expectations, docs, etc). This can be addressed in a subsequent PR.


ndom91 (approved):
This looks awesome! Great job cleaning all this up. Love the little details like the links to the docs for the various labelers, etc.

balazsorban44:
cc @lluia, @StefanSelfTaught, @ThangHuuVu. This hopefully will help with triaging.

PR #4002, opened 2022-02-17
Add refresh_token_expires_in field

hayoung0Lee:
## Reasoning 💡
I was trying to use Kakao Provider with Prisma based on the schema here. 
* https://next-auth.js.org/adapters/prisma

But it kept making errors related to the issue below. 

image

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## Affected issues 🎟
So I added a field `refresh_token_expires_in field`

### DIFF
--- a/docs/docs/adapters/prisma.md
+++ b/docs/docs/adapters/prisma.md
@@ -70,7 +70,7 @@ model Account {
   session_state      String?
   oauth_token_secret String?
   oauth_token        String?
-
+  refresh_token_expires_in Int?
   user User @relation(fields: [userId], references: [id], onDelete: Cascade)
 
   @@unique([provider, providerAccountId])


balazsorban44:
@ndom91 I think we have mentioned this on our last chat.

Hi, and thanks for the PR! `refresh_token_expires_in` is a field that we have seen on a couple of providers, but AFAIK, it is not part of the official OAuth/OIDC spec, thus we don't want to include it in the default schema:

OAuth: https://datatracker.ietf.org/doc/html/rfc6749#section-5.1
OIDC: https://openid.net/specs/openid-connect-core-1_0.html#TokenResponse

What we have done instead is notifying about this on individual provider docs like in case of GitHub:

https://next-auth.js.org/providers/github

We are thinking about how to make this experience better. Maybe a normalized table or TS interface that extends the default ones? cc @lluia

Another way to make this better is to better explain how to create/extend an adapter (like overriding a single method).

So for now, I would like to close the PR, but we welcome a docs update similar to GitHub's that mentions that Kakao Provider has this extra field.

PR #4003, opened 2022-02-17
fix: only warn when using Twitter + OAuth 2.0

balazsorban44:
@TheoBr reported this to me, thank you. We should only show the warning if the user opts into Twitter + OAuth 2.0, as their implementation is still somewhat subject to change.

In a future release, we can swap this warning to encourage people to use OAuth 2.0, so after that we can finally drop support for OAuth 1.0 support together with a 5-year-old 130kb dependency, decreasing the install size by ~40%! 

- https://packagephobia.com/result?p=oauth
- https://packagephobia.com/result?p=next-auth

I would like to entertain the idea that once we get the green light from Twitter to advertise their OAuth 2 over OAuth 1, we just bump to v5 and only require the user to install the `oauth` dependency only if they use Twitter 1.

The code is already pretty safe and would not block other users to decrease their bundle size.

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/packages/next-auth/src/core/lib/assert.ts
+++ b/packages/next-auth/src/core/lib/assert.ts
@@ -47,12 +47,13 @@ export function assertConfig(
   if (!req.host) return "NEXTAUTH_URL"
 
   let hasCredentials, hasEmail
-  let hasTwitterProvider
+  let hasTwitterOAuth2
 
   for (const provider of options.providers) {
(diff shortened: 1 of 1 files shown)


t3dotgg (approved):
Thank you for this!!

0ubbe (approved):

PR #4015, opened 2022-02-19
Switch BattleNet to OIDC

### LINKED ISSUE #4013: BattleNet Provider doesn't work (few issues)
### Provider type

Battlenet

### Environment

```
  System:
    OS: macOS 12.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 917.94 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    npm: 8.4.1 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Brave Browser: 98.1.35.103
    Chrome: 98.0.4758.102
    Firefox: 97.0.1
    Safari: 15.2
  npmPackages:
    next: 12.1.0 => 12.1.0 
    next-auth: 4.2.1 => 4.2.1 
```

### Reproduction URL

-

### Describe the issue

Hi, 
The Battlenet provider doesn't seem to work.  I've come across two issues so far.

The first is:
```
https://next-auth.js.org/errors#callback_oauth_error id_token detected in the response, you must use client.callback() instead of client.oauthCallback() RPError: id_token detected in the response, you must use client.callback() instead of client.oauthCallback()
```

This seems to be resolved by setting `idToken: true` on the provider.

Once past this part, we get:
```
[next-auth][error][OAUTH_CALLBACK_ERROR] 
https://next-auth.js.org/errors#oauth_callback_error unexpected iss value, expected undefined, got: https://eu.battle.net/oauth {
  error: {
    message: 'unexpected iss value, expected undefined, got: https://eu.battle.net/oauth',
    stack: 'RPError: unexpected iss value, expected undefined, got: https://eu.battle.net/oauth\n' +
      '    at Client.validateJWT (/Users/Karl.Stoney/git/github/gin/ui/node_modules/openid-client/lib/client.js:924:15)\n' +
      '    at Client.validateIdToken (/Users/Karl.Stoney/git/github/gin/ui/node_modules/openid-client/lib/client.js:765:60)\n' +
```

The code appears to be comparing the `iss` from the JWT response to the provider issuer, and they don't match:

```
{
  at_hash: '...',
  sub: '...',
  aud: '...',
  azp: '...',
  iss: 'https://eu.battle.net/oauth',
  exp: 1645371778,
  iat: 1645285411,
  battle_tag: 'Stono#2158',
  jti: '...'
} 

Issuer {
  authorization_endpoint: 'https://eu.battle.net/oauth/authorize',
  issuer: undefined,
  token_endpoint: 'https://eu.battle.net/oauth/token',
  userinfo_endpoint: 'https://us.battle.net/oauth/userinfo'
}
```

Not sure how to resolve that one...

If i just hack that part out, the next error I get is:

```
[next-auth][error][CALLBACK_OAUTH_ERROR] 
https://next-auth.js.org/errors#callback_oauth_error jwks_uri must be configured on the issuer TypeError: jwks_uri must be configured on the issuer
```

Which again looking at the Issuer makes sense?  It's almost like the `Issuer` object is missing key information that's required to validate the JWT.

### How to reproduce

Just configure the Battlenet provider:

```
export const battleNetProvider = BattleNetProvider({
  clientId: process.env.BATTLENET_CLIENT_ID,
  clientSecret: process.env.BATTLENET_CLIENT_SECRET,
  region: 'EU'
})
```

### Expected behavior

It should work

Stono:
## Reasoning 💡

The current battlenet provider is broken, this switches it to a OIDC provider instead.

It switches it to a typescript file from a js file too.

## Checklist 🧢

- [ ] Documentation
- [ ] Tests (tested locally)
- [x] Ready to be merged

## Affected issues 🎟

Fixes: https://github.com/nextauthjs/next-auth/issues/4013


Stono:
I'm new to next-auth btw so still finding my feet.  I've noticed account linking only works when you have the same email address (and you're logged in when the flow starts). 

Battlenet doesn't return an email address 🤷  

I assume to do account linking here i need to wait for the api (next release?)

packages/next-auth/src/providers/battlenet.ts
```diff
@@ -0,0 +1,32 @@
+import type { OAuthConfig } from "."
+import OAuthProvider from './auth0'
+export interface BattleNetProfile {
```
balazsorban44, line 3:
```suggestion
import type { OAuthConfig, OAuthUserConfig } from "."

export interface BattleNetProfile {
```

packages/next-auth/src/providers/battlenet.ts
```diff
@@ -0,0 +1,32 @@
...
+  sub: string
+  battle_tag: string
+}
+
+export default function BattleNet(
+  options: { clientId: string, clientSecret: string, region: 'US' | 'EU' | 'KR' | 'TW' | 'CN' }
+): OAuthConfig<BattleNetProfile> {
```
balazsorban44, line 10:
```suggestion
export default function BattleNet<
  P extends Record<string, any> = BattleNetProfile
>(
  options: OAuthUserConfig<P> & {
    /** Defaults to `"CN"` */
    region?: "US" | "EU" | "KR" | "TW" | "CN"
  }
): OAuthConfig<P> {
```

packages/next-auth/src/providers/battlenet.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  const wellKnown = options.region === 'CN' 
+    ? 'https://www.battlenet.com.cn/oauth/.well-known/openid-configuration' 
+    : `https://us.battle.net/oauth/.well-known/openid-configuration`
```
balazsorban44, line 14:
```suggestion
  const { region = "CN" } = options
  const domain =
    region === "CN"
      ? "https://www.battlenet.com.cn"
      : `https://${region}.battle.net`
```

packages/next-auth/src/providers/battlenet.ts
```diff
@@ -0,0 +1,32 @@
...
+): OAuthConfig<BattleNetProfile> {
+
+  const wellKnown = options.region === 'CN' 
+    ? 'https://www.battlenet.com.cn/oauth/.well-known/openid-configuration' 
+    : `https://us.battle.net/oauth/.well-known/openid-configuration` 
+
+  return OAuthProvider({
+    id: "battlenet",
+    name: "Battle.net",
+    wellKnown,
```
balazsorban44, line 19:
```suggestion
    wellKnown: `${domain}/oauth/.well-known/openid-configuration`,
```

packages/next-auth/src/providers/battlenet.ts
```diff
@@ -0,0 +1,32 @@
...
+    idToken: true,
```
balazsorban44, line 20:
```suggestion
```

packages/next-auth/src/providers/battlenet.ts
```diff
@@ -0,0 +1,32 @@
...
+    clientId: options.clientId,
+    clientSecret: options.clientSecret,
```
balazsorban44, line 22:
```suggestion
```

These are set implicitly

balazsorban44 (changes requested):
Thanks, I left some comments to clean the config up.

As proof, could you also take a screenshot of your logged-in user so we know it actually works? Good to have as a future reference.

balazsorban44:
> I've noticed account linking only works when you have the same email address (and you're logged in when the flow starts).

This is correct. The default linking probably won't change, but you can use the `adapter.linkAccout` method (with proper security in mind on your side since it would be user-land code) however you like to connect accounts.

packages/next-auth/src/providers/battlenet.ts
Stono: I assumed that we wouldn't want to allow people to override the "working" config :D

packages/next-auth/src/providers/battlenet.ts
Stono: Considering region isn't optional, why would we default it to CN here?

Stono:
Can't show you my signed in user because i'm not actually allowing people to sign in with this, i'm trying a convoluted way to get account linking to work.  

However here i've console logged out the details in the signin method so you can see it's working:

token
signin

Stono:
> > I've noticed account linking only works when you have the same email address (and you're logged in when the flow starts).
> 
> This is correct. The default linking probably won't change, but you can use the `adapter.linkAccout` method (with proper security in mind on your side since it would be user-land code) however you like to connect accounts.

Locally where would the most appropriate place for me to do that be?  For context i'm trying to sign in using Credentials but then link several oauth'd accounts.

packages/next-auth/src/providers/battlenet.ts
balazsorban44: Sorry, I assumed it was the case before, but it isn't true. https://github.com/nextauthjs/next-auth/blob/3d0c68d9e00c707f254d17a511c54af4ebb9eeb0/packages/next-auth/src/providers/battlenet.js#L3-L7

balazsorban44:
Should be a discussion, not a comment on this PR :wink:

balazsorban44 (approved):

balazsorban44:
I standardized the changes of this PR in a follow-up commit here: https://github.com/nextauthjs/next-auth/commit/163149b4243b857f73a2cee9afa62a958d57bb51

Basically, we should be using `issuer` cause that's the common reference in OAuth.

codal-Jamie:
Running into `id_token detected in the response, you must use client.callback() instead of client.oauthCallback()` and the already mentioned issue with the Facebook provider too. Can the Facebook Provider also be updated to address the issue?

PR #4023, opened 2022-02-20
docs: add SvelteKit playground

wobsoriano:
## Reasoning 💡

Adds a sveltekit + nextauth integration playground

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged


apps/playground-sveltekit/.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict=true
```
balazsorban44, line 1: Just curious if this file is necessary, if not, could we drop it to keep the playground minimal?

apps/playground-sveltekit/.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "useTabs": false,
+  "singleQuote": true,
+  "trailingComma": "none",
+  "printWidth": 100
+}
```
balazsorban44, line 6: Our default prettier config usually looks like this:

```js
  "prettier": {
    "semi": false,
    "singleQuote": false
  },
```

Let's keep the same for consistency's sake. :pray:

I would also prefer to just put it into the `package.json` file for fewer files in the root directory.

apps/playground-sveltekit/README.md
```diff
@@ -0,0 +1,85 @@
+# SvelteKit + NextAuth Playground
```
balazsorban44, line 1:
```suggestion
# SvelteKit + NextAuth.js Playground
```

apps/playground-sveltekit/README.md
```diff
@@ -0,0 +1,85 @@
...
+
```
balazsorban44, line 2:
```suggestion

NextAuth.js is committed to bringing easy authentication to other frameworks. https://github.com/nextauthjs/next-auth/issues/2294

SvelteKit support with NextAuth.js is currently experimental. This directory contains a minimal, proof-of-concept application. Parts of this is expected to be abstracted away into a package like `@next-auth/sveltekit`

```

Let's make this clear in case people will see this directory.

apps/playground-sveltekit/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "svelte-check": "^2.2.6",
+    "svelte-preprocess": "^4.10.1",
+    "svelte2tsx": "^0.5.5",
+    "tslib": "^2.3.1",
+    "typescript": "~4.5.4"
+  },
+  "type": "module",
+  "dependencies": {
+    "cookie": "0.4.1",
+    "next-auth": "^4.2.1"
```
balazsorban44, line 35:
```suggestion
```

I think this is unnecessary to define in a monorepo setup. Otherwise, you risk installing a version from npm, rather than reusing the local version.

apps/playground-sveltekit/pnpm-lock.yaml
```diff
@@ -0,0 +1,2780 @@
+lockfileVersion: 5.3
```
balazsorban44, line 1: This monorepo currently uses Yarn, so let's get rid of this file :+1:

apps/playground-sveltekit/src/global.d.ts
```diff
@@ -0,0 +1,7 @@
+/// <reference types="@sveltejs/kit" />
+
+interface ImportMetaEnv {
+  VITE_GITHUB_CLIENT_ID: string;
+  VITE_GITHUB_CLIENT_SECRET: string;
+  VITE_NEXTAUTH_URL: string;
```
balazsorban44, line 6:
```suggestion
  VITE_NEXTAUTH_URL: string;
  VITE_NEXTAUTH_SECRET: string;
```

apps/playground-sveltekit/.env.example
```diff
@@ -0,0 +1,3 @@
+VITE_GITHUB_CLIENT_ID=
+VITE_GITHUB_CLIENT_SECRET=
+VITE_NEXTAUTH_URL=
```
balazsorban44, line 3:
```suggestion
VITE_NEXTAUTH_URL=
VITE_NEXTAUTH_SECRET=
```

See: https://next-auth.js.org/configuration/options#nextauth_secret

apps/playground-sveltekit/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "sveltekit-nextauth",
+  "version": "0.0.1",
+  "scripts": {
+    "dev": "svelte-kit dev",
+    "build": "svelte-kit build",
+    "package": "svelte-kit package",
```
balazsorban44, line 7: Without knowing the details yet, I am fairly confident we will have our own way of packaging, or at least not as part of the playground.

If possible, could we remove anything related to packaging from this directory?

apps/playground-sveltekit/src/lib/next-auth.ts
```diff
@@ -0,0 +1,105 @@
...
+async function SKNextAuthHandler({ request, url, params }: RequestEvent, options: NextAuthOptions) {
+  const nextauth = params.nextauth.split('/');
+  let body = null;
+  try {
+    body = await request.formData();
+    body = getFormBody(body);
+  } catch {
+    // no formData passed
+  }
+
```
balazsorban44, line 55:
```suggestion
  options.secret = import.meta.env.VITE_NEXTAUTH_SECRET
```

Do you know if third-party libraries can read environment variables like this?

This would let the user configure fewer things. We would magically inject this value if they named the env var as we expected here.

apps/playground-sveltekit/src/routes/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,14 @@
...
+const nextAuthOptions: NextAuthOptions = {
+  providers: [
+    GithubProvider({
+      clientId: import.meta.env.VITE_GITHUB_CLIENT_ID,
+      clientSecret: import.meta.env.VITE_GITHUB_CLIENT_SECRET
+    })
+  ]
+};
+
+export const { get, post } = NextAuth(nextAuthOptions);
```
Comment (balazsorban44) [inline apps/playground-sveltekit/src/routes/api/auth/[...nextauth].ts:14]:
```suggestion
import GithubProvider from 'next-auth/providers/github';

export const { get, post } = NextAuth({
  providers: [
    GithubProvider({
      clientId: import.meta.env.VITE_GITHUB_CLIENT_ID,
      clientSecret: import.meta.env.VITE_GITHUB_CLIENT_SECRET
    })
  ]
});
```

balazsorban44 (changes requested):
This is looking good! I added some comments, but after addressing those, I think we could merge it as a base to go off of. Nice work so far! :green_heart:

apps/playground-sveltekit/package.json
wobsoriano: Yeah I think we can remove this since we'll have our own packaging

apps/playground-sveltekit/src/lib/next-auth.ts
wobsoriano: I think only vite reads this

wobsoriano:
Updated some of the suggestions except anything related to monorepo for now 👍🏼

balazsorban44 (approved):
Thanks! Let's merge it for now and keep working on it.

wobsoriano:
I think the next good feature to have is the client API

Issue #3970, opened 2022-02-13
Credentials Provider Not Firing Session Callback 


Suonx002:
### Description 🐜

Hello,

I'm having trouble with using Credentials Provider where it's not firing the session callback. I've logged the whole process from signin and output with logger functionality. 

I'm not sure if it's my [..nextauth] setting or it's a bug within next-auth and prisma adapter.

Can provide the repo link if needed. 

Please note that Github Provider are working fine where it generate session token.

### Is this a bug in your own project?

Yes

### How to reproduce ☕️

import NextAuth from 'next-auth';
import CredentialsProvider from 'next-auth/providers/credentials';
import GitHubProvider from 'next-auth/providers/github';
import { PrismaAdapter } from '@next-auth/prisma-adapter';
import { PrismaClient } from '@prisma/client';

import { verifyPassword, hashPassword } from '@functionHelpers/auth/passwords';

const prisma = new PrismaClient();

export default NextAuth({
	adapter: PrismaAdapter(prisma),
	secret: process.env.NEXTAUTH_SECRET,
	// session: {
	// 	jwt: true,
	// },
	pages: {
		// signIn: '/auth/signin',
		// signOut: "/auth/logout",
		// error: "/auth/error", // Error code passed in query string as ?error=
	},
	logger: {
		error(code, metadata) {
			console.log({ type: 'inside error logger', code, metadata });
		},
		warn(code) {
			console.log({ type: 'inside warn logger', code });
		},
		debug(code, metadata) {
			console.log({ type: 'inside debug logger', code, metadata });
		},
	},
	providers: [
		GitHubProvider({
			clientId: process.env.GITHUB_CLIENT_ID,
			clientSecret: process.env.GITHUB_CLIENT_SECRET,
		}),
		CredentialsProvider({
			id: 'app-login',
			name: 'App Login',
			credentials: {
				firstName: {
					label: 'First Name',
					type: 'text',
				},
				lastName: {
					label: 'last Name',
					type: 'text',
				},
				email: {
					label: 'Email Address',
					type: 'email',
				},
				password: {
					label: 'Password',
					type: 'password',
				},
			},
			async authorize(credentials, req) {
				console.log('Starting the signup/login process ---');
				try {
					let maybeUser = await prisma.user.findFirst({
						where: {
							email: credentials.email,
						},
					});

					// register process calling
					if (credentials.methodType === 'register') {
						if (!maybeUser) {
							if (
								!credentials.password ||
[...]

Issue #3729, opened 2022-01-25
Credentials does not work with `strategy` = `database`


KATT:
### Description 🐜

Credentials don't seem to work with strategy database and just tries to create a JWT token

https://github.com/nextauthjs/next-auth/blob/4824f8c02ad1b0799f4c15eea9c86e93fc5c72fc/src/core/routes/callback.ts#L398-L417

### Is this a bug in your own project?

No

### How to reproduce ☕️

- Use Credentials provider
- Use strategy database

### Environment 🖥

```bash

  System:
    OS: macOS 12.0.1
    CPU: (8) x64 Apple M1
    Memory: 86.30 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    Yarn: 1.22.15 - /usr/local/bin/yarn
    npm: 8.3.2 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 97.0.4692.99
    Firefox: 94.0.1
    Safari: 15.1
  npmPackages:
    next: ^12.0.7 => 12.0.7
    next-auth: ^4.1.2 => 0.0.0-semantically-released
    react: ^18.0.0-rc.0 => 18.0.0-rc.0

```

### Contributing 🙌🏽

Yes,  I am willing to help solve this bug in a PR

balazsorban44:
This is intended: https://next-auth.js.org/providers/credentials

We encourage the flow in general.

KATT:
It works with jwt though? And the other provider (Google) that I'm using works without any extra hoops.

How and where is a session created? I don't see any docs on it.

KATT:
Oh, it's disabled, I see.

KATT:
@balazsorban44 
If I want to make my own adapter that works with a user/password strategy which database sessions, what would the steps be? I've made it so it works in jwt mode already

jrmyio:
Just waited a couple hours on this because it silently fails.

@balazsorban44 This should be opened as the current behavior is clear bugged. 

Setting an adapter sets "session" as the default strategy, which in turn makes credentials login fail.
When you signin with username/password, it does create a 'next-auth.session-token'-cookie but the database does not contain that session.

But worse, when you do GET /api/auth/session, your 'next-auth.session-token' cookie is instantly destroyed.

jrmyio:
[...]

PR #4036, opened 2022-02-21
Add option to customize the handling of routes in order to extend or fully overwrite current behavior

jrmyio:
This PR adds an option to overwrite and extend the default route handling by adding a `routesHandler` prop in the `NextAuthOptions`. By default it uses the current behavior: https://github.com/nextauthjs/next-auth/blob/a72f1b6d21da89b4223542006c865862635c027b/packages/next-auth/src/core/index.ts#L89-L235

However, when a `routesHandler` is configured, one can completely overwrite or extend the above block and re-use the default implementation by calling the default handler function `import { defaultRoutesHandler } from 'auth-next` .

## Reasoning 💡

In its current form there are no ways to extend or change next-auth's routes. Doing this on the framework level (for example in Next itself) is not an option as you would have to re-implement tons of functionality that is called before and after the routes are handled, for example the initialization of the `InternalOptions`:
- https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/core/init.ts
- https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/core/index.ts#L49-L88 

The initialization of the session store:
https://github.com/nextauthjs/next-auth/blob/a72f1b6d21da89b4223542006c865862635c027b/packages/next-auth/src/core/index.ts#L83-L87

And all the wrapping done by the next.js implementation for authNextHandler:
https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/next/index.ts#L16-L60

By adding an option for devs to overwrite the handling of routes the core becomes a lot more flexible. It gives devs a way to start exploring more functionalities without having to maintain a fork of next-auth, like we already can with `adapters` and `providers`.

For example, this PR could help people with implementing:
- custom `provider types`
- database session handling for the `credentials`-type 
- more routes where they see fit, for example a  `change e-mail` - function.

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

If there is interest in this feature I can flesh out the details.

## Affected issues 🎟

This does give people a way to start handling `credentials` differently add possibly add `database`-session support themselves or as part as a different package / plugin:
https://github.com/nextauthjs/next-auth/issues/3729
https://github.com/nextauthjs/next-auth/issues/3970
https://github.com/nextauthjs/next-auth/discussions/3787

This could be a start for allowing custom types:
https://github.com/nextauthjs/next-auth/discussions/3275

### DIFF
--- a/packages/next-auth/src/core/handle-routes.ts
+++ b/packages/next-auth/src/core/handle-routes.ts
@@ -1,0 +1,164 @@
+import logger from "../lib/logger"
+import * as routes from "./routes"
+import renderPage from "./pages"
+
+import { RoutesHandler } from "./types"
+import type { ErrorType } from "./pages/error"
+
+export const defaultRoutesHandler: RoutesHandler = async ({
+  req,
+  cookies,
+  options,
+  sessionStore,
+  userOptions,
+  error,
+}) => {
+  const { method, action } = req
+  if (method === "GET") {
+    const render = renderPage({ ...options, query: req.query, cookies })
+    const { pages } = options
+    switch (action) {
+      case "providers":
+        return (await routes.providers(options.providers)) as any
+      case "session": {
+        const session = await routes.session({ options, sessionStore })
+        if (session.cookies) cookies.push(...session.cookies)
+        return { ...session, cookies } as any
+      }
+      case "csrf":
+        return {
+          headers: [{ key: "Content-Type", value: "application/json" }],
+          body: { csrfToken: options.csrfToken } as any,
+          cookies,
+        }
+      case "signin":
+        if (pages.signIn) {
+          let signinUrl = `${pages.signIn}${
+            pages.signIn.includes("?") ? "&" : "?"
+          }callbackUrl=${options.callbackUrl}`
+          if (error) signinUrl = `${signinUrl}&error=${error}`
+          return { redirect: signinUrl, cookies }
+        }
+
+        return render.signin()
+      case "signout":
+        if (pages.signOut) return { redirect: pages.signOut, cookies }
+
+        return render.signout()
+      case "callback":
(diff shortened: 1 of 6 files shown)


balazsorban44:
There are multiple ways of going about this, this is not the one. First of all, a feature request should be opened as an issue first, linking to any additional context, providing a good example of why the change is needed.


Also, you can already achieve custom behavior like this through https://next-auth.js.org/configuration/initialization#advanced-initialization

You can define your files in `api/auth/` that will completely override the NextAuth.js functionality that's defined here: https://next-auth.js.org/getting-started/rest-api

jrmyio:
> There are multiple ways of going about this, this is not the one. First of all, a feature request should be opened as an issue first, linking to any additional context, providing a good example of why the change is needed.

I am sorry I missed the part of an issue being required before a PR. 

> Also, you can already achieve custom behavior like this through https://next-auth.js.org/configuration/initialization#advanced-initialization

This is basically what I would call 'at framework level', meaning you cannot use the abstractions on top of the req/res that are being used by next-auth internally (IncomingRequest/OutgoingResponse). Additionally, in my PR text I've listed some links to code blocks and files, for example the initialization of the internal configuration, which is not finalized yet when you are handling own route handling inside `/api/auth`.

I can see how adding another way to deal with routes, while one could use https://next-auth.js.org/configuration/initialization#advanced-initialization, is not really the way to go.  

Maybe it makes more sense  that more internal parts are exported as simple functions so that they can be combined again for your custom functionality when using https://next-auth.js.org/configuration/initialization#advanced-initialization.

For example: https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/core/init.ts 

Would it make sense to make a Feature Request issue for this instead of this PR?

PR #4053, opened 2022-02-23
chore(docs): small css update

ndom91:
## Reasoning 💡

Fixes

image

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged


ThangHuuVu:
IMO we can keep the CSS unchanged and the actual fix should be for this line:
https://github.com/nextauthjs/next-auth/blob/1f07ff05d6c551ed87f40eaabd46f8286a95ceea/docs/src/pages/index.js#L96
It's semantically better because the start count `span` is put inside of the GitHub button instead of the NPM button. WDYT?

balazsorban44 (approved):

PR #4064, opened 2022-02-24
[TypeORM adapter] Widen refresh_token type to permit null

trptcolin:
This updates `refresh_token` in the TypeORM adapter's default entities (and associated tests and docs) to be `string |
null` rather than `string`. The field was already nullable in the database, so this only affects the TypeScript type, not anything in the database.

Because the type is being widened, not narrowed, this should also be fully backwards-compatible, since pre-existing custom entities of type `string` will be substitutable for the new default type of `string | null`. I've verified this locally by running the [sqlite] tests with `refresh_token` in `custom-entities.ts` set to both `string` *and* `string | null`, and it typechecks and passes in both cases.

## Reasoning 💡

The existing type of `refresh_token` here, `string`, does not allow for the possibility of `NULL`, but the database field itself is nullable, so it should. See #4055 for more details on rationale.

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## Affected issues 🎟

Fixes #4055 - see that ticket for more details on the user-facing impact.

### DIFF
--- a/docs/docs/adapters/typeorm.md
+++ b/docs/docs/adapters/typeorm.md
@@ -116,7 +116,7 @@ export class AccountEntity {
   providerAccountId!: string
 
   @Column({ type: "varchar", nullable: true })
-  refresh_token!: string
+  refresh_token!: string | null
 
   @Column({ type: "varchar", nullable: true })
   access_token!: string | null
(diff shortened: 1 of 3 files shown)


balazsorban44 (approved):
Thanks!

Issue #4030, opened 2022-02-21
Prisma Adapter on MongoDB errors out when updating user after login


ferrohd:
### Adapter type

@next-auth/prisma-adapter

### Environment

- next: 12.1.0
- next-auth: 4.2.1
- @next-auth/prisma-adapter: 1.0.1
- react: 17.0.2

### Reproduction URL

https://gist.github.com/ferrohd/d769b39194f03844d18e2bcfe85a61e6

### Describe the issue

I'm using PrismaAdapter with MongoDB as underlying database. I don't know if it's because of the prisma.model I'm using, but Prisma doesn't allow to update the "id" field of a document in a collection. When logging-in with an already registered user, the `updateUser` function is called, and tries to update the `emailVerified` field in the User model.

The `data` argument passed to the function has values:
```typescript
{
      id: '62057f64643a91216fcd9f73', // id of the user to update
      emailVerified: 2022-02-21T08:43:23.333Z
}
```
The implementation of updateUser in @next-auth/prisma-adapter ([here](https://github.com/nextauthjs/next-auth/blob/3ec3761634f7953acc52d460e8bb0b2a678d7703/packages/adapter-prisma/src/index.ts#L16)) passes all the data to the update method, including the id
```typescript
updateUser: (data) => p.user.update({ where: { id: data.id }, data })
```
Which causes the following error:
```
Invalid `prisma.user.update()` invocation:

{
  where: {
    id: '62057f64643a91216fcd9f73'
  },
  data: {
    id: '62057f64643a91216fcd9f73',
    ~~
    emailVerified: new Date('2022-02-21T10:58:02.657Z')
  }
}

Unknown arg `id` in data.id for type UserUpdateInput. Available args:

type UserUpdateInput {
  name?: String | NullableStringFieldUpdateOperationsInput | Null
  email?: String | NullableStringFieldUpdateOperationsInput | Null
  emailVerified?: DateTime | NullableDateTimeFieldUpdateOperationsInput | Null
  image?: String | NullableStringFieldUpdateOperationsInput | Null
  token?: Int | IntFieldUpdateOperationsInput
  accounts?: AccountUpdateManyWithoutUserInput
  sessions?: SessionUpdateManyWithoutUserInput
}
```

## Working Fix
```typescript
updateUser: (data) => {
    const { id, ...rest } = data;
    return p.user.update({ where: { id }, data: rest })
}
```

### How to reproduce

[...]

PR #4066, opened 2022-02-24
Prevent updateUser to add id twice

stophecom:
#### MongoDB Adapter
I noticed that when using the `updateUser` function, the `id` is basically copied and stored as an additional property. This seems unnecessary. This PR fixes this issue for the MongoDB adapter, I haven't checked if the same pattern exists within other adapters. In any case, it is not a big issue, but may save some bytes.

Screen_Shot_2022-02-23_at_22_01_10

## Checklist 🧢

~~- [ ] Documentation~~
~~- [ ] Tests~~
- [x] Ready to be merged


ThangHuuVu:
Thank you for opening this PR, I can confirm this symptom when using `mongodb-adapter`. However, I doubt this PR would fix #4030 since it's a problem with either Prisma or our `prisma-adapter`. We're still investigating together with the Prisma team on #4030; and depending on the outcome this PR might not be needed. I suggest we revisit this PR once #4030 is solved. cc @balazsorban44

balazsorban44:
Looks like we are not currently running the tests and build for this in GitHub Actions. After trying this branch out locally, unfortunately, the changes don't seem to be sufficient. There are both TS and test errors.

stophecom:
@balazsorban44 I didn't dig into the code. The PR would only solve a very narrow case. But I guess it will be obsolete with https://github.com/nextauthjs/next-auth/pull/4082/files. So feel free to close :)

stophecom:
Now when I look at it again, I see there is more to it: (`_id` vs. `id`) 
Closing this.

PR #4085, opened 2022-02-28
Update securing-pages-and-api-routes.md

DiamondeX:
I'm not sure, but it looks like the `null`  should be returned in case of server-side run, that is with `typeof window === "undefined"`

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged


balazsorban44 (approved):
You are correct, although this whole page needs a refresh or should be removed entirely in favor of our Middleware docs, which is a way cleaner method to protect pages and API routes!

https://next-auth.js.org/configuration/nextjs#middleware

PR #4097, opened 2022-03-02
fix(providers): change type of EmailConfig

nkg168:
## Reasoning 💡

This change allow to set the [well-known service](https://github.com/nodemailer/nodemailer/blob/master/lib/well-known/services.json) in the mail settings.

Types for argument of createTransport is here.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/nodemailer/index.d.ts#L54

## Checklist 🧢

- ~[ ] Documentation~
- ~[ ] Tests~
- [x] Ready to be merged


packages/next-auth/src/providers/email.ts
```diff
@@ -1,13 +1,13 @@
 import { createTransport } from "nodemailer"
 
 import type { CommonProviderOptions } from "."
-import type { Options as SMTPConnectionOptions } from "nodemailer/lib/smtp-connection"
+import type { Options as SMTPTransportOptions } from "nodemailer/lib/smtp-transport"
```
lluia, line 4: Does this type exist? I can't see it on **DefinitivelyTyped**:
https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/nodemailer/lib

Are we use both types have precisely the same options? We don't want this to be a breaking change 💅🏽

0ubbe (changes requested):

packages/next-auth/src/providers/email.ts
nkg168: this type is here.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/nodemailer/lib/smtp-transport.d.ts#L42

Options named SMTPTransportOptions extends Options named SMTPConnectionOptions.

lluia: I see, so this type [`SMTPTransportOptions`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/nodemailer/lib/smtp-transport.d.ts#L42-L46) extends [`SMTPConnectionOptions`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/nodemailer/lib/smtp-connection/index.d.ts#L131-L172) with few more things 🤔 

In that case it's ok for me 👍🏽

0ubbe (approved):

ThangHuuVu:
@lluia shall we merge this? 😄

PR #4147, opened 2022-03-09
docs(oauth): use Mermaid for diagram

0ubbe:
## Reasoning 💡

This is a bit of an experiment, use [`mermaid`](https://mermaid-js.github.io/mermaid/#/) for the diagram we have at:

- https://next-auth.js.org/configuration/providers/oauth

It's much nicer, as being a diagram it's easier to maintain and update instead of an image. 

I used a [sequence diagram](https://mermaid-js.github.io/mermaid/#/sequenceDiagram) to reflect the flow, as not sure I could replicate that image with the tooling mermaid has 🤔 . I think now it's clearer and easier to understand, although longer 😅 . I used **Github** as the auth provider in the example.

## Checklist 🧢

- [x] Documentation
- [ ] ~~Tests~~
- [ ] ~~Ready to be merged~~

## Affected issues 🎟

None


ndom91 (approved):
We went over this together in the call, right? Nice addition! The sequence diagram makes it more understandable than the preexisting diagram imo.

LGTM :+1:

balazsorban44:
Thanks!

PR #4149, opened 2022-03-09
fix: Reduce required node engine to ^16.10.0

MIreland:
This PR reduces the required node engine to 16.10.

Jest has memory leaks in Node 16.11.0+, as documented [in this ticket](https://github.com/facebook/jest/issues/11956#issuecomment-1034135234)

Our monorepo currently is pegged at Node 16.10, and we would like the ability to install next-auth without needing to run `yarn install --ignore-engines`.

## Reasoning 💡

When we upgrade past Node 16.10, the memory leak causes our test suite to crash.

Once Node 18 releases, we are optimistic that we can use that instead- the memory leak does not currently occur in Node 17.

## Checklist 🧢

~- [ ] Documentation~
~- [ ] Tests~
- [ ] Ready to be merged


balazsorban44:
We are following the settings based on our dependency https://github.com/panva/node-openid-client/blob/a315ad4e4d97ac8bfbaa2e7b41c5225240364df7/package.json#L74

for now, you can try to ignore the engines setting, but there is no guarantee for next-auth to work properly then.

PR #4167, opened 2022-03-12
chore(docs): 1st party guides section

ndom91:
## Reasoning 💡

As discussed, moved the 1st party guides/tutorials into their own sidebar section titled "Guides". I'd just reused the headers "Basics", "Fullstack", "Databases" as we had them in the tutorials page, but I'm open to anything else if you have a better idea :+1: 

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged


balazsorban44 (approved):
I thought we would also rewrite URLs like `/tutorials/usage-with-class-components` to `/guides/usage-with-class-components`. Probably good start anyway, so I'll approve & merge.

PR #4169, opened 2022-03-12
chore(docs): add sandpack to /configuration/pages examples

ndom91:
Based upon the initial `@codesandbox/sandpack` support, added it to our `/configuration/pages` page to allow users to customize and play with the custom pages support, as well as theming.

STILL WIP ATM - Having issues with the custom files / custom dependency setup. 

See: https://next-auth-git-ndom91-add-sandpack-custom-page-c606fc-nextauthjs.vercel.app/configuration/pages

Notes:
- You can specify a `template` (like this [`react-ts`](https://github.com/codesandbox/sandpack/blob/main/sandpack-react/src/templates/react-typescript.ts) one) or a custom `entry` point, at which point you have to provide all the deps and configs yourself..
- https://sandpack.codesandbox.io/docs/getting-started/custom-content
- https://sandpack.codesandbox.io/docs/api/react/interfaces/SandpackSetup
- https://sandpack.codesandbox.io/docs/api/react/#sandboxenvironment

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged


docs/src/components/Sandpack.jsx
```diff
@@ -6,11 +6,89 @@ import {
...
-  <SandpackProvider template="react">
-    <SandpackLayout>
-      <SandpackCodeEditor />
-      <SandpackPreview />
+  <SandpackProvider
+    customSetup={{
+      dependencies: {
+        react: "17.0.2",
+        "react-dom": "17.0.2",
+        next: "^12.0.11-canary.4",
```
danilowoz, line 28: Hey, I'm sorry if I didn't make myself very clear in [this commentary](https://github.com/codesandbox/sandpack/issues/359#issuecomment-1031513557) and [here](https://sandpack.codesandbox.io/docs/faq#does-sandpack-support-nextjs-or-other-server-side-applications). Sandpack doesn't support any dependency that needs a server to run (but only runtime projects). I thought you had different purposes for Sandpack here, so I'm afraid to say that CodeSandbox Embed is still the best option.

I also addressed this issue on Sandpack project, to avoid this kind of assumption again https://github.com/codesandbox/sandpack/pull/410

ndom91: Ah yes you're right, must have forgotten that piece. Thanks anyway for all the support so far!

ndom91:
Closing this for now!

PR #4170, opened 2022-03-13
chore(providers): Add typings for Discord Provider

ryanbarr:
## Reasoning 💡
The current Discord provider is not TypeScript compatible out-of-the-box and requires developers to declare their own typing and interface. This change does the following:

1. Renames `providers/discord.js` to `providers/discord.ts`.
2. Creates, implements, and exports a `DiscordProfile` interface.
3. Updates the `Discord()` provider to support `OAuthUserConfig` and `OAuthConfig`.

**Note:** These changes were modeled off of `providers/twitch.ts`.

## Checklist 🧢

- ~[ ] Documentation~
- ~[ ] Tests~
- [x] Ready to be merged

This is ready to be merged as it does not introduce breaking changes or API changes and is backwards compatible with the latest stable version. This has been tested by launching the `dev` app, updating the `.env.local` with Discord API credentials, and testing Discord login inside the test rig.

## Affected issues 🎟

None.

### DIFF
--- a/packages/next-auth/src/providers/discord.ts
+++ b/packages/next-auth/src/providers/discord.ts
@@ -1,5 +1,26 @@
-/** @type {import(".").OAuthProvider} */
-export default function Discord(options) {
+import type { OAuthConfig, OAuthUserConfig } from "."
+
+export interface DiscordProfile {
+  accent_color: number
+  avatar: string
+  banner: string
+  banner_color: string
+  discriminator: string
+  email: string
+  flags: number
+  id: string
+  image_url: string
+  locale: string
+  mfa_enabled: boolean
+  premium_type: number
+  public_flags: number
+  username: string
+  verified: boolean
+}
+
+export default function Discord<P extends DiscordProfile>(
+  options: OAuthUserConfig<P>
+): OAuthConfig<P> {
   return {
     id: "discord",
     name: "Discord",
--- a/packages/next-auth/src/providers/discord.ts
+++ b/packages/next-auth/src/providers/discord.ts
@@ -1,6 +1,6 @@
 import type { OAuthConfig, OAuthUserConfig } from "."
 
-export interface DiscordProfile {
+export interface DiscordProfile extends Record<string, any> {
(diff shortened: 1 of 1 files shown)


packages/next-auth/src/providers/discord.ts
```diff
@@ -1,5 +1,26 @@
...
+  image_url: string
+  locale: string
+  mfa_enabled: boolean
+  premium_type: number
+  public_flags: number
+  username: string
+  verified: boolean
+}
+
+export default function Discord<P extends DiscordProfile>(
```
lluia, line 21: I think the safest way to type this would be:
```suggestion
export default function Discord<P extends Record<string, any> = DiscordProfile>(
```
It's the pattern we use in other providers as well 👍🏽

0ubbe (changes requested):
Thanks 💚 , one small request 💬

packages/next-auth/src/providers/discord.ts
ryanbarr: Hey @lluia! Thanks for the suggestion -- perhaps you can provide some guidance here:

I attempted to provide this typing, but it actually causing a typing error on lines 14 and 17 (35 and 38 in the new code) where `profile.image_url` has its value set. Honestly, the error is a bit confusing because the typing should be correct.

Any idea what the reasoning is? Note that it's fine on line 23 (44 in the new code).

ryanbarr: @lluia Good catch! I've made the adjustments as requested.

If I find time this week, I'll look into adjusting and testing the other providers' interfaces as well to have the correct typing. Appreciate your help!

ryanbarr: Going to mark this as resolved, let me know if you need anything else!

ThangHuuVu:
@ryanbarr @lluia I think this one's ready, let's get it merged! 👍

mshd (approved):

balazsorban44 (approved):

PR #4182, opened 2022-03-14
docs(adapters): add dynamoDB cf ProjectionType

calinalexandru:
## Reasoning 💡

Hello,
Firstly, a huge thanks for creating and maintaining this awesome project!

I encountered a small error while using the DynamoDB adapter only related to the cloudformation.yaml example. If you choose to declare GlobalSecondaryIndexes then you must also specify ProjectionType to ALL, otherwise, it will default to KEYS_ONLY

Using KEYS_ONLY methods from the adapter that rely on that global index (eg: getUserByAccount) will fail. 

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/docs/docs/adapters/dynamodb.md
+++ b/docs/docs/adapters/dynamodb.md
@@ -119,6 +119,8 @@ NextAuthTable:
         KeyType: RANGE
     GlobalSecondaryIndexes:
       - IndexName: GSI1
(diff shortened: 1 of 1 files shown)


ndom91 (approved):
Good catch! :+1:

Issue #4183, opened 2022-03-14
Update @upstash/redis in adapter-upstash-redis


chronark:
### Description 📓

We just released [@upstash/redis v1.0.0](https://github.com/upstash/upstash-redis/releases/tag/v1.0.0) with lots of improvements to the typesystem and pipeline support.

Unfortunately the changes are not backwards compatible, but because the underlying REST API did not change, there is no rush in upgrading. The current version will continue to work.

I'm happy to creating a PR myself to upgrade everything in a few weeks but maybe someone beats me to it :)

### How to reproduce ☕️

Nothing to reproduce :)

### Contributing 🙌🏽

Yes, I am willing to help implement this feature in a PR

balazsorban44:
Yes, PR is welcome! :clap: We can even avoid breaking changes, if it's easy to infer the installed lib version, and use the correct methods under the hood. It might be useful if you want people to migrate slowly. Otherwise, we can just do a major release as well, up to you. :+1:

chronark:
yes it would be fairly easy to infer, because the new version exports different methods, I'll think about it.

Lucasvo1:
I'm using this adapter with Azure AD but the AD returns the profile picture as an ArrayBuffer and not as an url. So, the adapter cannot read that field.

PR #4201, opened 2022-03-16
fix(docs): pin @upstash/redis version to 0.2.1

chronark:
Until we migrate to v1 there might be people installing the latest
version of @upstash/redis and then encountering errors.
With this they will at least get a warning.

This is obsolete and should be updated once
https://github.com/nextauthjs/next-auth/issues/4183 is resolved.


balazsorban44 (approved):
Sounds good!

Issue #3350, opened 2021-12-03
profile callback argument type


estarossa0:
### Question 💬

When trying to use the profile callback (42-school provider) I couldn't use the profile type as it was `any` even that most the profiles are now typed in v4 and have `P extends Record<string, any> = ProviderProfile` unless you explicitly typed it with generic `Provider<ProviderProfileType>`
![image](https://user-images.githubusercontent.com/57002508/144659056-50bc14f8-fbc2-4754-ad81-5122885f6960.png)

### How to reproduce ☕️

By using any provider with a typed profile, and then adding the profile callback

```ts
import NextAuth from "next-auth";
import FortyTwoProvider from "next-auth/providers/42-school";

export default NextAuth({
  providers: [
    FortyTwoProvider({
      clientId: process.env.FORTY_TWO_CLIENT_ID,
      clientSecret: process.env.FORTY_TWO_CLIENT_SECRET,
      profile(profile) {
        return {
          id: profile.login,
          name: profile.usual_full_name,
          email: profile.email,
          image: profile.image_url,
          login: profile.login,
        };
      },
    }),
  ],
});
```

### Contributing 🙌🏽

No, I am afraid I cannot help regarding this

estarossa0:
any response considering this topic 😅 @balazsorban44

ThangHuuVu:
I think the reason is because in the source code `next-auth` define the `Provider` type of `providers: Provider[]` array as following:
```ts
Provider = OAuthConfig<any> | EmailConfig | CredentialsConfig
```

All elements in the `providers` array will take this type. In this case, to use callback type, I think the only option is to explicitly define the Generic Type for the provider function; I don't think there's a way for TS to automatically infer the Generic of array element:
 
```ts
import NextAuth from "next-auth";
import FortyTwoProvider,  { FortyTwoProfile } from "next-auth/providers/42-school";

export default NextAuth({
  providers: [
    FortyTwoProvider<FortyTwoProfile>({ // explicitly define the Generic Type for the provider function;
      clientId: process.env.FORTY_TWO_CLIENT_ID,
      clientSecret: process.env.FORTY_TWO_CLIENT_SECRET,
      profile(profile) { // profile is of type FortyTwoProfile
        return {
          id: profile.login,
          name: profile.usual_full_name,
          email: profile.email,
          image: profile.image_url,
          login: profile.login,
        };
      },
    }),
  ],
});
```

@estarossa0 Does this make sense?

[...]

PR #4202, opened 2022-03-16
fix(providers): profile types

0ubbe:
## Reasoning 💡

When reviewing this PR https://github.com/nextauthjs/next-auth/pull/4170 we noticed that the types in our provider files [were incorrect](https://github.com/nextauthjs/next-auth/pull/4170/files#r828419058).

When declaring the options for the provider profile, we were doing this:
```ts
export default function Discord<P extends Record<string, any> = DiscordProfile>(
```
but **TS** is not able to infer the types correctly as you can see:

Screenshot 2022-03-16 at 21 32 15

However changing it like this:
```ts
interface DiscordProfile extends Record<string, any> {
  // ...
}

export default function Discord<P extends DiscordProfile>(
```

enables TS to pick up the types correctly:

Screenshot 2022-03-16 at 21 32 00

I believe both syntaxes are more or less equivalent to what we want to achieve which is typing the profile object of the provider but allowing extra fields for user flexibility.

## Checklist 🧢

- [ ] ~~Documentation~~
- [ ] ~~Tests~~
- [x] Ready to be merged

## Affected issues 🎟

None

### DIFF
--- a/packages/next-auth/src/providers/42-school.ts
+++ b/packages/next-auth/src/providers/42-school.ts
@@ -134,8 +134,7 @@ export interface CampusUser {
   created_at: string
   updated_at: string | null
 }
-
-export interface FortyTwoProfile extends UserData {
+export interface FortyTwoProfile extends UserData, Record<string, any> {
   groups: Array<{ id: string; name: string }>
   cursus_users: CursusUser[]
   projects_users: ProjectUser[]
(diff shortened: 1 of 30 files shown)


ThangHuuVu (approved):
LGTM! I think this would solve https://github.com/nextauthjs/next-auth/issues/3350 too. I wasn't sure how to solve that issue back then. Nice work! 👍

ndom91 (approved):
Yeah great catch!

PR #4210, opened 2022-03-17
feat: custom jwt decode method for middleware

hinsxd:
## Reasoning 💡

As mentioned in #4181, having a custom `jwt.decode` method in `[...nextauth].ts` will cause errors in any `_middleware.ts` when the middleware tries to call `getToken()` internally without the correct decode method.

## Checklist 🧢
- [x] Added `jwt.decode` option to middleware and pass it to `getToken`
- [x] Added documentation to Configuration > Next.js > Middleware to address the problem
- [x] Added documentation to Configuration > Options > Override JWT encode and decode methods to link to the above section
- [x] Tested by hand

## Affected issues 🎟

Fixes #4181


ndom91:
Yeah this makes a lot of sense to me, @lluia what do you think?

docs/docs/configuration/options.md
```diff
@@ -482,6 +482,8 @@ Using a custom cookie policy may introduce security flaws into your application
 
 NextAuth.js uses encrypted JSON Web Tokens ([JWE](https://datatracker.ietf.org/doc/html/rfc7516)) by default. Unless you have a good reason, we recommend keeping this behaviour. Although you can override this using the `encode` and `decode` methods. Both methods must be defined at the same time.
 
+**IMPORTANT: If you middleware to protect routes, make sure the same method is also set in the [`_middleware.ts`](/configuration/nextjs#custom-jwt-decode-method)**
```
ThangHuuVu, line 485:
```suggestion
**IMPORTANT: If you use middleware to protect routes, make sure the same method is also set in the [`_middleware.ts` options](/configuration/nextjs#custom-jwt-decode-method)**
```

ThangHuuVu (approved):
LGTM, I found a typo 🙏

agustif:
Awesome! I will probably need this as I haven't yet got around implementing my own JWT callback but might want to do so in the future to allow for custom options passing to it, like extended or no expiry date, and some metadata (like a user provided name for the token). I plan to re-use sessions as Personal Access Tokens for an API. and that's how I plan to create such tokens.

Thanks for the pr @hinsxd let's hope it can get merged =)

hinsxd:
@ThangHuuVu Thanks for spotting the typo!

ndom91:
Hey so this looks great, however we're going to hold off on merging and releasing a new version for a bit until Balazs is back from break. We appreciate the contribution though and will keep you posted!

balazsorban44 (approved):
Thank you! Looks good! :raised_hands:

Husain010:
So I have a custom JWT method to encode and decode the JWT, but it's not able to decode the token, I tried by using the default JWT token that nextAuth provides as well as the JWT that I create, it only works with the one nextAuth creates! 

I also tried creating an API to fetch the token using the getToken function, and if I pass {raw: true} it gives the raw token, but when I pass it as false it gives null, what's the issue exactly?

PR #4225, opened 2022-03-20
Make getToken also support Basic HTTP Auth.

agustif:
## Reasoning 💡
Hey there! ✋🏻 I think it would be really nice to also support HTTP Basic Auth by passing the session token. It basically searches for either "Bearer or "Basic" instead of just supporting Bearer token. 

I think this doesn't add much code to support in the future and is a nice extra way to support auth besides cookies and Bearer token!
👍🏻 

```
curl "http://localhost:3000/api/jwt" \
  -u 'eyJ....Session token Here'

```

## Checklist 🧢

~- [ ] Documentation~
~- [ ] Tests~
- [x] Ready to be merged

### DIFF
--- a/packages/next-auth/src/jwt/index.ts
+++ b/packages/next-auth/src/jwt/index.ts
@@ -90,7 +90,7 @@ export async function getToken<R extends boolean = false>(
 
   let token = sessionStore.value
 
-  if (!token && req.headers.authorization?.split(" ")[0] === "Bearer") {
+  if (!token && req.headers.authorization?.split(" ")[0] === "Bearer" || "Basic") {
     const urlEncodedToken = req.headers.authorization.split(" ")[1]
     token = decodeURIComponent(urlEncodedToken)
   }
--- a/packages/next-auth/src/jwt/index.ts
+++ b/packages/next-auth/src/jwt/index.ts
@@ -89,10 +89,10 @@ export async function getToken<R extends boolean = false>(
   )
 
   let token = sessionStore.value
-
-  if (!token && req.headers.authorization?.split(" ")[0] === "Bearer" || "Basic") {
-    const urlEncodedToken = req.headers.authorization.split(" ")[1]
-    token = decodeURIComponent(urlEncodedToken)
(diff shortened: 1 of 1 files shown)


ThangHuuVu:
Although the code looks good, I'm not sure if we should merge it because of the security implication. From our POV, we want to encourage best practices. (As I understand, Basic/Digest Auth is more vulnerable because the password is sent over plain text repeatedly).
@ndom91 @lluia, what do you guys think?

agustif:
> Although the code looks good, I'm not sure if we should merge it because of the security implication. From our POV, we want to encourage best practices. (As I understand, Basic/Digest Auth is more vulnerable because the password is sent over plain text repeatedly). @ndom91 @lluia, what do you guys think?

Actually the code is missing the encoding, will add now.

I just wanted to support both Bearer token and basic auth for an API I'm building, much like stripe supports https://stripe.com/docs/api

Final user can decide which one to use, but I do agree security is important so let me know what you decide on this.

Best regards,
Agusti

ndom91:
So as far as I understand it, the HTTP Basic Auth is using the same header (`Authorization`) as the bearer auth we're all used to, but its just the `username:password` appended to one another like this and base64 encoded.

i.e. 
Basic Auth = `Authorization: 'Basic ${atob(username:password)}'`
Current version (Bearer) = `Authorization: 'Bearer eyABC123'`

Therefore, they're are both encrypted in transit when using HTTPS. They're also both sent on every request. So technically I think it doesn't make much of a difference if you want to use, for example, the username portion of the http basic auth to transmit your JWT or the "Bearer" variant of that header.

Seeing as this opens up a new "feature" / way of using the library, I dont feel comfortable merging this without Balazs. I'll ask him next time I talk to him.

agustif:
Hey, I will no longer be require suing the Basic Auth for now, so I'll be fine if it doesn't get merged upstream finally, if it does happy to have helped!

Thank you for the awesome lib ;)

lluia:
Looks good to me. However, we should wait on whether more people request this and then only add it, as it adds yet another branching logic in our code 😬 ( little by little, they pile up! 🧱 )

PR #4236, opened 2022-03-23
fix(fetch): fix fetch credentials default value for old browsers

maximevast:
## Reasoning 💡

Old browsers use a version of fetch that [does not send the cookies alongside the request](https://caniuse.com/mdn-api_request_credentials_default_same-origin) if you don't specify `credentials: 'same-origin'`.

If you try a basic next-auth project in one of these old browsers, you'll have a weird behaviour, no error or warning, and will not be able to login.

I think next-auth code is expecting the defaut value of [Request.credentials](https://developer.mozilla.org/fr/docs/Web/API/Request/credentials) to always be `same-origin` but it is set to `omit` before chrome 72 and firefox 61. 

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/packages/next-auth/src/client/_utils.ts
+++ b/packages/next-auth/src/client/_utils.ts
@@ -39,7 +39,10 @@ export async function fetchData<T = any>(
     const options = req?.headers.cookie
       ? { headers: { cookie: req.headers.cookie } }
       : {}
-    const res = await fetch(`${apiBaseUrl(__NEXTAUTH)}/${path}`, options)
+    const res = await fetch(`${apiBaseUrl(__NEXTAUTH)}/${path}`, {
+      ...options,
+      credentials: "same-origin",
+    })
     const data = await res.json()
     if (!res.ok) throw data
     return Object.keys(data).length > 0 ? data : null // Return null if data empty


ndom91:
Thanks for your interest in the project and contribution!

Can you define what you mean by 'older browsers' exactly though?

maximevast:
> Can you define what you mean by 'older browsers' exactly though?

Capture d’écran 2022-03-25 à 10 00 55

You can see which version of the browsers started implementing `same-origin` as a default value for the credentials parameter [here](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials#browser_compatibility)

ndom91:
Okay these really do look like super old browsers. According to [browserslist](https://browserslist.dev), all of these latest unsupported versions of chrome, edge, firefox, opera and safari all calculated together aren't even used by 0.05% of users.

image
> https://browserslist.dev/?q=Y2hyb21lIDcxLCBlZGdlIDE3LCBmaXJlZm94IDYwLCBzYWZhcmkgMTIuMCwgb3BlcmEgNDQ%3D

This is however a rather small and innocuous change, so I'm tempted to say lets just add it. 

We're currently on pause with new releases however until everyone on the team is back from break. Therefore, I will mark this PR as "ready" and the primary maintainer can make the final decision when he's back.

Thanks for your PR and interest in the project!

balazsorban44:
I can see how this is a small change, at the same time, as pointed out it targets a very small group (if less than 0.05% of users are on those old versions, NextAuth.js users will be a fraction of that). I wonder what our margin should be in the case of evergreen browsers' older versions.

In such peculiar cases, the users should probably be asked to upgrade anyway as it might even be a security risk at that point (see [Google Issues Emergency Security Update For 3.2 Billion Chrome Users](https://www.forbes.com/sites/daveywinder/2022/03/26/google-confirms-emergency-security-update-for-32-billion-chrome-users-attacks-underway))

Another option is if we make all the fetch options configurable instead, so anyone can pass whatever they want. I can see that having more useful implications.

maximevast:
Just to add some context; we are a BTB company and some of our users are forced to use an old browser because of a business software that only runs on Firefox 52.

However, even with the fix that I proposed, we still encountered differences of behaviour between recent and outdated browsers, and we decided to stop using Next-Auth for the time being, so this is no longer an issue for us.

That being said, I think that making fetch options configurable would be a suitable solution, but for this issue in particular, it also make sense to hardcode the `credentials` value since it is not the default value for all browsers.

I leave you to it, again, thanks for all the work done!

Best

ndom91:
A quick fix may be to just merge this PR as is then. Seeing as `credentials: same-origin` is the default value for modern browsers, this would just enable older browsers to continue to function with NextAuth.js and not break anything for newer browsers. 

In short, it's only a minor addition. Nothing breaking.

Being able to configure all fetch options could be a follow-up PR down the road if theres enough demand, etc. 

What do yall think?

PR #4248, opened 2022-03-24
Experimental Autodesk Forge authentication provider implementation.

thomas-gale:
## Reasoning 💡

Adding an Autodesk Forge authentication provider (https://forge.autodesk.com/)

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

Not ready to merge till the package dependencies (`axios` / `qs`) are reviewed (and the maintainers preferred approach incorporated).

### DIFF
--- a/docs/docs/providers/forge.md
+++ b/docs/docs/providers/forge.md
@@ -36,11 +36,11 @@ const options = {
     ForgeProvider({
       clientId: process.env.FORGE_CLIENT_ID,
       clientSecret: process.env.FORGE_CLIENT_SECRET
-		  authorization: {
-		    params: {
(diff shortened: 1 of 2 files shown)


ndom91:
Thanks for submitting a new provider @thomas-gale and including the documentation!

Regarding `axios` and `qs`, we generally prefer to use `fetch` instead of `axios`, and qs in this case could probably be replaced with `JSON.stringify()` as well, right?

You can checkout our `/packages/next-auth/src/providers/linkedin.ts` or `github.ts` provider code for an example for making additional requests in the provider configuration.

PR #4250, opened 2022-03-24
chore(docs): add carbon ads to docs sidebar

ndom91:
## Reasoning 💡

Adding [CarbonAds](https://www.carbonads.net/) to the docs page. 

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -136,13 +136,6 @@ module.exports = {
       respectPrefersColorScheme: true,
     },
   },
-  // scripts: [
-  //   {
-  //     src: '//cdn.carbonads.com/carbon.js?serve=CEAI6K3N&placement=next-authjsorg',
-  //     id: '_carbonads_js',
-  //     async: true,
-  //   },
-  // ],
   presets: [
     [
       "@docusaurus/preset-classic",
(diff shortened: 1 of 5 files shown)


ndom91:
Alright so the initial setup is complete. 

**(A)** The default setup looks like this:

image

What do yall think?

**(B)** I'm thinking about bumping it down a bit with a `margin-top`, something like this maybe:

image

**(C)** Orrrr with flex-box pin the ad to the bottom of the screen :thinking: 

image

I'm open to whichever, let me know what yall think (A, B, C).

@balazsorban44 @ThangHuuVu @lluia

ThangHuuVu:
I dig the (A) and (B) options 👍 would it be a problem if the ads are not visible by default in the mobile view since they're in the sidebar?

lluia:
**(B)** version looks the more clean to me 💅🏽 ; similar question as @ThangHuuVu about ads being in the sidebar.

ndom91:
Thanks for the quick responses guys! So according to them, they only need to `The ad should be visible above-the-fold (upper 800 pixels) for desktop browser.` So on mobile, being in the sidebar should be fine. In fact i'll test that and potentially hide it completely on mobile if it breaks the slide out navbar in docusaurus. 

I like **B** as well, lets go with that then!

ndom91:
I had to deactivate so much ad-blocking extensions and config and whitelist domains on my pihole for this PR :joy: :joy:

PR #4283, opened 2022-03-29
chore(docs): Fix API directory

wimsikal:
Update the API directory from `pages/api/[...nextauth].ts` to `pages/api/auth/[...nextauth].ts`  so it matches the typescript example.


ndom91 (approved):
Nice catch!

PR #4286, opened 2022-03-29
fix: dropbox provider (#3899)

eliasylonen:
## Reasoning 💡

This change should fix the currently broken Dropbox provider. Dropbox userinfo endpoint uses POST, not GET, which causes the 404 error mentioned in issue #3899.

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## Affected issues 🎟

Fixes #3899

PR #4299, opened 2022-04-01
feat(core): make `next-auth` optionally opt-into Web-compatible mode

ThangHuuVu:
## Reasoning 💡

> 🚧 WIP

This PR removes [openid-client](https://github.com/panva/node-openid-client) dependency and adds [oauth4webapi](https://github.com/panva/oauth4webapi). It would decouple NextAuth.js from Node.js and enable us to run NextAuth.js on other JS runtimes like [the edge runtime](https://nextjs.org/docs/api-reference/edge-runtime).

### Notes on using PKCE and state/nonce
PKCE is always used. See: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#section-2.1.1
If PKCE support is missing (either missing from AS metadata or we are not sure), use `state` in addition. Nonce support for OpenID is WIP at #4100 

## Checklist 🧢

- [x] Replace openid-client with oauth4webapi
- [ ] Handle auth on middleware 
- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## References
https://github.com/panva/oauth4webapi/blob/main/examples/code.ts
https://github.com/panva/oauth4webapi/blob/main/examples/oauth.ts


packages/next-auth/src/core/lib/oauth/state-handler.ts
```diff
@@ -16,7 +15,8 @@ export async function createState(
     return
   }
 
-  const state = generators.state()
+  // TODO: replace with something like generateState()
+  const state = generateRandomCodeVerifier()
```
ThangHuuVu, line 19: Seems like the library is missing this function

packages/next-auth/src/providers/oauth.ts
```diff
@@ -112,8 +98,8 @@ export interface OAuthConfig<P> extends CommonProviderOptions, PartialIssuer {
   version?: string
   profile?: (profile: P, tokens: TokenSet) => Awaitable<User & { id: string }>
   checks?: ChecksType | ChecksType[]
-  client?: Partial<ClientMetadata>
-  jwks?: { keys: JWK[] }
```
ThangHuuVu, line 116: We might not need this anymore, as the new [Client](https://github.com/panva/oauth4webapi/blob/main/docs/interfaces/Client.md) doesn't require it.

packages/next-auth/src/providers/oauth.ts
```diff
@@ -130,9 +116,6 @@ export interface OAuthConfig<P> extends CommonProviderOptions, PartialIssuer {
   region?: string
   // TODO: only allow for some
   issuer?: string
-  /** Read more at: https://github.com/panva/node-openid-client/tree/main/docs#customizing-http-requests */
-  httpOptions?: HttpOptions
```
ThangHuuVu, line 134: This is specific to Node.js

packages/next-auth/src/providers/oauth.ts
```diff
@@ -71,12 +57,12 @@ export type TokenEndpointHandler = EndpointHandler<
...
      * Contains params like `state`.
      */
-    params: CallbackParamsType
+    params: URLSearchParams
     /**
      * When using this custom flow, make sure to do all the necessary security checks.
-     * Thist object contains parameters you have to match against the request to make sure it is valid.
+     * This object contains parameters you have to match against the request to make sure it is valid.
      */
-    checks: OAuthChecks
```
ThangHuuVu, line 79: I wonder if this is still necessary. I couldn't find the equivalent in `oauth4webapi`

packages/next-auth/src/providers/facebook.ts
```diff
@@ -18,13 +18,13 @@ export default function Facebook<
       url: "https://graph.facebook.com/me",
       // https://developers.facebook.com/docs/graph-api/reference/user/#fields
       params: { fields: "id,name,email,picture" },
-      async request({ tokens, client, provider }) {
```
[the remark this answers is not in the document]
balazsorban44: Unfortunately, the `request` method was added (see #1846) to support providers that do not follow the spec properly... It's a shame that Facebook doesn't do it. (Note, it might have changed since, but I haven't checked in a while). If we can get rid of it, I'm all for though. But there might be other providers still using it.

We should discuss if we want to start discouraging (read: **drop support**/`request`) providers that do not implement the spec 100% correctly...

What I absolutely want to avoid is to go back to handling provider-specific stuff in our core. (eg `if (provider.id === "facebook")`, etc.)

packages/next-auth/src/providers/oauth.ts
balazsorban44: It's basically the way to tell if the provider wants to use `state`, `pkce`, both or `none`. (`nonce` support is WIP #4100, but I haven't had the time to have a look yet).

If omitted, our default is to use `state` currently.

packages/next-auth/src/providers/oauth.ts
balazsorban44: This exists to be able to override the usage of `openid-client`. I wonder if we can truly get rid of it and still be flexible. We should discuss.

balazsorban44: Example: https://github.com/nextauthjs/next-auth/pull/4286

How could we solve it otherwise?

packages/next-auth/src/core/lib/oauth/callback.ts
```diff
@@ -5,11 +5,10 @@ import { usePKCECodeVerifier } from "./pkce-handler"
 import { OAuthCallbackError } from "../../errors"
 import {
   authorizationCodeGrantRequest,
+  clientCredentialsGrantRequest,
```
panva, line 8:
```suggestion
```

packages/next-auth/src/core/lib/oauth/pkce-handler.ts
```diff
@@ -63,11 +71,17 @@ export async function createPKCE(options: InternalOptions<"oauth">): Promise<
...
+  authorizationServer: AuthorizationServer,
   options: InternalOptions<"oauth">
 ): Promise<{ codeVerifier: string; cookie: Cookie } | undefined> {
   const { cookies, provider } = options
+  const { code_challenge_methods_supported } = authorizationServer
 
-  if (!provider?.checks?.includes("pkce") || !codeVerifier) {
+  if (
+    !provider?.checks?.includes("pkce") ||
+    !codeVerifier ||
```
panva, line 82:
```suggestion
```

packages/next-auth/src/core/lib/oauth/pkce-handler.ts
```diff
@@ -63,11 +71,17 @@ export async function createPKCE(options: InternalOptions<"oauth">): Promise<
...
+    code_challenge_methods_supported?.length === 0
```
panva, line 83:
```suggestion
    Array.isArray(code_challenge_methods_supported) && code_challenge_methods_supported.includes('S256') === false
```

packages/next-auth/src/core/lib/oauth/callback.ts
```diff
@@ -65,81 +81,110 @@ export default async function oAuthCallback(params: {
...
 
-    // @ts-expect-error
-    if (provider.token?.request) {
-      // @ts-expect-error
+    const response = await authorizationCodeGrantRequest(
+      authorizationServer,
+      client,
+      callbackParameters,
+      provider.callbackUrl,
+      pkce?.codeVerifier as string
```
ThangHuuVu, line 126: I need to ask Filip about this, as I'm not sure how to call this method without the codeVerifier.

panva: You don't, that's the whole point. The API is designed so that you always use S256 PKCE. The discussion we had about AS not supporting PKCE means that you'd use nonce (if openid), or state (if not openid) in addition to PKCE since you don't know if PKCE is supported or not.

ThangHuuVu: Thanks for taking a look, @panva! 🙏

I thought that this API would also handle requesting the access token in the cases PKCE is not supported, like https://www.rfc-editor.org/rfc/rfc6749.html#section-4.1 and https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth. How would you suggest we make the token request in those cases?

panva: The client can (and as per latest security BCP MUST) always use PKCE.

panva: You just always use PKCE.

panva: The client always uses PKCE. And in cases where you're not sure if PKCE is supported on the AS you still use PKCE but add nonce (when using OIDC) or state (when not using OIDC).

ThangHuuVu: Got it, I just updated the PKCE handler to always return a value.  🙏

david58:
With the upcoming end of the active phase of node 16, is there a plan to move this forward as it seems to be the main blocker of official support of higher versions?

ThangHuuVu:
@david58 there is another prior to this PR: https://github.com/nextauthjs/next-auth/pull/4769 👀 Hope to ship this soon, too 🙌

balazsorban44:
Superseded by #5536

PR #4305, opened 2022-04-02
feat(middleware): add option to ignore specified paths

callumok2004:
## Reasoning 💡

This PR simply adds another option to middleware allowing people to specify path's to be ignored. Doing this allows paths that do not require middleware to not have to wait for any backend checks to see if the user is authenticated, nor does any custom middleware that isnt used on said pages have to be ran.

**Before**
jm25j 1

**After**
anr47 1

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

This is my first pull request in this repository, hopefully I did everything correctly :pray:

### DIFF
--- a/apps/example-nextjs/pages/admin/_middleware.ts
+++ b/apps/example-nextjs/pages/admin/_middleware.ts
@@ -1,8 +1,11 @@
-import { withAuth } from "next-auth/middleware"
+import { withAuth } from "next-auth/middleware";
 
 // More on how NextAuth.js middleware works: https://next-auth.js.org/configuration/nextjs#middleware
 export default withAuth({
   callbacks: {
     authorized: ({ token }) => token?.userRole === "admin",
   },
+  ignored: {
+    paths: ["/admin/ignored"]
+  }
 })
--- a/apps/example-nextjs/pages/admin/_middleware.ts
+++ b/apps/example-nextjs/pages/admin/_middleware.ts
@@ -1,4 +1,4 @@
-import { withAuth } from "next-auth/middleware";
+import { withAuth } from "next-auth/middleware"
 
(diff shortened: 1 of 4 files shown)


ndom91:
Hey @callumok2004 so first of all, thanks for the interest and new feature PR!

I just wanted to mention that middleware can also be applied to specific subroutes, instead of globally. i.e. if you have an `/pages/admin/*` set of routes, yuo can put the `_middleware.js` file under `/pages/admin/_middleware.js` and it will only apply to those routes. 

But I can definitely also see the use-case where you want more fine grained control..

We're currently taking a short break from merging new features or anything other than critical bug fixes and docs, so I will mark this as "ready" (to review) so the other maintainers can easily find it and take a look when they're back!

lluia:
I think to justify this we need to understand which benefits it brings over having middleware in sub-routes directly:
```
/pages/admin/_middleware.js
```
I don't like defining the ignored paths as strings within the options; any typo would cause the ignoring behaviour not to act as expected. Not sure middleware runs when requesting **static assets** 🤔 ( as per your code example ), so I think there's a need to ignore those manually.

balazsorban44:
I don't think this is an option to be added to our API.

You can already also use the many forms of `withAuth` invocation we support to achieve what you seem to want. I believe this should work:

```js

export async function middleware(req) {
  if(/*your condition*/) {
    return
  }

  await withAuth(req, {/*options*/})
}
```

I'm going to close this, as it would need a proper discussion first, anyway.

As pointed out https://github.com/nextauthjs/next-auth/pull/4305#issuecomment-1086613158, you can define a middleware for subpages only, which would also be better in terms of Edge Function invocations, and probably the most right thing to do. Example in our docs: https://next-auth.js.org/configuration/nextjs#default-withauth-export

PR #4329, opened 2022-04-05
Updated `providerAccount` to `account`

momotofu:
Changed `providerAccount` property to `account` in the message object received by the linkAccount event handler.

## Reasoning 💡

The message object contains `{ user,  account }` in v4, instead of `{ user, providerAccount }` which was in v3.

## Checklist 🧢

- [x] Documentation
- [ ] Tests
- [X] Ready to be merged


0ubbe (approved):
Thanks 💚

PR #4349, opened 2022-04-07
Fix inability to use custom decode with middleware

darknoon:
## Reasoning 💡

Thought my custom `encode()` / `decode()` wasn't working at all because the token was always null when checking for authorization in , but it turns out that this bug was affecting me.

I am trying to do this so I can have an unencrypted JWT client-side to query my separate non-nextjs graphql server (hasura).

## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## Affected issues 🎟

Fixes #4181


darknoon:
Ah, see there is already a PR attached to the task

PR #4352, opened 2022-04-07
fix(signin.tsx): set type of built-in email sign-in input to "email" and add "required" attribute for browser validation

raulmarindev:
## Reasoning 💡

Currently, the built-in sign-in page has the wrong type (text) for the input for emails when using
the magic link provider. Because of this, a user can enter whatever value they want in that input and click on the "Sign in with Email". By switching the type to be "email" we can make use of the browser-provided validation and users will be
forced to use properly formed emails addresses in order to sign in.

image

image

Additionally, the built-in sign-in page doesn't have the required attribute for the input for emails when using the magic link provider. Because of this, a user is able to submit the form without entering a value. By adding the "required" attribute we can make use of the browser-provided validation and users will be forced to provide a value in order to sign in.
## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/packages/next-auth/src/core/pages/signin.tsx
+++ b/packages/next-auth/src/core/pages/signin.tsx
@@ -123,7 +123,7 @@ export default function SigninPage(props: SignInServerPageParams) {
                 <input
                   id={`input-email-for-${provider.id}-provider`}
                   autoFocus
-                  type="text"
+                  type="email"
                   name="email"
                   value={email}
                   placeholder="email@example.com"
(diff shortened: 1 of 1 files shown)


balazsorban44 (commented):
Good catch, yeah, I think we can also make it required. 👍

0ubbe (approved):
Good catch 💚

lluia:
@raulmarindev lets make it `required` also as per @balazsorban44 hint 👍🏽

raulmarindev:
@lluia Thank you both, I've added the required attribute

0ubbe (approved):
Nice 💯

PR #4116, opened 2022-03-04
feat: introduce experimental `unstable_getServerSession` API

ThangHuuVu:
## Reasoning 💡

~Improve the `getServerSession` API to match the `NextAuth` API.~
- [x] Rename `getServerSession` to `unstable_getServerSession`
- [x] Update part of docs (rest is in #3982)

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

PR #4362, opened 2022-04-09
Add a new item to tutorials.md (Auth-wrapped getServerSideProps + InferGetServerSidePropsType HOC)

timfee:
I created a new repository showcasing an auth-wrapped getServerSideProps HOC that works with InferGetServerSidePropsType

### DIFF
--- a/docs/docs/tutorials.md
+++ b/docs/docs/tutorials.md
@@ -45,6 +45,10 @@ title: Tutorials and Explainers
 
 - Learn how to use Sign-In With Ethereum to authenticate your users with their existing Ethereum wallets - identifiers they personally control.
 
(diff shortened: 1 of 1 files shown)


ndom91:
Thanks for the additional tutorial! I wanted to mention that we've (silently for now) launched a `getServerSession` API that may make this more simple.

See draft docs here: https://github.com/nextauthjs/next-auth/pull/4116

timfee:
@ndom91 thanks for that!

is there any advantage to that approach? e.g. they appear functionally equivalent:
```tsx
export const getServerSideProps: GetServerSideProps = async (ctx) => {
  return {
    props: {
      session: await getSession(ctx), // or
      //session: await getServerSession(ctx, authOptions),
    },
  };
};
```

PR #4364, opened 2022-04-10
docs: scope example

alexandredev3:
## Reasoning 💡
I added the scope in the [...nextauth].js getting started example because I took some time to figure out where to add this scope, and I think that will be helpful for developers who are using NextAuth for the first time and wanna add this scope.

### DIFF
--- a/docs/docs/getting-started/example.md
+++ b/docs/docs/getting-started/example.md
@@ -25,6 +25,11 @@ export default NextAuth({
     GithubProvider({
       clientId: process.env.GITHUB_ID,
       clientSecret: process.env.GITHUB_SECRET,
+      authorization: {
+        params: {
+          scope: 'read:user'
+        },
+      },
     }),
     // ...add more providers here
   ],


lluia:
Thanks 💚

While I agree the customisation options for the `authorization`, `token` and `userinfo` requests are not obvious, I think we should document this some other way. Maybe it deserves a whole new section on the **Getting Started** page. What do you think @balazsorban44 @ndom91 ?

balazsorban44:
I think this should either be addressed/better documented under https://next-auth.js.org/configuration/providers/oauth#override-default-options

Or we should allow top-level scope again, as it seems to be a common point of confusion. :thinking: This would mean that we diverge from the path I wished though (enforcing stricter OAuth concepts by grouping `scope` under `authorization`)

lluia:
I'll open a PR trying to explain this better ✍🏼

balazsorban44:
I'm closing as I think this would be the wrong place to explain `scope`. We are thinking about restructuring our documentation anyway.

PR #4365, opened 2022-04-10
fix(providers): add optional chaining to avoid nullish reference errors

yisu-kim:
## Reasoning 💡

Hello, I encountered an error while using the Kakao provider.
If you don't select your nickname and profile image in Kakao settings, an 'undefined' error occurs because the profile object is not provided.

Please refer to the picture below.
Kakao developers
undefined error

As you can see in line 40 of the code, the profile is an optional property.
So I changed to get a nickname or profile image only if there is a profile object.

## Checklist 🧢

~- [ ] Documentation~
~- [ ] Tests~
- [x] Ready to be merged

### DIFF
--- a/packages/next-auth/src/providers/kakao.ts
+++ b/packages/next-auth/src/providers/kakao.ts
@@ -82,9 +82,9 @@ export default function Kakao<P extends Record<string, any> = KakaoProfile>(
     profile(profile) {
       return {
         id: profile.id,
-        name: profile.kakao_account?.profile.nickname,
+        name: profile.kakao_account?.profile?.nickname,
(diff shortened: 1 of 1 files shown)


0ubbe (approved):
Thanks 🖤 , Typescript should have caught this: `profile` is currently typed as `any` here, we have a PR attempting to fix it https://github.com/nextauthjs/next-auth/pull/4202

yisu-kim:
That's a fundamental solution to the problem. Thank you for letting me know.

balazsorban44 (approved):
Thank you!

PR #4383, opened 2022-04-13
Cleanup global `__NEXTAUTH` state after unmount

skirsten:
Re-mounting the component was not reloading the session because the global `__NEXTAUTH` was not cleaned up on unmount.

## Reasoning 💡

https://github.com/nextauthjs/next-auth/pull/4252#issuecomment-1082390355

## Checklist 🧢

- ~[] Documentation~
- ~[] Tests~
- [x] Ready to be merged

Note: If using `reactStrictMode: true` and using React 18 (https://github.com/vercel/next.js/issues/35822) you will see two `api/auth/session` requests during `next dev`

### DIFF
--- a/packages/next-auth/src/react/index.tsx
+++ b/packages/next-auth/src/react/index.tsx
@@ -359,6 +359,12 @@ export function SessionProvider(props: SessionProviderProps) {
     }
 
     __NEXTAUTH._getSession()
+
+    return () => {
+      __NEXTAUTH._lastSync = 0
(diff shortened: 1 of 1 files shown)


lluia:
Thanks for the contribution 💚

This is, AFAIK on purpose, so the session is stored in memory so long as the application is running. @balazsorban44 is the one who knows the details of this though, let's wait for his feedback ☕️

balazsorban44 (approved):
Thanks, this sounds good. :+1:

PR #4385, opened 2022-04-13
feat: Add the support of custom cookieName on the next-auth/middleware

42atomys:
## Reasoning 💡

When using the advanced custom cookie feature in next-auth, the middleware no longer works due to the fact that it looks for the default names of next-auth. The middleware calls `getToken` without allowing to override the cookie name

## Checklist 🧢

- [x] Add an option to custom the cookie name (and others cookies if needed in the future)
- [x] Add documention on Configuration > NextJS page
- [x] Test by hand

## Affected issues 🎟

No

## Notes

Try to following the work done in this PR to make more flexible merge in the future https://github.com/nextauthjs/next-auth/pull/4210


balazsorban44 (changes requested):
Thank you!

My end goal is to match the API of `NextAuth` within Middleware (so once we are fully edge-compatible, we can reduce the public API surface again, probably in v5), so I appreciate that you used the same API.

As for the docs, I think we should use what's already in https://next-auth.js.org/configuration/options#cookies

Removing the parts that aren't configurable right now.

It looks good to me otherwise. Were you actually be able to test it out?

42Atomys:
@balazsorban44 
> Removing the parts that aren't configurable right now.
Done 🎉

> Were you actually be able to test it out?
Don't find the middleware testing file to add more tests, any help about implementation of it will be appreciate

balazsorban44 (approved):
Thanks!

PR #4391, opened 2022-04-14
remove auth from keycloack url

dacitto:
in keycloak 17.0+ , There's a change in endpoints as per the official documentation (https://www.keycloak.org/docs-api/11.0/rest-api/index.html#_users_resource) .
so the new url well be https://my-keycloak-domain.com/realms/My_Realm

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/docs/docs/providers/keycloak.md
+++ b/docs/docs/providers/keycloak.md
@@ -37,5 +37,5 @@ providers: [
 ```
 
 :::note
-`issuer` should include the realm – e.g. `https://my-keycloak-domain.com/auth/realms/My_Realm`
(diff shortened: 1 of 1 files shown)


ndom91:
Hey thanks for this update! However, as keycloak is often a self-hosted auth solution, there will probably be a lot of people still on the older versions for a while, no? 

Maybe we make this configurable. 

What do you think?

balazsorban44:
I think this is good to merge, as it only changes documentation. `issuer` is not just configurable, but also required to be configured.

PR #4395, opened 2022-04-15
fix: node engine version move to "17.6.0"

UsboKirishima:
The precedents versions returns an error

## Checklist 🧢

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "typescript": "^4.5.2"
   },
   "engines": {
-    "node": "^12.19.0 || ^14.15.0 || ^16.13.0"
+    "node": "^17.6.0"
   },
   "prettier": {
     "semi": false,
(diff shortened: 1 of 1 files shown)


0ubbe (changes requested):
Thanks, can we get a bit more context on this contribution. Why is it needed? What error does it solve? Why do we need to ping **Node** to a specific version?

UsboKirishima:
> Thanks, can we get a bit more context on this contribution. Why is it needed? What error does it solve? Why do we need to ping **Node** to a specific version?

It returns the following error:

error openid-client@5.1.5: The engine "node" is incompatible with this module. Expected version "^12.19.0 || ^14.15.0 || ^16.13.0". Got "17.6.0"

error

PR #4402, opened 2022-04-17
Add additional type

jiyeol-lee:
## Reasoning 💡

I'm trying to use `getToken` function in `_middleware.ts` but it complains because req type only has `NextApiRequest | Pick<NextApiRequest, "cookies" | "headers">`.

Screen Shot 2022-04-17 at 8 06 11 AM

## Checklist 🧢

`NextRequst` type for nextjs `_middleware.ts`

Screen Shot 2022-04-17 at 8 01 32 AM

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

Yes

## Affected issues 🎟

I don't think it affects something else.


0ubbe (approved):
Thanks 💚

I think it makes sense as with the introduction of middleware, `getToken` can also be called with this context. 

Now that you're at it, do you mind updating this line here?

https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/next/middleware.ts#L84

We won't need the cast to `any` anymore 💆🏽‍♂️

kafelix496:
> 

Thank you for checking my PR :)

I agree with what you mentioned! I remove `as any` in there.

packages/next-auth/src/jwt/index.ts
```diff
@@ -37,7 +37,7 @@ export async function decode(params: JWTDecodeParams): Promise<JWT | null> {
 
 export interface GetTokenParams<R extends boolean = false> {
   /** The request containing the JWT either in the cookies or in the `Authorization` header. */
-  req: NextApiRequest | Pick<NextApiRequest, "cookies" | "headers">
+  req: NextRequest | NextApiRequest | Pick<NextApiRequest, "cookies" | "headers">
```
lluia, line 40: We'll need to import `NextRequest` from `next` above in order [for this to work](https://github.com/nextauthjs/next-auth/runs/6112902914?check_suite_focus=true):
```js
import type { NextApiRequest, NextRequest } from "next"
```

0ubbe (changes requested):

packages/next-auth/src/jwt/index.ts
kafelix496: Oh, yes... my bad. I updated it!

0ubbe (approved):

kafelix496:
@lluia I updated the code to import `NextRequest` from `next/server`.

PR #4420, opened 2022-04-19
chore: pnpm

ThangHuuVu:
## Reasoning 💡
Switch from yarn to [pnpm](https://pnpm.io/)
In this PR:
- Dependencies correction: Some adapter packages had the wrong/missing dependencies.
- `adapter-test `-> `@next-auth/adapter-test`
- `tsconfig` -> `@next-auth/tsconfig`
- In `release.yml`: Replace `actions/cache` with `actions/setup-node`
- dev app: Switch from npm-run-all to concurrently. It seems there is a bug in `npm-run-all` when using with `pnpm`. I couldn't run the command in parallel.
- remove `--no-workspaces` flag in `publish.ts`
## Checklist 🧢

- ~[ ] Documentation~
- ~[ ] Tests~
- [x] Ready to be merged

## Affected issues 🎟
None


.pnpm-debug.log
```diff
@@ -0,0 +1,21 @@
+{
```
balazsorban44, line 1: This file shoudl be gitignored

package.json
```diff
@@ -38,6 +39,7 @@
     "eslint-plugin-promise": "^6.0.0",
     "git-log-parser": "1.2.0",
     "husky": "^7.0.4",
+    "pnpm": "^6.32.8",
```
balazsorban44, line 42: Is this necessary? I think we can expect it to be available globally

ThangHuuVu: Yeah these can definitely be dropped, thanks for the catch

ThangHuuVu:
@balazsorban44 I replaced cache-node with setup-node v3. setup-node support built-in caching since pnpm v6.10+: https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data

PR #4436, opened 2022-04-21
docs: Use correct module in examples for augmenting TypeScript Sessio…

### LINKED ISSUE #671: How to augment Session using TypeScript?
**Your question**

Hi, would it be possible to add an example of type augmentation to the documentation?

**What are you trying to do**

I've modified what the `session` object returns and would like to augment the type defined in `next-auth` and `next-auth/client` to reflect the changes I've made. I've tried overriding the type definition by creating a `next-auth.d.ts` at the root of the project and included it in my `tsconfig.json`. However, VSCode uses the one defined by the module and not the one I've created (even though it recognizes that there are two type definitions). 

To show what I have, here's the type definition I've created:

```ts
// next-auth.d.ts

import 'next-auth' 
import * as auth from 'next-auth' 
// import * as client from 'next-auth/client' 
// import 'next-auth/client' 

declare module 'next-auth' {
  export * from 'next-auth'
  export type InitOptions = auth.InitOptions
  export default NextAuth
  interface Session {
    user: {
      name: string
      email: string
      image: string
      uid: string
    }
  }
}

declare module 'next-auth/client' {
  export * from 'next-auth/client'

  export interface Session {
    user: {
      name: string
      email: string
      image: string
      uid: string
    }
  }
}


```

Any help would be appreciated.

Thank you.

**Feedback**

vbudovski:
…n and User types

## Reasoning 💡

The examples in the documentation for overriding the TypeScript User and Session types did not work as intended. I believe this is due to the incorrect module path provided. The types themselves live inside the `next-auth/core/types` module rather than the `next-auth` module, even though they are being imported in the latter. This change allows TypeScript to correctly detect that the types have been augmented. Tested in TypeScript 4.6.2.

## Checklist 🧢

- [x] Documentation
- [ ] ~Tests~
- [X] Ready to be merged

## Affected issues 🎟

Fixes #671

### DIFF
--- a/docs/docs/getting-started/typescript.md
+++ b/docs/docs/getting-started/typescript.md
@@ -75,9 +75,9 @@ export default function IndexPage() {
 To extend/augment this type, create a `types/next-auth.d.ts` file in your project:
 
 ```ts title="types/next-auth.d.ts"
-import NextAuth from "next-auth"
+import NextAuth from "next-auth/core/types"
 
-declare module "next-auth" {
+declare module "next-auth/core/types" {
   /**
    * Returned by `useSession`, `getSession` and received as a prop on the `SessionProvider` React Context
    */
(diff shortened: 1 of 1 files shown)


balazsorban44:
This should not be necessary. :thinking: Is this a recent change in TypeScript? As you mentioned, all types are re-exported here:

https://github.com/nextauthjs/next-auth/blob/d4fb7af6f504e4abc02b28ba5f1fd23f58553249/packages/next-auth/src/index.ts#L1

vbudovski:
You are correct. My source code lives inside a `src` directory and I had set `"baseUrl": "./src/"` inside my tsconfig. This seems to have caused the type resolution to fail for the augmented types.

PR #4449, opened 2022-04-22
docs(tutorial): Adding role based login to database session strategy

GlennStreetman:
## Checklist 🧢

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged


ndom91:
Thanks for the addition @GlennStreetman! 

Can you rename the file to some appopriate title, like `role-based-login-strategy.md` or something?

Also can you link to it from the `docs/tutorials.md` page like all other tutorials?

ndom91:
Great thanks! 

Sorry to be so nit-picky, but yuo've got some odd spacing all over here as well. Can we clean this up?

```javascript
  const session = await getSession({ req });
  if (session && session.user.roll === "admin") {
    //Allow access
  } else {
    //Deny access
  }
  
```

To something like this maybe?

```javascript
const session = await getSession({ req });
if (session && session.user.roll === "admin") {
  //Allow access
} else {
  //Deny access
}
```

Also `roll` is spelled `role` :+1:

ndom91:
@GlennStreetman thanks for the updates! I hope you don't mind, but I just pushed a commit with a couple more small formatting / copy updates. Will merge this shortly then :rocket:

ndom91:
So the build failed. It also failed when I tried to build this branch locally.

The strange thing is, we're only changing two very minor markdown files here. No idea what could be causing this...

PR #4472, opened 2022-04-26
fix(ts): handle NextRequest type

0ubbe:
## 💡 Reasoning

Following the merge of this PR https://github.com/nextauthjs/next-auth/pull/4402, the build got broken 🔥 

That's because the `getToken()` implementation assumed the `req` object passed to it was an instance of `NextApiRequest`. However, the new middleware implementation:

https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/next/middleware.ts#L56-L84

is calling this function with a `req` object which is an instance of `NextRequest`.

At the end of the day:
```ts
const token = getToken({ req })
``` 
is only interested in `cookies` and `headers` in the `req` object supplied to it. It happens though that `headers` are different in `NextApiRequest` and `NextRequest`... in `NextApiRequest` it'll be [`IncomingHttpHeaders`](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules__types_node_http2_d_._http2_.incominghttpheaders.html) while in `NextRequest` it'll be [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers) and both are read differently...

### Random thought

Was [`getToken()`](https://next-auth.js.org/tutorials/securing-pages-and-api-routes#using-gettoken) working before when not called in the context of an API route?
```ts
// my-page.tsx
  const token = getToken({ req: context.req })
  // ...
```

## 🧢 Checklist 

- ~~[ ] Documentation~~
- ~~[ ] Tests~~
- [x] Ready to be merged

## 🎟 Affected issues

None

### DIFF
--- a/packages/next-auth/src/core/lib/cookie.ts
+++ b/packages/next-auth/src/core/lib/cookie.ts
@@ -120,7 +120,7 @@ export class SessionStore {
     option: CookieOption,
     req: {
       cookies?: Record<string, string>
-      headers?: Record<string, string> | IncomingHttpHeaders
+      headers?: Headers | IncomingHttpHeaders
     },
     logger: LoggerInstance | Console
   ) {
--- a/packages/next-auth/src/core/lib/cookie.ts
+++ b/packages/next-auth/src/core/lib/cookie.ts
@@ -120,7 +120,7 @@ export class SessionStore {
     option: CookieOption,
     req: {
       cookies?: Record<string, string>
-      headers?: Headers | IncomingHttpHeaders
+      headers?: Headers | IncomingHttpHeaders | Record<string, string>
     },
     logger: LoggerInstance | Console
   ) {
--- a/packages/next-auth/src/jwt/index.ts
+++ b/packages/next-auth/src/jwt/index.ts
@@ -2,8 +2,8 @@ import { EncryptJWT, jwtDecrypt } from "jose"
 import hkdf from "@panva/hkdf"
 import { v4 as uuid } from "uuid"
 import { SessionStore } from "../core/lib/cookie"
-import { NextRequest } from "next/server"
-import type { NextApiRequest} from "next"
+import type { NextApiRequest } from "next"
+import type { NextRequest } from "next/server"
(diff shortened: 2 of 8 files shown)


balazsorban44 (approved):
LGTM, thanks!

packages/next-auth/src/jwt/index.ts
```diff
@@ -91,8 +91,13 @@ export async function getToken<R extends boolean = false>(
 
   let token = sessionStore.value
 
-  if (!token && req.headers.authorization?.split(" ")[0] === "Bearer") {
-    const urlEncodedToken = req.headers.authorization.split(" ")[1]
+  let authorizationHeader =
```
balazsorban44, line 94:
```suggestion
  const authorizationHeader =
```

balazsorban44 (changes requested):

balazsorban44 (approved):

packages/next-auth/src/core/lib/cookie.ts
```diff
@@ -120,7 +120,7 @@ export class SessionStore {
     option: CookieOption,
     req: {
       cookies?: Record<string, string>
-      headers?: Record<string, string> | IncomingHttpHeaders
+      headers?: Headers | IncomingHttpHeaders | Record<string, string>
```
KATT, line 123: Doesn't `Headers` require node 18?

ThangHuuVu: Interesting. `Headers` and other Fetch API interfaces are also available on Node 16 with the `--experimental-fetch` flag: https://nodejs.org/en/blog/release/v16.15.0/

KATT: I had some testing in jest of some API helpers which started breaking with the new patch. Fixed it by adding `import 'next';` in the top.

balazsorban44: This one is old, but came up recently so. Next.js polyfills `Headers` for you, if it's missing:

https://github.com/vercel/next.js/blob/canary/packages/next/server/node-polyfill-fetch.js

PR #4496, opened 2022-05-01
update Prisma adapter to handle session not existing when attempting to delete

### LINKED ISSUE #4495: Prisma throws on attempted to delete record that does not exist but next-auth core does not expect that
### Adapter type

@next-auth/prisma-adapter

### Environment

  System:
    OS: macOS 12.3.1
    CPU: (8) x64 Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz
    Memory: 81.57 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.2 - ~/Library/Caches/fnm_multishells/75640_1651423362851/bin/node
    Yarn: 1.22.18 - ~/Library/Caches/fnm_multishells/75640_1651423362851/bin/yarn
    npm: 8.5.0 - ~/Library/Caches/fnm_multishells/75640_1651423362851/bin/npm
  Browsers:
    Chrome: 100.0.4896.127
    Firefox: 99.0
    Safari: 15.4
  npmPackages:
    next: ^12 => 12.1.5
    next-auth: ^4.1.2 => 4.3.3
    react: 18 => 18.0.0

  npmPackages:
    @next-auth/prisma-adapter: ^1.0.1 => 1.0.3

### Reproduction URL

n/a

### Describe the issue

Prisma throws an exception if `delete` is called for a record which does not exist. The adapter does not handle that exception so it breaks the login. Here is where the session is being deleted in the adapter:

https://github.com/nextauthjs/next-auth/blob/main/packages/adapter-prisma/src/index.ts#L33-L34

Prisma has some surprising/rough edges and this is one of them -- ongoing issue about it here:

https://github.com/prisma/prisma/issues/4072

### How to reproduce

- create an app with both Google and email-only logins (using Prisma database adapter)
- login via Google
- login via Email-only w/ same email as above
- truncate Accounts (removes recorded created by Google login)
- attempt to login via Email
- requesting login URL works
- get to login page but get error message related to Callback

Logs:

```
https://next-auth.js.org/errors#callback_email_error An operation failed because it depends on one or more records that were required but not found. Record to delete does not exist. Error: An operation failed because it depends on one or more records that were required but not found. Record to delete does not exist.
    at Object.request (/var/sites/fasmussen/node_modules/@prisma/client/runtime/index.js:45578:15)
    at async PrismaClient._request (/var/sites/fasmussen/node_modules/@prisma/client/runtime/index.js:46405:18) {
  name: 'DeleteSessionError',
  code: 'P2025'
}
```

### Expected behavior

The Prisma adapter should factor in the odd Prisma API that throws on attempting to delete a record that does not exist. I worked around it for now by changing:

https://github.com/nextauthjs/next-auth/blob/main/packages/adapter-prisma/src/index.ts#L33-L34

```js
    deleteSession: (sessionToken) =>
      p.session.delete({ where: { sessionToken } }),
```

To:

```js
    deleteSession: (sessionToken) => {
      try {
        return p.session.delete({ where: { sessionToken } });
      } catch (e) {
        console.error("Failed to delete session", e);
        return null;
      }
    },
```

cymen:
## ☕️ Reasoning

Prisma throws an exception when attempting to delete a record that does not exist. The `useVerificationToken` function already handles this -- this PR updates the adapter so that the `deleteSession` function also handles it.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: https://github.com/nextauthjs/next-auth/issues/4495


balazsorban44:
See my https://github.com/nextauthjs/next-auth/issues/4495#issuecomment-1142264353

It's not expected that a user session stops existing unless you manipulate database records outside `next-auth`.

PR #3649, opened 2022-01-16
fix(core): detect Vercel without `NEXTAUTH_URL`

balazsorban44:
Fixes #3419

Based on https://github.com/nextauthjs/next-auth/issues/3419#issuecomment-1003004601

PR #4509, opened 2022-05-05
Prefer User-Defined Host (NEXTAUTH_URL) Over Forwarded Host in Vercel

### LINKED ISSUE #4507: pathname of `NEXTAUTH_URL` is ignored in when deployed to vercel
### Environment

  System:
    OS: Linux 5.13 Ubuntu 21.10 21.10 (Impish Indri)
    CPU: (16) x64 AMD Ryzen 7 2700 Eight-Core Processor
    Memory: 4.38 GB / 31.30 GB
    Container: Yes
    Shell: 5.1.8 - /bin/bash
  Browsers:
    Firefox: 99.0
  npmPackages:
    next: 12.1.x => 12.1.6 
    next-auth: 4.3.x => 4.3.4 
    react: 17.0.x => 17.0.2 

### Reproduction URL

---

### Describe the issue

We have `NEXTAUTH_URL` set to "https://ourdomain.com/api/v1/auth" with the modified pathname since we're versioning our api, and we're correctly setting `basePath` in `<SessionProvider>`. This is working in our development environment, however when deploying to Vercel, the "login with google" oauth path no longer works. I have tracked this down in next-auth:
1. `"providers"` is populated in `parseProviders`, which gets [a `"url"` in `init()`](https://github.com/nextauthjs/next-auth/blob/4f29d39521451e859dbdb83179756b372e3dd7aa/packages/next-auth/src/core/init.ts#L47)
3. `init` is called in [NextAuthHandler](https://github.com/nextauthjs/next-auth/blob/4f29d39521451e859dbdb83179756b372e3dd7aa/packages/next-auth/src/core/index.ts#L72) which uses `req.host`
4. And `req.host` is provided [in `NextAuthNextHandler`](https://github.com/nextauthjs/next-auth/blob/4f29d39521451e859dbdb83179756b372e3dd7aa/packages/next-auth/src/next/index.ts#L26) but is constructed via `detectHost`
5. `detecthost` will [return the "x-forwarded-host" header](https://github.com/nextauthjs/next-auth/blob/4f29d39521451e859dbdb83179756b372e3dd7aa/packages/next-auth/src/next/utils.ts#L18) if it detects it's in vercel
6. [Back in the `init`](https://github.com/nextauthjs/next-auth/blob/4f29d39521451e859dbdb83179756b372e3dd7aa/packages/next-auth/src/core/init.ts#L43) `parseUrl` is called on the host
7. Since [`parseUrl`](https://github.com/nextauthjs/next-auth/blob/4f29d39521451e859dbdb83179756b372e3dd7aa/packages/next-auth/src/lib/parse-url.ts#L15) is passed the "x-forwarded-host" header instead of the url in `NEXTAUTH_URL` it cannot parse any custom set pathname, and the signinUrl and callbackUrl in the providers call become incorrect.

This seems to have been broken by https://github.com/nextauthjs/next-auth/pull/3649 . I think if the two approaches could be used together, that would be ideal: 
1. use the `x-forward-host` + the pathname from `NEXTAUTH_URL`
2. Or, allow `basePath` to be set in the nextauth config like it is for `SessionProvider` 

### How to reproduce

1. use a custom pathname in `NEXTAUTH_URL`, eg: `http://custom.vercel.app/api/v1/auth`
2. have a login with google
4. deploy to vercel
5. try logging in

The call to the /providers url will return a `signinUrl`, and `callbackUrl` that don't have the "/v1/" part of the url and will default to "/api/auth"

### Expected behavior

The redirect url provided to the `signIn` call should have the custom pathname from the `NEXTAUTH_URL`

jjorissen52:
This PR affects behavior that can only be observed in Vercel deployments.

The commit reverts behavior introduced by https://github.com/nextauthjs/next-auth/pull/3649, where the user-defined environment variable `NEXTAUTH_URL` is ignored in favor of the value of `x-forwarded-host`.

## ☕️ Reasoning
If a user deliberately sets a `NEXTAUTH_URL` environment variable, it means they wanted Vercel to use this as the canonical URL. next-auth should consider this deliberately defined value first and then fallback to automatic behavior otherwise. The current behavior, which is to use `x-forwarded-host` when available and fallback to user-defined `NEXTAUTH_URL` (on top of being surprising and not consistent with the [documentation](https://next-auth.js.org/configuration/options#nextauth_url)) breaks in multiple situations, including our current staging environment:

1. If a proxy sits between the user and the deployment on Vercel, wherein the actual site host and Vercel's deployment URL do no match, next-auth will end up using Vercel's deployment URL. For example, we have nginx listening at example.dev and our staging deployment at staging.example.com. Even though NEXTAUTH_URL=example.dev, next-auth sets `redirect_uri` to staging.example.com.
2. If you are using a Preview deployment, e.g. example.vercel.app, the current behavior is to set `redirect_uri=<preview_url>` which is practically guaranteed to not be whitelisted by your provider (since you can't know the auto-generated preview URL ahead of time).

## 🧢 Checklist

- [X] ~Documentation~ consistent with existing
- [X] Tests
- [X] Ready to be merged (behavior tested locally)

## 🎫 Affected issues

Fixes: https://github.com/nextauthjs/next-auth/issues/4507

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/next/utils.ts
+++ b/packages/next-auth/src/next/utils.ts
@@ -16,8 +16,11 @@ export function setCookie(res, cookie: Cookie) {
 
 /** Extract the host from the environment */
 export function detectHost(forwardedHost: any) {
+  /* If the user has explicitly set canonical URL using NEXTAUTH_URL,
+   we respect that setting. */
+  if (process.env.NEXTAUTH_URL) return process.env.NEXTAUTH_URL
   // If we detect a Vercel environment, we can trust the host
   if (process.env.VERCEL) return forwardedHost
-  // If `NEXTAUTH_URL` is `undefined` we fall back to "http://localhost:3000"
-  return process.env.NEXTAUTH_URL
+  // Fallback to "http://localhost:3000"
+  return undefined
 }


jmosley5 (approved):

Naddiseo:
There is a tangent situation, which I didn't mention in my original bug report: if there really is a proxy.

We have two apps that connect to the same backend api (where `[...nextauth].ts` is located), and access it using nextjs's `rewrite` on the two apps, which behaves as a proxy and has the x-forwarded-host header. 
```javascript
///next.config.js in app1 and app2
module.exports = {
   rewrites() {
      return [
          {source: "/api/:path*", destination: `${process.env.API_URL}/api/:path*` }
      ]
   }
};
```

Our architecture is like this:
```
+-------------------------+     +--------------------------------+
| app1.com/api/auth       | ->  |     api.backend.com/api/auth   | 
+-------------------------+     |                                |
                                |                                |
+-------------------------+     |                                |
|  app2.com/api/auth      |  -> |                                |
+-------------------------+     +--------------------------------+
 
```

So for each request, we need to know which of the two apps the request came from so that next-auth will redirect back to the correct one. In this particular case, the x-forward-host header is actually very useful, except when the request comes from an oauth callback.
I currently have to detect the x-forwarded-host header AND the `next-auth.callback-url` cookie THEN set `NEXTAUTH_URL` explicitly.

If this PR were changed slightly to pull the `pathname` from `NEXTAUTH_URL` if it exists, and also took the callback url in cookie into account, next-auth's long standing request for a dynamic  `NEXTAUTH_URL` would be satisfied

jjorissen52:
@Naddiseo Just reading through it, I think the callback url in that cookie is set based on the return value of `detectHost`. It's unclear to me what changes would be required to accomplish what you are trying to do in a way that doesn't break our use case, which is to proxy the entire Vercel deployment.

Naddiseo:
@jjorissen52, I _think_ I understand what you're saying: In a non-proxy environment the `x-forwarded-host` refers to the deployment url instead of the assigned domain? In which case, I see what you're saying about my use case being at odds. 

I think I may have originally misunderstood how/what the `x-forward-host` actually was, because how it works, works in our favour *if* it also picked up the custom `pathname` something like:
```typescript

if (process.env.VERCEL) {
    if (process.env.NEXTAUTH_ENV) { // http://example.com/custom/api/auth/path
        const {pathname} = new URL(process.env.NEXTAUTH_ENV);
        return `${forwardedHost}${pathname}`; // forwarded.com/custom/api/auth/path
    }
    return forwardedHost
}
```

Given that, I do think this PR is the correct fix in that it makes the code match what the documentation says it'll do. 

For reference, how we're solving the multi app issue is something like:
```typescript
/// [...nextauth].ts
export default (req,res) =>{ 
    const forwardedHost = req.headers['x-forwarded-host'];
    const callbackUrl = req.cookies['next-auth.callback-url'];
    const baseUrl = isAllowedUrl(forwardedHost)
        ? forwardedHost 
        : isAllowedUrl(callbackUrl) 
            ? callbackUrl
            : undefined;
    assert(typeof baseUrl !== "undefined");
    const url = new Url(baseUrl);
    url.pathname = "/custom/api/auth";
    
    process.env.NEXTAUTH_URL = url.toString();

    
    return NextAuth(req,res, { ...options});
}

**Comment by jjorissen52** [comment]
Yes, you seem to understand the issue now. And that code snippet is very interesting. We hadn't considered making modifications like yours. We might be able to do something similar...

**Comment by jjorissen52** [comment]
@Naddiseo We haven't had a chance to completely verify that this fixes our problem (I will come back with an update) but it looks like it works in the preview deployment:

```js
export default (req, res) => {
  /* next-auth prioritizes x-forwarded-host over NEXTAUTH_URL
    when running in Vercel https://github.com/nextauthjs/next-auth/pull/4509 .
    We need that to not happen, since the actual domain from *our* nginx proxy will differ
    from x-forwarded-host set by Vercel's proxy. E.g.:
      1. Our Nginx listening on example.dev proxies staging.example.com
      2. Vercel's Nginx listening at staging.example.com sets x-forwarded-host to staging.example.com
      3. next-auth will now set redirect_uri to staging.example.com instead of example.dev
  */
  if (process.env.VERCEL) {
    // prefer NEXTAUTH_URL, fallback to x-forwarded-host
    req.headers["x-forwarded-host"] = process.env.NEXTAUTH_URL || req.headers["x-forwarded-host"]
  }
  return NextAuth(req, res, options) // eslint-disable-line new-cap
}
```

jjorissen52:
Please re-open this, thanks.

PR #4539, opened 2022-05-10
next.config.js

haris18896:
Hello there, Haris here.

while working with Next JS and Next-Auth-JS, I have came across that in most of the situation there is a required changes in the `next.config.js` file which aren't been described properly in the documentation, and the new comers waste too much time finding what's wrong with the app as it's the same as given in the documentation.

so kindly add those configurations of  `next.config.js` file to the documentation, so that people can find the solution easily.

thankyou
kindregards

## ☕️ Reasoning

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources

### DIFF
--- a/docs/docs/getting-started/example.md
+++ b/docs/docs/getting-started/example.md
@@ -63,6 +63,30 @@ Instances of `useSession` will then have access to the session data and status. 
 Check out the [client documentation](/getting-started/client) to see how you can improve the user experience and page performance by using the NextAuth.js client.
 :::
 
+## Setting Enviroment Variables
+
+To set your enviroment variables, we will need to create `.env.local` file in the root directory. We have to add the required enviroment variables that we are using for authentication.
+
+```.env title=".env.local" showLineNumbers
+  GITHUB_ID="< your github ID >"
+  GITHUB_SECRET="< your github secret key >"
+```
+
+Next JS is a server side language, so our enviroment variables will not be accessed anywhere in the app, to access those enviroment variables that we have defined in `.env.local` we have to configure `next.config.js`
+
+```jsx title="next.config.js" showLineNumbers
+/** @type {import('next').NextConfig} */
+const nextConfig = {
+  reactStrictMode: true,
+  env: {
+    GITHUB_ID: process.env.GITHUB_ID,
+    GITHUB_SECRET: process.env.GITHUB_SECRET
+  }
+}
+
+module.exports = nextConfig
+```
+
 ### Frontend - Add React Hook
 
 The [`useSession()`](/getting-started/client#usesession) React Hook in the NextAuth.js client is the easiest way to check if someone is signed in.


ndom91:
Hey Harris, thanks for adding this. It is mentioned in each providers docs, and probably elsewhere. But your right, makes sense to have directly there in getting started too imo

ndom91:
If we're going to have this section in getting started though, would be helpful to also mention the `NEXTAUTH_URL` and `NEXTAUTH_SECRET` required env vars here. 

Also can you make it clear that the github (for exampel) env vars are only needed if they're using an oauth provider 👍

haris18896:
@[ndom91](https://github.com/ndom91), No, NEXTAUT_URL and NEXTAUTH_SECRET aren't required at this point in the go through section, my point is that I have gone through the documentation and haven't found any thing related configuring NEXT JS configuration which is an important thing to add or link some resource so that a person working with it shouldn't search for the solutions on third party, and can configure it easily.

## **Extra read**

also if possible add all the required configurations sections in respective docs, like adding File System configuration to the Next JS (example) docs [Next Js Start Learning](https://nextjs.org/learn/foundations/about-nextjs),  as this kind of section is also missing from there. in that section you will have to configure `next.config.js` file for working with `fs`.

```js title='next.config.js'

config.node = {
  global: true,
  __filename: true,
  __dirname: true
}
```

you will have to add these settings to your `next.config.js` because setting `fs: false` will completely turn off the file system and as Next Js reuires Node JS so we cannot turn it off completely because while 'false' the Webpack wouldn't touch your '__filename' code and your '__dirname' code.

ndom91:
Hey Harris, so I'm not sure what you're getting at with the `next.config.js` `globa: true ...` stuff, that shouldn't be necessary at all.

Also we noticed a few other small things regarding your copy.

> `Next JS is a server side language, so our enviroment variables will not be accessed anywhere in the app, to access those enviroment variables that we have defined in '.env.local' we have to configure 'next.config.js'`.

Few notes:

- Next.js is a hybrid client-side / server-side framework, not language. 
- The env variables added in `.env.local` should automatically be available in your application. No need to add them again in the `next.config.js`, for example ([Source](https://nextjs.org/docs/basic-features/environment-variables)). 
- However, by default they're only available to server-side code, i.e. API routes and `getServerSideProps`, etc. If you want them to be available to your FE code as well (like pages and normal react components), they must be prefixed with `NEXT_PUBLIC_`. **However, this is not necessary for any NextAuth.js related environment variables - they are all only needed server side and should not be made available on the client**

I still think it makes sense to mention the env variable setup stuff in the `Getting Started` section, as you've suggested here. But, can you clean up some of these details?

Thanks!

balazsorban44:
I am closing as the content of this PR is not correct in its current form. Feel free to re-open/address Nico's suggestions though.

Also, see: https://nextjs.org/docs/basic-features/environment-variables

PR #4540, opened 2022-05-10
chore(tutorial): fixing the `expires_in` method of `account` object to `expires_at`

Palanikannan1437:
## ☕️ Reasoning

account.expires_in method is no longer supported and as a result, the comparison at line number 98 always returns true which leads to generation of a refresh token every time.

On line numbers 64 and 91, I've updated the expires_in method to expires_at method and now the refresh token is generated only if the primary access token expires!


0ubbe (approved):
Thanks 💚

ndom91 (approved):
LGTM :+1:

socketopp:
I tried the refresh token example at 23:38 this evening. 
The expired_at field was set to `expires_at: 1675895896` which converts to the same minute 23:38:16. 
How is this supposed to be used? 
What does this calculation actually do?
`accessTokenExpires: Date.now() + refreshedTokens.expires_at * 1000,`

PR #4546, opened 2022-05-11
feat(provider): add United Effects provider

cbetz:
## ☕️ Reasoning

Adding United Effects to a Next.js + NextAuth.js project and thought others would benefit as well. Full transparency: I work for United Effects :)

## 🧢 Checklist

- [X] Documentation
~Tests~ (tested manually)
- [X] Ready to be merged


packages/next-auth/src/providers/united-effects.ts
```diff
@@ -0,0 +1,29 @@
...
+    authorization: {
+      params: { scope: "openid email profile", resource: options.issuer },
+    },
+    checks: ["pkce", "state"],
+    idToken: true,
+    profile(profile) {
+      return {
+        id: profile.sub,
+        email: profile.email,
+      }
```
lluia, line 25: We want all providers to return the following shape for the **profile callback**:
```ts
      return {
        id: profile.sub,
        name: profile.name,
        email: profile.email,
        image: profile.avatar,
      }
```
so we're missing `name` and `image` that should be filled from the response of the United Effects API.

packages/next-auth/src/providers/united-effects.ts
```diff
@@ -0,0 +1,29 @@
+import type { OAuthConfig, OAuthUserConfig } from "."
+
+export interface UnitedEffectsProfile extends Record<string, any> {
+  sub: string
+  email: string
+}
+
+export default function UnitedEffects<P extends UnitedEffectsProfile>(
+  options: OAuthUserConfig<P>
```
lluia, line 9: If `issuer` is an option you should mark it here like so:
```suggestion
  options: OAuthUserConfig<P> & { issuer: string } 
```

0ubbe (changes requested):
Thanks 💚, a couple of suggestions 💬

packages/next-auth/src/providers/united-effects.ts
cbetz: Gotcha! United Effects doesn't return this information by design. Instead there is a concept of Secured Profiles. 

I can change this to add the `name` and `image` (as null or empty string?) if the shape should be maintained but not sure if that introduces further issues.  

Happy to chat more about Secure Profiles btw :)

lluia: Thanks @cbetz. Can you then mark `name` and `image` as `null` in the `profile()` response. 

Let us add a note as well [on the provider readme](https://github.com/nextauthjs/next-auth/pull/4546/files#diff-2a007578ddb2d2677121868605379941b4b279caadc570516a41a61072916a7c) mentioning these two fields will be missing when using this provider, something like:
```
:::note
`issuer` should be the fully qualified URL including your Auth Group ID – e.g. `https://auth.unitedeffects.com/YQpbQV5dbW-224dCovz-3`
:::

:::warning
You won't be able to get the user name and image from NextAuth when using this provider. This is by design in the United Effects API.
::: 
```
Otherwise, users using this provider might grow frustrated trying to render the user avatar once logged in and not see the image and name in the data.

cbetz: Thank you @lluia--all set!

0ubbe (approved):

Issue #600, opened 2020-08-24
Need to set NEXTAUTH_URL dynamically as an Option for multi-domain/multi-tenant use


SharadKumar:
**Your question**
How to dynamically work with Passwordless/Email auth, without setting NEXTAUTH_URL.

**What are you trying to do**
I am working on a use-case where NEXTAUTH_URL is not fixed at deploy-time or build-time, but run-time, for a multi-domain (single codebase) scenario. This is to have Email passwordless only. 

I have had good success with next-auth other providers for usual scenarios, and absolutely love the simplicity.

**Feedback**
I tried to browse around the code to get a sense of dependency of the deploy-time NEXTAUTH_URL, and it seems that it is use only to define the Url for `sendVerificationRequest`. If there was a way to pass it as an option, it would do it.

Please advise, whats the best approach.

* [x] Found the documentation helpful
* [ ] Found documentation but was incomplete
* [ ] Could not find relevant documentation
* [x] Found the example project helpful
* [ ] Did not find the example project helpful

SharadKumar:
Server has a `todo` flagged:
https://github.com/nextauthjs/next-auth/blob/12a5d6b1f42707e9b5da289e58b2ec5b4de2a60c/src/server/index.js#L47-L50

JWT. GetToken can be custom though:
https://github.com/nextauthjs/next-auth/blob/8115a7c66cdb1d86b0d2a0d76b9aa33c2bfaa33b/src/lib/jwt.js#L94-L103

And, for the client:
https://github.com/nextauthjs/next-auth/blob/8115a7c66cdb1d86b0d2a0d76b9aa33c2bfaa33b/src/client/index.js#L24-L26

These are the references I found. @iaincollins I can try to fork and attempt a hack on my side, since I really need to solve this. If you can advise how to go about it, it would be awesome. It seems you intend to refine/refactor this at some point, as stated above. Above three are the only references I found.

Any help is really appreciated.

SharadKumar:
I am attempting to work on a fork, to see if I can do this 💇‍♀️
I've followed the Contributing guide and linked the module. Local next-auth is loaded locally, until I add `useSession` or `getSession` in `_app.js`. Yes I've noted the comment in the guide about linking `react` module, and done that too.

```js
error - ../../next-auth/node_modules/require_optional/index.js:2:0
Module not found: Can't resolve 'fs'
null
```

Not able to resolve this and spent many hrs on this trying to figure.

SharadKumar:
[...]

PR #4556, opened 2022-05-13
break free of NEXTAUTH_URL using http header

devinrhode2:
We need to support multiple domain names pointing to the same next.js server (multi-tenancy). Other parts of our infrastructure will pass an `X-Forwarded-Host` header to this next.js server, which we need to use in place of the NEXTAUTH_URL.

This is just a hack to get working for our use case.
I noticed the `version:pr` npm script might actually publish this PR as a package, which is what I am trying to do by opening this PR. If it doesn't publish a package, then I will close and re-open later once this is ready for review

<sub>I tried using `patch-package` but was getting this `Can't find lockfile entry for next-auth` error :/</sub>

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

#600

## 📌 Resources

### DIFF
--- a/apps/dev/pages/protected-ssr.js
+++ b/apps/dev/pages/protected-ssr.js
@@ -30,7 +30,8 @@ export async function getServerSideProps(context) {
   let content = null
 
   if (session) {
-    const hostname = process.env.NEXTAUTH_URL || "http://localhost:3000"
+    if (!process.env.NEXTAUTH_URL) throw new Error('getServerSideProps does not have NEXTAUTH_URL')
+    const hostname = process.env.NEXTAUTH_URL
     const options = { headers: { cookie: context.req.headers.cookie } }
     const res = await fetch(`${hostname}/api/examples/protected`, options)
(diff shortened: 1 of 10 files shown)


0ubbe (changes requested):
Thanks for taking the time to open a PR ☕️

> This is just a hack to get working for our use case.

Can you close this PR and open an issue explaining the problem and the proposed solution in-depth? We can then define how the API is going to look there 👍🏽

devinrhode2:
I linked to issue #600 in the Affected Issues section

lluia:
@devinrhode2 So the core team is considering adding a config option to allow trusting certain headers (which will solve your issue and part of #600). 

We don't have a timeline for this yet (we have many fronts open); we have to be very clear about the messaging and call it an advanced option because it has serious security implications. 

We'll update on this in #600, for now, I suggest the best is to close PR and we'll take it as a reference when working on the solution 📙

PR #4558, opened 2022-05-13
Fix font size for small devices

matiaguilera:
I have never seen any documentation using such a small font size before. It's readable, but still I think this media query is unnecessary.


0ubbe (approved):
Thanks ☕️

As you mentioned, yes the `font-size` in small viewports ends up computed to `~12px` which is rather small.

PR #4619, opened 2022-05-25
[docs] signIn() 'email' or 'credential' provider

ArthurPedroti:
I'm submitting this addition for the docs, but I'm not sure if it is a lack of email and credentials provider docs, or if it is an error in the types.

But basically, if you pass this args types, the response will be correctly typed with 'SignInResponse | undefined', but if you don't pass, the types will be only 'undefined'.

So, I'm assuming that the purpose of who made the types, It's to pass this args, for the ternary condition at the result of the signIn promise, to get the 'SignInResponse | undefined' type.

If someone can check and give me an opinion about that, I will appreciate it.

e.g.
image
image

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/docs/docs/getting-started/client.md
+++ b/docs/docs/getting-started/client.md
@@ -329,6 +329,19 @@ e.g.
 - `signIn('credentials', { redirect: false, password: 'password' })`
 - `signIn('email', { redirect: false, email: 'bill@fillmurray.com' })`
 
+### Using the `email` or `credentials` as redirectable provider type
+
+:::note
+The redirectable provider type is only available for `credentials` and `email` providers.
+:::
+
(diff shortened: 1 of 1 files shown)


balazsorban44:
We should be able to infer the return type from the first argument. See:

https://github.com/nextauthjs/next-auth/blob/cfc0a5508074cf779b1a4ccd3a22a8a6d2ba36f0/packages/next-auth/src/react/index.tsx#L175-L183

https://github.com/nextauthjs/next-auth/blob/cfc0a5508074cf779b1a4ccd3a22a8a6d2ba36f0/packages/next-auth/src/providers/index.ts#L39

It sounds like a bug with the typings then.

Let's open a separate PR. Feel free to have a look. :+1:

PR #4633, opened 2022-05-30
feat(providers): Adding HubSpot Provider

Miciurash:
## ☕️ Reasoning

I've added the HubSpot Provider 

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources


Miciurash:
@ndom91 @balazsorban44 - anything else I need to provide?

packages/next-auth/src/providers/hubspot.ts
```diff
@@ -0,0 +1,104 @@
...
+      const { userInfo } = profile
+
+      return {
+        id: userInfo.user_id,
+        name: userInfo.user,
+        email: userInfo.user,
+
+        // TODO: get image from profile once it's available 
+        // Details available https://community.hubspot.com/t5/APIs-Integrations/Profile-photo-is-not-retrieved-with-User-API/m-p/325521
+        image: "https://avatars.hubspot.net/default-100",
```
ThangHuuVu, line 99: If it's not available I think it's fine to leave it as `null`. Some other providers do that as well.
```suggestion
        image: "https://avatars.hubspot.net/default-100",
```

packages/next-auth/src/providers/hubspot.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+
+const HubSpotConfig = {
+  clientId: process.env.HUBSPOT_CLIENT_ID ?? "DEFAULT_CLIENT_ID",
+  clientSecret: process.env.HUBSPOT_CLIENT_SECRET ?? "DEFAULT_CLIENT_SECRET",
+  redirectUri: process.env.HUBSPOT_REDIRECT_URI ?? "DEFAULT_REDIRECT_URI",
+  authorizationUrl: "https://app.hubspot.com/oauth/authorize",
+  tokenUrl: "https://api.hubapi.com/oauth/v1/token",
+  profileUrl: "https://api.hubapi.com/oauth/v1/access-tokens/"
```
ThangHuuVu, line 22: Shouldn't it be [this endpoint](https://legacydocs.hubspot.com/docs/methods/get-account-details)?

packages/next-auth/src/providers/hubspot.ts
```diff
@@ -0,0 +1,104 @@
...
+  user: string,
+  user_id: string,
+
+  hub_domain: string,
+  hub_id: string,
+}
+
+
+const HubSpotConfig = {
+  clientId: process.env.HUBSPOT_CLIENT_ID ?? "DEFAULT_CLIENT_ID",
```
ThangHuuVu, line 17: I find these "DEFAULT_" values don't help much. Please just remove them 🙏

ThangHuuVu (changes requested):
Thanks for the PR! I've added some comments

packages/next-auth/src/providers/hubspot.ts
Miciurash: @ThangHuuVu the `/me` endpoint provides even less info as compared to `/access-tokens` and it does not contain the `user_id` which is useful

packages/next-auth/src/providers/hubspot.ts
Miciurash: @ThangHuuVu I've outsourced them to the provider instantiation with the required fields. Updated docs for that as well.

packages/next-auth/src/providers/hubspot.ts
Miciurash: updated to null

packages/next-auth/src/providers/hubspot.ts
ThangHuuVu: So I just revisited their updated docs and see that [this endpoint](https://developers.hubspot.com/docs/api/settings/user-provisioning) contains the user ID, I suggest you use it instead of the access-tokens one

Miciurash: @ThangHuuVu not sure what is the benefits of using it. It still does not provide additional info related to the user.

SECURITY.md
```diff
@@ -15,7 +15,7 @@ If you contact us regarding a serious issue:
 
 The best way to report an issue is by contacting us via email at info@balazsorban.com or me@iaincollins.com and yo@ndo.dev, or raise a public issue requesting someone get in touch with you via whatever means you prefer for more details. (Please do not disclose sensitive details publicly at this stage.)
 
-> For less serious issues (e.g. RFC compliance for unsupported flows or potential issues that may cause a problem in the future) it is appropriate to submit these these publically as bug reports or feature requests or to raise a question to open a discussion around them.
+> For less serious issues (e.g. RFC compliance for unsupported flows or potential issues that may cause a problem in the future) it is appropriate to submit these publicly as bug reports or feature requests or to raise a question to open a discussion around them.
```
ndom91, line 18: That was spelled correctly initially. `Publically` :+1:

packages/next-auth/src/providers/hubspot.ts
```diff
@@ -0,0 +1,105 @@
...
+
+
+const HubSpotConfig = {
+  authorizationUrl: "https://app.hubspot.com/oauth/authorize",
+  tokenUrl: "https://api.hubapi.com/oauth/v1/token",
+  profileUrl: "https://api.hubapi.com/oauth/v1/access-tokens"
+}
+
+export default function HubSpot<P extends HubSpotProfile>(
+  options: OAuthUserConfig<P> & { "redirectURL": string }
```
ThangHuuVu, line 23: You don't need `redirectURL` here:
- In `authorization`: no need `redirect_uri`, I've just tested without it and it still works
- In `token`: you can use `context.provider.callbackUrl`

SECURITY.md
Miciurash: @ndom91 I could update if needed. The original spelling is `publicly` according to https://capitalizemytitle.com/publicly-vs-publically/ and also
image

packages/next-auth/src/providers/hubspot.ts
Miciurash: @ThangHuuVu I can keep an eye on the update they have in the future and update this as soon as they have a better endpoint for the user info. Would that work to have this resolved?

packages/next-auth/src/providers/hubspot.ts
Miciurash: @ThangHuuVu this is more to make the developer aware that it's a required field as per HubSpot. The request won't work without it. Callback is more of a client set redirect from my understanding of these terms. I also wanted it part of the configuration and environment variables  for the same reason.

SECURITY.md
Miciurash: @ndom91 are we good here?

packages/next-auth/src/providers/hubspot.ts
ThangHuuVu: @Miciurash sorry for the delayed response 👀
Following is what I suggested, just want to make sure we're on the same page:
```ts
    authorization: {
      url: HubSpotConfig.authorizationUrl,
      params: {
        scope: "oauth",
        // redirect_uri: options.redirectURL, You don't need to pass redirect_uri here, 
// NextAuth.js automatically appends it for you in the authorization request.
        client_id: options.clientId,
      },

    },
    token: {
      url: HubSpotConfig.tokenUrl,
      async request(context) {

        if (!context.params.code) {
          throw new Error("Missing HubSpot authentication code in the callback request parameters")
        }
        const urlParams = new URLSearchParams({
          client_id: options.clientId,
          client_secret: options.clientSecret,
          grant_type: "authorization_code",
          redirect_uri: context.provider.callbackUrl, // instead of options.redirectURL
          code: context.params.code,
        });
        
```

The value of the callback URL will be `http://localhost:3000/api/auth/callback/hubspot`
I think it's best to limit the number of configuration and instead reuse what's already there.

ThangHuuVu (commented):
Thanks again for opening this PR! Let's finalize it soon together 🙌
Could you also attach a video for proof of completion? (This is for formality when integrating with a new provider. I tested your code already and it works fine for me)

packages/next-auth/src/providers/hubspot.ts
Miciurash: Got it @ThangHuuVu, thank you for the example! I fixed it.

Miciurash:
@ThangHuuVu the video proof of the HubSpot login is available here: https://somup.com/c3jY0AUc97

Miciurash:
@ndom91 @ThangHuuVu - are we good to merge here?

packages/next-auth/src/providers/hubspot.ts
```diff
@@ -0,0 +1,104 @@
...
+      url: HubSpotConfig.authorizationUrl,
+      params: {
+        scope: "oauth",
+        client_id: options.clientId,
+      },
+
+    },
+    token: {
+      url: HubSpotConfig.tokenUrl,
+      async request(context) {
```
ThangHuuVu, line 43: @Miciurash, sorry for not suggesting this earlier, but it seems like we can use:
```

    client: {
      token_endpoint_auth_method: "client_secret_post",
    },
    token: HubSpotConfig.tokenUrl,
```

And `node-openid-client` will take care of sending the request. I've also just learned about this in [the other PR](https://github.com/nextauthjs/next-auth/pull/5236) 😅
I've tested it and it works for me locally.

Miciurash: Great suggestion @ThangHuuVu! I learned something new today, updated it and it works like a charm!

ThangHuuVu (approved):
LGTM! 🚀

PR #4623, opened 2022-05-26
fix: signIn infer provider type

ArthurPedroti:
The "P" type it's not passed in any props, so the result type doesn't understand and returns the "false type" always, Adding the "P" at provider type props.

My local test implementation:

```d.ts
// next-auth.d.ts
import 'next-auth'
import {
  BuiltInProviderType,
  RedirectableProviderType
} from 'next-auth/providers'
import {
  LiteralUnion,
  SignInAuthorizationParams,
  SignInOptions,
  SignInResponse
} from 'next-auth/react'

declare module 'next-auth/react' {
  export * from 'next-auth/react'

  export declare function signIn<
    P extends RedirectableProviderType | undefined = undefined
  >(
    provider?: LiteralUnion<P | BuiltInProviderType>,
    options?: SignInOptions,
    authorizationParams?: SignInAuthorizationParams
  ): Promise<
    P extends RedirectableProviderType ? SignInResponse | undefined : undefined
  >
}
```

before:
image

after:
image

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

Issue #4513, opened 2022-05-06
`signIn` with `RedirectableProviderType` type: `response` is possibly `undefined`


b-novikov-ipersonality:
### Environment

```
System:
    OS: Linux 5.17 EndeavourOS
    CPU: (12) x64 AMD Ryzen 5 4600H with Radeon Graphics
    Memory: 5.85 GB / 15.07 GB
    Container: Yes
    Shell: 5.8.1 - /bin/zsh
Binaries:
    Node: 16.14.2 - ~/.nvm/versions/node/v16.14.2/bin/node
    Yarn: 1.22.18 - ~/.nvm/versions/node/v16.14.2/bin/yarn
    npm: 8.7.0 - ~/.nvm/versions/node/v16.14.2/bin/npm
npmPackages:
    next: ^12.1.5 => 12.1.5 
    next-auth: ^4.3.4 => 4.3.4 
    react: ^18.1.0 => 18.1.0 
```

### Reproduction URL

https://codesandbox.io/s/objective-mendel-ixpsco?file=/src/App.tsx

### Describe the issue

According to [documentation](https://next-auth.js.org/getting-started/client#using-the-redirect-false-option) when you supply `email` or `credentials` provider to `signIn` function it returns promise of type `SignInResponse` but in fact return type of the `signIn` function is `Promise<SignInResponse | undefined>`. 

This typing results in typescript error `Object is possibly 'undefined'` if you don't check if `response` exists beforehand.

### How to reproduce

1. Specify `signIn` type to be `RedirectableProviderType`
2. Chain `then` to function call
3. Check for `response.error`
4. See typescript error

### Expected behavior

When specifying `signIn` type to be one of `RedirectableProviderType` the return type should be `Promise<SignInResponse>` and not `Promise<SignInResponse | undefined>`

balazsorban44:
The return type **can** be `undefined` (and it is by default):

https://github.com/nextauthjs/next-auth/blob/641d917175026ec974d1d255064b2591db0c56e6/packages/next-auth/src/react/index.tsx#L227-L233

It depends on

b-novikov-ipersonality:
@balazsorban44 with `redirect: false` option it still warns that `response` is possibly `undefined`.

ArthurPedroti:
The fact of redirect is false doesn't infer the response can be undefined, it's only used to show a use case of this feature.

But I agree with you, the docs were not clear about the response can be undefined, I think it's good to improve and make it clear.

[...]

PR #4679, opened 2022-06-06
fix: signIn infer provider type

ArthurPedroti:
## ☕️ Reasoning

Re-opening #4623

Sorry, I really don't saw this string validation, I put a ternary validation to check it's really a RedirectableProviderType, and if is, it's will pass the "P" at the provider props. I checked and the build it's passing now.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged


ArthurPedroti:
@balazsorban44 Commenting the issue #4513 that you have mentioned, I think this is exactly what you said, the response of signIn really can be "undefined", and this PR doesn't change that.

But I think will be interesting to improve the docs to be more understandable these possibly at it, because the wat that are now, it's really confusing, because it affirms the responses is "SignInResponse" type and not "SignInResponse | undefined"

If you think it's a good idea, I can do this little improvement on the docs and PR too.

balazsorban44 (approved):
Thanks!

Issue #974, opened 2020-12-21
End Session on Browser Close


jacklimwenjie:
**Summary of proposed feature**
Expose an option to _not_ persist a session, which the session will be terminated on browser close. 

**Purpose of proposed feature**
Some system has a "remember me" or "keep me logged in" option to specify if a session should be persisted to avoid a scenario such as forgot to log out from a public computer.

**Detail about proposed feature**
Make the `session` options to support `maxAge:0` to set the cookie to expire on `'Session'`. Currently, setting `maxAge` to value `0` will cause an error.  

Alternatively, expose the cookie as an API from the module so that developer can use it to modify the cookie as needed, such as setting the expiry to `'Session'`.

balazsorban44:
Although it would be possible to expire the cookie when  losing the browser but without federated logout, the user will still potentially be logged in on the computer.

ricardasjak:
Same issue for me, I would like to set expire time to browser session.

narutodev123:
Want user to get logged out if user close the browser. Please guide us on this.

Ronnehag:
> Want user to get logged out if user close the browser. Please guide us on this.

I ended up changing the GET route in /api/auth/[...nextauth]/route.ts.
You also need to remove the auth wrapper in middleware.ts if you have any, that will override this cookie as it renews the expiration on every execution. We're using a basePath, so you need to modify the code to fit your setup.

```
import { handlers } from '@/auth';
import { NextRequest, NextResponse } from 'next/server';

//Handles __Secure prefix in production and no prefix for local dev
const cookieRegex =
  /(?:__Secure-|__Host-)?authjs\.[^=]+=(?:[^,]|,(?!\s*(?:__Secure-|__Host-)?authjs\.))+/g;

function rewriteRequest(req: NextRequest) {
  const {
    headers,
    nextUrl: { protocol, host, pathname, search }
  } = req;

  const detectedHost = headers.get('x-forwarded-host') ?? host;
  const detectedProtocol = headers.get('x-forwarded-proto') ?? protocol;
  const _protocol = `${detectedProtocol.replace(/:$/, '')}:`;
  const url = new URL(
    _protocol + '//' + detectedHost + process.env.NEXT_PUBLIC_BASE_PATH + pathname + search
  );

  return new NextRequest(url, req);
}

function modifyResponseCookies(res: Response): NextResponse {
  const setCookieHeader = res.headers.get('set-cookie');

  if (!setCookieHeader) {
[...]

PR #4693, opened 2022-06-09
WIP Implement session-cookie with Session expiration

### LINKED ISSUE #2534: Support for Session Cookies
### Description 📓

Per [MDN/Cookies#define_the_lifetime_of_a_cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#define_the_lifetime_of_a_cookie):
1. A _Session Cookie_ will expire at the end of a session, e.g when the browser is closed. This is due to lacking an _Expires_ field in the cookie.
1. In contrast, a _Permanent Cookie_ has an Expires field and will persist across sessions (assuming it has not expired yet).

To the best of my understanding, NextAuth _always_ generates a _Permanent Cookie_ as it always adds an _Expires_ field to the cookie:

https://github.com/nextauthjs/next-auth/blob/0fae0c7a8eb224964e38da409ecb9ba0497e8c50/src/server/routes/callback.js#L273-L280

Assuming a Sign-In page that allows users to uncheck _Remember Me_, a _Session Cookie_ should be generated for the user instead of a _Permanent Cookie_. Is it possible to indicate to NextAuth that a _Session Cookie_ should be created?

### How to reproduce ☕️

N/A

### Contributing 🙌🏽

aleehedl:
## ☕️ Reasoning

Session cookies which are cleared upon closing the browser (with `expires: Session`) is a safety feature which has been requested and discussed multiple times (see https://github.com/nextauthjs/next-auth/issues/2534 and https://github.com/nextauthjs/next-auth/issues/974).  This pull request is more of a proof-of-concept and is lacking tests & documentation but I hope it well serve as a starting point for further discussion in order to get the feature properly implemented. 

You can now try the feature by starting the development app (`npm run dev:app`) and setting the `maxAge: false` in the next-auth options.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: #2534

### DIFF
--- a/packages/next-auth/src/adapters.ts
+++ b/packages/next-auth/src/adapters.ts
@@ -11,12 +11,12 @@ export interface AdapterSession {
   sessionToken: string
   /** Used to connect the session to a particular user */
   userId: string
-  expires: Date
+  expires: Date | false
 }
 
 export interface VerificationToken {
   identifier: string
-  expires: Date
+  expires: Date | false
   token: string
 }
 
@@ -79,7 +79,7 @@ export interface Adapter {
   createSession: (session: {
     sessionToken: string
     userId: string
-    expires: Date
+    expires: Date | false
   }) => Awaitable<AdapterSession>
   getSessionAndUser: (
     sessionToken: string
--- a/packages/next-auth/src/core/lib/callback-handler.ts
+++ b/packages/next-auth/src/core/lib/callback-handler.ts
@@ -26,6 +26,9 @@ export default async function callbackHandler(params: {
   options: InternalOptions
 }) {
   const { sessionToken, profile, account, options } = params
+  const { session: { maxAge } } = options;
+  const newExpires = maxAge === false ? maxAge: fromDate(maxAge);
+
   // Input validation
   if (!account?.providerAccountId || !account.type)
     throw new Error("Missing or invalid provider account")
@@ -115,7 +118,7 @@ export default async function callbackHandler(params: {
       : await createSession({
           sessionToken: generateSessionToken(),
           userId: user.id,
-          expires: fromDate(options.session.maxAge),
+          expires: newExpires,
         })
 
     return { session, user, isNewUser }
@@ -145,7 +148,7 @@ export default async function callbackHandler(params: {
         : await createSession({
             sessionToken: generateSessionToken(),
             userId: userByAccount.id,
-            expires: fromDate(options.session.maxAge),
+            expires: newExpires,
           })
 
       return { session, user: userByAccount, isNewUser }
@@ -211,7 +214,7 @@ export default async function callbackHandler(params: {
         : await createSession({
             sessionToken: generateSessionToken(),
             userId: user.id,
-            expires: fromDate(options.session.maxAge),
+            expires: newExpires,
           })
 
       return { session, user, isNewUser: true }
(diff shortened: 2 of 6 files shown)


balazsorban44:
Heads up, this might not work: https://github.com/nextauthjs/next-auth/issues/2534#issuecomment-989041887

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#:~:text=Warning%3A%20Many%20web%20browsers%20have%20a%20session%20restore%20feature%20that%20will%20save%20all%20tabs%20and%20restore%20them%20the%20next%20time%20the%20browser%20is%20used.%20Session%20cookies%20will%20also%20be%20restored%2C%20as%20if%20the%20browser%20was%20never%20closed.

aleehedl:
> Heads up, this might not work: [#2534 (comment)](https://github.com/nextauthjs/next-auth/issues/2534#issuecomment-989041887)

While this might be true, I still think session-expiration is a security enhancement. I'm thinking about those not-so-tech-savvy people using public computers who might not remember to log out. I don't think session-restoration in public computers is commonplace, and users shouldn't be able to access those settings.

balazsorban44:
Please refer to the MDN documentation I linked to. Modern browsers simply don't delete sessions even when closing the browser instance, so this will have no effect on security.

aleehedl:
> Please refer to the MDN documentation I linked to. Modern browsers simply don't delete sessions even when closing the browser instance, so this will have no effect on security, nor will it log out users, unfortunately.

I did read it, and in my opinion, stating that modern browsers don't delete sessions upon close is an over-interpretation of the cited warning. Yes, if the "continue where I left off" feature is enabled, this happens. But in public computers this shouldn't be enabled, and in the browsers that I tested (latest Chrome on Linux and macOS, Safari on macOS), cookies with `expires: Session` are deleted. On macOS you have to close the browser from the dock. But when you do, session cookies are deleted. Have you tried it yourself?

Even if there are some corner cases with cookie deletion in certain browsers / OSs, it's an extra layer of security and doesn't make anything less secure. Or do you think, that implementing a "remember me" feature encourages users to not tick it & to not log out?

balazsorban44:
I've tried both Firefox and Chrome on my Linux machine and I believe the only way not letting the browser restore the cookies was to kill the browser process, which a normal user won't do.

Admittedly this was a few months ago the last time I checked this (around when the first request about this appeared)

FWIW, I am not against this, only in my experience, this wasn't stable enough in modern browsers to be useful but we'll revisit it then to see if things have changed lately.

haase1020:
I would like to check on the status of this PR. I would love to have session-cookie functionality with NextAuth...

newerton (approved):

jamaybyrone:
@balazsorban44 iron-session allows this to be configurable, https://github.com/vvo/iron-session#session-cookies I'm not sure I'm understanding the reasoning of it may not work on all browsers, surely that's for who ever is configuring it to risk

ThangHuuVu:
Hi @aleehedl @balazsorban44, let's find closure for this PR 🤝 in summary, I think the net benefit of this feature won't outweigh the complexity/maintenance burden it brings because of session cookie (without `Expires`) would still be kept by modern browsers, leading to unexpected behaviors.

I am going to close this PR for now due to a long period of inactivity. If we see enough interests on the feature, we shall revisit it again.

PR #4701, opened 2022-06-11
Add node v17 as a supported version

callumok2004:
## ☕️ Reasoning

Updated supported node engines to include v17. I have been using next-auth with v17 since it was released, there have been no issues with it. Unless there is a reason this is not currently marked as supported, I fail to see why it should not be added!

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged


balazsorban44:
Will be fixed by #4299

PR #4702, opened 2022-06-11
allow user set canonical_urls on request with nextjs.

awitherow:
## ☕️ Reasoning

I am using a single Vercel instance for multiple domains, which generate dynamic content based on the domain. This would include which callback/canonical url to use for next-auth.

This is just my assumption of what might work with nextjs.

I wanted to submit early to get some eyes on the approach!

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/packages/next-auth/src/next/index.ts
+++ b/packages/next-auth/src/next/index.ts
@@ -20,12 +20,14 @@ async function NextAuthNextHandler(
 ) {
   const { nextauth, ...query } = req.query
 
+  options.canonical_url = options.canonical_url ?? req.headers["x-forwarded-host"]
+
   options.secret =
     options.secret ?? options.jwt?.secret ?? process.env.NEXTAUTH_SECRET
 
   const handler = await NextAuthHandler({
     req: {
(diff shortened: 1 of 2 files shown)


packages/next-auth/src/next/index.ts
```diff
@@ -20,12 +20,14 @@ async function NextAuthNextHandler(
...
   const { nextauth, ...query } = req.query
 
+  options.canonical_url = options.canonical_url ?? req.headers["x-forwarded-host"]
+
   options.secret =
     options.secret ?? options.jwt?.secret ?? process.env.NEXTAUTH_SECRET
 
   const handler = await NextAuthHandler({
     req: {
-      host: detectHost(req.headers["x-forwarded-host"]),
```
lluia, line 28: Why not use the `'x-forwarded-host'` header? We have that already set in place. If you're using **Vercel** I think it'll fit your use case. Did you try it?

lluia:
@awitherow, are you still interested in this, or are you ok if we close it?

PR #4704, opened 2022-06-12
feat: allow standard `Request` in `NextAuthHandler`

balazsorban44:
Using [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request) will make it easier to migrate `next-auth` to other frameworks/runtimes.

This PR also adds initial tests for our core and replaces Jest transforms to use SWC

In a follow up PR, I plan to change the result of `NextAuthHandler` to a standard [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response)


ndom91 (approved):
Lgtm but I haven't tested it

PR #3302, opened 2021-11-29
feat(providers): ensure GitHub provider always gives an email

nkalupahana:
## Reasoning 💡

Currently, if a GitHub user doesn't have an email public, it's not given in their profile. This is rather meaningless -- since email scope is requested, all you have to do is make another request to GitHub in order to get it. This functionality might as well be built directly into Next Auth -- and that's what this PR does. 

#374 shows that this is a common issue, and there are several workarounds listed in the issue. It was said that this functionality would be included in v4, so I've taken the initiative to get it done, since my organization would benefit greatly from it being built in directly.

## Checklist 🧢

~- [ ] Documentation~ Note that [documentation](https://next-auth.js.org/v3/providers/github) will have to be updated when this is merged in.
~- [ ] Tests~
- [x] Ready to be merged

## Affected issues 🎟
#2524

PR #4708, opened 2022-06-13
fix : Github provider docs

lukaakhalbedashvili:
by default, GitHub does not return email, at least not anymore.

### DIFF
--- a/docs/docs/providers/github.md
+++ b/docs/docs/providers/github.md
@@ -42,5 +42,5 @@ Only allows one callback URL per Client ID / Client Secret.
 :::
 
 :::tip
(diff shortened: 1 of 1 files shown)


ndom91:
Email is a critical part of our flow, maybe see which scope is newly required to also return it? 

Btw my next-auth 4.4.0 works with github out of the box still, i.e. with email address 👍

balazsorban44:
This is true, that's why we are making an extra request to get it: https://github.com/nextauthjs/next-auth/blob/8c5d9faad65931e478147412fd905bdd70aa0a4c/packages/next-auth/src/providers/github.js#L15-L29

Added in https://github.com/nextauthjs/next-auth/pull/3302

PR #4744, opened 2022-06-21
Implement updateSession & isUpdate parameter

### LINKED ISSUE #596: getSession call does not trigger session object update until window loses focus

### LINKED ISSUE #2129: Session is not renewed on client side right after SSR state changes
## Description 🐜

The problem is described here: https://github.com/nextauthjs/next-auth/discussions/2123#discussion-3395145

Seems like when changing the session at server side, the client will still use the old session states. 
This leads to false informations at the client side. 

## How to reproduce ☕️

https://codesandbox.io/s/4gokt?file=/pages/_app.tsx

Changes is at `[...nextauth]`, `_app` and `client`.
_app has getinitialprops, which uses the getSession method. 
It will at the server side and client side use the same state, but i expect the client side to use new the new state. 

## Environment 🖥

```
npx: installed 1 in 1.676s

  System:
    OS: macOS 11.4
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 854.63 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.0 - ~/.nvm/versions/node/v14.16.0/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 6.14.11 - ~/.nvm/versions/node/v14.16.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 91.0.4472.77
    Safari: 14.1.1
```

### LINKED ISSUE #2208: useSession/Provider does not update when a new session is created until page reload.

christoph-pflueger:
## ☕️ Reasoning

Implements `updateSession` to be able to update the `SessionProvider`'s state without having to reload the page. The update event is propagated to the `jwt` and `session` callbacks via an `isUpdate` parameter. This helps to avoid having to check if the token and session is up to date on every request.

As there are currently no tests implemented for callbacks, I skipped that step.

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: #596 
Fixes: #3941
Fixes: #2208
Might fix: #2129

## 📌 Resources

### DIFF
--- a/docs/docs/configuration/callbacks.md
+++ b/docs/docs/configuration/callbacks.md
@@ -22,10 +22,10 @@ You can specify a handler for any of the callbacks below.
     async redirect({ url, baseUrl }) {
       return baseUrl
     },
-    async session({ session, user, token }) {
+    async session({ session, user, token, isUpdate }) {
       return session
     },
-    async jwt({ token, user, account, profile, isNewUser }) {
+    async jwt({ token, user, account, profile, isNewUser, isUpdate }) {
       return token
     }
 ...
@@ -107,12 +107,14 @@ The redirect callback may be invoked more than once in the same flow.
 This callback is called whenever a JSON Web Token is created (i.e. at sign
 in) or updated (i.e whenever a session is accessed in the client). The returned value will be [encrypted](/configuration/options#jwt), and it is stored in a cookie.
 
-Requests to `/api/auth/signin`, `/api/auth/session` and calls to `getSession()`, `useSession()` will invoke this function, but only if you are using a [JWT session](/configuration/options#session). This method is not invoked when you persist sessions in a database.
+Requests to `/api/auth/signin`, `/api/auth/session` and calls to `getSession()`, `useSession()` and `updateSession()` will invoke this function, but only if you are using a [JWT session](/configuration/options#session). This method is not invoked when you persist sessions in a database.
 
 - As with database persisted session expiry times, token expiry time is extended whenever a session is active.
 - The arguments _user_, _account_, _profile_ and _isNewUser_ are only passed the first time this callback is called on a new session, after the user signs in. In subsequent calls, only `token` will be available.
 
 The contents _user_, _account_, _profile_ and _isNewUser_ will vary depending on the provider and on if you are using a database or not. You can persist data such as User ID, OAuth Access Token in this token. To make it available in the browser, check out the [`session()` callback](#session-callback) as well.
+
+The _isUpdate_ parameter is only `true` for requests initiated with `updateSession()`.
 
 ```js title="pages/api/auth/[...nextauth].js"
 ...
@@ -136,15 +138,17 @@ Use an if branch to check for the existence of parameters (apart from `token`). 
 
 The session callback is called whenever a session is checked. By default, only a subset of the token is returned for increased security. If you want to make something available you added to the token through the `jwt()` callback, you have to explicitly forward it here to make it available to the client.
 
-e.g. `getSession()`, `useSession()`, `/api/auth/session`
+e.g. `getSession()`, `useSession()`, `updateSession()`, `/api/auth/session`
 
 - When using database sessions, the User object is passed as an argument.
 - When using JSON Web Tokens for sessions, the JWT payload is provided instead.
 
+The _isUpdate_ parameter is only `true` for requests initiated with `updateSession()`.
+
 ```js title="pages/api/auth/[...nextauth].js"
 ...
 callbacks: {
-  async session({ session, token, user }) {
+  async session({ session, token, user, isUpdate }) {
     // Send properties to the client, like an access_token from a provider.
     session.accessToken = token.accessToken
     return session
--- a/docs/docs/getting-started/client.md
+++ b/docs/docs/getting-started/client.md
@@ -196,6 +196,28 @@ The tutorial [securing pages and API routes](/tutorials/securing-pages-and-api-r
 
 ---
 
+## updateSession()
+
+- Client Side: **Yes**
+- Server Side: No
+
+The `updateSession()` method can be used client-side to update the SessionProvider's state. This is especially useful, when the token or session has to be updated without a full page reload.
+
+#### Example
+
+```js
+import { updateSession } from "next-auth/react"
+
+const updateUser = async () => {
+  // Update the user, e.g. via a POST request
+
+  // Then update the SessionProvider's state
+  await updateSession()
+}
+```
+
+---
+
 ## getCsrfToken()
 
 - Client Side: **Yes**
--- a/packages/next-auth/src/core/index.ts
+++ b/packages/next-auth/src/core/index.ts
@@ -130,7 +130,14 @@ export async function NextAuthHandler<
       case "providers":
         return (await routes.providers(options.providers)) as any
       case "session": {
-        const session = await routes.session({ options, sessionStore })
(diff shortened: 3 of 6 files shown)


DitlevHB:
Any ETA for the release of this?

qiqo:
Can I gently nudge this pull request? I hope if gets merged already as this is also a blocker in our application - I think this should be considered critical

git-chrisp:
> Can I gently nudge this pull request? I hope if gets merged already as this is also a blocker in our application - I think this should be considered critical

The fact that this pull request hasn't received as much as a reply for almost three months even though it has been requested multiple times is mind-blowing to me. It's not like other pull requests are not being worked on.

qiqo:
> > Can I gently nudge this pull request? I hope if gets merged already as this is also a blocker in our application - I think this should be considered critical
> 
> The fact that this pull request hasn't received as much as a reply for almost three months even though there appears to be genuine interest is mind-blowing to me. It's not like other pull requests are not being worked on either.

and the fix isn't even outrageously complex as well it's practically  just assistive code to extend session's functionality, it just needs a quick nod/verify from them. The inaction is like compelling users to fork and manually implement the fix.

devrsi0n:
@balazsorban44 Could you help review this PR? It seems like a common feature should be built into next-auth. Thank you for your time.

qiqo:
@balazsorban44 hoping for your time and consideration on this fixes. Thanks

ThangHuuVu (commented):
Thanks for the PR @git-chrisp  and sorry for the delayed response, I reviewed your PR
It seems like the added method  `updateSession` is exactly the same as `getSession`, the only difference is in the callback `session` and `jwt`
Could you give an example case where the extra `isUpdate` param is helpful? 🤔

git-chrisp:
> Thanks for the PR @git-chrisp and sorry for the delayed response, I reviewed your PR It seems like the added method `updateSession` is exactly the same as `getSession`, the only difference is in the callback `session` and `jwt` Could you give an example case where the extra `isUpdate` param is helpful? 🤔

The implemented `updateSession` doesn't call `getSession`. It calls the `__NEXTAUTH._getSession` which is encapsulated inside the `SessionProvider` and cannot be accessed at the moment.

Assume someone changes something in his profile settings that is stored in e.g. the token's custom claims. There is no way to update this change in the `SessionProvider` at the moment without reloading the entire page.

The `isUpdate` parameter helps in the backend to differentiate between normal `getSession`, `_getSession` and `updateSession` calls.

shahryarjb:
Is there any deadline to fix this problem? For now the only way to fix this issue is reloading page and this way clean the state or the action user wants to do
Thank you.

mohamedGamalAbuGalala:
Is any update on this pr soon, we are having the same problem authenticating with a custom PKCE provider, and when refreshing the token and have the new token back. the client keeps calling JWT with the old invalid token. and we use one-time refresh tokens so, it's pretty impossible to get around this setup without any compromise with the rules we already have.

rustunooldu:
May I kindly ask for an update? I had this PR bookmarked months ago, and it's a little disheartening to see that there hasn't been any progress... It's a blocker for our application, and there seems to be a lot of interest from other people as well.

raulmarindev:
I'm sorry if I'm mistaken but this related to this workaround? https://stackoverflow.com/a/70405437/9123724

lemcii:
Currently using `window.refresh`/`window.location.pathname` since I cannot await session update after updating a users profile. The changes here would likely solve that.

shahryarjb:
> window.location.reload()

It is not good, I need to update my auth 2 tokens {refresh token, access token} every 1 hour, but for now I am forced to reload my site: [it is my next auth API](https://github.com/mishka-group/mishka-cms-front/blob/master/pages/api/auth/%5B...nextauth%5D.js), and it is my client:

https://github.com/mishka-group/mishka-cms-front/blob/master/apps/mishka_user/helper/authHelper.ts#L121-L123

rolanday:
Happy to send reviewers a case of beer to get this PR approved (and pulled). As others have noted, this is basic/needed functionality. Thanks!

docs/docs/configuration/callbacks.md
```diff
@@ -22,10 +22,10 @@ You can specify a handler for any of the callbacks below.
     async redirect({ url, baseUrl }) {
       return baseUrl
     },
-    async session({ session, user, token }) {
+    async session({ session, user, token, isUpdate }) {
```
devrsi0n, line 25: Could you use `isUpdated` instead? `isUpdate` has a grammar issue.

kblizeck:
> 

I'd pitch in for a second case - this would be super beneficial!

balazsorban44:
Closing in favor of #7056

PR #4745, opened 2022-06-21
[edge] support request.cookies as a map

Schniz:
## ☕️ Reasoning

in next Next.js versions, NextRequest.cookies will be an instance of NextCookies which is
some kind of a Map, instead of a plain object.

This commit checks whether there's a `get` function in req.cookies, and acts accordingly,
to make sure we will support newer Next.js versions with Edge Functions/Middleware

I don't see any tests here for that. I can introduce some, but not sure if it makes sense. Please let me know what to do with this!

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/core/lib/cookie.ts
+++ b/packages/next-auth/src/core/lib/cookie.ts
@@ -120,7 +120,7 @@ export class SessionStore {
   constructor(
     option: CookieOption,
     req: {
-      cookies?: Record<string, string>
+      cookies?: Record<string, string> | { get: (key: string) => string }
       headers?: Headers | IncomingHttpHeaders | Record<string, string>
     },
     logger: LoggerInstance | Console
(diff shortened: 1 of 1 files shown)


balazsorban44 (approved):

packages/next-auth/src/core/lib/cookie.ts
```diff
@@ -132,7 +132,10 @@ export class SessionStore {
 
     for (const name in req.cookies) {
```
patelnav, line 133: @Schniz & @balazsorban44 
I think this needs to be a for-of loop for Map to work.
I'm running into this issue where the cookie isn't being read in the middleware
The for-in loop just iterates the properties of the Map object and not the actual constants

PR #4747, opened 2022-06-21
Add logic to pass newUserInfo when creating a new user

### LINKED ISSUE #1069: Collect profile info in passwordless email signIn() flow
I want to collect a new user's full name using: `signIn("email", { email })`

I encountered this use case.

A createProfile option:
`signIn("email", { email: "lawrence@example.com", createProfile: { name: "Lawrence" } })`
If createProfile exists, name will be persisted to the database. If omitted, standard behavior occurs.

**Potential problems**
Couldn't think of any.

Collecting username after authenticated, but this is inconvenient.

**Additional Context**
Would probably involve changing the `{ email }` placeholder in [line 159 of src/server/routes/callback.js](https://github.com/nextauthjs/next-auth/blob/canary/src/server/routes/callback.js#L159) if createProfile exists.

Yes! (But I'm new to this so might need guidance later on...)

dgrahamlong:
## ☕️ Reasoning

Added logic that allows the passing of additional info when creating a new user. I wanted the ability to give prospective users a verification code that they could use to allow them to create accounts because I don't want anyone to be able to create an account at the moment. I can then check this verification code in the signup callback before the user is created. That being said, I believe there are more use cases for this such as, allowing the user to save things like their display name when creating an account. I think it's pretty common to ask for a little bit more info when signing up other than just wanting a user's email or Gmail/apple/etc account.

Note: This is my very first pull request so if there is anything I'm missing, please lmk.

## 🧢 Checklist

- [x] Documentation
- [x] Tests (passed pnpm test)
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: https://github.com/nextauthjs/next-auth/issues/1069

## 📌 Resources

### DIFF
--- a/docs/docs/configuration/callbacks.md
+++ b/docs/docs/configuration/callbacks.md
@@ -62,6 +62,8 @@ callbacks: {
 
   You can check for the `verificationRequest` property to avoid sending emails to addresses or domains on a blocklist (or to only explicitly generate them for email address in an allow list).
 
+- If you pass additional info for new users to the `newUserInfo` option of the client `signIn()` method, that information can be found in `user.newUserInfo`. This can help you determine if the new user can be created based on the additional information you provide.
+
 * When using the **Credentials Provider** the `user` object is the response returned from the `authorize` callback and the `profile` object is the raw body of the `HTTP POST` submission.
 
 :::note
--- a/docs/docs/getting-started/client.md
+++ b/docs/docs/getting-started/client.md
@@ -126,10 +126,10 @@ function Auth({ children }) {
   // if `{ required: true }` is supplied, `status` can only be "loading" or "authenticated"
   const { status } = useSession({ required: true })
 
-  if (status === 'loading') {
+  if (status === "loading") {
     return <div>Loading...</div>
   }
-  
+
   return children
 }
 ```
@@ -355,6 +355,25 @@ e.g.
   url: string | null
 }
 ```
+
+### Using the `newUserInfo` option
+
+You can pass additional data to save when creating a new user by passing a `newUserInfo` option to the second argument of `signIn()`.
+
+`newUserInfo` can take a string or an object
+
+e.g.
+
+- `signIn("email", {email, newUserInfo: {displayName: 'Bill Johnson'}})`
+- `signIn("google", {email, newUserInfo: 'Bill Johnson'})`
+
+:::note
+ Your database adapter has to be configured to accept the new `newUserInfo` parameter when creating a user or an error will be triggered.
+:::
+
+:::note
+`newUserInfo` will be ignored when the user already exists
+:::
 
 ### Additional parameters
 
--- a/packages/next-auth/src/adapters.ts
+++ b/packages/next-auth/src/adapters.ts
@@ -3,6 +3,7 @@ import { Account, User, Awaitable } from "."
 export interface AdapterUser extends User {
   id: string
   emailVerified: Date | null
+  newUserInfo?: Record<string, any> | string | null
 }
 
 export interface AdapterSession {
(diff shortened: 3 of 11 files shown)


HummingBird24:
Hi, @balazsorban44 and @ndom91, I know you two are pretty busy but I was wondering if I can get an update about this proposal? Thanks for all your work!

balazsorban44:
Thanks, our top priority is elsewhere right now, I'll revisit other PRs like this when #4299 and #4769 are merged and tested. :slightly_smiling_face:

I will most likely want to iterate on this PR later.

HummingBird24:
> Thanks, our top priority is elsewhere right now, I'll revisit other PRs like this when #4299 and #4769 are merged and tested. 🙂
> 
> I will most likely want to iterate on this PR later.

Understood. Thanks for the quick response and thanks again for all the hard work you've done!  It's truly appreciated 🙌🏽

sakibulalam:
are there any plans to merge this? is there a workaround currently to enable this?

michal-kapala:
> are there any plans to merge this? is there a workaround currently to enable this?

For user creation you probably have to make a separate request as there is no space in the core types to add your custom data into as of now. 
For getting custom user data from a database to a client you could use [`session()` callback](https://next-auth.js.org/configuration/callbacks#session-callback) to pass any extension properties to your `User` model and add them to the returned session object for the data to make it to the client. This method currently has two drawbacks for Typescript:
- it's type-unsafe because it forces the use of `[...nextauth].js` rather than typed config
- `useSession()` and `unstable_getServerSession` return a session object with nullable `user` property of unnamed type which is not extendable when using typed nextauth config:

```js
(property) DefaultSession.user?: {
    name?: string | null | undefined;
    email?: string | null | undefined;
    image?: string | null | undefined;
} | undefined
```
Accessing a custom property of `session.user` returned by the session hooks results in a type error:
```
`Property X does not exist on type '{ name?: string | null | undefined; email?: string | null | undefined; image?: string | null | undefined; }'.
```

To get around that right now you can define an extension for `User` or `AdapterUser` type and cast the user returned by a hook to the extension.

# Example

Say you'd want to add a `role` property and return it to the client:
```ts
// CustomUser.ts
import { User } from "next-auth"

export interface CustomUser extends User {
	role: string
}
```
To pass the value down to the client, you define `session` callback:
```js
// [...nextauth].js

export const authOptions = {
  adapter: yourAdapter,
  providers: yourProviders,
  callbacks: {
    // invoked on a session hook call
    async session({ session, token, user }) {
      // With JS we can add a 'role' prop on session.user (currently impossible with TS)
      session.user.role = user.role;
      return session;
    }
  },
}
export default NextAuth(authOptions);

```

Lastly, to access `role` in a component we cast the user object returned by a session hook to our custom type to avoid a type error:
```ts
// Component.tsx
import { useSession } from "next-auth/react"
import type { CustomUser } from './CustomUser'

export default function Component() {
  const { data: session} = useSession();

  // Cast user data to include role injected in session() callback
  var userCast = session?.user as CustomUser;

  // Errors out
  console.log(session?.user.role)

  // Ok
  return (
    <div>{userCast.role}</div>
  )
}
```

This PR would allow you to define your extension object using `AdapterUser`, which is a possible arg type inside of `session` callback, however I think then `DefaultSession.?user` needs to become `AdapterUser` too so you can use `[...nextauth].ts` and pass your custom user with `useSession`/`getSession`/`unstable_getServerSession` without that much worry about the core types.

For reference here's a [dashboard page](https://github.com/michal-kapala/overact/blob/main/pages/dashboard.tsx) that utilizes the above trick. Looking forward to see this merged.

ThangHuuVu:
Hi there @stoickeyboard I think we're ready to review this PR again - could you please help resolve the conflicts and move the changes under `packages/core` instead? 🙌 In `packages/next-auth` we will only be accepting critical fix moving on.

AlexErrant:
I build on this PR here https://github.com/nextauthjs/next-auth/pull/7234

ThangHuuVu:
Let's close this and move the discussion to https://github.com/nextauthjs/next-auth/pull/7234 🙏

PR #4780, opened 2022-06-28
chore: fix builds/tests/eslint

balazsorban44:
ESLint has not been working for me for a while now, I finally had a look at it to fix it. Builds also started failing locally, because of some missing types, which should be fixed now as well.

I also turned on linting on test files.


.eslintrc.js
```diff
@@ -0,0 +1,33 @@
+const path = require('path')
```
ThangHuuVu, line 1: I moved eslint config to here to use the `path.resolve` in `parserOptions.project`

tsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
+  "include": [
+    ".eslintrc.js",
```
ThangHuuVu, line 3: This file is included to run lint on `eslintrc.js`

.eslintignore
```diff
@@ -0,0 +1 @@
+.eslintrc.js
```
balazsorban44, line 1: Can we make this not an extra file? :sweat_smile:

.eslintrc.js
```diff
@@ -0,0 +1,32 @@
...
+
+module.exports = {
+  root: true,
+  parser: "@typescript-eslint/parser",
+  parserOptions: {
+    project: [
+      path.resolve(__dirname, './packages/**/tsconfig.eslint.json'),
```
balazsorban44, line 8: Would be happy if this did not need to move out of `package.json`, but I see you are using `path`. I won't have strong feelings about it if there is no workaround. :thinking:

ThangHuuVu: I moved it inside `pacakge.json` again! I thought missing `path` and `__dirname` was the problem but it turned out weren't

.eslintignore
ThangHuuVu: Removed because eslintConfig is restored in `package.json`

ndom91 (approved):
Tests pass.. LGTM 😊👍

ThangHuuVu:
@ndom91 great! @balazsorban44 and Nico, could you check if it works on your local machine? 🙌

ndom91:
> @ndom91 great! @balazsorban44 and Nico, could you check if it works on your local machine? 🙌

Yeah, can check later this evening when I'm back home 👍

ThangHuuVu:
In your IDE, make sure to add Prettier `configPath` for auto-save:
```
{
  "prettier.configPath": "package.json"
}
```
Merging this now, I pulled a fresh repo and it still works 🤞

ThangHuuVu (approved):

Issue #4234, opened 2022-03-22
Infinite redirect with middleware and basePath


nicks6853:
### Question 💬

I have a NextJS app setup to run under the `/admin` basepath. Whenever I use the middleware, I am having problems with it sending me to the wrong url for the login page. If I try passing a custom sign in page to the options, it redirects me infinitely. Can anyone point me in the right direction? Is this even possible?

### How to reproduce ☕️

I am running `next-auth@4.3.1`, `next@12.1.0`

`_app.js`

```javascript
import { SessionProvider } from 'next-auth/react';
import '../styles/globals.css';

function App({ Component, pageProps: { session, ...pageProps } }) {
  return (
    <SessionProvider
      session={session}
      basePath="/admin/api/auth"
    >
      <Component {...pageProps} session={session} />
    </SessionProvider>
  );
}

export default App;
```

`_middleware.js `

> Note that without the custom signin page, the app redirects me to `/api/auth/signin` which does not exists (I can see the page if I manually go to `/admin/api/auth/signin`) and with the custom signin page I get an infinite redirect error.

```javascript
import withAuth from 'next-auth/middleware';

export default withAuth({
  pages: {
    signIn: '/admin/signin',
  },
});
```

`next.config.js`
```javascript
/** @type {import('next').NextConfig} */

const nextConfig = {
  reactStrictMode: true,
  env: {
    COGNITO_CLIENT_ID: '<removed>',
    COGNITO_CLIENT_SECRET: '<removed>',
    COGNITO_ISSUER: '<removed>',
    COGNITO_DOMAIN: '<removed>',
    NEXTAUTH_URL: 'http://localhost:3000/admin/api/auth',
    NEXTAUTH_SECRET: '<removed>',
  },
  basePath: '/admin',
};

module.exports = nextConfig;
```

`pages/signin.js`

```javascript
import { Button } from '@mantine/core';
import { signIn } from 'next-auth/react';

export default function SignIn({ providers }) {
  return (
    <Button onClick={() => signIn('cognito')}>
      Sign in with Cognito
    </Button>
  );
}
```

`pages/api/auth`

```javascript
import NextAuth from 'next-auth/next';
import CognitoProvider from 'next-auth/providers/cognito';

export default NextAuth({
  providers: [
    CognitoProvider({
      clientId: process.env.COGNITO_CLIENT_ID,
      clientSecret: process.env.COGNITO_CLIENT_SECRET,
[...]

PR #4788, opened 2022-06-29
Middleware unauthorized callback option

### LINKED ISSUE #4787: Infinite loop Middleware isAuth Handling
### Environment

 System:
    OS: macOS 12.4
    CPU: (10) arm64 Apple M1 Pro
    Memory: 75.20 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.1 - ~/.nvm/versions/node/v16.15.1/bin/node
    Yarn: 1.22.15 - ~/.nvm/versions/node/v16.15.1/bin/yarn
    npm: 8.11.0 - ~/.nvm/versions/node/v16.15.1/bin/npm
  Browsers:
    Chrome: 103.0.5060.53
    Safari: 15.5
  npmPackages:
    next: ^12.2.0 => 12.2.0 
    next-auth: ^4.7.0 => 4.7.0 
    react: 17.0.2 => 17.0.2 

### Reproduction URL

https://next-auth-example.vercel.app/me

### Describe the issue

##Stuck Infinite loop when using middleware for page authorizing. 

If your authorization callback is anything more complex than checking for users to be logged in for example role permissions.

Then it is very easy to get stuck in infinite loop, and current withAuth Config does not allow for easy configuration to change behavior. 

### How to reproduce

Optional steps:

1. Go to https://next-auth-example.vercel.app
2.  Log in

Required Steps:

3. Go to https://next-auth-example.vercel.app/me 
4. Sign in

### Expected behavior

Currently behaves as expected, but not user intuitive. Many people will see this as bug. 

 #4234

HXavS:
# Discussions
#4791 

## ☕️ Reasoning

Added callback option for middleware isAuthorized returning false. 

###Previous behavior was to always go to get stuck in signIn redirect loop. 
 Changing Pages.signIn option in middleware is inadequate because a user can be logged in but not authorized to a page. Changing the sign in page would cause user's not signed in to end up on the same page. 
 
Current workaround is to Create a middleware function and pass as first arg of withAuth. This issue with this is that you would have to validate only user login with isAuthorized and handle actual permissions check in the function being fired "onSuccess".
This is not intuitive and more complex then needed for most use cases. Callbacks stream line the handling of different authentication scenarios.
 
Default Behavior remains the same with noUser actions converted to configurable callback with default behavior working as current release. 

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: #4787 #4234

## 📌 Resources

### DIFF
--- a/apps/dev/middleware.ts
+++ b/apps/dev/middleware.ts
@@ -4,6 +4,35 @@ export const config = { matcher: ["/middleware-protected"] }
 
 // Other ways to use this middleware
 
+
+
+// import { NextResponse } from 'next/server';
+// import { withAuth } from "next-auth/middleware"
+// import { JWT } from 'next-auth/jwt';
+
+
+// //Just done here fore quick example of middleware changes. In practice would recommend extending JWT type in types/next-auth.d.ts
+// interface TokenMiddle extends Partial<JWT>{ 
+//   home: URL | null
+// }
+
+// export default withAuth(
+//   function middleware(req, ev) { //function here can run custom logic and return next response.
+//     console.log(req, ev)
+//     return undefined // NOTE: `NextMiddleware` should allow returning `void`
+//   },
+//   {
+//     pages: {
+//       signIn: "/unauthorized"
+//     },
+//     callbacks: {
+//       authorized: ({ token }) => !!token && false, //User isAuthorized should different then logged in.
+//       unAuthorized: ({ req, token }) => NextResponse.redirect(new URL((token as TokenMiddle).home ?? "/unauthorized", req.nextUrl.origin)), 
+//     }
+//   }
+// )
+
+
 // import withAuth from "next-auth/middleware"
 // import { withAuth } from "next-auth/middleware"
 
--- a/packages/next-auth/src/next/middleware.ts
+++ b/packages/next-auth/src/next/middleware.ts
@@ -12,6 +12,11 @@ type AuthorizedCallback = (params: {
   req: NextRequest
 }) => Awaitable<boolean>
 
+type UnAuthorizedCallback = (params: {
+  token: JWT | null
+  req: NextRequest
+}) => NextResponse | null
+
 export interface NextAuthMiddlewareOptions {
   /**
    * Where to redirect the user in case of an error if they weren't logged in.
@@ -83,6 +88,8 @@ export interface NextAuthMiddlewareOptions {
      * [Documentation](https://next-auth.js.org/getting-started/nextjs/middleware#api) | [`signIn` callback](configuration/callbacks#sign-in-callback)
      */
     authorized?: AuthorizedCallback
+
+    unAuthorized?: UnAuthorizedCallback
   }
 }
 
@@ -94,6 +101,16 @@ async function handleMiddleware(
   const signInPage = options?.pages?.signIn ?? "/api/auth/signin"
   const errorPage = options?.pages?.error ?? "/api/auth/error"
   const basePath = parseUrl(process.env.NEXTAUTH_URL).path
+
+  const noUser = () => {
+    const signInUrl = new URL(signInPage, req.nextUrl.origin)
+    signInUrl.searchParams.append(
+      "callbackUrl",
+      `${req.nextUrl.pathname}${req.nextUrl.search}`
+    )
+    return NextResponse.redirect(signInUrl)
+  }
+
   // Avoid infinite redirect loop
   if (
     req.nextUrl.pathname.startsWith(basePath) ||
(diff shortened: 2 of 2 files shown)


balazsorban44:
Thanks. so there is `pages.signIn` that you can set: https://next-auth.js.org/configuration/nextjs#pages

I think we could add a query parameter if the user was not authorized and show that info in the signin page to inform the user correctly,instead of adding another callback method

HXavS:
I originally went the route of using the pages.signIn. 
I thought a call back was more appropriate in some cases for a few reasons. 
- I do not want users who are not authorized to access admin only areas to have to log in again. 
- I would rather return them to where they were originally or default to their "homepage" with the error message. 
- If a user was not signed in and you redirected to a page without the sign in it creates a messy user experience. 
Let's say an admin goes to direct link, but is not logged in. Middleware Redirects to default user dashboard, which triggers another not authorized redirect to pages.signIn configured in  api/auth/[...next-auth].ts. When the admin does sign in the callback will be to the default user dashboard and not original admin request.

Also the default login for next-auth does not included meaningful navigation for users to get out of the loop.
If a user was to go to http://example.com/admin directly, then they would only be able to see the login screen with error message saying they are not authorized even if they did have an account.

balazsorban44:
the reason is that I don't want to introduce another callback method for this

HXavS:
Okay then maybe something simple like this to prevent infinite sign in loops for users that exist but are not authorized for the route the middleware is evaluating. 

`
  // the user is not logged in, redirect to the sign-in page
  const signInUrl = new URL(signInPage, req.nextUrl.origin)
  if(!token)
    signInUrl.searchParams.append(
      "callbackUrl",
      `${req.nextUrl.pathname}${req.nextUrl.search}`
    )

`

MrPumpking:
How about allowing the user to return `NextResponse`  or `boolean` from `authorized`?

If the return type is boolean then we would apply current logic (redirect to sign in if false, if true then allow the request to proceed). Otherwise, the middleware would just return the provided `NextResponse` allowing for custom redirects

```ts

export default withAuth({
  callbacks: {
    authorized: ({ req, token }) => {
      const path = req.nextUrl.pathname;

      if (path.startsWith('/admin') && token !== null && token.role !== 'admin') {
        const url = req.nextUrl.clone();
        url.pathname = '/redirect-to'
        return NextResponse.redirect(url)
      }

      return token !== null;
    }
  }
})

export const config = {
  matcher: [
    "/profile/:path*",
    "/admin/:path*"
  ]
}
```

HXavS:
@MrPumpking. I personally like this idea. In my opinion it simplifies user role handling logic. The examples in documentation show the admin role check. Yet, the current implementation seems more geared towards should user be allowed to log in. and not page specific permission handling. 

Currently, the withAuth function from "next-auth/middleware" can be passed a function that is called onSuccess of Authorization. Custom role routing logic can be handled there. 

The main thing I want to have fixed is that default behavior of infinitely looping logged in users constantly being brought to a login screen if authentication is ever returned false.

pybuche:
👋 Is there any news on this PR? I'm also interested in being able to inform the user the requested page is not authorized, instead of redirect them to the sign in page (that makes no sense if the user is correctly logged in)

Please tell me if I can help make this move forward 🤗

ThangHuuVu:
IMO we could also allow returning a `URL` instance so that we can pass in our `NextResponse.redirect` call, would be simpler for developers. 

```ts

export default withAuth({
  callbacks: {
    authorized: ({ req, token }) => {
      const path = req.nextUrl.pathname;

      if (path.startsWith('/admin') && token !== null && token.role !== 'admin') {
        const url = req.nextUrl.clone();
        url.pathname = '/redirect-to'
        return url
      }

      return token !== null;
    }
  }
})

export const config = {
  matcher: [
    "/profile/:path*",
    "/admin/:path*"
  ]
}
```

The internal would look something like this
```ts

  const isAuthorizedOrRedirectURL =
    (await options?.callbacks?.authorized?.({ req, token })) ?? !!token

  // the user is authorized, let the middleware handle the rest
  if (isAuthorizedOrRedirectURL === true) return await onSuccess?.(token)
  // the user is not authorized and a custom redirect URL is defined, use that
  if (isAuthorizedOrRedirectURL instanceof URL) {
    return NextResponse.redirect(isAuthorizedOrRedirectURL)
  }
```

Let me know what you think @balazsorban44 @MrPumpking 👀

pencilcheck:
Middlware still doesn't work

nextjs 13.1.1
next-auth 4.18.3

Issue #1913, opened 2021-05-03
Passwordless login without DB (using emailed code)


benderillo:
## Summary 💭

Add support for passwordless authentication flow that involves sending a random code to the user's email and asking them to type it into the login form on the frontend in order to get a JWT. This flow does not require DB.

## Description 📓
_DISCLAIMER: I am not a security expert and my proposal could be flawed with security concerns._

Presently next-auth can be used without a DB for OAuth flows that social login providers offer. But if one opts for passwordless(i.e. email provider), then the DB is required to store the requests.
I understand that this is necessary requirement because the "magic link" is issued and when user clicks it, we need to match it with the request on the backend.

There is another popular passwordless login flow that requires a user to type in a code that is sent to their email. 
Sites that use this UX: workflowy.com, stripe.com

It is my understanding that this flow can be done without a Database.
This is how I envision this designed:
Frontend code (further a Client) interacts with the server-side function (further Backend).

1. A Client sends an email address to the Backend
2. The Backend generates random code, hashes the object that contains user email and the code all salted with some secret key and returns the hash back to the client (as HttpOnly secure cookie with configured expiration time).

```js
const hash = hash({email: "johndoe@void.io", code: "3849-0928-1237"}, salt);
```

3. The Backend sends an email with the code value to the given email address
4. The Client receives the hash renders the view asking a user to enter the code
5. The User enters the code that arrives in their mailbox
6. The Client sends to the Backend the following data: {code, email}
7. The Backend verifies that cookie with the hash has not expired;
8. The Backend validates the {code, email} by comparing hash with the one in the cookie.
9. If hash matches it issues access_token (JWT)

- **purpose of the feature**
Passwordless auth is modern way of dealing with Auth.
If NextAuth support it without the DB, then the site can be deployed on Vercel or Netlify without the need to provision a database hosting.

One of the purposes of this request is to get it reviewed by someone who understands security and authentication well enough to tell if this will fly or crash and burn.

- **potential problems**
As I mentioned, I am not a security expert and there maybe severe security concerns that I am not aware of.

- **potential alternatives**
The existing email provider and the database.

balazsorban44:
#1051 touched on this same idea, and @iaincollins gave a good answer why we don't support it: https://github.com/nextauthjs/next-auth/issues/1051#issuecomment-792942113

That said, he also mentioned that we could provide it as an alternative. My problem with it would be that it will only work on the same device, while the current solution will work across devices.

[...]

PR #4824, opened 2022-07-03
feat: lazily import nodemailer

mmkal:
I'm roughly following/adapting this guide for sending a six-digit code via email: https://www.ramielcreations.com/nexth-auth-magic-code

I would like to use it in a stack where we're using Amazon SES as an emailing abstraction, rather than an SMTP server directly. And while developing, it has been useful to just `console.log` the verification code in a custom `sendVerificationRequest` implementation. One snag I hit, though, was that not having the `nodemailer` dependency installed caused an error just by requiring the `EmailProvider`, even though it's only used in one place, that I was overriding anyway. I've seen in https://github.com/nextauthjs/next-auth/issues/1913 and some other related issues/discussions that there may be willingness to create a non-email-biased `PasswordlessProvider` in future. But for now this step is a really simple code change to make and decreases the `nodemailer` dependency without needing to create a whole new provider.

### DIFF
--- a/packages/next-auth/src/providers/email.ts
+++ b/packages/next-auth/src/providers/email.ts
@@ -1,5 +1,3 @@
-import { createTransport } from "nodemailer"
-
 import type { CommonProviderOptions } from "."
 import type { Options as SMTPConnectionOptions } from "nodemailer/lib/smtp-connection"
 import type { Awaitable } from ".."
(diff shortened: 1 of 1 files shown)


balazsorban44:
Please see https://github.com/nextauthjs/next-auth/pull/3168#issuecomment-965821531

balazsorban44:
I recommend creating your own email provider for this case for now.

PR #4844, opened 2022-07-05
feat(adapters): update `typeorm` support

balazsorban44:
We haven't kept up with the recent TypeORM changes, and since they are still <1, it's likely that users kept upgrading, even if there were breaking changes.

BREAKING CHANGE:

[`typeorm`](https://github.com/typeorm/typeorm) is still in active development and has not yet published a stable release. Because of this, you can expect breaking changes in minor versions. This release of the adapter expects `typeorm@0.3.7` and is not validated against previous or future releases.

Run `npm i typeorm@latest` and make sure to read the [release notes](https://github.com/typeorm/typeorm/releases) for breaking changes in TypeORM


ndom91 (approved):
LGTM :+1: 

Worked locally with the Prisma sqlite db!

PR #4877, opened 2022-07-09
Added `getServerProviders` handler

NuroDev:
#### What's been changed?
 - Moved `PublicProvider` from providers route to core types module
 - Added new `getServerProviders` to next module

### Why have these changes been made?

This is a very niche use case but I have a custom login page set up with SSG (`getStaticProps`), instead of using SSR (`getServerSideProps`) as it's slower, to dynamically fetch the providers using `getProviders` & pass them to the client for rendering.

However, using the current recommended `getProviders` handler relies on an active endpoint fetch in order to retrieve the providers at build time which is possible for production builds, though outdated, but impossible for local builds.

As such I have added this new `getServerProviders` handler designed for dedicated backend / server side use, such as `getStaticProps`, to retrieve providers.

## 🧢 Checklist

- [ ] Documentation
    - [ ] Add `getServerProviders` section to `/configuration/nextjs` docs page
- [x] Tests
    - [x] Test locally in development environment
- [ ] Ready to be merged

### DIFF
--- a/packages/next-auth/src/core/routes/providers.ts
+++ b/packages/next-auth/src/core/routes/providers.ts
@@ -1,13 +1,5 @@
 import type { OutgoingResponse } from ".."
-import type { InternalProvider } from "../types"
-
-export interface PublicProvider {
-  id: string
-  name: string
-  type: string
-  signinUrl: string
-  callbackUrl: string
-}
+import type { InternalProvider, PublicProvider } from "../types"
 
 /**
  * Return a JSON object with a list of all OAuth providers currently configured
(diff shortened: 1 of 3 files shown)


0ubbe (changes requested):
Thanks for the PR 💚

To add a new top-level client helper like this, we need to see more interest from the community and users of the library. Otherwise, if we add a helper for every corner case a user finds, the library would end up bloated.

To get this in: 
1. First, open an issue explaining the problem
2. Explain the proposed helper there and get feedback on its API 

If we then see activity from more users requesting this, it's then the time to open a PR and integrate it into the library 👍🏽

balazsorban44:
thanks, but we are rethinking the core API. For now I recommend just copying the parseProviders code from the core.

PR #4901, opened 2022-07-11
fix: Stop JWT maxAge from being overridden by Session maxAge

### LINKED ISSUE #3563: JWT maxAge is not working
### Description 🐜

When using `maxAge` option for JWT (refer: https://next-auth.js.org/configuration/options#json-web-token-options), the `exp` value of the `next-auth.session-token` is not reflected properly. I am also having custom callbacks for `jwt` and `session`, but I dont think it should affect the JWT expiry.

My custom callbacks
```
callbacks:{
        async jwt({ token, user, account, isNewUser }){
            if (user) {
               // token={...user}
               token={status:user.status}
            }
            return token;
        },
        // That token store in session
        async session({ session, token }) { // this token return above jwt()
            session.tokenStatus = token.status;
            // //if you want to add user details info
            // session.user = { name: "name", email: "email" };//this user info get via API call or decode token. Anything you want you can add
            return session;
        },
    },
```

It still uses the `maxAge` for the session option (refer: https://next-auth.js.org/configuration/options#session) . Ideally it should use the `jwt.maxAge` instead of `session.maxAge`

No

### How to reproduce ☕️

Add this to your `NextAuth` options (`pages/api/auth/[...nextauth].js`)
```
jwt:{
        maxAge:120,
        secret:'secret2'
    },
    session:{
        strategy:'jwt',
        // maxAge:60,
    }
```

You can view the decrypted and decoded JWE token via creating a new route using the `getToken` (refer: https://next-auth.js.org/v3/tutorials/securing-pages-and-api-routes#using-gettoken)
Example file `my-nextjs-project/pages/api/getTokenDetails.ts`

```
import { getToken } from 'next-auth/jwt';

const secret = 'secret2';

    const token = await getToken({ req, secret });


    console.log(token)
    res.status(200).json({...token});
};
```

### Screenshots / Logs 📽

The API response which contains the `exp` of the JWT
```
{
  status: 'success',
  iat: 1641311170,
  exp: 1643903170,
  jti: '65698cae-75e8-4831-a121-b2f0d02d4b8f'
}

```
Also attached screenshot
Screenshot from 2022-01-04 21-16-18

### Environment 🖥

```
  System:
    OS: Linux 5.11 Ubuntu 20.04.3 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i7-7560U CPU @ 2.40GHz
    Memory: 2.64 GB / 15.36 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 14.17.4 - ~/.nvm/versions/node/v14.17.4/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v14.17.4/bin/yarn
    npm: 8.3.0 - ~/.nvm/versions/node/v14.17.4/bin/npm
  Browsers:
    Brave Browser: 96.1.33.106
    Chrome: 96.0.4664.110
    Firefox: 95.0.1

```

### Contributing 🙌🏽

Yes,  I am willing to help solve this bug in a PR

### LINKED ISSUE #4924: JWTEncodeParams maxAge type definition
### Question 💬

In the type definition for JWTEncodeParams, maxAge is defined as optional. 
image

```
export interface JWTEncodeParams {
    /** The JWT payload. */
    token?: JWT;
    /** The secret used to encode the NextAuth.js issued JWT. */
    secret: string | Buffer;
    /**
     * The maximum age of the NextAuth.js issued JWT in seconds.
     * @default 30 * 24 * 30 * 60 // 30 days
     */
    maxAge?: number; // line being referred to in this issue!
}
```

Since a default value is always provided by nextauth (30 days), the maxAge field is always defined, making the question mark incorrect. Ran into this while overriding the encode function. Will fix if it is decided that I am correct. Thanks!

### How to reproduce ☕️

Look at type definition for JWTEncodeParams

### Contributing 🙌🏽

roberte777:
## ☕️ Reasoning
Fixes #3563, fixes #4924

Essentially, the session maxAge was overriding the JWT maxAge. This meant that the JWT maxAge options did not do anything and was never applied.

I fixed this by removing the default maxAge for the JWT option. This part might be controversial. But, by letting it be undefined, I applied the logic to only apply the session maxAge to the JWT expiration if JWT maxAge is undefined. This prevents the JWT maxAge from being overridden and allows it to be applied. 

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: https://github.com/nextauthjs/next-auth/issues/3563

## 📌 Resources


roberte777:
This is my first open source PR, please let me know if there is anything I need to do in order to improve this :D

packages/next-auth/src/core/init.ts
```diff
@@ -91,7 +91,6 @@ export async function init({
     // JWT options
     jwt: {
       secret, // Use application secret if no keys specified
-      maxAge, // same as session maxAge,
```
balazsorban44, line 94: Let's do `userOptions.jwt?.maxAge ?? maxAge`

packages/next-auth/src/core/routes/session.ts
```diff
@@ -71,7 +71,7 @@ export default async function session(
       const newToken = await jwt.encode({
         ...jwt,
         token,
-        maxAge: options.session.maxAge,
+        maxAge: options.jwt.maxAge ?? options.session.maxAge,
```
balazsorban44, line 74:
```suggestion
        maxAge: options.jwt.maxAge,
```

Because of https://github.com/nextauthjs/next-auth/pull/4901/files#r918432349, this should now be safe.

balazsorban44 (changes requested):
Thanks, added small comments, let me know what you think!

packages/next-auth/src/core/init.ts
roberte777: Agreed, I’ll get to this today

roberte777:
> We have a custom backend where we want to authenticate against using the custom credentials provider and I was just trying to figure out how we can provide an expiry that we get back from the server and pass to JWT expiry as well. Otherwise how would it know if the session is still valid or not?
> 
> Is this fix required for the above use case to work?

disclaimer: I am not a next auth expert

This change is only needed if you want to set the JWT expiry time to anything other than the default 30 days through the next auth settings. 

Someone more experienced can correct me if I’m wrong, but it is my understanding that what you are trying to do is provide a dynamic expiry time that is returned from the server, not determined by your next auth settings. It is also my understanding that to do this, my change doesn’t really help you. You would need to override the default encode function and determine your own expiry time dynamically from your api call there. 

You could also set the maxAge option in your next auth settings, and just make sure it matches whatever your backend has set it to. This requires less setup, but if you change the expiry time on your backend you would have to  change it in your next auth settings to match if you so desire. More maintenance, less setup, and doing this WOULD require my change

viperfx:
Thanks for the clarification. @roberte777

roberte777:
@balazsorban44 could I get some feedback on my comments to your review?

PR #4908, opened 2022-07-12
chore(providers): rewrite github provider in typescript

mshd:
I rewrote the provider from js to ts and added the profile types.

I've tested it with two accounts, but please test it further as it's a critical provider.

## ☕️ Reasoning

Embrace typescript

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources


packages/next-auth/src/providers/github.ts
```diff
@@ -0,0 +1,108 @@
...
+    token: "https://github.com/login/oauth/access_token",
+    userinfo: {
+      url: "https://api.github.com/user",
+      async request({ client, tokens }) {
+        // Get base profile
+        const profile = await client.userinfo(tokens.access_token!)
+
+        // If user has email hidden, get their primary email from the GitHub API
+        if (!profile.email) {
+          const emails: GithubEmail[] = await (
```
mshd, line 81: I've added to the types to the email request too.

balazsorban44 (approved):
Awesome, thanks!

balazsorban44 (changes requested):
Linting failed, other than that, looks good!

packages/next-auth/src/providers/github.ts
```diff
@@ -0,0 +1,108 @@
...
+    authorization: {
+      url: "https://github.com/login/oauth/authorize",
+      params: { scope: "read:user+user:email" },
+    },
+    token: "https://github.com/login/oauth/access_token",
+    userinfo: {
+      url: "https://api.github.com/user",
+      async request({ client, tokens }) {
+        // Get base profile
+        const profile = await client.userinfo(tokens.access_token!)
```
balazsorban44, line 77: https://github.com/nextauthjs/next-auth/runs/7305230206?check_suite_focus=true#step:7:22

This throws as well. I believe it should always be `string` though, so maybe a `// @ts-expect-error` is forgivable here. If not, we can change it to throw an error in the future.

balazsorban44 (changes requested):

packages/next-auth/src/providers/github.ts
mshd: I've added the same behavior of https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/providers/facebook.ts

balazsorban44 (approved):
Thanks!

PR #4925, opened 2022-07-13
fix(ts):docs

toddjudd:
## ☕️ Reasoning

When following the [Advanced usage](https://next-auth.js.org/configuration/nextjs#advanced-usage) example for wrapping middleware in next.js the example as is throws type validation errors.

This may not be the correct fix, though for the docs to be functional in TS in their current state, the `nextauth` property on the type intersection with the `NextRequest` type for the middleware function parameter must be optional. 

Alternatively a fix to the Type definitions to include `[NextRequest & { nextauth?: { token: JWT | null } }]` as a valid type may be in order.

## 🧢 Checklist

- [ x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

## 📌 Resources

### DIFF
--- a/docs/docs/configuration/nextjs.md
+++ b/docs/docs/configuration/nextjs.md
@@ -183,7 +183,7 @@ import { withAuth } from "next-auth/middleware"
 
 export default withAuth(
   // `withAuth` can augment your Request with the user's token.
-  function middleware(req: NextRequest & { nextauth: { token: JWT | null } }) {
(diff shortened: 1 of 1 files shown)


balazsorban44:
After some tinkering, I came up with a way to solve this without requiring the user to type the inner middleware function at all. The request will always receive the `nextauth` object since we are wrapping it. I opened a PR which will close this. Thanks for letting me realize this!

toddjudd:
Excellent!!

PR #4951, opened 2022-07-16
chore: turbo cleanup

balazsorban44:
Turbo should take care of dependencies in dev now.


balazsorban44:
I'm not sure why this is necessary though:

https://github.com/nextauthjs/next-auth/blob/18728558cc95cee8559ae4abd47bf42b635f301b/turbo.json#L11-L26

I thought this should be enough:

```json
"dev": {
  "cache": false,
  "dependsOn": ["^build"]
},
```

But I have to explicitly add the build of the adapters, and define that each adapter `dependsOn` `next-auth#build` (because of the types)

balazsorban44:
Not sure why the test fails... It works locally, but not on GitHub, even after https://github.com/nextauthjs/next-auth/commit/4ff836a8cfbf39cdcecfffa8455996e3a90dca51. :thinking: 

image

turbo.json
```diff
@@ -5,44 +5,35 @@
     "build": {
       "dependsOn": ["^build"]
     },
-    "next-auth#build": {
-      "dependsOn": ["^build"],
-      "outputs": [
```
ThangHuuVu, line 10: This option was added to ensure that there will be a cache hit if the output files are deleted in https://github.com/nextauthjs/next-auth/pull/4025
If we remove this, then there will be a cache hit on running `build` after `clean` on `next-auth` which is not expected 🤔

package.json
```diff
@@ -6,16 +6,16 @@
   "scripts": {
     "build:app": "turbo run build --filter=next-auth-app --include-dependencies",
     "build": "turbo run build --filter=next-auth --filter=@next-auth/* --no-deps",
-    "lint": "turbo run lint --filter=!next-auth-docs",
-    "lint:fix": "turbo run lint -- --fix",
-    "test": "turbo run test --concurrency=1 --filter=!@next-auth/pouchdb-adapter --filter=!next-auth-* --filter=[HEAD^1]",
-    "setup": "turbo run setup",
+    "clean": "turbo run clean --no-cache",
+    "dev:app": "turbo run dev --parallel --filter=next-auth-app",
```
ThangHuuVu, line 10: > I'm not sure why this is necessary though:
> 
> https://github.com/nextauthjs/next-auth/blob/18728558cc95cee8559ae4abd47bf42b635f301b/turbo.json#L11-L26
> 
> I thought this should be enough:
> 
> ```json
> "dev": {
>   "cache": false,
>   "dependsOn": ["^build"]
> },
> ```
> 
> But I have to explicitly add the build of the adapters, and define that each adapter `dependsOn` `next-auth#build` (because of the types)

Try this: 
```suggestion
    "dev:app": "turbo run dev --filter=next-auth-app...",
```

Two things:
- Remove `--parallel` because `turborepo` ignores the dependency graph with this option. Not sure why it was added though, maybe because we tried to run two packages at the same time then 😁 
- Filter dependent packages with `...` as described [here](https://turborepo.org/docs/core-concepts/filtering#include-dependencies-of-matched-packages)

ThangHuuVu (commented):
I've got some suggestions 🚀

balazsorban44:
@ThangHuuVu what's remaining here is to clean up `apps/dev`:

https://github.com/nextauthjs/next-auth/blob/ba20974b5f8fe2ee8986c3a8b4aed35e0f9b571d/apps/dev/package.json#L7-L11 I think this should be simplified. `pnpm dev` should watch `next-auth` (both js and CSS) and Turbo should make sure that the dev app gets the updates from there with HMR, no copying or https://github.com/nextauthjs/next-auth/blob/ba20974b5f8fe2ee8986c3a8b4aed35e0f9b571d/apps/dev/next.config.js#L8 should be necessary.

ThangHuuVu:
> @ThangHuuVu what's remaining here is to clean up `apps/dev`:
> 
> https://github.com/nextauthjs/next-auth/blob/ba20974b5f8fe2ee8986c3a8b4aed35e0f9b571d/apps/dev/package.json#L7-L11
> I think this should be simplified. `pnpm dev` should watch `next-auth` (both js and CSS) and Turbo should make sure that the dev app gets the updates from there with HMR, no copying or
> 
> https://github.com/nextauthjs/next-auth/blob/ba20974b5f8fe2ee8986c3a8b4aed35e0f9b571d/apps/dev/next.config.js#L8
> should be necessary.

@balazsorban44 I added next-auth as apps/dev dependency now.
It introduces some new changes in the setup:
- `build` is no longer required for `dev`, instead `dev` now requires  topological `dev`
- Remove `copy:css`
- Bring back `--parallel` and add `--continue` for `dev`

apps/dev/package.json
```diff
@@ -4,11 +4,8 @@
   "description": "NextAuth.js Developer app",
   "private": true,
   "scripts": {
-    "clean": "rm -rf .next",
-    "copy:css": "cpx \"../../packages/next-auth/css/**/*\" src/css --watch",
-    "watch:css": "cd ../../packages/next-auth && pnpm watch:css",
-    "dev": "concurrently \"pnpm dev:next\" \"pnpm watch:css\" \"pnpm copy:css\"",
-    "dev:next": "next dev",
+    "clean": "rm -rf .next src/css",
```
ThangHuuVu, line 7:
```suggestion
    "clean": "rm -rf .next",
```

PR #4953, opened 2022-07-16
fix(ts): match TS types better with implementation

balazsorban44:
Historically, we have been lying in some places about the public interface to simplify the user-facing types, but this gave false security. This PR attempts to remove many `// @ts-expect-error` comments which have uncovered these mistakes.

Notable changes:

- `Account` is re-exported as `AdapterAccount` from `next-auth/adapters` for convenience
- There have been some TS issues on adapters because of how `Omit<User, "id">` obscured the underlying `User` interface when it extended `Record<string, unknown>`. Locally it seems like this was a bug in TS, and is no longer needed, but this will need another pair of eyes. to verify :eyes:. (Direct solution to part of the below-mentioned tweet :crossed_fingers:)
- When using the Email provider, we should be failing fast (i.e. in dev) when required methods aren't present.
- `AdapterSession` has required `id`, although in practice we never used it, so this PR removes it from the default interface (module augmentation will still let the user define it if needed, but we use our own `sessionToken` anyway to not leak the DB through the session cookie)
- TIL: Prisma [`@@id`](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#id-1), so we don't need to have explicit `id`s. (MongoDB still requires it, so we need the ugly `delete foo.id` things :disappointed:  )
- Anything else, WIP

Some of these changes are inspired by this report from Twitter: https://twitter.com/_clem/status/1505305187614920709

~(Might split this PR into multiple ones.)~ Reverted any unrelated change so it should be slightly easier to review. :pray:

TODO: Add tests!


packages/adapter-upstash-redis/tests/index.test.ts
```diff
@@ -11,6 +11,14 @@ if (!process.env.UPSTASH_REDIS_URL || !process.env.UPSTASH_REDIS_KEY) {
   process.exit(0)
 }
 
+if (process.env.CI) {
+  // TODO: Fix this
+  test('Skipping UpstashRedisAdapter tests in CI because of "Request failed" errors. Should revisit', () => {
```
balazsorban44, line 16: :shrug:

packages/next-auth/src/core/routes/signin.ts
```diff
@@ -59,7 +59,6 @@ export default async function signin(params: {
 
     // Check if user is allowed to sign in
     try {
-      // @ts-expect-error
```
lluia, line 62: Could we add a `TODO:` comment explaining why this was skipped?

packages/next-auth/src/core/lib/oauth/callback.ts
```diff
@@ -161,25 +159,19 @@ export interface GetProfileParams {
...
   provider,
   logger,
-}: GetProfileParams): Promise<GetProfileResult> {
+}: GetProfileParams) {
   try {
     logger.debug("PROFILE_DATA", { OAuthProfile })
-    // @ts-expect-error
     const profile = await provider.profile(OAuthProfile, tokens)
     profile.email = profile.email?.toLowerCase()
+    if (!profile.id) throw new TypeError("Profile id is missing")
```
lluia, line 173: We could log which Provider it's here:
```
Profile ID is missing in Twitter OAuth profile response
```
Thinking in case the user has configured multiple OAuth providers, it wouldn't be obvious 🤔

packages/next-auth/src/core/lib/oauth/callback.ts
```diff
@@ -57,7 +54,8 @@ export default async function oAuthCallback(params: {
         profile = JSON.parse(profile)
       }
 
-      return await getProfile({ profile, tokens, provider, logger })
+      const newProfile = await getProfile({ profile, tokens, provider, logger })
+      return { ...newProfile, cookies: [] }
```
lluia, line 58: I believe ignoring the `cookies` here is for security?

0ubbe (commented):
Had a scan, looking good overall 💯 ✨

ThangHuuVu:
I committed a few more improvements:
- refactored `Adapter` interface -> type, got rid of the "verified in assertConfig" comments & `ts-expected-error`
- improved `parseProvider` to get rid more `ts-expected-error`. Added `InternalOAuthConfig`

packages/next-auth/src/core/routes/signin.ts
ThangHuuVu: @lluia I don't get this comment, could you help clarify it a bit? 🙌

packages/next-auth/src/core/lib/oauth/callback.ts
ThangHuuVu: @lluia This is just for conforming with the check later I think, the logic remains the same:
https://github.com/nextauthjs/next-auth/blob/ffcf2b06d54f096d5d9ad8cd92c63b00210c4b77/packages/next-auth/src/core/routes/callback.ts#L55

ThangHuuVu (approved):

baptisteArno:
So, if I understand this correctly, email is not nullable anymore?

Prisma schema example on the doc still shows `email String? @unique`.

Still, the Facebook provider doesn't save an `email` to the database, so I feel like this is an incorrect typing?

ifightcrime:
Running into the same issue as @baptisteArno above. Shouldn't we be allowed to have a null email if we're using a non-email provider? e.g. I want to build a custom provider & adapter for logging in via SMS.

PR #4975, opened 2022-07-20
Remove confusing and unnecessary throw from useSession React hook.

davemackintosh:
## ☕️ Reasoning

An error is thrown that is confusing to users if a session has no value in the context (the default value is undefined) then it throws an error saying that useSession needs to be wrapped in the SessionProvider (even if it already is) I don't believe this is correct behaviour

## 🧢 Checklist

- [X] Documentation - Not needed
- [X] Tests - None needed
- [X] Ready to be merged

## 🎫 Affected issues

Fixes: No known issues but I believe it is an unnecessary error to be thrown

## 📌 Resources


davemackintosh:
Ignore me this opens a slue of other issues it turns out 👎

PR #4983, opened 2022-07-20
fix(jwt): make `getToken` work in node 14/16

mmkal:
## ☕️ Reasoning

My team use node v14.17.0 and this line was throwing because `Headers` is undefined. Looks like it was [added in v17.5.0](https://nodejs.org/api/globals.html#class-headers).

Verified that this fix worked, but since package.json declares `"engines": {    "node": "^12.19.0 || ^14.15.0 || ^16.13.0"   }` might be worth using an eslint rule to make it easier to spot this kind of thing, since there will probably be more new nodejs features that don't exist in older node versions.

## 🧢 Checklist

Haven't done docs or testing, because this is just a thing that's broken that would become not-broken. A test could be sensible but since we've already worked around this will leave that to someone else!

### DIFF
--- a/packages/next-auth/src/jwt/index.ts
+++ b/packages/next-auth/src/jwt/index.ts
@@ -92,7 +92,7 @@ export async function getToken<R extends boolean = false>(
   let token = sessionStore.value
 
   const authorizationHeader =
-    req.headers instanceof Headers
+    typeof Headers !== 'undefined' && req.headers instanceof Headers
       ? req.headers.get("authorization")
       : req.headers.authorization
 


balazsorban44:
This isn't necessary, as Next.js polyfills `Headers` if it isn't available. 🤔

https://github.com/vercel/next.js/blob/canary/packages/next/server/node-polyfill-fetch.js#L20

Could you open a bug report with a reproduction first so we can investigate what's going on in your case?

mmkal:
I'm trying to use `getToken` from a normal express server that isn't next.js. Everything works except this!

PR #4985, opened 2022-07-21
Update custom reddit provider to v4

dgrcode:
## ☕️ Reasoning
The docs around Reddit provider were obsolete because they were using a previous version of NextAuth. That resulted in a weird error that led me to the discussion https://github.com/nextauthjs/next-auth/discussions/3204. It seems there was a bug open in a previous repo to update the docs, but the bug seems to have been lost on a repo migration.

In the scopes I've added the same scopes that were previously in the docs, but I've personally only tested with the "identity" scope. With this I mean that the code I've written might not work out of the box. However, given the previous code didn't work either, at worst we're still improving the docs.

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

I haven't found any open issue searching for "reddit"

## 📌 Resources

### DIFF
--- a/docs/docs/providers/reddit.md
+++ b/docs/docs/providers/reddit.md
@@ -49,13 +49,9 @@ providers: [
   RedditProvider({
     clientId: process.env.REDDIT_CLIENT_ID,
     clientSecret: process.env.REDDIT_CLIENT_SECRET,
-    version: '2.0',
     authorization: {
-      url: 'https://www.reddit.com/api/v1/authorize',
       params: {
-        scope: 'identity',
         duration: 'permanent',
-        response_type: 'code',
       },
     },
   }),
--- a/docs/docs/providers/reddit.md
+++ b/docs/docs/providers/reddit.md
@@ -46,30 +46,19 @@ This Provider template only has a one hour access token to it and only has the "
 
 ```js
 providers: [
-  {
-    id: 'reddit',
-      name: 'Reddit',
-      clientId: process.env.REDDIT_CLIENT_ID,
-      clientSecret: process.env.REDDIT_CLIENT_SECRET,
-      authorization: {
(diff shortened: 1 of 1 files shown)


docs/docs/providers/reddit.md
```diff
@@ -47,25 +47,28 @@ This Provider template only has a one hour access token to it and only has the "
...
-      "https://www.reddit.com/api/v1/authorize?response_type=code&duration=permanent",
-    profileUrl: "https://oauth.reddit.com/api/v1/me",
-    profile: (profile) => {
-      return {
-        id: profile.id,
-        name: profile.name,
-        email: null,
-      }
-    },
+    id: 'reddit',
```
balazsorban44, line 50: Let's use `RedditProvider`, so you only need to override what's different from the default config.

dgrcode: sounds good 👍

dgrcode:
@balazsorban44 I just updated my changes to reflect your comment. What do you think?

docs/docs/providers/reddit.md
```diff
@@ -46,27 +46,19 @@ This Provider template only has a one hour access token to it and only has the "
...
-        email: null,
-      }
+    version: '2.0',
+    authorization: {
+      url: 'https://www.reddit.com/api/v1/authorize',
+      params: {
+        scope: 'identity',
+        duration: 'permanent',
+        response_type: 'code',
+      },
```
balazsorban44, line 59:
```suggestion
    authorization: {
      params: {
        duration: 'permanent',
      },
```
Could be even simpler :grin:

balazsorban44 (commented):
Since we deep-merge user configs, I could simplify this even more, but you were onto what I intended, thanks!

PR #5027, opened 2022-07-26
s/OAUTH_CALLBACK_ERROR/CALLBACK_OAUTH_ERROR/

sato11:
## ☕️ Reasoning
There is an inconsistent use of two of the error codes, `OAUTH_CALLBACK_ERROR` and `CALLBACK_OAUTH_ERROR`. Especially in `routes/callback.ts` the two sit next to each other and emit subtly discrepant errors. They apparently do not contain asymmetrical information, and practically look interchangeable. But since errors page only mentions the latter, let's align them and make use of it: https://next-auth.js.org/errors#callback_oauth_error. With this fix I hope anybody like me will no longer navigate unfortunately through the different link with a broken anchor.

Skimming through the history, I found `OAUTH_CALLBACK_ERROR` had been replaced in #235, reintroduced perhaps accidentally in #332 and #1698. Let me know if I'm missing any context!

Below is the like of what I've seen in the log:

```
[next-auth][error][OAUTH_CALLBACK_ERROR] 
https://next-auth.js.org/errors#oauth_callback_error connect ECONNREFUSED 127.0.0.1:80 {
  error: {
    message: 'connect ECONNREFUSED 127.0.0.1:80',
    stack: 'Error: connect ECONNREFUSED 127.0.0.1:80\n' +
      '    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1237:16)',
    name: 'Error'
  },
  providerId: 'some-provider',
  message: 'connect ECONNREFUSED 127.0.0.1:80'
}
[next-auth][error][CALLBACK_OAUTH_ERROR] 
https://next-auth.js.org/errors#callback_oauth_error connect ECONNREFUSED 127.0.0.1:80 Error: connect ECONNREFUSED 127.0.0.1:80
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1237:16) {
  name: 'OAuthCallbackError',
  code: 'ECONNREFUSED'
}
```

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Please [scout and link issues](https://github.com/nextauthjs/next-auth/issues) that might be solved by this PR.

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources

- [Contributing guidelines](./CONTRIBUTING.md)
- [Code of conduct](./CODE_OF_CONDUCT.md)
- [Contributing to Open Source](https://kcd.im/pull-request)

PR #5046, opened 2022-07-29
fix: Use consistent error type between doc, logger and error class

### LINKED ISSUE #5045: Use consistent error type between doc, logger and error class

There is an inconsistent use of two of the error codes, `OAUTH_CALLBACK_ERROR` and `CALLBACK_OAUTH_ERROR`, of which only the latter is documented as errors: https://next-auth.js.org/errors.

In https://github.com/nextauthjs/next-auth/pull/5027#issuecomment-1197053995, @ThangHuuVu has suggested that we should:

1. merge docs of the two types mentioned above,
2. rename them more generic something like `CALLBACK_ERROR`, and
3. keep from the same error logged twice.

As a result we can merit from getting rid of subtly discrepant errors, both from docs page and from the source.

In the PR mentioned above, I had attempted to use only `CALLBACK_OAUTH_ERROR`, but after that I found there is the `OAuthCallbackError` class. Consequently I'm thinking to use `OAUTH_CALLBACK_ERROR` so that documentation better represents the implementation.

After receiving a comment in #5027 I've researched a bit and found some things that might affect the way we sort our docs.

So currently we have in the errors page a headline that says:

> oauth_callback_error expected 200 OK with body but no body was returned

while in fact it seems this `expected 200 OK with body but no body was returned` message never happens in callback phase, but in discovery phase. That is, openid-client emits this message when (and apparently only when) `/.well-known/openid-configuration` returns no body. cf. https://github.com/panva/node-openid-client/blob/v5.1.8/lib/helpers/process_response.js#L59

In next-auth's side, things described below in order happens

1. triggers issuer discovery
2. fails to get authorization url
4. logs `SIGNIN_OAUTH_ERROR` with`expected 200 OK with body but no body was returned`

So I believe the original content which attributes this error message to `OAUTH_CALLBACK_ERROR` was not true. I guess associating it with `SIGNIN_OAUTH_ERROR` can be of help and so I'm planning to create a PR.

https://next-auth.js.org/errors

sato11:
## ☕️ Reasoning
Following up this comment: https://github.com/nextauthjs/next-auth/pull/5027#issuecomment-1197053995.

I've abstained from further renaming it like `CALLBACK_ERROR`, since my original intention is only to make them consistent. I'm feeling happy enough with `OAUTH_CALLBACK_ERROR`, partly because I am not really sure over the intention of renaming. So I'd like to leave that to you.

About the doc, two changes have been made: merging `CALLBACK_OAUTH_ERROR` and `oauth_callback_error`, and relocating `openid-client` stuff to `SIGNIN_OAUTH_ERROR` section. Background of the latter change is reasoned in #5045. 

Commits are separately made, but if you want me to squash them let me know.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Closes: #5045

## 📌 Resources


packages/next-auth/src/core/routes/callback.ts
```diff
@@ -190,7 +190,7 @@ export default async function callback(params: {
       }
     } catch (error) {
       if ((error as Error).name === "OAuthCallbackError") {
-        logger.error("CALLBACK_OAUTH_ERROR", error as Error)
+        logger.error("OAUTH_CALLBACK_ERROR", error as Error)
```
ThangHuuVu, line 193:
```suggestion
        logger.error("OAUTH_CALLBACK_ERROR", {
          error: error as Error,
          providerId: provider.id,
        })
```
Let's provide the `providerId` here 🙌

sato11: Sure. Committing your suggestion 🙂

sato11:
@ThangHuuVu Thanks for the feedback. How does it look to you now?

ThangHuuVu (approved):
LGTM 🚀

ThangHuuVu:
@ndom91 @lluia I was wondering if changing the error code requires a bump in semver 🤔 WDYT?

Issue #4716, opened 2022-06-15
Email provider bug with case sensitivity & custom magic code implementation


ifightcrime:
### Provider type

Email

### Environment

N/A

### Reproduction URL

N/A

### Describe the issue

The bug looks like it actually exists in core, but impacts `EmailProvider`. When building out a custom flow to send a verification code instead of a link, I noticed that when `useVerificationToken()` is called it's not down-casing the email address. Therefore logging in doesn't work if you type any uppercase characters (e.g. on iPhone where it auto-capitalizes the first letter)

### How to reproduce

This doesn't seem like an issue with vanilla email provider as it looks like the link is generated with a downcased version here:

https://github.com/nextauthjs/next-auth/blob/2469e44572f23f709fa8c5c65c6b7a4eb2383e9f/packages/next-auth/src/core/routes/signin.ts#L38

So likely not super high priority. However, this caused a bug in my custom implementation as I was taking the email address as written from the form. My workaround is to down-case it myself before I send to next-auth, but it seems like the down-casing should be more centralized. Maybe here:

https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/core/routes/callback.ts#L205-L211

### Expected behavior

Email is always down-cased and user can login no matter how they case their email when inputting. Not sure if there are any security implications of this, but might be a nice lead-up for issues like #1465

balazsorban44:
Thanks, do you want to open a PR? It looks like we just need to call `.toLowerCase()` before passing the `identifier` to `useVerificationToken`.

PR #5067, opened 2022-08-02
Transform email to lowercase letters for setVerfication hook input(issue #4716)

hamidrezaghanbari:
for #4716 issue
@ThangHuuVu  (good-first-issue)

### DIFF
--- a/packages/next-auth/src/core/routes/callback.ts
+++ b/packages/next-auth/src/core/routes/callback.ts
@@ -203,7 +203,7 @@ export default async function callback(params: {
       const { useVerificationToken, getUserByEmail } = adapter!
 
       const token = query?.token
-      const identifier = query?.email
+      const identifier = query?.email?.toLowerCase()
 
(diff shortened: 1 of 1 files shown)


ThangHuuVu (approved):
Thanks, LGTM 🙌

hamidrezaghanbari:
> Thanks, LGTM 🙌

balazsorban44 (changes requested):
A more appropriate solution would probably be to run the identifier through the same `normalizeIdentifier` method that was just introduced here:

https://github.com/nextauthjs/next-auth/blob/a3c6786f7829590689a714d5fcfd3d90c8569744/packages/next-auth/src/core/routes/signin.ts#L38-L55

This will lowercase by default, but will also take the user's config into consideration if they define a `normalizeIdentifier`.

I would probably move the default implementation into the `EmailProvider` so we can just assume `provider.normalizeIdentifier` is always defined.

PR #5080, opened 2022-08-03
Add error for oauth denial

cderszteler:
## ☕️ Reasoning

Adds another error type to oauth2 abortion. When an oauth2 request is actively cancelled by the user, this error type will be used.

With that, it is possible to identify an cancellation as intended and react properly. This is useful if you have for example only one provider and redirect on `/api/auth/signin` directly to this provider. That would currently end in an infintie loop, which
can be avoid by checking if the user canceled the request intentionally.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: [https://github.com/nextauthjs/next-auth/discussions/3682](https://github.com/nextauthjs/next-auth/discussions/3682)

## 📌 Resources

### DIFF
--- a/docs/docs/configuration/pages.md
+++ b/docs/docs/configuration/pages.md
@@ -54,6 +54,8 @@ The following errors are passed as error query parameters to the default or over
 - **EmailSignin**: Sending the e-mail with the verification token failed
 - **CredentialsSignin**: The `authorize` callback returned `null` in the [Credentials provider](/providers/credentials). We don't recommend providing information about which part of the credentials were wrong, as it might be abused by malicious hackers.
 - **SessionRequired**: The content of this page requires you to be signed in at all times. See [useSession](/getting-started/client#require-session) for configuration.
+- **Configuration**: The provider was configured wrongly, check the debug log to see detailed information from the provider.
+- **UnavailableService**: The provider is currently not reachable or has internal server problems.
 - **Default**: Catch all, will apply, if none of the above matched
 
 Example: `/auth/signin?error=Default`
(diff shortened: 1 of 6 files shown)


balazsorban44 (changes requested):
Thanks, for looking into this! There have been previous attempts (#1820), but they stalled.

I think this should be handled in: https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/core/lib/oauth/callback.ts

And we should probably handle other error responses as well:

https://www.oauth.com/oauth2-servers/authorization/the-authorization-response/#:~:text=the%20redirect%20URL%3A-,error,-a%20single%20ASCII

We should consider `error_description` and maybe log it alongside `error_uri` if exists, for debugging purposes, but I am not sure if these usually contain any sensitive information. We have to be careful about what we log as errors in production (https://github.com/nextauthjs/next-auth/security/advisories/GHSA-p6mm-27gq-9v3p).

The above page says:
> There are two different kinds of errors to handle. The first kind of error is when the developer did something wrong when creating the authorization request. The other kind of error is when the user rejects the request (clicks the “Deny” button).

So for the first type, we could redirect to `error?error=Configuration` and expand the documentation, and for when the user cancels the login flow, it's probably OK to show them a specific message as you did in this PR.

Qetzing:
Hey,

I wasn't quite sure, if this update is what you meant. I added the other oauth2 error responses and tried to sort them by "fault" (developer, user, etc.).

Depending on the fault, the user sees a different message and the error is logged. 

Furthermore, I tried to add a bit more documentation, however I don't know if that's what you meant.

balazsorban44:
Good/bad news, this will be handled in the new `@auth/core` package: https://github.com/nextauthjs/next-auth/blob/2cbf81544523044596125d2c87175ee4ec61df06/packages/core/src/lib/oauth/callback.ts#L80C7-L87, but not planning to backport it for `next-auth` at the moment. Thanks for the PR anyway!

PR #5000, opened 2022-07-22
fix: avoid redirect on always public paths

balazsorban44:
Fixes #4848

## ☕️ Reasoning

What changes are being made? What feature/bug is being fixed here?

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Please [scout and link issues](https://github.com/nextauthjs/next-auth/issues) that might be solved by this PR.

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources

- [Contributing guidelines](./CONTRIBUTING.md)
- [Code of conduct](./CODE_OF_CONDUCT.md)
- [Contributing to Open Source](https://kcd.im/pull-request)

PR #5085, opened 2022-08-04
fix(middleware): use `includes()` for NextAuth pages

Juneezee:
## ☕️ Reasoning

Some users could be setting their `signIn` and `error` pages option to
`/` to disable the automatically generated pages, as suggested in https://github.com/nextauthjs/next-auth/discussions/2330#discussioncomment-1678298.

This PR reverts the behaviour for matching `signIn` and `error`
pages in `handleMiddleware` to pre-v4.10.3.

EDIT:
```
const signInPage = "/"
const errorPage = "/"
const publicPaths = [signInPage, errorPage, "/_next", "/favicon.ico"]

// pathname = "/protected" will always return true
publicPaths.some((p) => pathname.startsWith(p))
```

Fixes: aedabc8d ("fix: avoid redirect on always public paths")

## 🧢 Checklist

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/next/middleware.ts
+++ b/packages/next-auth/src/next/middleware.ts
@@ -106,12 +106,13 @@ async function handleMiddleware(
   const signInPage = options?.pages?.signIn ?? "/api/auth/signin"
   const errorPage = options?.pages?.error ?? "/api/auth/error"
   const basePath = parseUrl(process.env.NEXTAUTH_URL).path
-  const publicPaths = [signInPage, errorPage, "/_next", "/favicon.ico"]
+  const publicPaths = ["/_next", "/favicon.ico"]
 
   // Avoid infinite redirects/invalid response
   // on paths that never require authentication
   if (
     pathname.startsWith(basePath) ||
+    [signInPage, errorPage].includes(pathname) ||
     publicPaths.some((p) => pathname.startsWith(p))
   ) {
     return


ThangHuuVu:
Hi 👋 So in #5000 we made that change to avoid the infinite re-direct loop when you have a custom error page that is also require authentication. I wasn't aware that setting `/` to all custom pages is a way to disable all built in page. 🤔 Seems like there is a conflict here. cc @balazsorban44 

We would be appreciated if you could open a new issue so that we can investigate and think of a solution together there instead 🙌 I'm going to close this one in the mean time. 
Your PR doesn't seem to change the output at all, BTW 🤔:
```js
const signInPage = "/"
const errorPage = "/"
const publicPaths = [signInPage, errorPage, "/_next", "/favicon.ico"]

const pathname = "/"
// current
publicPaths.some((p) => pathname.startsWith(p)) // return true

// your change
[signInPage, errorPage].includes(pathname) // also return true
```

balazsorban44:
I think this should be OK, but in that case, you have to ensure middleware is not invoked on `/`. :thinking:

Juneezee:
> Your PR doesn't seem to change the output

@ThangHuuVu Sorry, I made a typo in the code snippet. It should be

```javascript
const signInPage = "/"
const errorPage = "/"
const publicPaths = [signInPage, errorPage, "/_next", "/favicon.ico"]

// pathname = "/protected/pathA" will always return true
publicPaths.some((p) => pathname.startsWith(p))
```

So if the middleware has the `pages` option configured as
```javascript
pages: {
  signIn: '/',
  error: '/',
}
```
and
```javascript
export const config = { matcher: ["/protected/:path*"] }
```

Visiting `/protected/pathA` will not redirect to the `signIn` page.

Juneezee:
> I think this should be OK, but in that case, you have to ensure middleware is not invoked on `/`. 🤔 I don't think this needs a PR.

@balazsorban44 See https://github.com/nextauthjs/next-auth/pull/5085#issuecomment-1205432217. Middleware is not invoked on `/`. It works fine on `v4.10.2` with `signIn` and `error` set as `/`.

ThangHuuVu:
Thanks for the explanation, this makes more sense to me now 👍 
Would you be able to add a test so that we don't regress again in future changes? 🙏

Juneezee:
@ThangHuuVu Not sure what went wrong with GitHub, it's not showing the new commit that I've pushed. Should I open another PR that adds the test case after this PR is merged?

image

balazsorban44:
Might be because we closed it. It would be nice to have the test and the code in the same PR so we can verify it first. Maybe we should close this PR and just create a new one. Sorry 💚

PR #5109, opened 2022-08-06
fix(middleware): improve handling of custom nextjs basePath

psperber:
## ☕️ Reasoning

The middleware did not respect the nextjs base path correct. It was not prepended to the path at the redirect.
Within this PR I introduce a function to get the nextjs base path from the `req.nextUrl` object, which is then used in the middleware to perform a correct validation and redirect.

## 🧢 Checklist

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

## 📌 Resources


ThangHuuVu:
Hi @psperber, could you provide a simple reproduction of the issue first? We need to verify if its really an issue first before accepting this PR 🙌

psperber:
Hi, of course. Hope that's sufficient: https://github.com/psperber/next-auth-issues

psperber:
Hey, anything I can do to help you going forward with my PR?

packages/next-auth/src/next/middleware.ts
```diff
@@ -103,16 +104,18 @@ async function handleMiddleware(
 ) {
   const { pathname, search, origin } = req.nextUrl
 
-  const signInPage = options?.pages?.signIn ?? "/api/auth/signin"
-  const errorPage = options?.pages?.error ?? "/api/auth/error"
+  const nextJsBasePath = getNextJsBasePath(req.nextUrl)
+  const fullPathname = nextJsBasePath + pathname
+  const signInPage = options?.pages?.signIn ?? `${nextJsBasePath}/api/auth/signin`
+  const errorPage = options?.pages?.error ?? `${nextJsBasePath}/api/auth/error`
   const basePath = parseUrl(process.env.NEXTAUTH_URL).path
```
ThangHuuVu, line 111: If your `NEXTAUTH_URL` contains a custom base path, for example, "https://example.com/custom-base", this line will extract the correct `basePath` as `/custom-base`, otherwise it will return `/api/auth` as default. Can't we use this to get the custom base path if provided instead of adding a new function? 🤔
I noticed that in your reproduction,  the `NEXTAUTH_URL` contains `/api/auth` which is incorrect.

ThangHuuVu (commented):
Thanks for raising the PR 🙌 I added a comment, let me know what you think. 
While you're at it, please help resolve the conflicts, too

packages/next-auth/src/next/middleware.ts
psperber: Hey, I configured the reproduction like it states in the documentation about the custom base path: https://next-auth.js.org/configuration/options#nextauth_url

psperber:
> Thanks for raising the PR 🙌 I added a comment, let me know what you think. While you're at it, please help resolve the conflicts, too

Conflict is resolved, verified if it is still working with the current version. Also checked with `NEXTAUTH_URL=http://localhost:3000/custom-base` with the latest next-auth version released as you suggested, but it is not working.

packages/next-auth/src/next/middleware.ts
ThangHuuVu: Oh sorry, I totally missed that!

psperber: I updated the packages to the latest version and add a version with `NEXTAUTH_URL=http://localhost:3000/custom-base`.
https://github.com/psperber/next-auth-issues

packages/next-auth/src/utils/nextjs-base-path.ts
```diff
@@ -0,0 +1,15 @@
+import type { NextURL } from "next/dist/server/web/next-url";
+
+export default function (nextUrl: NextURL): string {
```
lavcraft, line 3: Why do you use this function? NextUrl now contain  basePath prop. It's more transparent and supported way to extract basePath in next.js based projects. Are you agree? Or i miss something?

lavcraft: Also, i think pathname is not an optimal option, let me explain via example:

1. Go to protected page like /orders/[orderId]/view and concrete /orders/1234/view
2. Redirect to signIn page
3. see callbackUrl like callbackUrl=/orders/[orderId]/view!!! Instead of /orders/1234/view

psperber: > Why do you use this function? NextUrl now contain basePath prop. It's more transparent and supported way to extract basePath in next.js based projects. Are you agree? Or i miss something and didn't cover all cases?

Pretty good catch. I adapted to req.nextUrl.basePath

lavcraft: @psperber can't reproduce step3 or don't understand what i mean?

I mean: after redirect from protected url to signIn page, value contained previous url of protected page will write into query param as invalid value, without replaced value. But after few test a can't reproduce this in my implementation #5491 . Sorry it's my fault

packages/next-auth/src/next/middleware.ts
```diff
@@ -101,19 +101,20 @@ async function handleMiddleware(
   options: NextAuthMiddlewareOptions | undefined,
   onSuccess?: (token: JWT | null) => Promise<NextMiddlewareResult>
 ) {
-  const { pathname, search, origin } = req.nextUrl
+  const { pathname, search, origin, basePath: nextJsBasePath } = req.nextUrl
 
-  const signInPage = options?.pages?.signIn ?? "/api/auth/signin"
-  const errorPage = options?.pages?.error ?? "/api/auth/error"
+  const fullPathname = nextJsBasePath + pathname
+  const signInPage = options?.pages?.signIn ?? `${nextJsBasePath}/api/auth/signin`
```
lavcraft, line 107: Adding basePath into signIn and error page from start this function is bad idea in my opinion, because code for check urls for skip is getting complicated after this.

I advice to remove it from this code and add when construct resulted url via new URL when error happened or call redirect
@ThangHuuVu what do you think about this?

ThangHuuVu (commented):
Noticed that you've removed the tests, can you add back at least one? 🙌

packages/next-auth/src/next/middleware.ts
psperber: I agree removing it here, but I have to add it at least in line 129 and 149 to have a proper redirection. We have to prepend the nextjs to the sign in and error page on redirect.

psperber: I adapted accordingly

psperber:
> Noticed that you've removed the tests, can you add back at least one? 🙌

I added the test from @lavcraft branch, thanks for that :)

packages/next-auth/src/utils/nextjs-base-path.ts
psperber: > How about using `parseUrl(process.env.NEXTAUTH_URL).path` ? It should work fine, I remember tested it when I first review this PR

Use it for what? The problem with `parseUrl(process.env.NEXTAUTH_URL).path` is, that it contains the nextjs basepath. The nextUrl.pathname does not. The logic about publicPaths in line 115 and 116 are based on the nextUrl.pathname, which does not contain the nextjs basepath.
I think the current state it be least complex variant to do the job

lavcraft:
@ThangHuuVu could i help with sth for merging this? Looks like it's good now, but i don't understand what we are waiting for :)

packages/next-auth/src/next/middleware.ts
```diff
@@ -101,7 +101,7 @@ async function handleMiddleware(
   options: NextAuthMiddlewareOptions | undefined,
   onSuccess?: (token: JWT | null) => Promise<NextMiddlewareResult>
 ) {
-  const { pathname, search, origin } = req.nextUrl
+  const { pathname, search, origin, basePath: nextJsBasePath } = req.nextUrl
```
ThangHuuVu, line 104: I suggest we use the variable name`basePath` here. I don't want to use `nextJsBasePath` since we're in `next` specific logic already.
To avoid conflict, we could rename the existing `basePath` at line 108 to `authPath` 🤔 this value is the pathname of the `NEXTAUTH_URL`.

ThangHuuVu (commented):
Looks pretty good to me now, thanks @lavcraft  @psperber! @psperber could you format the files again? 🙌 Added a comment, let me know what you think
cc @balazsorban44 maybe you want to add any comments. Otherwise I think we can ship this 🚀

ThangHuuVu (approved):

ThangHuuVu:
🙌

lavcraft:
@ThangHuuVu thanks. Already throw it to production, it works :) Many thanks!

PR #5173, opened 2022-08-17
feat: Add AppDirect provider

autumn-hoerr:
## ☕️ Reasoning

- Adds [AppDirect](https://www.appdirect.com) to the available providers.

## 🧢 Checklist

- [ x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/providers/appdirect.js
+++ b/packages/next-auth/src/providers/appdirect.js
@@ -7,7 +7,7 @@ export default function AppDirect(options) {
         id: "appdirect",
         name: "AppDirect",
         type: "oauth",
-        wellKnown: `https://${marketplaceUrl}/.well-known/openid-configuration`,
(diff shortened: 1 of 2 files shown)


packages/next-auth/src/providers/appdirect.js
```diff
@@ -0,0 +1,27 @@
+/** @type {import(".").OAuthProvider} */
```
ThangHuuVu, line 1: It would be awesome if you could provide this file as a `.ts` file
We're trying to convert all the providers from `.js` to `.ts`, so naturally, all new providers should be in `.ts` 😄

ThangHuuVu (commented):
@dan-hoerr Thanks for the PR!
Could you attach a video for proofing that the AppDirect provider works?

Issue #5180, opened 2022-08-17
nextjs nested middleware is not allowed anymore


LuisCadillo:
### What is the improvement or update you wish to see?

[This section](https://next-auth.js.org/tutorials/securing-pages-and-api-routes#nextjs-middleware) of the `next-auth` docs contains outdated info about  [nested middlewares](https://nextjs.org/docs/messages/nested-middleware) on nextjs

### Is there any context that might help us understand?

Tried to follow the advice of putting the _middleware file under a nested directory and runned under this [Nested Middleware](https://nextjs.org/docs/messages/nested-middleware) error

### Does the docs page already exist? Please link to it.

https://next-auth.js.org/tutorials/securing-pages-and-api-routes#nextjs-middleware

balazsorban44:
Good call, care to update in a PR? There's a edit this page button at the end of that page that should take you directly to the file that needs changing. Here is the correct documentation, using `matcher` instead of nested Middleware https://next-auth.js.org/configuration/nextjs#basic-usage

PR #5181, opened 2022-08-17
fix: remove outdated nested middleware info #5180

luiisca:
[This section](https://next-auth.js.org/tutorials/securing-pages-and-api-routes#nextjs-middleware) of the next-auth docs contains outdated info about [nested middlewares](https://nextjs.org/docs/messages/nested-middleware) on nextjs


balazsorban44 (changes requested):
Let's mention `matcher` for filtering instead of removing this info

balazsorban44 (approved):

PR #5188, opened 2022-08-19
🐛 Save updated tokens from a new sign-in

dglsparsons:
## ☕️ Reasoning

When a user signs in, the response may contain an updated set of
access/id/refresh tokens. When these are present, they need
updating in the stored Account information. This is important as
some auth providers will only allow one refresh token at a time
for a given application (looking at you twitter).

Failure to update them can leave the tokens in an non-working state.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [X] Ready to be merged

## 🎫 Affected issues

I'm happy to raise an issue if necessary. I hit this with the twitter provider (oauth 2.0), complete with the offline.access scope, which causes access/refresh/id tokens to be returned on a successful sign in.

### DIFF
--- a/packages/next-auth/src/core/lib/callback-handler.ts
+++ b/packages/next-auth/src/core/lib/callback-handler.ts
@@ -149,6 +149,13 @@ export default async function callbackHandler(params: {
             expires: fromDate(options.session.maxAge),
           })
 
+      if (account.access_token ?? account.refresh_token ?? account.id_token) {
+        // If we have any tokens in the request, then these need updating in the stored Account information.
+        // This is important as some auth providers will only allow one refresh token at a time for a given
+        // application (looking at you twitter). Failure to update them can leave the tokens in an non-working state.
+        await linkAccount({ ...account, userId: userByAccount.id })
+      }
+
       return { session, user: userByAccount, isNewUser }
     } else {
       if (user) {


packages/next-auth/src/core/lib/callback-handler.ts
```diff
@@ -149,6 +149,13 @@ export default async function callbackHandler(params: {
             expires: fromDate(options.session.maxAge),
           })
 
+      if (account.access_token ?? account.refresh_token ?? account.id_token) {
+        // If we have any tokens in the request, then these need updating in the stored Account information.
+        // This is important as some auth providers will only allow one refresh token at a time for a given
+        // application (looking at you twitter). Failure to update them can leave the tokens in an non-working state.
+        await linkAccount({ ...account, userId: userByAccount.id })
```
ThangHuuVu, line 156: I'm not sure if `linkAccount` is the right one to use here 🤔 In most of the implementation of adapters, it's an INSERT query, but we want UPDATE on Account instead
Which adapter are you using in your case?

dglsparsons: Oh no. 

I couldn't see any method that did an update on the `Account` - only link/unlink. I'm using the dynamodb adapter, so this was an upsert. 

I could opt for adding an updateAccount method to the [adapter interface](https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/adapters.ts#L58), but that might be pretty involved, as it would require adding this to every provider 🤔

ThangHuuVu: > I could opt for adding an updateAccount method to the [adapter interface](https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/adapters.ts#L58), but that might be pretty involved, as it would require adding this to every provider 🤔


Yeah exactly. I think it's a bit early now to invest in adding `updateAccount` for all adapters, especially when not many OAuth providers will rely on this code. Although, it could be useful in edge cases. cc @ndom91 WDYT? 
For the time being, I think you'll have to keep this at the application code. I think  either [Advanced Initialization](https://next-auth.js.org/configuration/initialization#advanced-initialization) or updating the account directly through the `session` callback would sufficient

dglsparsons: Yeah. I think that may be sensible. I suspect it's probably _only_ the twitter oauth2 provider, and only when `offline.access` is passed as a scope. 

Sadly I think the tokens are swallowed before they are returned, so I'm not sure Advanced Initialization will work. However, I have a work-around at the moment using `pnpm patch` on `next-auth`, so this isn't essential for my use-case.

dglsparsons:
Closing this PR as it looks like the behaviour is broken, but with no intention for a fix to be incorporated any time soon. No point keeping this open forever :)

PR #5328, opened 2022-09-11
Add option for custom generateSessionToken

halfmatthalfcat:
## ☕️ Reasoning

Instead of a random UUID, I want to leverage `cuid` as the session token. This PR allows the ability to define a custom `generateSessionToken` function and return the custom session token. Leverage original logic (random UUID) by default.

## 🧢 Checklist

- [x] Documentation
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: N/A

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/core/init.ts
+++ b/packages/next-auth/src/core/init.ts
@@ -1,3 +1,4 @@
+import { randomBytes, randomUUID } from "crypto"
 import { NextAuthOptions } from ".."
 import logger from "../utils/logger"
 import parseUrl from "../utils/parse-url"
@@ -86,6 +87,10 @@ export async function init({
       strategy: userOptions.adapter ? "database" : "jwt",
       maxAge,
       updateAge: 24 * 60 * 60,
+      generateSessionToken: () => {
+        // Use `randomUUID` if available. (Node 15.6++)
+        return randomUUID?.() ?? randomBytes(32).toString("hex")
+      },
       ...userOptions.session,
     },
     // JWT options
--- a/packages/next-auth/src/core/init.ts
+++ b/packages/next-auth/src/core/init.ts
@@ -88,7 +88,7 @@ export async function init({
       maxAge,
       updateAge: 24 * 60 * 60,
       generateSessionToken: () => {
-        // Use `randomUUID` if available. (Node 15.6++)
+        // Use `randomUUID` if available. (Node 15.6+)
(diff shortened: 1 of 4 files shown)


halfmatthalfcat:
@lluia @ThangHuuVu @ndom91 thoughts?

packages/next-auth/src/core/types.ts
```diff
@@ -362,6 +362,7 @@ export interface CookiesOptions {
   pkceCodeVerifier: CookieOption
   state: CookieOption
   nonce: CookieOption
+  generateSessionToken: () => string
```
ThangHuuVu, line 365: This `CookiesOptions` is more like for cookies-related configurations, as you can see they are all `CookieOption`
We should move this new config to somewhere else. I suggest it be inside of `session`.

ThangHuuVu (changes requested):

packages/next-auth/src/core/types.ts
halfmatthalfcat: Ah good suggestion! Made the changes.

packages/next-auth/src/core/types.ts
```diff
@@ -468,6 +468,13 @@ export interface SessionOptions {
    * @default 86400 // 1 day
    */
   updateAge: number
+  /**
+   * Generate a custom session token for database-based sessions.
+   * By default, a random UUID or string is generated depending on Node version
+   * however, you can specific your own custom string (such as cuids) to be used.
+   * @default randomUUID or randomBytes.toHex depending on Node version
```
balazsorban44, line 475:
```suggestion
   * Generate a custom session token for database-based sessions.
   * By default, a random UUID or string is generated depending on the Node.js version.
   * However, you can specify your own custom string (such as CUID) to be used.
   * @default `randomUUID` or `randomBytes.toHex` depending on the Node.js version
```

packages/next-auth/src/core/init.ts
```diff
@@ -86,6 +87,10 @@ export async function init({
       strategy: userOptions.adapter ? "database" : "jwt",
       maxAge,
       updateAge: 24 * 60 * 60,
+      generateSessionToken: () => {
+        // Use `randomUUID` if available. (Node 15.6++)
```
balazsorban44, line 91:
```suggestion
        // Use `randomUUID` if available. (Node 15.6+)
```

PR #5398, opened 2022-09-21
fix(jwt): sort chunk cookies by key num index

stephenway:
## ☕️ Reasoning

I'm adding in a sort function on cookie chunks in the JWT `SessionStore` value to ensure that when these chunks are received in the wrong order from the browser that they are joined in the correct order and do not cause null sessions or JWE decryption errors in production.

## 🧢 Checklist

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

- No issue on this repo was found

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/core/lib/cookie.ts
+++ b/packages/next-auth/src/core/lib/cookie.ts
@@ -111,7 +111,7 @@ export function defaultCookies(useSecureCookies: boolean): CookiesOptions {
         path: "/",
         secure: useSecureCookies,
       },
-    }
+    },
   }
 }
 
@@ -120,6 +120,18 @@ export interface Cookie extends CookieOption {
 }
 
 type Chunks = Record<string, string>
+
+/**
+ * Sort cookies by chunk name to avoid joining them in the wrong order
+ */
+function sortedObject(obj) {
+  const index = Object.keys(obj).sort()
+  const sorted = {}
+  index.forEach((key) => {
+    sorted[key] = obj[key]
+  })
+  return sorted
+}
 
 export class SessionStore {
   #chunks: Chunks = {}
@@ -152,7 +164,7 @@ export class SessionStore {
   }
 
   get value() {
-    return Object.values(this.#chunks)?.join("")
+    return Object.values(sortedObject(this.#chunks))?.join("")
   }
 
   /** Given a cookie, return a list of cookies, chunked to fit the allowed cookie size. */
--- a/packages/next-auth/tests/jwt.test.ts
+++ b/packages/next-auth/tests/jwt.test.ts
@@ -1,0 +1,23 @@
+import { SessionStore } from "../src/core/lib/cookie"
+
+it("should result in the correct cookie order", async () => {
+  const cookieName = "__Secure-next-auth.session-token"
+  const logger = console
+  const sessionStore = new SessionStore(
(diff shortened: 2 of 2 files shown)


balazsorban44 (changes requested):
This change sounds unnecessary. We haven't gotten reports about it. Have you encountered a problem? If so, could you add a reproduction so we can investigate?

balazsorban44:
Let's open an issue with a reproduction. We can re-open the PR if the issue is verified. :pray:

stephenway:
> Let's open an issue with a reproduction. We can re-open the PR if the issue is verified. :pray:

I can look into creating a reproduction for you but I'm not sure next-auth controls the headers coming in from the browser. We were seeing intermittent ordering issues with headers on session calls which caused next-auth to unset the chunk cookies in return. The best way for us to repro was to reorder headers in a curl copied from a session request in the browser.

I was thinking that my unit test would suffice showing the issue because it fails on your current branch without my code change.

Appreciate the attention and help. Thanks!

stephenway:
^ @balazsorban44

janhoogeveen:
> Let's open an issue with a reproduction. We can re-open the PR if the issue is verified. 🙏

@balazsorban44 

I've been using `@sidebase/nuxt-auth` for a project which uses next-auth under the hood, and yes, I can reproduce this for you.

There's an issue which describes the behaviour in more detail:
https://github.com/sidebase/nuxt-auth/issues/293

codetheorist:
I've also had this issue using @sidebase/nuxt-auth and have traced the root cause to this module.  This PR should be re-opened with a view to merging this in.

packages/next-auth/src/core/lib/cookie.ts
```diff
@@ -121,6 +121,18 @@ export interface Cookie extends CookieOption {
 
 type Chunks = Record<string, string>
 
+/**
+ * Sort cookies by chunk name to avoid joining them in the wrong order
+ */
+function sortedObject(obj) {
```
codetheorist, line 127: Probably wouldn't be an issue, but I'm not sure this would work correctly if there where more than 10 chunks, where the order would be 0,1,10,2,3,4... using this function.

packages/next-auth/tests/jwt.test.ts
```diff
@@ -0,0 +1,23 @@
+import { SessionStore } from "../src/core/lib/cookie"
+
+it("should result in the correct cookie order", async () => {
+  const cookieName = "__Secure-next-auth.session-token"
+  const logger = console
+  const sessionStore = new SessionStore(
```
codetheorist, line 6: This should be expanded to have 11 chunks to ensure the string re-ordering works as expected.

codetheorist (commented):
Nice work on the PR, hope the comments don't seem too picky.

codetheorist:
> Let's open an issue with a reproduction. We can re-open the PR if the issue is verified. 🙏

Could we re-open this now the issue is werified?

codetheorist:
I see this has been fixed in another PR and merged in the latest versions. Tested this in the `@sidebase/nuxt-auth` package and everything works as expected with regards to the chunked JWT session cookie.

christine927t:
@codetheorist In which version of `@sidebase/nuxt-auth ` is this working correctly?

PR #5431, opened 2022-09-26
Fix type of Adapter.updateUser

ysulyma:
## ☕️ Reasoning

This fixes the type of `Adapter::updateUser`, which was noted to be incorrect in the DynamoDB Adapter source.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged


packages/next-auth/src/adapters.ts
```diff
@@ -84,7 +84,7 @@ export interface DefaultAdapter {
   getUserByAccount: (
     providerAccountId: Pick<AdapterAccount, "provider" | "providerAccountId">
   ) => Awaitable<AdapterUser | null>
-  updateUser: (user: Partial<AdapterUser>) => Awaitable<AdapterUser>
+  updateUser: (user: Partial<Omit<AdapterUser, "id">> & Pick<AdapterUser, "id">) => Awaitable<AdapterUser>
```
ThangHuuVu, line 87: This doesn't make much sense to me, it's the same type, no? 🤔
`Omit<AdapterUser, "id">> & Pick<AdapterUser, "id">`

ThangHuuVu (changes requested):

packages/next-auth/src/adapters.ts
ysulyma: This is `AdapterUser` with `id` mandatory and all other fields optional

ThangHuuVu:
sorry for the delay @ysulyma I continued the code here https://github.com/nextauthjs/next-auth/pull/8044 to solve the conflicts & move the code the `packages/core`
I also added you as a co-author for that commit, thank you for contributing 🙏

PR #5442, opened 2022-09-28
fix(core): return JSON for non-HTML server route errors

balazsorban44:
In certain cases when the documentation is not followed, deployments on serverless platforms like Vercel can lead to unexpected rate limiting, since the code ends up in an infinite loop.

It is important to always follow the documentation (Eg. for Vercel, we have a dedicated section here: https://next-auth.js.org/deployment#vercel).

But in this case, we can bail out early from unexpected redirect loops by sending a proper JSON response for requests that expect a non-HTML response.

Let's look at the following example:

```js
// `pages/api/auth/[...nextauth].js` with custom pages

import GitHub from "next-auth/providers/github"

export default NextAuth({
  pages: { error: "/auth/login", signIn: "/auth/login" },
  providers: [
    GitHub({
      clientId: process.env.GOOGLE_ID,
      clientSecret: process.env.GOOGLE_SECRET,
    }),
  ],
})

// Custom login page at `pages/auth/login.jsx`.

import { getSession, signIn } from "next-auth/react"

export default function Login() {
  return <button onClick={() => signIn("github")}>Log in</button>
}

export async function getServerSideProps(ctx) {
  const session = await getSession(ctx)
  if (!session) return { props: {} }
  return { redirect: { destination: "/logged-in" } }
}
```

This looks innocent enough, but if the developer forgets to set `NEXTAUTH_SECRET`, the `getSession()` call until now would have returned with a redirect to the same error page (since NextAuth.js identified the missing secret as a misconfiguration). But that error page will call `getSession()` again, and an infinite loop can start.

To avoid this, we now terminate `getSession()` (and similar methods that would expect a JSON response) with a `500` status response and an error message:

> There is a problem with the server configuration. Check the server logs for more information.

The error cause will be logged in the server-side so the developer can fix the issue and after a redeploy, everything should be working fine.

### DIFF
--- a/packages/next-auth/src/core/index.ts
+++ b/packages/next-auth/src/core/index.ts
@@ -96,7 +96,7 @@ export async function NextAuthHandler<
     // Bail out early if there's an error in the user config
     logger.error(assertionResult.code, assertionResult)
 
-    const htmlPages = ["signin", "verify-request", "signout"]
+    const htmlPages = ["signin", "signout", "error", "verify-request"]
     if (!htmlPages.includes(req.action) || req.method !== "GET") {
       const message = `There is a problem with the server configuration. Check the server logs for more information.`
       return {
@@ -108,7 +108,7 @@ export async function NextAuthHandler<
     const { pages, theme } = userOptions
 
     const authOnErrorPage =
-      pages?.error && req.query?.callbackUrl.startsWith(pages.error)
+      pages?.error && req.query?.callbackUrl?.startsWith(pages.error)
 
     if (!pages?.error || authOnErrorPage) {
       if (authOnErrorPage) {
--- a/packages/next-auth/src/core/index.ts
+++ b/packages/next-auth/src/core/index.ts
@@ -94,13 +94,21 @@ export async function NextAuthHandler<
     assertionResult.forEach(logger.warn)
   } else if (assertionResult instanceof Error) {
     // Bail out early if there's an error in the user config
-    const { pages, theme } = userOptions
     logger.error(assertionResult.code, assertionResult)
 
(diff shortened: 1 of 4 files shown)


packages/next-auth/src/core/index.ts
```diff
@@ -94,13 +94,21 @@ export async function NextAuthHandler<
     assertionResult.forEach(logger.warn)
   } else if (assertionResult instanceof Error) {
     // Bail out early if there's an error in the user config
-    const { pages, theme } = userOptions
     logger.error(assertionResult.code, assertionResult)
 
+    const htmlPages = ["signin", "verify-request", "signout"]
+    if (!htmlPages.includes(req.action) || req.method !== "GET") {
```
ThangHuuVu, line 100: Should we check for header like `Accept: application/json` instead? 🤔 That way we'll be sure that we don't miss any requests in the check

packages/next-auth/src/next/index.ts
```diff
@@ -122,8 +122,10 @@ export async function unstable_getServerSession(
 
   cookies?.forEach((cookie) => setCookie(res, cookie))
 
-  if (body && typeof body !== "string" && Object.keys(body).length)
+  if (body && typeof body !== "string" && Object.keys(body).length) {
+    if (session.status === 500) throw new Error((body as any).message)
```
ThangHuuVu, line 126: trivial, but we can destruct status from session above 🙌

packages/next-auth/src/core/index.ts
balazsorban44: Some requests will set something like `Accept: */*` so we cannot rely on that, unfortunately.

ThangHuuVu (approved):
LFG 🚀

PR #5444, opened 2022-09-28
fix: safe use of localStorage API

etienne-martin:
## ☕️ Reasoning

The `localStorage` API can be traitorous. Making sure that it exists on the `window` object is not enough. In some cases browsers will define the `localStorage` property but will set it to `null`. Some will throw an error when trying to use its methods.

We are using `next-auth` in production on a high traffic website and it results in hundreds of thousands of errors (`Cannot read properties of null (reading 'setItem')`) per minute due to the lack of `localStorage` on some browsers.

For this reason, I added a try/catch to handle cases where the API might not be available.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/packages/next-auth/src/client/_utils.ts
+++ b/packages/next-auth/src/client/_utils.ts
@@ -94,10 +94,18 @@ export function BroadcastChannel(name = "nextauth.message") {
     /** Notify other tabs/windows. */
     post(message: Record<string, unknown>) {
       if (typeof window === "undefined") return
-      localStorage.setItem(
-        name,
-        JSON.stringify({ ...message, timestamp: now() })
-      )
+      try {
+        localStorage.setItem(
+          name,
+          JSON.stringify({ ...message, timestamp: now() })
+        )
+      } catch {
+        /**
+         * The localStorage API isn't always available.
+         * It won't work in private mode prior to Safari 11 for example.
+         * Notifications are simply dropped if an error is encountered.
+         */
+      }
     },
   }
 }


packages/next-auth/src/client/_utils.ts
```diff
@@ -94,10 +94,18 @@ export function BroadcastChannel(name = "nextauth.message") {
     /** Notify other tabs/windows. */
     post(message: Record<string, unknown>) {
       if (typeof window === "undefined") return
```
balazsorban44, line 96:
```suggestion
      if (typeof localStorage !== "object") return
```

Should this not be enough? :thinking:

etienne-martin: This is not enough unfortunately, some browsers throw when calling the `setItem` and `getItem` methods, even if they both exist. Sadly the only reliable way to test if the API is available (and working as expected) is to do something like this:

```
function isSupported(getStorage) {
  try {
    const key = "something-random";
    getStorage().setItem(key, key);
    getStorage().removeItem(key);

    return true;
  } catch (e) {
    return false;
  }
}
```

https://michalzalecki.com/why-using-localStorage-directly-is-a-bad-idea/

balazsorban44: :exploding_head: Thanks for the context!

balazsorban44 (approved):
Thanks!

darius208:
Hi, I got the same  issue right now
Any solution for this?

Issue #5049, opened 2022-07-29
[firebase adapter] - Firebase adapter requires open security rules we cannot allow in production


Chibionos:
### Adapter type

@next-auth/firebase-adapter

### Environment

  System:
    OS: Windows 10 10.0.22000
    CPU: (24) x64 AMD Ryzen 9 3900X 12-Core Processor
    Memory: 9.27 GB / 31.95 GB
  Binaries:
    Node: 16.13.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 8.1.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (103.0.1264.62)
    Internet Explorer: 11.0.22000.120
  npmPackages:
    next: ^12.2.3 => 12.2.3
    next-auth: ^4.10.2 => 4.10.2
    react: 18.2.0 => 18.2.0

  npmPackages:
    @next-auth/firebase-adapter: ^1.0.1 => 1.0.1

### Reproduction URL

https://github.com/nextauthjs/next-auth-example

### Describe the issue

**Firebase adapter is using client side library and it is expecting the security rules to be open for it to edit sessions,  accounts and other collections managed by it. We cannot allow those collections to have an open security rule.**

Is there a way we can safely open security rules for the firebase adapter to write and access these files as part of user context ensuring that a malicious user cannot access the records.

eg:
 
```
match /users/{uid} {
   allow read, write:  if request.auth.uid == uid;
}

match /sessions/{id} {
   allow read, write:  if request.auth.uid == resource.data.userId;
}

match /accounts/{id} {
  allow read, write: if request.auth.uid == resource.data.userId;
}
```

I tried a few with no luck. If you have the documentation for the rules the adapter is very viable to be used in production and would solve a lot of manual work that need to be done. 

The ideal fix here would be move the firebase adapter to require to use firebase-admin package as it would then have the right permissions without us opening the security rules.

Great work on the adapter !

-----

### How to reproduce

Setup the adapter and lock down security rules

```
rules_version = '2';
service cloud.firestore {
  match /databases/{database}/documents {
    match /{document=**} {
      allow read, write: if false;
    }
  }
}
```

### Expected behavior

We should have a clear security rule setup ensuring there is no security vulnerabilities exposed by opening up permissions.

balazsorban44:
[...]

Issue #4927, opened 2022-07-14
Firebase Adapter Fails with App Name Already exist error


satish860:
### Adapter type

@next-auth/firebase-adapter

### Environment

yarn dev

### Reproduction URL

https://github.com/satish860/codegist

### Describe the issue

I am getting the following error. 

`FirebaseError: Firebase: Firebase App named '[DEFAULT]' already exists with different options or config (app/duplicate-app).`

![image](https://user-images.githubusercontent.com/817360/178895566-c45bc524-6108-4a6a-acbe-a6168f86635f.png)

In my initial debugging, I found Firebase adapter is trying to initialize the app, which is already initialized. 

What if we can add the similar code 
`// Initialize Firebase
const app = !getApps().length ? initializeApp(firebaseConfig) : getApp();`

### How to reproduce

Step 1:  Run the app in the repo using Yarn dev
Step 2: Click the Sign link
Step 3: You will be getting the following auth error screen

![image](https://user-images.githubusercontent.com/817360/178896099-387195b1-e132-4da5-9b9f-49a170d5938e.png)

### Expected behavior

It should redirect me to the Provider Login screen

balazsorban44:
@chanceaclark let's revisit and allow passing an app, as people might have already set up one outside next-auth. 👍

ifeoluwaayo:
Going through this same issue please kindly fix or if there's a workaround I would love to know.

Chibionos:
Any scope of passing an firestore object created from firebase-admin ? that would solve our security rules issues #5049

Chibionos:
@ifeoluwaayo how are you setting up your security rules, I am curious as I see I cannot open the security rules for the collections this adapters requires. Did you find a work around ?

chanceaclark:
So I think there is a big gotcha here: **If you initialize a firebase app on the client, you need to initialize a new one on the server as well.**

I went ahead and added some logic in #5055 to make sure we get the default app or initialize a new one to prevent that duplicate app issue. Along with that I also added some documentation about reusing your client app inside the server config.

27lakshay:
I'm facing this same issue, I hope this gets merged quickly

itsmylife:
@27lakshay @chanceaclark I am waiting for that [PR](https://github.com/nextauthjs/next-auth/pull/5055) too. It's been very long and I can imagine the workload that people have. So let's cross our fingers 🤞 that will have some attention soon.

Sanjay-B:
[...]

PR #5449, opened 2022-09-29
Add Support for FirebaseApp passing to FirestoreAdapter

sbhadr:
Provides a single source of truth for instancing a FirestoreApp.

☕️ Reasoning

Previously, the adapter would automatically instance an app as '[DEFAULT]' which would conflict / break other firebase instances. If you only had one app that was instanced through the adapter, the issue was not directly noticeable since there was no real issue. It only happened if you had more than one app. Adapters are really supposed to bridge / add a compatibility layer. The developer should choose how they'd like to interpret / pass the actual FirestoreApp since it covers a variety of use-cases than just one. 

Instance a ``FirestoreApp`` how you usually would and then, pass it to ``FirestoreAdapter``. 

```js
// [...nextauth].js
import { initializeApp, getApps, getApp  } from 'firebase/app';

const firebaseApp = !getApps().length ? initializeApp({
    // Keys
}, "NEXT_AUTH") : getApp("NEXT_AUTH");

export default NextAuth({
    providers: [],
    adapter: FirestoreAdapter(firebaseApp),
    secret: process.env.AUTH_SECRET,
    jwt: {
        encryption: true
    },
    callbacks: {}
});
```
If you want to use this in the file directly. 

```js
// A separate file
import { initializeApp, getApps, getApp  } from 'firebase/app';

export const firebaseApp = !getApps().length ? initializeApp({
    // Keys
}, "NEXT_AUTH") : getApp("NEXT_AUTH");
```
If you want to use this in a different file. Export the variable. 

🧢 Checklist
 Documentation
 Tests
 Ready to be merged

🎫 Affected issues
Firebase Adapter Fails with App Name Already exist error - https://github.com/nextauthjs/next-auth/issues/4927

📌 Resources
[Contributing guidelines](https://github.com/nextauthjs/next-auth/pull/CONTRIBUTING.md)
[Code of conduct](https://github.com/nextauthjs/next-auth/pull/CODE_OF_CONDUCT.md)
[Contributing to Open Source](https://kcd.im/pull-request)

❓ Questions
I disabled the emulator. Do enough people use the emulator where it would be important to keep? I'd argue that creating a dev-version of the database is free and easy to do. The emulator seems like added work for no real reason that can't already be accomplished. 

I feel like the discussion for this https://github.com/nextauthjs/next-auth/issues/5049 should be started. 

Tested this and seems to work on my end. If someone could repro for me to see if it works for them.


drunkenvalley:
Better practice imo would be to check that the specific app actually exists, not just that there are apps existing and assuming the one you made exists.

```typescript
const name = 'NEXT_AUTH' // arbitrary name
const app = !!appList.length && !!appList.find(a => a.name === name)
        ? getApp(name)
        : initializeApp(options, name)
```

Realistically, if they're using Firestore they may very plausibly have a default app that they're using *as well*, so just checking if an app exists would still produce the error in that situation.

I feel like we could be accomplishing the matter of all this without dropping support for passing options though.

Sanjay-B:
> Better practice imo would be to check that the specific app actually exists, not just that there are apps existing and assuming the one you made exists.
> 
> ```ts
> const name = 'NEXT_AUTH' // arbitrary name
> const app = !!appList.length && !!appList.find(a => a.name === name)
>         ? getApp(name)
>         : initializeApp(options, name)
> ```
> 
> Realistically, if they're using Firestore they may very plausibly have a default app that they're using _as well_, so just checking if an app exists would still produce the error in that situation.
> 
> I feel like we could be accomplishing the matter of all this without dropping support for passing options though.

The snippet I provided checks to see if a firebase app with the identifier "NEXT_AUTH" exists before initializing a new one. It's not an assumption because, we're actively checking before we pass it along to the adapter. The current adapter just creates a new default app which effectively breaks any codebases that do not have special IDs for their instances. The goal here is to make sure that *we have a non-default app*. We could accomplish this without dropping options but, passing in options and having the adapter init the app for us goes against what an adapter functionally means. An adapter is meant to take two items that are naturally *incompatible* and then, bridges them along a common interface given that the two things that were naturally incompatible are supplied. First thing would be NextAuth. Second thing would be Firebase's SDKs. The common interface here would be the firebase adapter. 

Just because it's one way, doesn't mean that way is correct. 

@drunkenvalley

PR #5459, opened 2022-09-29
fix(providers): change EVE Online to OAuth2

### LINKED ISSUE #3760: scope changes causing authentication to fail for EVEOnline provider
### Description 🐜

Scope parameter in EVEOnlline provider is being overwritten with "**openid**"

The EVEOnline provider requires there to be NO scope parameter. However in the provider code it is specified as "publicData" but it's not even using that?

`    ..., wellKnown:
      "https://login.eveonline.com/.well-known/oauth-authorization-server",
    authorization: {
      params: {
        scope: "publicData",
      },
    },
`

However, when the login URL is constructed we get this:

`url: 'https://login.eveonline.com/oauth/authorize?client_id=*redacted*&scope=openid&response_type=code&redirect_uri=http%3A%2F%2Flocalhost%3A3000%2Fapi%2Fauth%2Fcallback%2Feveonline&nextauth=signin%2Ceveonline&state=GejPhUb_qWw8yj03eDsaDqmuN-tL1BLmXy3iRYW2Z-E`

See in the URL that **scope=openid** .. does anyone know why this is happening? EVEOnline reports a authentication error saying the scope is invalid. Is there a way to completely remove the **scope** parameter from the login URL?

Yes

### How to reproduce ☕️

Use the auth provider for EVEOnline. You'll need to get a developer API key which you can get from https://developers.eveonline.com/applications

### Screenshots / Logs 📽

image
image

### Environment 🖥

  System:
    OS: Windows 10 10.0.19043
    CPU: (8) x64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz   
    Memory: 4.19 GB / 15.93 GB
  Binaries:
    Node: 17.0.1 - E:\Program Files\nodejs\node.EXE
    npm: 8.1.2 - E:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 97.0.4692.99
    Edge: Spartan (44.19041.1266.0), Chromium (97.0.1072.69)
    Internet Explorer: 11.0.19041.1202
  npmPackages:
    next: ^12.0.9 => 12.0.9
    next-auth: ^4.0.0-beta.7 => 4.0.0-beta.7
    react: 17.0.2 => 17.0.2

### Contributing 🙌🏽

Yes,  I am willing to help solve this bug in a PR

brownoxford:
## ☕️ Reasoning

This PR fixes #3760, which is being caused by missing `id_token` in EvE authentication response as well as missing `userinfo_endpoint` in `.well-known` OIDC configuration.

This PR updates the eve configuration to manually specify endpoints for auth, token, and userinfo.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: #3760

### DIFF
--- a/packages/next-auth/src/providers/eveonline.ts
+++ b/packages/next-auth/src/providers/eveonline.ts
@@ -17,14 +17,9 @@ export default function EVEOnline<P extends EVEOnlineProfile>(
     id: "eveonline",
     name: "EVE Online",
     type: "oauth",
-    wellKnown:
-      "https://login.eveonline.com/.well-known/oauth-authorization-server",
-    authorization: {
-      params: {
-        scope: "publicData",
-      },
-    },
-    idToken: true,
+    authorization: "https://login.eveonline.com/v2/oauth/authorize?scope=publicData",
+    token: "https://login.eveonline.com/v2/oauth/token",
+    userinfo: "https://login.eveonline.com/oauth/verify",
     profile(profile) {
       return {
         id: String(profile.CharacterID),


joaomlneto:
Can confirm this fixes the current issues with the provider — LGTM.

While noting I'm probably the least-knowledgeable person in OAuth around here, I must ask: is this a compliance issue on Eve Online's side that should be reported, or just a slight out-of-the-box compatibility issue with NextAuth?

balazsorban44 (approved):
Thanks!


This should probably be taken up by the EVE Online developers too since their implementation seems non-spec conformant.

I also see https://login.eveonline.com/.well-known/openid-configuration, which also does not contain the userinfo endpoint. :thinking: (the two are just aliases :open_mouth:)

balazsorban44:
Hmm... Might have spoken too fast: https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata

It is only recommended, but I guess would be still nice if they included it. The `idToken: true` setting was likely wrong since we did not use the `openid-configuration` well-known endpoint.

PR #4893, opened 2022-07-10
feat(core): add allowedAccountTokens option

skirsten:
## ☕️ Reasoning

Just came about the problem described in https://github.com/nextauthjs/next-auth/discussions/4516 when trying to use the GitLab provider.
I could simply add the missing fields to my Prisma schema or remove it in a custom prisma adapter but I would rather not bloat it with unnecessary fields.

This PR introduces a new optional option to configure which fields from the provider will be stored on the `Account`:

> By default NextAuth.js stores all information about an OAuth account gathered by the provider on the [Account Model](https://next-auth.js.org/adapters/models#account).
> These tokens can include fields that are not defined on the database schema. With this option the tokens can be filtered before being stored on the Account.
> 
> _For example:_
> 
> ```js
> allowedAccountTokens: [
>   "refresh_token",
>   "access_token",
>   "expires_at",
>   "token_type",
>   "scope",
>   "id_token",
>   "session_state",
> ],
> ```
> 
> ensures only the fields that are supported by the [default Prisma schema](https://next-auth.js.org/adapters/prisma) are stored.

I am not fixed on a proper name yet. Maybe `allowedAccountFields` makes more sense to not confuse this with anything related to the JWT tokens.

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes #3818 (only the original issue, not the second part of the discussion)
Fixes #4515
Fixes #3823
Fixes #3067

PR #5461, opened 2022-09-30
feat(adapters): allow `Account` mapping before DB

balazsorban44:
This is an alternative proposal to #4893

Right now, we pass through all the values of a provider account to the DB, and it causes confusion/unforeseen breaks when a provider decides to return a new property.

To solve this, we can add an `account` property to the provider configuration, similar to the `profile` callback, to control what part of the account should be saved to the database.

Example:

```ts
// pages/api/auth/[...nextauth].ts
import AzureAD from "next-auth/providers/azure-ad"

// this can be in `next-auth/providers/azure-ad`
interface AzureADAccount {
  ext_expires_in: number
}

declare module "next-auth" {
  interface Account extends AzureADAccount {}
}


export default NextAuth({
  providers: [
    AzureAD({
      ...
      // this would show a type error, because we required `ext_expires_in`
      account({ ext_expires_in, ...a }) {
        return a
      }
      // account: "minimal" // conversely, we can also only allow the most common properties
    })
  ]
})
```

Closes #4893

### DIFF
--- a/packages/next-auth/src/core/lib/callback-handler.ts
+++ b/packages/next-auth/src/core/lib/callback-handler.ts
@@ -151,9 +151,14 @@ export default async function callbackHandler(params: {
       return { session, user: userByAccount, isNewUser }
     } else {
       if (user) {
+        const newAccount = await getAccount(
+          (options as InternalOptions<"oauth">).provider,
+          { ...account, userId: user.id }
+        )
+
         // If the user is already signed in and the OAuth account isn't already associated
         // with another user account then we can go ahead and link the accounts safely.
-        await linkAccount({ ...account, userId: user.id })
+        await linkAccount(newAccount)
         await events.linkAccount?.({ user, account, profile })
 
         // As they are already signed in, we don't need to do anything after linking them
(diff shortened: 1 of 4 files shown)


balazsorban44:
cc @palminha curious what you think. This would still introduce a new config option but would be already familiar since it's similar to the `profile()` callback and gives full control to the developer.

We could change the default value to `"minimal"` in a future major release.

It would align with the `profile` callback, which also only maps  `id`, `name`, `image`, and `email` by default. :+1:

packages/next-auth/src/core/lib/callback-handler.ts
```diff
@@ -218,3 +228,35 @@ export default async function callbackHandler(params: {
...
+    return {
+      ...required,
+      access_token: rest.access_token,
+      token_type: rest.token_type,
+      id_token: rest.id_token,
+      refresh_token: rest.refresh_token,
+      scope: rest.scope,
+      expires_at: rest.expires_at,
+      session_state: rest.session_state,
+    }
```
balazsorban44, line 259: These are the values currently in the default schemas.

balazsorban44:
Closing as discussed with @ThangHuuVu. With some minor changes, this will be part of `@auth/core`, so the intention is to still simplify the default account model.

PR #5485, opened 2022-10-03
feat(providers): Add Pinterest Provider

ranjan-purbey:
## ☕️ Reasoning
Add a provider for [Pinterest](https://www.pinterest.com)

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/docs/docs/providers/pinterest.md
+++ b/docs/docs/providers/pinterest.md
@@ -21,8 +21,8 @@ You can override any of the options to suit your own use case.
 
 ## Example
 
-```js
-import PinterestProvider from "next-auth/providers/pinterest";
+```ts
+import PinterestProvider from "next-auth/providers/pinterest"
 ...
 providers: [
   PinterestProvider({
@@ -31,7 +31,6 @@ providers: [
   })
 ]
 ...
-```
 
 :::tip
 To use in production, make sure the app has standard API access and not trial access
--- a/docs/docs/providers/pinterest.md
+++ b/docs/docs/providers/pinterest.md
@@ -1,0 +1,38 @@
+---
+id: pinterest
+title: Pinterest
+---
+
+## Documentation
+
+https://developers.pinterest.com/docs/getting-started/authentication/
+
+## Configuration
+
+https://developers.pinterest.com/apps/
+
+## Options
+
+The **Pinterest Provider** comes with a set of default options:
+
+- [Pinterest Provider options](https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/providers/pinterest.ts)
+
+You can override any of the options to suit your own use case.
+
+## Example
+
+```js
+import PinterestProvider from "next-auth/providers/pinterest";
+...
+providers: [
+  PinterestProvider({
+    clientId: process.env.PINTEREST_ID,
+    clientSecret: process.env.PINTEREST_SECRET
+  })
+]
+...
+```
+
+:::tip
+To use in production, make sure the app has standard API access and not trial access
+:::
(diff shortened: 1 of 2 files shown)


packages/next-auth/src/providers/pinterest.ts
```diff
@@ -0,0 +1,37 @@
...
+    authorization: {
+      url: "https://www.pinterest.com/oauth",
+      params: {
+        scope: "user_accounts:read",
+      },
+    },
+    checks: ["state"],
+    token: {
+      url: "https://api.pinterest.com/v5/oauth/token",
+    },
```
balazsorban44, line 26:
```suggestion
    token: "https://api.pinterest.com/v5/oauth/token",
```

packages/next-auth/src/providers/pinterest.ts
```diff
@@ -0,0 +1,37 @@
...
+    userinfo: {
+      url: "https://api.pinterest.com/v5/user_account",
+    },
```
balazsorban44, line 29:
```suggestion
    userinfo: "https://api.pinterest.com/v5/user_account",
```

packages/next-auth/src/providers/pinterest.ts
```diff
@@ -0,0 +1,37 @@
...
+    profile({ username, profile_image, ...profile }) {
+      return { id: username, name: username, image: profile_image, ...profile  };
```
balazsorban44, line 31:
```suggestion
      return {
        id: username,
        name: username,
        image: profile_image,
        email: null
      }
```

packages/next-auth/src/providers/pinterest.ts
```diff
@@ -0,0 +1,37 @@
...
+    },
+    options
+  };
+};
+
+export default PinterestProvider;
```
balazsorban44, line 37:
```suggestion
export default PinterestProvider
```

docs/docs/providers/pinterest.md
```diff
@@ -0,0 +1,38 @@
...
+import PinterestProvider from "next-auth/providers/pinterest";
+...
+providers: [
+  PinterestProvider({
+    clientId: process.env.PINTEREST_ID,
+    clientSecret: process.env.PINTEREST_SECRET
+  })
+]
+...
+```
```
balazsorban44, line 34:
```suggestion
```ts
import PinterestProvider from "next-auth/providers/pinterest"
...
providers: [
  PinterestProvider({
    clientId: process.env.PINTEREST_ID,
    clientSecret: process.env.PINTEREST_SECRET
  })
]
...
```
```

File: packages/next-auth/src/providers/pinterest.ts [hunk 6]
```diff
@@ -0,0 +1,34 @@
...
+    },
+    checks: ["state"],
+    token: "https://api.pinterest.com/v5/oauth/token",
+    userinfo: "https://api.pinterest.com/v5/user_account",
+    profile({ username, profile_image }) {
+      return {
+        id: username,
+        name: username,
+        image: profile_image,
+        email: null,
```
balazsorban44, line 29: We don't want to expose more than these by default. Do you know if the e-mail can be retrieved from somewhere? Otherwise, this provider won't work with our database strategy.

balazsorban44 (changes requested):
I just requested app access for us, I'll check it out once it is approved. The provider looks generally ok, I applied some stylistic changes to better comply with the code base, but thank you!

packages/next-auth/src/providers/pinterest.ts
ranjan-purbey: No, Pinterest doesn't provide email 
See docs: https://developers.pinterest.com/docs/api/v5/#tag/user_account

ranjan-purbey: I noticed a lot of existing providers such as Instagram do not provide user email as well

ranjan-purbey:
Is anything required from my end before the PR can be merged?

balazsorban44 (approved):
Thanks, I finally got access, tested it, works great! It's sad that we cannot retrieve the email address, but it looks like there is not much we can do about it. :shrug:

PR #5556, opened 2022-10-11
fix: TS Module Augmentation

### LINKED ISSUE #5542: Module Augmentation doesn't work since last update
### Question 💬

After upgrade next-auth from 4.12.2 to 4.13.0. Type augmentation doesn't work as expected regarding the docs.

In [...nextauth].ts in signIn callback function, user type is now:

```typescript
user: User | {
  email: string;
};
 ```
When I want access user in my current signIn callback implementation, TS shows me error bellow:
```TS error: Property 'id' does not exist on type 'User | { email: string; }'.```

I didn't found any info regarding this change. I would like to avoid to create helper function which check if user parameter is User type. I also tried to typecast the parameter but with it brings more errors...

Can you check please if it's a bug or expected behaviour which is not covered yet in docs and I need to check the type by myself now?

### How to reproduce ☕️

```typescript
async signIn({user}) {
      const userISID = user.id.toUpperCase()
      // rest of the code
 }
 ```

### Contributing 🙌🏽

### PR DESCRIPTION
**Pull request by ThangHuuVu**
## ☕️ Reasoning

This fixes the Module Augmentation issue:
- [x] Remove `| { email: string }` part which breaks the Module Augmentation. I don't think we need it for anything 🤔. Matches that type with `AdapterUser` instead. 
- [x] Added a quick module augmentation in the dev app for easier review, will remove it afterward 🙌

## 🎫 Affected issues

Fixes: #5542 

## 📌 Resources

- [Security guidelines](./Security.md)

### DIFF
--- a/apps/dev/types/nextauth.d.ts
+++ b/apps/dev/types/nextauth.d.ts
@@ -1,0 +1,18 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import NextAuth, { DefaultSession } from "next-auth"
+
+declare module "next-auth" {
+  /**
+   * Returned by `useSession`, `getSession` and received as a prop on the `SessionProvider` React Context
+   */
+  interface Session {
+    user: {
+      /** The user's postal address. */
+      address: string
+    } & DefaultSession["user"]
+  }
+
+  interface User {
+    foo: string
+  }
+}
--- a/apps/dev/types/nextauth.d.ts
+++ b/apps/dev/types/nextauth.d.ts
@@ -1,5 +1,5 @@
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-import NextAuth, { DefaultSession } from "next-auth"
+import NextAuth from "next-auth"
 
 declare module "next-auth" {
   /**
@@ -9,7 +9,7 @@ declare module "next-auth" {
     user: {
       /** The user's postal address. */
       address: string
-    } & DefaultSession["user"]
+    } & User
   }
 
   interface User {
--- a/packages/next-auth/src/core/lib/email/getUserFromEmail.ts
+++ b/packages/next-auth/src/core/lib/email/getUserFromEmail.ts
@@ -1,4 +1,4 @@
-import type { AdapterUser } from "../../adapters"
+import type { AdapterUser } from "../../../adapters"
 import type { InternalOptions } from "../../types"
 
 /**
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: apps/dev/types/nextauth.d.ts [hunk 1]
```diff
@@ -0,0 +1,18 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import NextAuth, { DefaultSession } from "next-auth"
```
balazsorban44, line 2:
```suggestion
import NextAuth from "next-auth"
```

packages/next-auth/src/core/types.ts
```diff
@@ -265,7 +265,7 @@ export interface CallbacksOptions<P = Profile, A = Account> {
    * [Documentation](https://next-auth.js.org/configuration/callbacks#sign-in-callback)
    */
   signIn: (params: {
-    user: User | { email: string }
+    user: User
```
balazsorban44, line 268: From our source code, it looks like this can also be `AdapterUser`

balazsorban44 (changes requested):
I believe events and the signIn callback both can receive the user from the database.

ThangHuuVu:
> I believe events and the signIn callback both can receive the user from the database.
That's true, I matched the type of `{ email: string }` to `AdapterUser` 🙌

packages/next-auth/src/core/lib/email/getUserFromEmail.ts
```diff
@@ -1,19 +1,20 @@
+import { AdapterUser } from "src/adapters"
```
balazsorban44, line 1:
```suggestion
import type { AdapterUser } from "src/adapters"
```

balazsorban44 (approved):

packages/next-auth/src/core/lib/email/getUserFromEmail.ts
```diff
@@ -1,19 +1,20 @@
+import type { AdapterUser } from "../../adapters"
```
balazsorban44, line 1:
```suggestion
import type { AdapterUser } from "../../../adapters"
```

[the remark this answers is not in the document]
ThangHuuVu: Gotcha 👀

packages/next-auth/src/providers/credentials.ts
```diff
@@ -17,7 +17,7 @@ export interface CredentialsConfig<
   authorize: (
     credentials: Record<keyof C, string> | undefined,
     req: Pick<RequestInternal, "body" | "query" | "headers" | "method">
-  ) => Awaitable<(Omit<User, "id"> | { id?: string }) | null>
```
emrysal, line 20: @ThangHuuVu We use SERIAL id's as we have a PostgreSQL database coming from v3, the removal of this line triggers a type error as number is not assignable to string.

zomars: Do we need to extend the `User` interface as well? @emrysal

PR #5586, opened 2022-10-17
chore(gitlab): add example of redirect uri

### LINKED ISSUE #5031: Add a example for callback url in docs for all providers

I recently noticed this while working with the auth0 provider that 

Only the docs for google provider tell us about what callback / allowed redirect to use    

while it is something pretty simple and is mentioned pretty early in the docs 

In my humble opinion I believe that it'd be so much more usefull and convenient if we could list them out on the docs for each provider

not only that but if we do this it'll make the docs look a lot more consistent as well

With all due respect, Please let me know you're thought's 
I'd be really happy to contribute this one  

E.g. 

Here on the doc for google provider 
in the configuration section we list what callback's to use for the allowed redirect url's

Screenshot 2022-07-26 at 20-29-07 Google NextAuth js

Where as on the docs for auth0, gitlab , etc.. we didn't mention it at all

Screenshot 2022-07-26 at 20-29-28 Auth0 NextAuth js

Screenshot 2022-07-26 at 20-51-23 GitLab NextAuth js

https://next-auth.js.org/providers/google

paul-vd:
## ☕️ Reasoning

It adds clear and concise redirect url examples for gitlab.

## 🧢 Checklist

- [X] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: https://github.com/nextauthjs/next-auth/issues/5031

### DIFF
--- a/docs/docs/providers/gitlab.md
+++ b/docs/docs/providers/gitlab.md
@@ -11,6 +11,15 @@ https://docs.gitlab.com/ee/api/oauth2.html
 
 https://gitlab.com/-/profile/applications
 
+The "redirect URIs" used when creating the credentials must include your full domain and end in the callback path. For example;
+
+- For production: `https://{YOUR_DOMAIN}/api/auth/callback/gitlab`
+- For development: `http://localhost:3000/api/auth/callback/gitlab`
+
+:::tip
+Gitlab allows multiple redirect URI's for the same application, one line per URI
+:::
+
 ## Options
 
 The **Gitlab Provider** comes with a set of default options:


balazsorban44 (changes requested):
Thanks, but I think this should be addressed holistically for all providers instead.

Admittedly, the docs here is very thin: https://next-auth.js.org/getting-started/rest-api#getpost-apiauthcallbackprovider

We should explain what `:provider` means, and maybe link to this from all provider docs.

PR #5618, opened 2022-10-23
stricter types for email provider

brunocrosier:
## ☕️ Reasoning

Currently the type for `server` is `string` or an incorrect type from nodemailer.
By using template literal types we can ensure a better developer experience when using this provider

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged


packages/next-auth/src/providers/email.ts
```diff
@@ -16,8 +15,14 @@ export interface SendVerificationRequestParams {
 
 export interface EmailConfig extends CommonProviderOptions {
   type: "email"
-  // TODO: Make use of https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html
-  server: string | SMTPTransportOptions
+  server: `smtp://${string}:${string}@${string}.${string}:${number}` | {
```
ThangHuuVu, line 18: This seems stricter than before. For the object, I think it's best to keep using the TS type from `nodemailer`. There are other properties there that could be useful for the developer 🤔 https://nodemailer.com/smtp/

For the string: 
- We could simplify the `host` bit to `string` instead of `${string}.${string}`. That's because sometimes the host can have more than one dot (subdomains etc...).
- Can we have optional query params at the end?
- Can we add support for `smtps` too? 
Let me know what you think. 🙌

ThangHuuVu (changes requested):

packages/next-auth/src/providers/email.ts
brunocrosier: great feedback! unfortunately at the moment the types from `nodemailer` don't seem to work - i will check again though

re: `${string}.${string}` - this should already accept multiple dots ([link to TS Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAysBOBLAdgcwCIHthyWqAvFAAYAkA3gM4IqoC+AdBdXvcQLABQXAxpstSgBDAFywaaLDgmpCUAOQAbTPAgBbBojCUArmvm9+ggEZjctKefxF5AE0zL4DSomAMhaiMAOcA9L6gAAWBKAFoIAA9IHmBw+HgVQwFgKB4zGUsZOXlkbFCUHkUdW1pQllofIA))

definitely good point re: query params at the end and `smtps` - i will add these

reconbot: hostnames aren't always public FQDN's sometimes so I wouldn't ensure a hostname has any dot

reconbot: I have a usecase for using the SES transport for nodemailer, so I took a dive into the types. It's impossible to do cleanly with how the types exist in nodemailer. It is possible to derive input type but the overloads that define the input aren't a proper type union and show up as unknown.

```ts
type TransportOptions = Parameters<typeof createTransport>[0]
// turns into 
TransportOptions | Transport<unknown> // which has no possible inputs
```

So what do we do? We make our own union. I'd worry it's a maintenance burden but the options haven't changed in a few years and we can always drop it for an `any`. It turned out to be a lot of work so I can't really drop it here in a comment, and instead I opened https://github.com/nextauthjs/next-auth/pull/7950 which solves the nodemailer options but not the template literal.

ndom91:
Closing this as its mostly been taken care of by https://github.com/nextauthjs/next-auth/pull/7950

PR #5626, opened 2022-10-23
feat(providers): add Dribbble provider

nawok:
## ☕️ Reasoning

This pull request adds the new `DribbbleProvider` which allows authenticating with [Dribbble](https://dribbble.com/).

Demo: https://next-auth-drbl.vercel.app

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

No issues found.

## 📌 Resources

- [Security guidelines](../blob/main/SECURITY.md)
- [Contributing guidelines](../blob/main/CONTRIBUTING.md)
- [Code of conduct](../blob/main/CODE_OF_CONDUCT.md)


ndom91:
Please include a `light`/`dark` SVG of the provider logo. We've recently redesigned the built-in Signin and other pages, and the new social login buttons now include the logo of the provider as well as the colors. 

See: https://github.com/nextauthjs/next-auth/pull/5802

ndom91:
@balazsorban44 @ThangHuuVu I've added a dribbble test application `client_id`/`client_secret` to our password manager list :+1:

nawok:
@ndom91, I pushed the logos.

Wasn’t sure about the `style` object but I still provided it in a similar way to other providers because I wanted to include the colors from [Dribbble Brand Resources, Logo Downloads & Guidelines](https://dribbble.com/media-kit).

The implementation looks a bit strange for two reasons:

- For example, the Apple logo is referenced as `https://raw.githubusercontent.com/nextauthjs/next-auth/main/packages/next-auth/provider-logos/apple.svg`. What happens if NextAuth is used in an environment where `raw.githubusercontent.com` is inaccessible?
- When submitting a PR contributors need to provide a link that is 404 before the merge.

Shouldn’t the SVGs be «baked into» the package?

ndom91:
> @ndom91, I pushed the logos.
> 
> Wasn’t sure about the `style` object but I still provided it in a similar way to others because I wanted to include the colors from [Dribbble Brand Resources, Logo Downloads & Guidelines](https://dribbble.com/media-kit).
> 
> The implementation looks a bit strange for a couple of reasons:
> 
> * For example, the Apple logo is referenced as `https://raw.githubusercontent.com/nextauthjs/next-auth/main/packages/next-auth/provider-logos/apple.svg`. What happens if NextAuth is used in an environment where `raw.githubusercontent.com` is inaccessible?
> * If the image location is changed (due to refactoring, repository URL change, main branch renaming, or even rebranding of the provider) all versions in production will have broken icons.
> * When submitting a PR, contributors need to provide a link that is 404 before the merge.
> 
> Shouldn’t the SVGs be «baked into» the package and use relative paths?

Hey so first of all, thanks for updating that so quickly!

And all great points - we didn't want ship a bunch of SVG's with the library, _but_ you're right on these counts. I don't think the `main` branch will change and we (now) can't move the `provider-logos` directory without breaking this, right haha, so I see those as less of an issue. But the shipping a 404 in the PR is really annoying imo (until its merged to main). 

Do you have any other suggestions on how we could handle these better without actually shipping the SVG's? Or is that really the only other smart solution? :/ :thinking:

nawok:
> Do you have any other suggestions on how we could handle these better without actually shipping the SVG's?

As I can see there is just a plain image tag with the URL at the moment (if URL is provided):

```tsx
{provider.style?.logo && (
  <img id="provider-logo" src={provider.style.logo} />
)}
```

1. I would think about a gracious fallback if the image URL doesn’t resolve. This would allow to move the images around and do a fallback to a newer location. In the worst case just don’t render the `` at all.
2. Create a separate package for UI? This would allow to decouple the library code from its design. We can still keep the `style` API for customization but the new package can provide sensible defaults.
3. Use a naming convention for styles lookup in the UI package.

What do you think?

ndom91:
> > Do you have any other suggestions on how we could handle these better without actually shipping the SVG's?
> 
> As I can see there is just a plain image tag with the URL at the moment (if URL is provided):
> 
> ```tsx
> {provider.style?.logo && (
>   
> )}
> ```
> 
> 1. I would think about a gracious fallback if the image URL doesn’t resolve. This would allow to move the images around and do a fallback to a newer location. In the worst case just don’t render the `` at all.
> 2. Create a separate package for UI? This would allow to decouple the library code from its design. We can still keep the `style` API for customization but the new package can provide sensible defaults.
> 3. Use a naming convention for styles lookup in the UI package.
> 
> What do you think?

Great ideas! 

1. This can definitely be done here relatively quickly, right 
2. @balazsorban44 was just talking about potentially splitting up the ui components into their own exports that people could pick and chose from when building their own signin pages, for example. So this could align with that 
3. I'm not entirely sure what you mean with this one 🤔

nawok:
> 3\. I'm not entirely sure what you mean with this one 🤔

I mean that, let’s say, we have some providers that don’t have `style` object yet. Or it might be as well just some newly merged provider.

When the sign-in page is rendered, we can look up by provider `id` if there is a corresponding `style` in the UI package (hence the naming convention: the style ID and package ID should match).
- If not, we don't render the image and use the default colors for the text and background.
- When a new UI package is released, the new icon and colors will be found and used on the page.

ndom91:
@nawok ah okay, makes sense!

Yeah I like that path forward a lot actually. @balazsorban44 @ThangHuuVu what do you guys think about the separate UI package thing?

nawok:
After updating this PR I noticed that the `legacy` label was applied. What should I update/improve to modernize it?

nawok:
Updated to use the new package scructure.

ndom91:
So generally this looks good to me!

ThangHuuVu:
Closed via #8531 , thanks @nawok 🙏

PR #5644, opened 2022-10-25
Export default next pages

tobilen:
> _NOTE_:
>
> - Please make sure that you are _NOT_ opening a PR to fix a potential security vulnerability. Instead, please follow the [Security guidelines](../Security.md) to disclose the issue to us confidentially.

## ☕️ Reasoning

I would like to customize the next-auth ui, while keeping the basic functionality in place. An example use case would be a redirect in case of errors:

```typescript
import * as React from "react";
import { GetServerSideProps } from "next";
import SignInPage, { SignInServerPageParams } from "next-auth/core/pages/signin";
import { useRouter } from "next/router";
import { getProviders } from "next-auth/react";

export const getServerSideProps: GetServerSideProps = async () => {
  const providers = await getProviders();
  return {
    props: { providers },
  };
};

const SignIn: React.FC<SignInServerPageParams> = ({ error, ...props }) => {
  const router = useRouter();
  React.useEffect(() => {
    if (error === "OAuthCallback") {
      router.replace("/");
    }
  }, [error, router]);

  return <SignInPage error={error} {...props} />;
};
export default SignIn;
```

Currently, this is not possible when using ESM because the pages are not exported from the package.

## 🧢 Checklist

~- [ ] Documentation~
~- [ ] Tests~
- [x] Ready to be merged

## 🎫 Affected issues
-

### DIFF
--- a/packages/next-auth/package.json
+++ b/packages/next-auth/package.json
@@ -31,6 +31,7 @@
     "./jwt": "./jwt/index.js",
     "./react": "./react/index.js",
     "./core": "./core/index.js",
+    "./core/pages/*": "./core/pages/*.js",
     "./next": "./next/index.js",
     "./middleware": "./middleware.js",
     "./client/_utils": "./client/_utils.js",


balazsorban44:
Since we cannot guarantee that these pages will stay the same, it's safer if you just paste the code directly from source, or create your own pages from scratch.

Issue #5647, opened 2022-10-26
Support Next.js 13 `app` directory 


mm-webx:
### Description 📓

Add some hook to use like that:
```ts
const session = use(getSession());
```

### How to reproduce ☕️

.

### Contributing 🙌🏽

Yes, I am willing to help implement this feature in a PR

MoltenCoffee:
Would this already work with the experimental `unstable_getServerSession`?

mm-webx:
I will try it today.On 26 Oct 2022, at 08:27, Thijs-Jan ***@***.***> wrote:﻿
Would this already work with the experimental unstable_getServerSession?

mm-webx:
```ts
import { use } from "react";
import { unstable_getServerSession } from "next-auth/next";
export default function Page(props) {
  console.log({ props }); // { props: { params: {}, searchParams: {} } }

  const session = use(unstable_getServerSession());  //TS2345: Argument of type '[]' is not assignable to parameter of type '[IncomingMessage & { cookies: Partial<{ [key: string]: string; }>; }, ServerResponse<IncomingMessage>, NextAuthOptions] | [...]'.   Type '[]' is not assignable to type '[NextApiRequest, NextApiResponse<any>, NextAuthOptions]'.     Source has 0 element(s) but target requires 3.

  return (
    <>
      <h1>Hi</h1>
    </>
  );
}
```

There is no access to req and response. Using getSession also not working, even with `use client`

mm-webx:
I created `layout`:
```ts
"use client";

import { SessionProvider } from "next-auth/react";

export default function RootLayout({
  children,
  ...props
}: {
  children: React.ReactNode;
}) {
  console.log("layout", { props });
  return (
    <html>
      <head></head>
      <body>
        <SessionProvider session={props.session}>{children}</SessionProvider>
      </body>
    </html>
  );
}
```

And some "ClientSide" component:
```ts
"use client";

import { useState } from "react";
import { useSession } from "next-auth/react";

export default function Counter() {
  const session = useSession();
  const [count, setCount] = useState(0);
  console.log({ session });
  return (
    <div>
      <p>You clicked {count} times</p>
      <button onClick={() => setCount(count + 1)}>Click me</button>
    </div>
  );
}
```
Where `session` is empty
```
But cannot obtain session details.

**Comment by mm-webx** [comment]
Current output is: 
```
web:dev: error - ../node_modules/next-auth/react/index.js (108:27) @ eval
[...]

PR #5718, opened 2022-11-04
fix(react): support importing "next-auth/react" in Server Component environment

zenflow:
## ☕️ Reasoning

Currently, importing "next-auth/react" in a Server Component module (or in a module imported by a Server Component module) will fail with `Uncaught TypeError: React.createContext is not a function`.

This is because React Context is unavailable in Server Components.

With this change, "next-auth/react" can be imported without error, and an informative error is thrown if there's an attempt to actually *use* React Context.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [X] Ready to be merged

## 🎫 Affected issues

Related issue: #5647 (this PR doesn't add complete support, but improves support)

### DIFF
--- a/packages/next-auth/src/react/index.tsx
+++ b/packages/next-auth/src/react/index.tsx
@@ -95,9 +95,9 @@ export type SessionContextValue<R extends boolean = false> = R extends true
       | { data: Session; status: "authenticated" }
       | { data: null; status: "unauthenticated" | "loading" }
 
-export const SessionContext = React.createContext<
-  SessionContextValue | undefined
->(undefined)
+export const SessionContext = React.createContext
+  ? React.createContext<SessionContextValue | undefined>(undefined)
+  : undefined
 
 /**
  * React Hook that gives you access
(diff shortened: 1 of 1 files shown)


balazsorban44:
Thanks, we plan to design a better API for the Next.js app directory, but as for now, the recommendation is to add the `'use client'` directive to the file where you wrap your components with `SessionProvider`.

zenflow:
> as for now, the recommendation is to add the 'use client' directive to the file where you wrap your components with SessionProvider.

That's basically what I'm doing...

```
"use client";
// This module exists for the sole purpose of cloning SessionProvider & adding the compiler directive above

import React from "react";
import { SessionProvider } from "next-auth/react";

export const ClientSessionProvider: typeof SessionProvider = (props) => {
  return <SessionProvider {...props} />;
};
```

... but it doesn't solve this separate problem (described under Reasoning above).

> we plan to design a better API for the Next.js app directory

@balazsorban44 Since that will probably take some time, would you consider merging this small change as a stop-gap?

balazsorban44:
Upon revisiting, I think this is a good solution and does not introduce problems, even if we want to change APIs in the future. Given `app/` is experimental in Next.js, we are also free to do breaking changes if needed.

PR #5730, opened 2022-11-04
fix(client): prevent the session from being out-of-sync between tabs

etienne-martin:
Hi there, we noticed a small bug that resulted in sessions becoming out-of-sync between browser tabs.

Steps to reproduce:

1. Create a page that prints the result of `useSession`
2. Open the page in two browser tabs
3. Wait for the session to change on the backend (The change in our case was a token that got refreshed with new data).
4. Call the `getSession` function manually in one of the tabs
5. See the changes get broadcasted to the other tab
6. **The issue:** the tab that triggered the call didn't get the update

This is due to the way localStorage events work, storage events won't be delivered on the page that made the changes. 

So in this case, the `getSession` call broadcasts the event to other tabs, the other tabs listen to the event and update their session but the main tab is stuck with an outdated state.

The way I fixed the issue is by making the `getSession` function act as the source of truth for session data. Whenever a call is made to `getSession`, no matter who calls it (could be an internal call or one made by consumers of the library), the `getSession` function will emit the latest result of the function call. The provider listens for the result of those calls and updates its state. This ensures that what you see from (`getSession`) is what you get from (`useSession`).

## 🧢 Checklist

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged


zenflow (approved):

elvinasv:
Hi guys, thanks for your work 🙌 
Is there a plan to release this in the near future?

etienne-martin:
We ended up ditching the client-side portion of next-auth and rewrote it ourselves by talking directly to the next-auth API.

ThangHuuVu:
hi @etienne-martin,  would you please explain more about your case? I think it is pretty rare, calling `getSession` outside of the React context after the session being updated on the server. 🤔 FYI, we have a new `update` function that you could use on the client to update the session, and it will sync the session between tabs.
For now, I'll close the PR. If there is anyone still interested in shipping this as part of `next-auth` client, please open a new issue _with proper reproduction_ so that we can discuss the issue in details. Thanks 🙏

PR #5739, opened 2022-11-06
feat(provider): Mastodon

### LINKED ISSUE #5669: Mastodon as provider
### Description 📓

Because Elon Musk now owns Twitter, lot's of people switched to mastodon. Adding mastodon as auth-provider would be great.

[Mastodo docs for oauth](https://docs.joinmastodon.org/client/token/) 

Mastodon is a federated service, so the app has to be registered at login-time.

### How to reproduce ☕️

-

### Contributing 🙌🏽

leifarriens:
## ☕️ Reasoning

This pull request adds a `MastodonProvider` and the corresponding documentation which allows authenticating with [Mastodon](https://mastodon.social/). 

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Resolves: #5669 

## 📌 Resources


Jak2k:
Mastodon is federated. So Client-Id and -Secret are unique on every instance. The app must be registered dynamicly for every instance.

Instead of Client-Id/-Secret, only the name, adress and redirect urls should be passed.

leifarriens:
> Mastodon is federated. So Client-Id and -Secret are unique on every instance. The app must be registered dynamicly for every instance.

Generating `clientId` and `clientSecret` has to be handled outside of next-auth. You could solve this by [creating an app](https://docs.joinmastodon.org/client/token/#app) elsewhere in your application and securely store `id` and `secret` (if not present) for the given instance. Then redirect to `[...nextauht].ts` where you asynchronously populate the config.

```ts
export default async function auth(req: NextApiRequest, res: NextApiResponse) {
  const mastodonProviders = await db.getMastodonInstances();

    providers: [
      GoogleProvider({
        clientId: process.env.GOOGLE_ID ?? '',
        clientSecret: process.env.GOOGLE_SECRET ?? '',
      }),
      ...mastodonProviders.map((instance) =>
        Mastodon({
          clientId: instance.client_id,
          clientSecret: instance.client_secret,
          issuer: instance.issuer,
        }),
      ),
    ],
  });
}
```

Jak2k:
There are currently more than 300 instances. I think, that would slow down to much.

ndom91:
I would assume the developer / admin just manually creates the app in their dev settings in their mastodon instance once. No need to create it on multiple. 

Once you have the clientId/secret from your mastodon instance, you can pass those two and the URL for your instance to this provider, then you can authenticate against it. That's it. 

You don't expect to be able to login with your one mastodon instances' credentials to every single mastodon server, right? 

Sidenote: to create an app in the mastodon dev settings, in order to get a clientId/secret, you need a login for that mastodon instance yourself. So that would mean to do this dynamic generation of clientId on every mastodon instance, that your app would need a login for every one too. Just doesn't seem feasible..

dsaltares:
@ndom91 could you not create apps programmatically for the requested instance using this API? https://docs.joinmastodon.org/methods/apps/#create

ndom91:
> @ndom91 could you not create apps programmatically for the requested instance using this API? https://docs.joinmastodon.org/methods/apps/#create

Yeah but you'd still need an account at each instance to get an API key, right?

dsaltares:
https://docs.joinmastodon.org/methods/apps/#create returns the `clientId` and the `clientSecret` for your app against that instance. That seems to be all you need to be able to get tokens via oauth https://docs.joinmastodon.org/methods/oauth/#authorize and then create a post https://docs.joinmastodon.org/methods/statuses/#create, for example.

Jak2k:
> > @ndom91 could you not create apps programmatically for the requested instance using this API? https://docs.joinmastodon.org/methods/apps/#create
> 
> Yeah but you'd still need an account at each instance to get an API key, right?

You don't need an account on every instance. The Client-id/-secret is automaticly created over that api-method and then comes the oauth-stuff.

The client-id/-secret are usually created when the user tries to log in. Then it could be saved with the token or saved in a database with functions, that are provided by the developer, who uses the lib.

Creating client-id/-secret should be made by the lib.

Jak2k:
Is this still active?

leifarriens:
> Is this still active?

I think the state of this PR is like @ndom91 mentioned a working version of "login with an mastodon insance" instead of "login with mastodon". There is a variety of other providers that follow this paradigm of using an issuer.

> The client-id/-secret are usually created when the user tries to log in. Then it could be saved with the token or saved in a database with functions, that are provided by the developer, who uses the lib.

The adapter logic exceeds my knowledge of nextauth, but in my understadning this would intervene in the business logic of the different database adapters and I am unsure if the maintainers would want to support it just for a single provider.

ndom91:
Hmm alright so correct me if im wrong, but you can create client_id's / client_secret's on the fly with a public API on each mastodon instance? 

So the flow would be something like:
1. User enters their mastodon username + password
2. User enters the URL of their mastodon server
3a. Auth.js checks if we have a `client_id` for that server yet, if so use it
3b. If not, generate one on the fly with the (authenticated? unauthenticated?) api 
4. Use that client_id to sign in the user

Is that about right?

So yeah that sounds a lot more doable to me than registering a dummy user at every server imaginable and to create client_id's ahead of time 😅 

As @leifarriens mentioned, however, we do not have anywhere in the database to save `client_id`'s / `client_secret`'s, as these should normally be passed via environment variables by the user. Also databases are only optional with our oauth providers, so there will be many auth.js users who do not even have a database to save those in.. 

Would writing them to a local json file on the server make sense? Maybe in self-hosted instances of the auth.js-containing app, but on providers like vercel and netlify the hosting environment is wiped for every new deploy so it'd be gone then as well.

If I understood everything correctly (^^), then this still sounds like its not a good fit for auth.js. At least the way its been currently planned / suggested 🤔

Jak2k:
> Hmm alright so correct me if im wrong, but you can create client_id's / client_secret's on the fly with a public API on each mastodon instance?
> 
> So the flow would be something like:
> 
> 1. User enters their mastodon username + password
> 2. User enters the URL of their mastodon server
>    2a. Auth.js checks if we have a `client_id` for that server yet, if so use it
>    2b. If not, generate one on the fly with the (authenticated? unauthenticated?) api
> 3. Use that client_id to sign in the user
> 
> Is that about right?
> 
> So yeah that sounds a lot more doable to me than registering a dummy user at every server imaginable and to create client_id's ahead of time 😅
> 
> As @leifarriens mentioned, however, we do not have anywhere in the database to save `client_id`'s / `client_secret`'s, as these should normally be passed via environment variables by the user. Also databases are only optional with our oauth providers, so there will be many auth.js users who do not even have a database to save those in..
> 
> Would writing them to a local json file on the server make sense? Maybe in self-hosted instances of the auth.js-containing app, but on providers like vercel and netlify the hosting environment is wiped for every new deploy so it'd be gone then as well.
> 
> If I understood everything correctly (^^), then this still sounds like its not a good fit for auth.js. At least the way its been currently planned / suggested 🤔

Maybe it could be saved in the client.

Storing the client-* only prevents duplicates in the *authorized-apps*-list. So it is optional. Therefore next-auth can also make it optional.

ndom91:
Came across a nice open source mastodon client from the vue/nuxt folks. 

Looks like they support mastodon oauth with any server by just adding the server URL. Probably has some good ideas in it: https://github.com/elk-zone/elk/blob/main/components/user/UserSignIn.vue

Jak2k:
Is this still active?

jibla:
> Hmm alright so correct me if im wrong, but you can create client_id's / client_secret's on the fly with a public API on each mastodon instance?
> 
> So the flow would be something like:
> 
> 1. User enters their mastodon username + password
> 2. User enters the URL of their mastodon server
>    2a. Auth.js checks if we have a `client_id` for that server yet, if so use it
>    2b. If not, generate one on the fly with the (authenticated? unauthenticated?) api
> 3. Use that client_id to sign in the user
> 
> Is that about right?
> 
> So yeah that sounds a lot more doable to me than registering a dummy user at every server imaginable and to create client_id's ahead of time sweat_smile
> 
> As @leifarriens mentioned, however, we do not have anywhere in the database to save `client_id`'s / `client_secret`'s, as these should normally be passed via environment variables by the user. Also databases are only optional with our oauth providers, so there will be many auth.js users who do not even have a database to save those in..
> 
> Would writing them to a local json file on the server make sense? Maybe in self-hosted instances of the auth.js-containing app, but on providers like vercel and netlify the hosting environment is wiped for every new deploy so it'd be gone then as well.
> 
> If I understood everything correctly (^^), then this still sounds like its not a good fit for auth.js. At least the way its been currently planned / suggested thinking

I don't think 1st step needs to be there.

1. User does just enters the mastodon server's url that he/she wants to sign-in with.
2. Auth.js first goes to /api/v1/apps endpoint and grabs the client_id & secret_id
3. Auth.js dynamically creates the standard oauth 2 provider with the client/secret and then everythin must continue pretty standard oauth way.

There can be an optional step where we save the client/secret and check if we already have for a given server, but we also could ask to new client/secret every time we need them. Of course its super ineffective but in theory it can also work this way.

jibla:
I have published an article about Auth.js, showing how it can be used outside its supported frameworks, purely with Node & Express. I show this with Mastodon OAuth example. If anyone is interested you can check it here: https://medium.com/@jibla/auth-js-exploration-1b6c27cf076f

ThangHuuVu:
hi folks, I moved the code in this PR to https://github.com/nextauthjs/next-auth/pull/8516 and merge it to core to provide the function to login to a Mastodon instance.

PR #5741, opened 2022-11-06
feat(next): allow `unstable_getServerSession` in Server Components

### LINKED ISSUE #5647: Support Next.js 13 `app` directory

balazsorban44:
Next.js 13 shipped with the new [`app/` directory](https://beta.nextjs.org/docs/app-directory-roadmap) that implements [React Server Components](https://reactjs.org/blog/2020/12/21/data-fetching-with-react-server-components.html). With the recently introduced change of allowing `async/await` directly, there is a need for accessing the session in server components in a clean way.

With this PR, `unstable_getServerSession` will now work in Server Components as well:

```tsx
// app/page.tsx
import { authOptions } from "pages/api/auth/[...nextauth]"

  const session = await unstable_getServerSession(authOptions)
  return <pre>{JSON.stringify(session, null, 2)}</pre>
}
```

**Notes:**

1. Currently, the underlying Next.js `cookies()` method does [only provides read access](https://beta.nextjs.org/docs/api-reference/cookies) to the request cookies. This means that the `expires` value is stripped away from `session` in Server Components. Furthermore, there is a hard expiry on sessions, after which the user will be required to sign in again. (The default expiry is 30 days).

2. Since `app/` is experimental in Next.js, we are considering this to be experimental in NextAuth.js so we can iterate on the implementation if upstream changes require it.

Fixes #5647

### DIFF
--- a/apps/dev/app/layout.tsx
+++ b/apps/dev/app/layout.tsx
@@ -1,0 +1,12 @@
+export default function RootLayout({
+  children,
+}: {
+  children: React.ReactNode
+}) {
+  return (
+    <html>
+      <head></head>
+      <body>{children}</body>
+    </html>
+  )
+}
--- a/apps/dev/app/server-component/page.tsx
+++ b/apps/dev/app/server-component/page.tsx
@@ -1,0 +1,7 @@
+import { unstable_getServerSession } from "next-auth/next"
+import { authOptions } from "pages/api/auth/[...nextauth]"
+
+export default async function Page() {
+  const session = await unstable_getServerSession(authOptions)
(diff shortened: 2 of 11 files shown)


ThangHuuVu (approved):
🚢

glenarama:
While this appears to work in dev - i get no session when hosted on Vercel. Is anyone else able to confirm if this is true or perhaps related to my setup?

PR #5763, opened 2022-11-09
fix(next): improve dev environment variable handling

balazsorban44:
1. We should not show a warning locally, as we can default to `http://localhost:3000`
2. We should not use `process.env.VERCEL` or `process.env.AUTH_TRUST_HOST` in development by default. Introduce a new `trustHost` option to `NextAuthOptions` in case you still want to. (Not documented yet on purpose) (Implementation is verbose for `detectHost` params so we can avoid referencing `process.env` anywhere in core)

Slack discussion for @ThangHuuVu: https://vercel.slack.com/archives/D031RDYDJUF/p1666172865143889

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources


packages/next-auth/src/next/index.ts
```diff
@@ -141,12 +146,18 @@ export async function unstable_getServerSession(
...
 
   const session = await NextAuthHandler<Session | {} | string>({
     options,
     req: {
-      host: detectHost(req.headers["x-forwarded-host"]),
+      host: detectHost(
+        options.trustHost,
+        req.headers["x-forwarded-host"],
+        process.env.NEXTAUTH_URL ??
+          (process.env.NODE_ENV !== "production" && "http://localhost:3000")
```
HaNdTriX, line 159: @balazsorban44 

I am not sure if hardcoding the port makes sence here. This might break the experience since Next.js might choose another port if:

* the `-p` command line switch is used (e.g.: `next dev -p 4000`)
* or port `3000` is already in use. 

Afaik Node.js & the Edge Env add an Host header anyway, when using the http protocol. Is there any reason, we are not using this `req.host` or `req.headers.get('host')` as a fallback? I have just tested this using `next dev` and it seems to work in edge and node.js envs.

### Docs

- [`http.host`](https://nodejs.org/api/http.html#requesthost)

PR #5782, opened 2022-11-10
Do not throw exception if callbackUrl is malformed

SferaDev:
## ☕️ Reasoning

In our application we incorrectly had `callbackUrl: String(url)` which was producing `"undefined"` as `callbackUrl` sent to next-auth. Next auth was trying to do a `new URL` with it that was throwing an exception.

Even if the error was on our side, it would be nice from next auth to do not throw trying to parse the callback url and if it's malformed just ignore it.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/core/lib/default-callbacks.ts
+++ b/packages/next-auth/src/core/lib/default-callbacks.ts
@@ -5,9 +5,13 @@ export const defaultCallbacks: CallbacksOptions = {
     return true
   },
   redirect({ url, baseUrl }) {
-    if (url.startsWith("/")) return `${baseUrl}${url}`
-    else if (new URL(url).origin === baseUrl) return url
-    return baseUrl
+    try {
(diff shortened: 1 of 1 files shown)


balazsorban44 (changes requested):
Technically, as you noticed too, this is a userland error, and we should not cloak it.

Rather, we should surface the issue better by doing a check in https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/core/lib/assert.ts to assert that the result of `redirect` can correctly be turned into a `URL`, eg: something like:


```js
try {
  new URL(callbacks.redirect({url, baseUrl}))
} catch {
  throw new InvalidRedirectCallback("`callbacks.redirect` returned an invalid URL.")
}
```

PR #5789, opened 2022-11-10
made prisma adapter function a generic

keepri:
## ☕️ Reasoning

Transformed the PrismaAdapter function into a generic in order to be able to pass in any PrismaClient type. This change should allow the adapter to not throw an error when importing and using a prisma client that has been generated elsewhere.

### DIFF
--- a/packages/adapter-prisma/src/index.ts
+++ b/packages/adapter-prisma/src/index.ts
@@ -1,7 +1,7 @@
-import type { PrismaClient, Prisma } from "@prisma/client"
+import type { Prisma, PrismaClient } from "@prisma/client"
 import type { Adapter, AdapterAccount } from "next-auth/adapters"
 
-export function PrismaAdapter(p: PrismaClient): Adapter {
+export function PrismaAdapter<T extends PrismaClient>(p: T): Adapter {
   return {
     createUser: (data) => p.user.create({ data }),
     getUser: (id) => p.user.findUnique({ where: { id } }),


balazsorban44 (changes requested):
I'm curious, what does "elsewhere" refer to here?

I haven't seen this problem surfaced before, could you create a bug report first with a reproduction so we can verify before doing this change?

cpitul:
> I'm curious, what does "elsewhere" refer to here?
> 
> I haven't seen this problem surfaced before, could you create a bug report first with a reproduction so we can verify before doing this change?

Elsewhese: in the [...nextAuth].ts file, when using the prisma adapter, if I pass in a prisma client that was not generated inside of the node_modules folder it thows a type error saying the prisma client generated has a different _fetcher_ declared than it is expecting.

I've encountered this problem only when trying to use a prisma client that has been generated elsewhere, ex: inside of the "database" internal package of a monorepo.

davguij:
FWIW, I found this issue yesterday as well, and I'm pretty sure that I was trying to do the same thing as @cpitul: set up Prisma and Next Auth in a Turborepo, taking advantage of the build caching capabilities of the tool.

The issue was first reported at https://github.com/vercel/turbo/pull/2390#issuecomment-1310190331

NuroDev:
Agee with @davguij & @cpitul here.
Tried doing the same thing & ran into issues when caching generated client types related to this / the PR linked above

cpitul:
> 

You mean this: https://github.com/vercel/turbo/pull/2390#issuecomment-1310190331 ?

Dr1Ku:
This sounds like a nice improvement, I'd merely suggest sticking with the `prisma` 'convention' of naming the `Client` `prisma`, `p` here is too terse.

Slightly off-topic but also related to making the `prisma` `adapter` more generic -- how about having the option to pass in different `model` name(s) so as to be able to vary the otherwise hardcoded `user`, `account`, `session` and `verificationToken`? 

I'm proposing this since this would make the adapter more 'drop-in' than it is now. In our (incipient) codebase, we had to, for now, rename our own `user` and `account` models, mark `next-auth`'s `user` and `account` as 'not ours' via `[@@map](https://next-auth.js.org/adapters/prisma#naming-conventions)` and run with having two sets of `user`s and `account`s. 

This, of course, has a disadvantage of having to link our own tables to next-auth's. I've tried unifying by marking all the non-`next-auth` fields as optional but that didn't work, I'll let you know how/if it improved when I revisit this issue in our codebase.

ThangHuuVu:
> This sounds like a nice improvement, I'd merely suggest sticking with the `prisma` 'convention' of naming the `Client` `prisma`, `p` here is too terse.
> 
> Slightly off-topic but also related to making the `prisma` `adapter` more generic -- how about having the option to pass in different `model` name(s) so as to be able to vary the otherwise hardcoded `user`, `account`, `session` and `verificationToken`?
> 
> I'm proposing this since this would make the adapter more 'drop-in' than it is now. In our (incipient) codebase, we had to, for now, rename our own `user` and `account` models, mark `next-auth`'s `user` and `account` as 'not ours' via [@@map](https://next-auth.js.org/adapters/prisma#naming-conventions) and are 'stuck' with having two sets of `user`s and `account`s.
> 
> This, of course, has the disadvantage of having to link our own tables to next-auth's. I've tried unifying by marking all the non-`next-auth` fields as optional but that didn't work, I'll let you know how/if it improved when I revisit this issue in our codebase.

You can open a discussion separately on the option to pass in different `model` names. 👍

ThangHuuVu:
hey @keepri , if you are still interested, please help to file a proper reproduction for this issue - I will reopen the PR after checking the reproduction 😃 FWIW, I've tried to clone the [example repo](https://turbo.build/repo/docs/handbook/tools/prisma) at https://github.com/ThangHuuVu/turbo-prisma-nextauth but I don't see the build issue 🤔

PR #5791, opened 2022-11-11
feat(jwt): allow `getToken` in Server Components

### LINKED ISSUE #5754: Allow `getToken` in Server Components
### Description 📓

It is now possible to use `unstable_getServerSession` from a server component (in Next.js 13) thanks to #5741.

It should be possible to use the same _trick_ to allow using `getToken` as well I guess?

I’m available to draft a PR with a POC in the next couple of days.

### How to reproduce ☕️

N/A

### Contributing 🙌🏽

scastiel:
## ☕️ Reasoning

In Next.js, with new `app` folder, it might be necessary to call `getToken` from a server component, with no access to the request `req`.

This PR aims to reproduce what was done on #5741 for `unstable_getServerSession`, to allow `getToken` to be called without `req`.

```tsx

  const session = await unstable_getServerSession()
  const token = await getToken()
  return (
    <>
      <h2>Session</h2>
      <h2>Token</h2>
      <pre>{JSON.stringify(token, null, 2)}</pre>
    </>
  )
}
```

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes #5754

### DIFF
--- a/apps/dev/app/server-component/page.tsx
+++ b/apps/dev/app/server-component/page.tsx
@@ -1,6 +1,15 @@
+import { getToken } from "next-auth/jwt"
 import { unstable_getServerSession } from "next-auth/next"
 
 export default async function Page() {
   const session = await unstable_getServerSession()
-  return <pre>{JSON.stringify(session, null, 2)}</pre>
+  const token = await getToken()
+  return (
+    <>
+      <h2>Session</h2>
+      <pre>{JSON.stringify(session, null, 2)}</pre>
+      <h2>Token</h2>
+      <pre>{JSON.stringify(token, null, 2)}</pre>
+    </>
+  )
 }
--- a/apps/dev/app/server-component/page.tsx
+++ b/apps/dev/app/server-component/page.tsx
@@ -1,9 +1,18 @@
-import { getToken } from "next-auth/jwt"
+import { Session } from "next-auth"
+import { JWT } from "next-auth/jwt"
 import { unstable_getServerSession } from "next-auth/next"
 
+type SessionWithToken = Session & { token: JWT }
+
 export default async function Page() {
-  const session = await unstable_getServerSession()
-  const token = await getToken()
+  const { token, ...session } =
+    await unstable_getServerSession<SessionWithToken>({
+      providers: [],
+      callbacks: {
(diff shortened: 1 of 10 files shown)


scastiel:
**Note:** this PR should have more checks, update documentation, and probably display some warning. But before going further, if I could get some kind of _go signal_ that’d be terrific 😉

balazsorban44:
I actually think if we should just document to let the `session` (and/or maybe the `jwt) callback out from `authOptions` when you use `unstable_getServerSession`. That's essentially the only difference.

Both reads the session cookie, but `getToken` cannot handle if the session is using `strategy: "database"`. `getToken` is basically just reading the cookie and decrypt it. `unstable_getServerSession` does that too, without any options passed to it.

I want our APIs to start converging more, cause right now we have `getToken`, `getSession`, `useSession`, `unstable_getServerSession`... It would be nice to drop `getToken`.

scastiel:
Updated the PR to use `unstable_getServerSession` with `session` callback instead to get the token.

> And to satisfy TypeScript, maybe modify things a bit so the return type of unstable_getServerSession is either Session or the return type of callbacks.session if defined.

I tried something in the PR. Not fully convinced (seems a bit complex, lots of generics, I wonder if a deeper changer in the API wouldn’t be better).

At this point I’m keeping the PR for the suggestion, but feel free to close if you don’t want to follow this way 😉

balazsorban44:
What are your thoughts on #5792 as an alternative to this? Could you review that PR?

scastiel:
> What are your thoughts on https://github.com/nextauthjs/next-auth/pull/5792 as an alternative to this?

I prefer your solution! Just tested the experimental release by replacing `getToken` with `unstable_getServerSession` on a personal project, works like a charm 👍 

I wonder if the API isn’t starting to get weird though. Encouraging to call an `unstable` method that is supposed to return a session but that you can configure to return something else… 😅

balazsorban44:
Hence unstable. We might change the name before finalizing. My dream is to have `NextAuth` work _anywhere_. I'm closing this in favor of that PR then. 👍

PR #5792, opened 2022-11-11
fix(ts): improve `unstable_getServerSession` return type

### LINKED ISSUE #5754: Allow `getToken` in Server Components

balazsorban44:
Currently, `unstable_getServerSession` assumes that it always returns a `Session`. However, in practice, it should match the `session` callback's return type. This can be used to create feature parity with `getToken` like this:

```ts
// app/page.tsx
import { headers, cookies } from "next/headers"

  const session = await unstable_getServerSession({
    callbacks: { session: ({ token }) => token },
  }) // session will take the shape of JWT

    req: {
      headers: Object.fromEntries(headers()),
      cookies: Object.fromEntries(cookies().getAll().map((c) => [c.name, c.value])),
    } as any,
  })

  // These two should look equal
  return (
    <>
      <pre>{JSON.stringify(token, null, 2)}</pre>
    </>
  )
}
```

This is an alternative to #5791, reducing the number of APIs needed in the future (no need for `getToken`). (See https://github.com/nextauthjs/next-auth/pull/5791#issuecomment-1311120142)

Closes #5791, fixes #5754


packages/next-auth/src/next/index.ts
```diff
@@ -85,17 +86,25 @@ export default NextAuth
 
 let experimentalWarningShown = false
 let experimentalRSCWarningShown = false
-export async function unstable_getServerSession(
+
+type GetServerSessionOptions = Partial<Omit<NextAuthOptions, "callbacks">> & {
+  callbacks?: Omit<Partial<CallbacksOptions>, "session"> & {
```
ThangHuuVu, line 91:
```suggestion
  callbacks?: Omit<NextAuthOptions['callbacks'], "session"> & {
```

ThangHuuVu (commented):
I really like this approach, save us from exposing another API! 👍 Added 2 comments

packages/next-auth/src/next/index.ts
```diff
@@ -85,17 +86,25 @@ export default NextAuth
...
+    session?: (...args: Parameters<CallbacksOptions["session"]>) => any
```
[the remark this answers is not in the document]
balazsorban44: Yes it is, that's the difference and the reason for `GetServerSessionOptions` to exist :D

packages/next-auth/src/next/index.ts
```diff
@@ -9,6 +9,7 @@ import type {
 } from "next"
 import type { NextAuthOptions, Session } from ".."
 import type {
+  CallbacksOptions,
```
balazsorban44, line 12:
```suggestion
```

packages/next-auth/src/next/index.ts
```diff
@@ -143,7 +153,7 @@ export async function unstable_getServerSession(
   } else {
     req = args[0]
     res = args[1]
-    options = args[2]
+    options = Object.assign(args[2], { providers: [] })
```
ammmze, line 156: Is it intentional that we are modifying `args[2]` here? I recently updated next-auth and started getting some odd behavior where I've got places in my app using my `authOptions` object and it's fine at first, but eventually the providers got cleared out...this is the reason for that.

I suspect we meant to do `options = Object.assign({}, args[2], { providers: [] })` instead?

ammmze: I've opened a [PR](https://github.com/nextauthjs/next-auth/pull/5973) that should fix this.

PR #5798, opened 2022-11-11
Added an example app to demonstrate sign-in and sign-up

Issaminu:
When I first tried out Next Auth, I found it really difficult to find a full example of using Next Auth for authentication in a Next app.
So today I made an example app that contains account creation front-end and back-end logic, Sign-in, and persisting user session locally with Recoil and Recoil Persist.
The example app is available here: https://github.com/Issam-Boubcher/credentials-next-auth-example.

Feel free to check it out and give feedback.

### DIFF
--- a/docs/docs/providers/credentials.md
+++ b/docs/docs/providers/credentials.md
@@ -73,6 +73,15 @@ providers: [
 
 See the [callbacks documentation](/configuration/callbacks) for more information on how to interact with the token.
 
+## Example - Sign-in with email and password
+
+The credentials provider can be used to set-up authentication with email and password.
+
+For more information, check out the example app below:
+
+- [Example App Repo](https://github.com/Issam-Boubcher/credentials-next-auth-example).
+- [Example App Demo](https://credentials-next-auth-example.vercel.app/).
+
 ## Example - Web3 / Signin With Ethereum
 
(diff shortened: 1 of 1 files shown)


balazsorban44:
Thanks, but we are actively not pursuing to showcase this flow, as generally not recommend and such would like to keep the docs minimal around it.

Issam-Boubcher:
I understand your sentiment around this, but please consider the fact that for many, their provider choice will mainly be credentials, with maybe an option to sign in with socials.
The lack of comprehensive documentation guiding the developer to using this provider **legitimately** leads to a horrible experience.

My main motivation behind this is that I want others to not have the same experience I had. When I just started using next-auth, I spent upwards of 20 hours trying to set up the credentials provider to no avail. And eventually just ended up switching to Auth0.

I understand that you must desensitize the use of credentials because they're an outdated concept at this point, but there's still a mountain of developers that wish/are required to use it. And being aggressive towards pushing against it would only lead to them making bad decisions surrounding authentication/authorization _(like I did with Auth0)_.

ldenblyd:
Agree with Issam ! 
users sometimes needs credentials login, and not including it just because i'ts more complicated to implement is not a good solution.

PR #5802, opened 2022-11-12
feat: add signin button styles

ndom91:
### Changes

- Changed default provider button styling
- Added provider svgs and brand colors

### Todos
- [ ] Improve default darkmode black button styling

Preview: 
<table>
<tr>
<td>
next-auth light mode btn preview
</td>
<td>
next-auth dark mode btn preview
</td>
</tr>
</table>

### Notes
- If you're involved with any of these projects and we've used your brand colors wrong or anything like that - we encourage you to reach out here and we'll be happy to fix it!
- Some providers really do provide some strict rules around what they'd like the btn to look like. Here are Google's, for example - [developers.google.com/identity/branding-guidelines#incorrect](https://developers.google.com/identity/branding-guidelines#incorrect)


packages/next-auth/src/core/pages/signin.tsx
```diff
@@ -1,4 +1,5 @@
 import type { InternalProvider, Theme } from "../types"
+import React from "react"
```
balazsorban44, line 2: Could we mark this with `type`.
```suggestion
import type React from "react"
```

packages/next-auth/src/providers/github.ts
```diff
@@ -98,6 +98,22 @@ export default function Github<P extends GithubProfile>(
         image: profile.avatar_url,
       }
     },
+    style: {
```
balazsorban44, line 101: I think we should rather accept a link and host the defaults in our repo next to the providers, maybe. Ideally we should make sure the logo sizes are the same as well.
Can we have some strict guidelines like transparent no shadow backgrounds required (saw this on the Google icon)? Then one logo would be enough too. optionally have a colored and monochrome version and add a toggle in `theme.signInButtons: "colored"|"monochrome" |"none"`

ndom91: Yeah so I was debating, for example with the Github one, you could reuse the same svg and then just adjust the `fill` color based on `light | dark` mode.

If we're just saving the links to svgs (in the repo for example) then I also don't think it'd be a big deal to allow options like `colored | monochrome | none`, etc. Dig it!

And regarding icon sizes and stricter requirements. I think these might be a good starting point, thoguhts?
- 32x32
- No whitespace around the svg (so the logo itself really is 32x32)
- ?

ndom91: The thing with the using the same svg file and adjusting it for light/dark automatically is, its going to be different for each provider, right? So we'd have to provide some more config options/fields where people can specify which color the `fill` should be in light/dark mode. And what if theres multiple shapes / paths that need different fills for each theme.. I think we might just be better off providing two (or more) svgs. Like `apple.svg` and `apple-dark.svg`.

packages/next-auth/src/core/pages/signin.tsx
```diff
@@ -101,8 +102,28 @@ export default function SigninPage(props: SignInServerPageParams) {
...
                   <input type="hidden" name="callbackUrl" value={callbackUrl} />
                 )}
-                <button type="submit" className="button">
-                  Sign in with {provider.name}
+                <button
+                  type="submit"
+                  className="button"
+                  style={
+                    {
+                      "--provider-bg": provider.style?.bg ?? "",
```
balazsorban44, line 110: Let's use default values, so we can remove them in the default provider configs.

ndom91: Yeah so theres default values set in the CSS already actually.

image

Gawdfrey:
I am happy to see an initiative to improve the visual components, but I still feel that there can be some improvements to the aesthetics.

Looking at services that offer login infrastructure and UI, like [Supabase Auth](https://supabase.com/auth) and [Clerk](https://clerk.dev/components#auth-user-components), I would say that they are aesthetically pleasing and more modern looking.

I am by no means a designer, but I quickly drew up an alternative design which does not drastically change anything but at least adds some colors. Based the color on one of the purple colors in the logo.
Auth

I am of course biased, but I feel that a more modern looking and more colorful design would make it unnecessary for people to override the base design.

balazsorban44:
Keep in mind that colors are dictated by the [`theme`](https://next-auth.js.org/configuration/options#theme), but otherwise this is useful feedback and we will discuss it! Thanks!

Exerra:
I like @Gawdfrey's mockup more as it is more in line with the modern design of Supabase Auth & Auth0.
My suggestions for the redesign are:
* Make the background a grey-ish colour and remove the border from the login box (will make it look cleaner & for dark mode probably reverse? background is black and the login box is grey-ish)
* Put the logo inside the login box & make it smaller (sorta like @Gawdfrey's suggestion, but even smaller)
* Use theme colour for the login button (idk about dark mode)
* 30px rounding
* Smaller provider icons
* Grey border instead of shadow

Gawdfrey:
I made some alterations according to some of the suggestions @Exerra had.
The button background color is controlled by the theme and t
Desktop - 1
Desktop - 2
he button text is controlled by the button text field.

ndom91:
> I made some alterations according to some of the suggestions @Exerra had. The button background color is controlled by the theme and the button text is controlled by the button text field. 

I like the mockups a lot in general, however, are you allergic to the standard login button layout? i.e. Left aligned button and center aligned text in the remaining space? :joy: `</sarcasm>`

Gawdfrey:
> 
> I like the mockups a lot in general, however, are you allergic to the standard login button layout? i.e. Left aligned button and center aligned text in the remaining space? 😂 `</sarcasm>`

Hehe, did not know that I was allergic of that until today!
I looked at some examples and I have to agree that it certainly looks much better.
Desktop - 7
Desktop - 8

ndom91:
> > I like the mockups a lot in general, however, are you allergic to the standard login button layout? i.e. Left aligned button and center aligned text in the remaining space? joy `</sarcasm>`
> 
> Hehe, did not know that I was allergic of that until today! I looked at some examples and I have to agree that it certainly looks much better. Desktop - 7 Desktop - 8

Haha so this is still a bit off imo, you have the text `left-aligned` to a certain point. I'd be curious to see your mockup with something like this:

image

Basically you can make the container `flex`, with two children - the image and the `<span>` with the text. Then you can set on the flex parent `justify-content: flex-start` which will push all of it to the left. Then set the text `span` to `flex-grow: 1` which will cause it to take up the rest of the space dynamically. Finally, in that span you can set `text-align: center`. That should do the trick

Gawdfrey:
> Basically you can make the container `flex`, with two children - the `img` and the `span` with the text. Then you can set the parent to `justify-content: flex-start` which will push all of it to the left. Then set the `span` to `flex-grow: 1` which will cause it to take up the rest of the space dynamically. Finally, in that `span` you can set `text-align: center`. That should do the trick

Aah, my bad I misunderstood. Thought you were talking about the "Sign in" button..

Desktop - 7
Desktop - 8

Exerra:
My horrible photoshop skills aside, I feel like making the background a grey colour really helps draw the focus to the login box (+ with no border looks nice and modern 👀).

But that is all I could nitpick from @Gawdfrey's design without going into the accursed realm of designing for dark mode 🥴

githubmockup

Gawdfrey:
I took the liberty to implement these design suggestions in code as well, with @Exerra latest sketch in mind. I am not allowed to create a pull request, how could I submit the code for review? Fork the repo?

ndom91:
> I took the liberty to implement these design suggestions in code as well, with @Exerra latest sketch in mind. I am not allowed to create a pull request, how could I submit the code for review? Fork the repo?

Yeah so thats generally how it works, you make a fork, make your changes there then create a PR from the fork into the main one :+1:

Gawdfrey:
I added a new PR with the aforementioned designs [here](https://github.com/nextauthjs/next-auth/pull/5817). Should maybe just be a draft for now as it is also lacking documentation changes.

raulmarindev:
I like the new design. Thank you all for the work you do in this amazing library.

ndom91:
Last update to this PR, add darkmode button border back

image

ndom91:
I'd recommend move forward with this button logo + provider styling PR here. And then follow up shortly thereafter with another more general pages styling refresh, for example with [gawdfrey's PR](https://github.com/nextauthjs/next-auth/pull/5817).

balazsorban44 (approved):
This looks good to me! :shipit:

adrianbienias:
I think a little visual bug was introduced by this PR.

"Sign in" isn't vertically centered.

Screenshot from 2023-03-21 16-42-05

PR #5804, opened 2022-11-13
fix(next-auth): update docstring for jwt field in NextAuthOptions

Collatty:
## ☕️ Reasoning

The docs say that JWTs are encrypted by default:

image

While the docstring begs to differ:

image

From the looks of the default implementation of `encode` in [jwt](https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/jwt/index.ts), the docstring is incorrect and remnants of v3.

This led to me having to double check if the JWT was actually encrypted by default, so maybe this PR can save someone from doing the same digging.

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Found no related issues for this.

### DIFF
--- a/packages/next-auth/src/core/types.ts
+++ b/packages/next-auth/src/core/types.ts
@@ -58,10 +58,8 @@ export interface NextAuthOptions {
    */
   session?: Partial<SessionOptions>
   /**
-   * JSON Web Tokens are enabled by default if you have not specified a database.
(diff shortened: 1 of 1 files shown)


balazsorban44 (approved):
Correct! Thanks for catching this!

Issue #5616, opened 2022-10-22
Promisify normalizeIdentifier


shennan:
### Description 📓

Hi, I'd like to propose that we promisify `normalizeIdentifier` in the `EmailProvider`.

So...

```typescript
normalizeIdentifier?: (identifier: string) => Promise<string>;
```

My use case is that I have a large subscriber base of emails that have been invited to the site by others. That is, they've been invited by people that don't necessarily know the users preferred email address. In the context of a world where many emails are gmail, and these emails can interchange freely between `@gmail` or `@googlemail.com` endings, I find myself wanting to improve user experience by allowing both emails to be looked up in the database before a decision is made on authentication.

My first attempts were to hook into the `sendVerificationRequest()` action; something like this:

```typescript
sendVerificationRequest: async ({ identifier: email, url, provider: { server, from }, theme }) => {

    const emailExists = await checkUserEmailExistsInUsersTableAndAlsoCheckGoogleAlternate(email)

    if (!emailExists) throw new Error(`Nah`)

   /* ...Magic link sending logic */
}
```

But the problem here is that the resulting email that gets authenticated is still the original email that was typed into the login. This hook only has two options: succeed or fail by throwing. There is no way to alter the course of the email itself. And by not actually modifying the email to suit what is matched in the database; a new database entry is created for that email - and suddenly I have two user accounts which are for the same person.

So my thinking is that the best place to make a check like this is actually in `normalizeIdentifier`, as the email that is auth'd then becomes whatever is returned from this method. The issue, though, is that this method is synchronous and won't allow me to make a database check.

Perhaps there is a much better and more efficient way of handling the alternate email issue? If so I'm all ears. 👂 

### How to reproduce ☕️

N/A

### Contributing 🙌🏽

Yes, I am willing to help implement this feature in a PR

balazsorban44:
This sounds like a etter use case for the `signIn` callback https://next-auth.js.org/configuration/callbacks#sign-in-callback

shennan:
Fair enough, although I believed I passed on the `signIn` callback because it wasn't allowing me to _**only**_ authorise users that existed within the DB (i.e not allow new users to signIn) which is the primary need for me as my site is invite-only. But I made this decision when I was pretty new to NextAuth so it may be I need to revisit that callback.

PR #5812, opened 2022-11-14
Allow EmailAdapter.normalizeIdentifier() to be awaitable

mPaella:
## ☕️ Reasoning

Allows `EmailAdapter.normalizeIdentifer()` to be awaitable.  I don't see a reason why this shouldn't be awaitable? But the specific reason necessitating this is that we want to rate limit based on email, without changing `sendVerificationRequest`.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Relevant: https://github.com/nextauthjs/next-auth/issues/5616 https://github.com/nextauthjs/next-auth/discussions/687


balazsorban44:
Please see: https://github.com/nextauthjs/next-auth/issues/5616#issuecomment-1288032819

The callback name might be confusing, `authorized` would probably be more clear, it's already the case for `next-auth/middleware` and will likely converge in a future release: https://next-auth.js.org/configuration/nextjs#callbacks.

`normalizeIdentifier` is only meant for normalizing an e-mail.

mPaella:
I think theres a need for a function which is run after normalizeIdentifier, but before sending the email

PR #5823, opened 2022-11-15
fix: added function overload to getToken

JohnMarsden24:
## ☕️ Reasoning

Removes the need for `// @ts-expect-error` comments and also does a check if `secret` has not been set / passed to satisfy the required type for `_decode`.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

N/A

## 📌 Resources


packages/next-auth/src/jwt/index.ts
```diff
@@ -82,6 +85,8 @@ export async function getToken<R extends boolean = false>(
   } = params
 
   if (!req) throw new Error("Must pass `req` to JWT getToken()")
+  if (!secret)
+    throw new Error("Must pass `secret` if not set to JWT getToken()")
```
balazsorban44, line 89: Let's throw this instead: https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/core/errors.ts#L45

balazsorban44 (changes requested):
Thanks, small comment

packages/next-auth/src/jwt/index.ts
JohnMarsden24: Done in latest commit.

balazsorban44 (approved):

PR #5889, opened 2022-11-25
chore: update mongodb.md with typescript filetypes

Ojself:
The previous code would not work in a .js-file since it has type annotations. Either the type needs to be removed, or the .ts-file-ending needs to be added. This PR chose the latter 

- [x ] Documentation
- [x] Tests
- [x] Ready to be merged


ndom91 (approved):
Makes sense :+1:

PR #5991, opened 2022-12-08
fix(core): handle `Request` -> `Response` regressions 

### LINKED ISSUE #5953: host detection/`NEXTAUTH_URL` breaks in some cases
### Provider type

Google

### Environment

  System:
    OS: Linux 6.0 Arch Linux
    CPU: (16) x64 12th Gen Intel(R) Core(TM) i7-1260P
    Memory: 20.61 GB / 31.06 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 16.18.1 - ~/.nvm/versions/node/v16.18.1/bin/node
    npm: 8.19.2 - ~/.nvm/versions/node/v16.18.1/bin/npm
  Browsers:
    Chromium: 108.0.5359.94
    Firefox: 107.0.1
  npmPackages:
    next: 13.0.2 => 13.0.2
    next-auth: ^4.18.0 => 4.18.0
    react: 18.2.0 => 18.2.0

### Reproduction URL

https://next-auth-google-issue-repro.vercel.app/
https://github.com/mchataigner/next_auth_google_issue_repro

### Describe the issue

I have deployed on Vercel a simple app with next-auth and Google Oauth provider.
It works well locally but I get an error when trying to login in production.
It seems the error is coming from /api/auth/providers returning an empty object despite having multiple providers set in my configuration:
* Discord alone => works fine
* Google alone => error (empty providers object)
* Google + Discord => error (empty providers object)

I tried with next-auth 4.15.1 and 4.18.0 with no success
screenshot20221205_102909
Trying again this morning, and it worked a couple of times and then stopped working again.
It seems to work on cold start but not on hot ones.

### How to reproduce

Create simple app and add Google Oauth provider.

### Expected behavior

Adding Google provider should expose google login for the application or at least avoid crashing other providers.
screenshot20221205_101539

### LINKED ISSUE #5986: 4.18.1 causes app to render in a <pre> tag when using `unstable_getServerSession`
### Environment

  System:
    OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    Memory: 25.39 GB / 31.32 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 17.8.0 - ~/.nvm/versions/node/v17.8.0/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 8.5.5 - ~/.nvm/versions/node/v17.8.0/bin/npm
  npmPackages:
    next: 13.0.6 => 13.0.6
    next-auth: 4.18.1 => 4.18.1
    react: 18.2.0 => 18.2.0

### Reproduction URL

https://stackblitz.com/edit/github-61ga3k-prdgby?file=pages%2Fserver.tsx

### Describe the issue

After upgrading from 4.18.0 to 4.18.1 I'm seeing a very strange issue where my app no longer renders. Instead, Next renders the HTML of my app in a `<pre>` tag, so it looks like this:

image

What I've determined so far:

- 4.18.0 doesn't have this bug
- 4.18.1 and 4.18.3 have this bug
- The issue only happens when using `unstable_getServerSession`
- The issue only happens when you have logged in and have a session defined

### How to reproduce

Check out the Stackblitz link I created. The issue is somewhat reproducible there. The "server" page there has a `unstable_getServerSession` used in `getServerSideProps`.

To reproduce the bug:

- Run that Stackblitz example
- Go to the "Server" page in the UI
- Click on "Sign in" and try to sign in to one of the providers (doesn't mattern if you complete this or not)
- Press the browser Back button to go back to the app
- BUG: The UI no longer renders. Instead you get a `<pre>` tag with all the code of your app

### Expected behavior

Bug doesn't happen. Next Auth should render the app as normal.

### LINKED ISSUE #5989: Set-Cookie header is split in half by comma in Expires section
### Environment

System:
    OS: Linux 6.0 Arch Linux
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
    Memory: 10.86 GB / 31.10 GB
    Container: Yes
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.12.1/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v18.12.1/bin/npm
  Browsers:
    Chromium: 108.0.5359.94
    Firefox: 107.0.1
  npmPackages:
    next: 13.0.6 => 13.0.6 
    next-auth: 4.18.3 => 4.18.3 
    react: 18.2.0 => 18.2.0

### Intŕoduced by

https://github.com/nextauthjs/next-auth/commit/7e91d7df54f9d8ca782c93a63f583ce89c2d178d#r92403804

### Describe the issue

`set-cookie` header is split in Expires section containing comma resulting in incomplete cookie being rejected

image

### How to reproduce

call client `signIn` method

### Expected behavior

`set-cookie` header is not split in half in Expires section of the cookie

balazsorban44:
#4769 was a major internal refactoring that caused some bugs to surface:

- [x] `Set-Cookie` headers have been incorrectly split, which is now handled correctly (see: https://github.com/whatwg/fetch/issues/973). Fixes #5989
- [x] `unstable_getServerSession` overrode the `Content-Type` header, which broke the rendering of SSR pages (using `getServerSideProps`). Fixes #5986
~- [ ] Parsing the host header has introduced some unexpected behavior in production. #5953~ Needs more work, will follow up in a separate PR

### DIFF
--- a/packages/next-auth/src/next/index.ts
+++ b/packages/next-auth/src/next/index.ts
@@ -1,5 +1,5 @@
 import { AuthHandler } from "../core"
-import { getURL, getBody } from "../utils/node"
+import { getURL, getBody, setHeaders } from "../utils/node"
 
 import type {
   GetServerSidePropsContext,
@@ -37,10 +37,7 @@ async function NextAuthHandler(
   const { status, headers } = response
   res.status(status)
 
-  for (const [key, val] of headers.entries()) {
-    const value = key === "set-cookie" ? val.split(",") : val
-    res.setHeader(key, value)
-  }
+  setHeaders(headers, res)
 
   // If the request expects a return URL, send it as JSON
   // instead of doing an actual redirect.
@@ -157,6 +154,8 @@ export async function unstable_getServerSession<
   )
 
   const { status = 200, headers } = response
+
+  setHeaders(headers, res)
 
   // This would otherwise break rendering
   // with `getServerSideProps` that needs to always return HTML
--- a/packages/next-auth/src/next/index.ts
+++ b/packages/next-auth/src/next/index.ts
@@ -158,10 +158,9 @@ export async function unstable_getServerSession<
 
   const { status = 200, headers } = response
 
-  for (const [key, val] of headers.entries()) {
-    const value = key === "set-cookie" ? val.split(",") : val
-    res.setHeader(key, value)
-  }
+  // This would otherwise break rendering
+  // with `getServerSideProps` that needs to always return HTML
+  res.removeHeader?.("Content-Type")
 
   const data = await response.json()
 
--- a/packages/next-auth/src/utils/node.ts
+++ b/packages/next-auth/src/utils/node.ts
@@ -1,4 +1,4 @@
-import type { IncomingMessage } from "http"
+import type { IncomingMessage, ServerResponse } from "http"
 import type { GetServerSidePropsContext, NextApiRequest } from "next"
 
 export function setCookie(res, value: string) {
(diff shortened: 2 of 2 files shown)


ammmze:
I'm still having some issues with `unstable_getServerSession` after switching to `next-auth@0.0.0-pr.5991.5c72424c`. 

I get an error page with the error `FetchError: invalid json response body at  reason: Unexpected token E in JSON at position 0`.

To debug further, I modified the `unstable_getServerSession` function in `node_modules` adding:

ammmze:
> Edit 2: But it does look like the base path is likely the main trigger here. I jumped in the `AuthHanderInternal` and added a `console.log({method, action});` which resulted in `{ method: 'GET', action: 'auth' }`. So it looks like it is picking up the wrong action.

Looks like one option is to update these lines in the [getURL](https://github.com/nextauthjs/next-auth/blob/eddd8fd7f9da86bc3ceab3244a4d3cf7120064cb/packages/next-auth/src/utils/node.ts#L48-L51) method with the following to strip any path from the `host` variable (which in my case is coming from the `NEXTAUTH_URL`:

```js
    if (host.startsWith("http://") || host.startsWith("https://")) {
      return new URL(`${host.replace(/(https?:\/\/[^\/]+)\/?.*/, '$1')}${url}`)
    }
    return new URL(`https://${host.replace(/([^\/]+)\/?.*/, '$1')}${url}`)
```

ammmze:
> > Edit 2: But it does look like the base path is likely the main trigger here. I jumped in the `AuthHanderInternal` and added a `console.log({method, action});` which resulted in `{ method: 'GET', action: 'auth' }`. So it looks like it is picking up the wrong action.
> 
> Looks like one option is to update these lines in the [getURL](https://github.com/nextauthjs/next-auth/blob/eddd8fd7f9da86bc3ceab3244a4d3cf7120064cb/packages/next-auth/src/utils/node.ts#L48-L51) method with the following to strip any path from the `host` variable (which in my case is coming from the `NEXTAUTH_URL`:
> 
> ```js
>     if (host.startsWith("http://") || host.startsWith("https://")) {
>       return new URL(`${host.replace(/(https?:\/\/[^/]+)\/?.*/, '$1')}${url}`)
>     }
>     return new URL(`https://${host.replace(/([^/]+)\/?.*/, '$1')}${url}`)
> ```

Hmm...though another issue that crops up from stripping the path from the host is that anywhere that we call that `getURL` with the path we then lose our base path which can lead to some invalid redirects. So I think we also need some updates to that `toInternalRequest` method that was added that extracts things like the action, which in gist is effectively doing `action = url.pathname.split("/").slice(3)[0]`. This works fine without a base path, but then add the base path and we get the wrong action...

```
❯ node
Welcome to Node.js v16.14.2.
Type ".help" for more information.
> '/api/auth/signin'.split('/').slice(3)
[ 'signin' ]
> '/sso/api/auth/signin'.split('/').slice(3)
[ 'auth', 'signin' ]
```

ammmze:
> I think we also need some updates to that `toInternalRequest` method

Perhaps we should do something like this in [there](https://github.com/nextauthjs/next-auth/blob/eddd8fd7f9da86bc3ceab3244a4d3cf7120064cb/packages/next-auth/src/utils/web.ts#L48)? So basically first we take just the part of the path starting with the last `/api/auth` and then do the split and slice. This will give us the consistency we need in extracting the "nextauth" array.

```ts
const nextauth = url.pathname.substring(url.pathname.lastIndexOf("/api/auth")).split("/").slice(3);
```

balazsorban44:
The custom basepath handling is on my todo list, it's not checked off in the description as you can see in https://github.com/nextauthjs/next-auth/pull/5991#issue-1484873050 :+1: Thanks!

balazsorban44:
I'll merge this PR for now to fix the two other issues, I did not manage to fix this today, postponing the custom base path fix tomorrow, will follow up in a separate PR.

ammmze:
> I'll merge this PR for now to fix the two other issues, I did not manage to fix this today, postponing the custom base path fix tomorrow, will follow up in a separate PR.

Sounds good! Thank you!

noclat:
Thanks guys, literally spent the complete past day just trying to debug CredentialsProvider (jwt ofc) not creating session cookie in prod and had absolutely no clue why, and woke up to this fix. At least I know the full NextAuth doc by heart now.

PR #6004, opened 2022-12-09
fix(core): correct status code when returning redirects

### LINKED ISSUE #6001: 4.18.4 no longer respects `signIn(..., {redirect: false})`
### Environment

  System:
    OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    Memory: 24.48 GB / 31.32 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 17.8.0 - ~/.nvm/versions/node/v17.8.0/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 8.5.5 - ~/.nvm/versions/node/v17.8.0/bin/npm
  npmPackages:
    next: 13.0.6 => 13.0.6
    next-auth: 4.18.0 => 4.18.0
    react: 18.2.0 => 18.2.0

### Reproduction URL

https://stackblitz.com/edit/github-61ga3k-pvybxx?file=pages%2Findex.tsx,pages%2Fapi%2Fauth%2F[...nextauth].ts

### Describe the issue

After upgrading from 4.18.0 to 4.18.4 I noticed that `signIn(..., {redirect: false})` is no longer respecting the `redirect: false` option.

In my app I have a `CredentialsProvider` and I issue a signIn request like this:

```js
const result = await signIn('myapp-login', {
    {username, password},
    redirect: false,
});
```

In 4.18.0, after a successful login, it this would return:

```js
{
    error: null,
    ok: true,
    status: 200,
    url: "http://localhost:3000/login"
}
```

But in 4.18.4 it returns:

```js
{
    error: null,
    ok: false,
    status: 302,
    url: "http://localhost:3000/login"
}
```

### How to reproduce

Create an example app with `next-auth@4.18.4` and update the `signIn()` call in the header to use the `redirect: false` option. Proceed to login and check the response of the `signIn()` request.

### Expected behavior

The `signIn()` call should resolve with an `ok: true` object upon successful login.

balazsorban44:
Continues #5991, Fixes #6001

The refactor in #4769 changes how we set status codes in the Node.js environment, and it set `302` instead of `200` even if the redirection was turned into a JSON response.

In this PR, I changed the way we handle redirection by relying on a `X-Auth-Return-Redirect` header rather than a `json: true` property in the request body.

### DIFF
--- a/packages/next-auth/src/core/index.ts
+++ b/packages/next-auth/src/core/index.ts
@@ -277,9 +277,12 @@ export async function AuthHandler(
   // If the request expects a return URL, send it as JSON
   // instead of doing an actual redirect.
   const redirect = response.headers.get("Location")
-  if (request.headers.get("X-Auth-Return-Redirect") && redirect) {
+  if (request.headers.has("X-Auth-Return-Redirect") && redirect) {
+    response.headers.delete("Location")
+    response.headers.set("Content-Type", "application/json")
     return new Response(JSON.stringify({ url: redirect }), {
-      headers: { "Content-Type": "application/json" },
+      headers: response.headers,
+      status: 200,
     })
   }
   return response
--- a/packages/next-auth/src/core/index.ts
+++ b/packages/next-auth/src/core/index.ts
@@ -271,5 +271,16 @@ export async function AuthHandler(
 ): Promise<Response> {
   const req = await toInternalRequest(request)
   const internalResponse = await AuthHandlerInternal({ req, options })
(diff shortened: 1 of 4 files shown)


EvHaus:
Confirming that `next-auth@0.0.0-pr.6004.b96e4fbf` fixes the issue. Thank you!

ThangHuuVu (approved):

Issue #5101, opened 2022-08-05
Navigating back after successful Azure AD B2C sign on throws OAuthCallback error for 'state' and 'pkce' check types


jcbtheo:
### Provider type

Azure Active Directory B2C

### Environment

```
System:
    OS: Windows 10 10.0.22000
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
    Memory: 6.21 GB / 15.77 GB
  Binaries:
    Node: 16.13.1 - C:\Program Files\nodejs\node.EXE
    npm: 8.1.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22000.120.0), Chromium (103.0.1264.77)
    Internet Explorer: 11.0.22000.120
  npmPackages:
    next: 12.2.4 => 12.2.4
    next-auth: ^4.10.3 => 4.10.3
    react: 18.2.0 => 18.2.0
```

### Reproduction URL

https://github.com/jcbtheo/azure-ad-b2c-issue

### Describe the issue

Using the out of the box Azure AD B2C provider, using the default provider `check` value to `'state'` or `'pkce'` will allow the user to login successfully. However, once the login is resolved, immediately navigating back in the browser will cause an OAuthCallback error. 

The root of the issue seems to be that after logging in, the Azure AD B2C authorize URL is still in the browser history, with the previously used `state` or `pkce` query params in the URL, and next auth is unable (or does not provide a way) to handle the resulting error. 

When `'state'` is selected, the error is 
```
[next-auth][error][OAUTH_CALLBACK_ERROR] 
https://next-auth.js.org/errors#oauth_callback_error checks.state argument is missing {
  error: {
    message: 'checks.state argument is missing',
    stack: 'TypeError: checks.state argument is missing\n' +
      '    at Client.callback (C:\\Dev\\AzureB2cNextIssue\\azure-ad-b2c-issue\\node_modules\\openid-client\\lib\\client.js:385:13)\n' +
      '    at oAuthCallback (C:\\Dev\\AzureB2cNextIssue\\azure-ad-b2c-issue\\node_modules\\next-auth\\core\\lib\\oauth\\callback.js:114:29)\n' +
      '    at runMicrotasks (<anonymous>)\n' +
      '    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n' +
[...]

PR #6013, opened 2022-12-09
Add Beyond Identity Provider

Will5:
## ☕️ Reasoning

This pull request adds a BeyondIdentityProvider and the corresponding documentation which allows authenticating with [Beyond Identity](https://www.beyondidentity.com/).

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged


Will5:
This originally only contained an "oauth" provider, but I added an "oidc" provider as well. This required a fix for the next-auth.pkce.code_verifier cookie, which should also address Issue #5101 as well.

Will5:
I merged in the latest master, resolved the latest merge conflicts, and retested via the example app. Is there anything else required? Thanks!

packages/core/src/lib/oauth/pkce-handler.ts
```diff
@@ -80,7 +80,7 @@ export async function usePKCECodeVerifier(
     codeVerifier: pkce?.value ?? undefined,
     cookie: {
       name: cookies.pkceCodeVerifier.name,
-      value: "",
+      value: pkce?.value ?? undefined,
```
ThangHuuVu, line 83: This should not be changed, did you run into an issue? if so, please open a separate PR for this

ThangHuuVu (changes requested):
Hi there @Will5 thanks for sending the PR! To keep it up to date and ready to merge, here are a few things that need to be done:
- Revert all changes in `packages/next-auth` and move them to `packages/core`
- Don't use `.md` doc, instead use JSDoc like https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/providers/auth0.ts
- Add the provider to the drop down at https://github.com/nextauthjs/next-auth/blob/main/.github/ISSUE_TEMPLATE/2_bug_provider.yml
LMK if you need any help 🙌

packages/core/src/lib/oauth/pkce-handler.ts
Will5: Yes, I don't remember the details, but PKCE was not working correctly without it. If I recall correctly, the code was empty in the token exchange, which seemed to be because the cookie was empty. However, I tried just now with the latest, and I did not encounter an issue. I will revert, and open an issue if I come across it again.

Will5: Oh, I think I encountered the issue again. The pkce challenge was not validated. It seems to be related to checking both "pkce" and "state". I did not encounter the issue when I just checked "pkce". That being said, I don't think this is the most appropriate solution.

Will5:
Hi @ThangHuuVu, just checking in. Is there anything else needed from me? Would love to get this in. Thanks!

ThangHuuVu:
@Will5 sorry for the late response, could you update the branch before I merge it? All LGTM 🙌

Will5:
Hi @ThangHuuVu, done. Let me know if you need anything else. Thanks 🙌

ThangHuuVu (approved):

packages/core/src/providers/beyondidentity.ts
```diff
@@ -0,0 +1,99 @@
...
+): OIDCConfig<BeyondIdentityProfile> {
+  return {
+    id: "beyondidentity",
+    name: "Beyond Identity",
+    type: "oidc",
+    profile(profile) {
+      return {
+        id: profile.sub,
+        name: profile.name,
+        preferred_username: profile.preferred_username,
```
balazsorban44, line 84:
```suggestion
        image: null,
```

packages/core/src/providers/beyondidentity.ts
```diff
@@ -0,0 +1,99 @@
...
+        email: profile.email,
+      }
+    },
+    checks: ["pkce"],
```
balazsorban44, line 88:
```suggestion
```

ThangHuuVu:
Can you attach a video for proof? @Will5 🙌

Will5:
@ThangHuuVu, below is a video using the dev app, though it seems the header remains in signed out state. This is the same when I use Auth0 as well, so I suspect it's a dev app bug. Should I submit a bug for this?

https://user-images.githubusercontent.com/4763133/220440859-7fe884ff-dc83-496f-b40d-17db5fa3ca8d.mov

ThangHuuVu:
@Will5 I see, we are actually not sure if it's a bug - it only happens on the local development. If you un-comment this line, it should work: https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/lib/web.ts#L81

Will5:
@ThangHuuVu Thanks for the tip! That worked for me 🙂 


https://user-images.githubusercontent.com/4763133/220812702-16b54628-c148-43f7-a8bd-320bacba479f.mov

Will5:
Hi @ThangHuuVu Thanks for your help! I notice the provider isn't showing up in the docs. Does it need to wait for the next release? If so, is there an expected time frame? Thanks!

Will5:
Never mind, I see it now! 🎉

PR #6066, opened 2022-12-15
fix(sveltekit): correctly reference SvelteKit as peer dependency

theetherGit:
You added wrong peer dependency svelte-kit@^1.0.0  instead of @sveltejs/kit": "^1.0.0


balazsorban44 (approved):
Thanks!

PR #6111, opened 2022-12-18
chore: clean up dev deps

balazsorban44:
The goal of this PR is:

- have `.eslintrc.js` , `.eslintignore` , `.prettierrc` , `.prettierignore` in the root
- none of the above in `packages/*` or `docs`
- have a single `tsconfig.json` in `packages/*` instead of 3 (no `tsconfig.dev.json`, `tsconfig.eslint.json` etc, just a single `tsconfig.json` for each package)
- `apps/example-*` should manage their own ESLint, Prettier stuff (if at all, my opinion is to remove it TBH), as they are being used as standalone templates like [this](https://github.com/nextauthjs/next-auth-example).
- any `packages/*` wanting formatting or linting should be able to add `"lint": "eslint"` and `"format": "prettier"` respectively in their `package.json`. No flags should be. If for some reason a package would require extra plugins, handle that in the root configs with overrides. It should be the exception though, we want the same style for everything.

Feel free to work on this by pushing commits directly.


balazsorban44:
Done directly on `main`... :muscle: <picture data-single-emoji=":hackerman:" title=":hackerman:">:hackerman:</picture>

PR #6115, opened 2022-12-18
docs: /reference sidebar improvements

balazsorban44:
Check out https://auth-docs-git-chore-reference-sidebar-authjs.vercel.app/reference/core/modules

**Before:**
image

**After:**
image


ndom91:
Small nit, the `Reference` index page and the "experimental" emojis in the sidebar are a bit confusing.

I know we're tryign to say that the `@auth/nextjs` package itself is experimental int he sidebar, and with the index page saying that Next.js support generally (via the "old" package) is stable. But an average user isn't going to know that..

image

ndom91 (approved):

PR #6148, opened 2022-12-22
Update Neo4j Adapter docs in order to allow compatibility with Neo4j v5

pedsm:
## ☕️ Reasoning

Neo4j v5 has been out for a month now and from my limited testing, Next-Auth seems compatible with it. The only breaking change is the syntax used for constraints. This PR will update the docs so that users can use Neo4j v5 as their DB.

Neo4j v5 dropped support for the following syntax

```cypher
CREATE CONSTRAINT ON ... ASSERT ...
// Replaced by:
CREATE CONSTRAINT FOR ... REQUIRE ...
```

[Source](https://neo4j.com/docs/cypher-manual/current/deprecations-additions-removals-compatibility/#_removed_features)

## 🧢 Checklist

- [x] Documentation
- Tests (Docs only)
- [x] Ready to be merged

## 📌 Resources

### DIFF
--- a/docs/docs/reference/06-adapters/neo4j.md
+++ b/docs/docs/reference/06-adapters/neo4j.md
@@ -1,3 +1,38 @@
+---
+id: neo4j
+title: Neo4j
+---
+
+This is the Neo4j Adapter for [`next-auth`](https://authjs.dev). This package can only be used in conjunction with the primary `next-auth` package. It is not a standalone package.
+
+## Getting Started
+
+1. Install the necessary packages
+
+```bash npm2yarn
+npm install next-auth @next-auth/neo4j-adapter neo4j-driver
+```
+
+2. Add this adapter to your `pages/api/[...nextauth].js` next-auth configuration object.
+
+```javascript title="pages/api/auth/[...nextauth].js"
+import neo4j from "neo4j-driver"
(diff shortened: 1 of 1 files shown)


pedsm:
@EarthlingDavey I saw in the other thread you are using this adapter in production. Any thoughts on this?

EarthlingDavey:
@pedsm I appreciate the mention, and of course the update. I'm still on v4 in prod btw.

I haven't run the commands for v5 here, but they look good to me. If you've run them on your v5 db, and they work, then I have no probs with the merge 👍

balazsorban44 (changes requested):
Should we add v5 to peer dependencies too?:

https://github.com/nextauthjs/next-auth/blob/54561a1231c3f9576b2a787f1ce4cf9d76ae179c/packages/adapter-neo4j/package.json#L36

I am fuzzy about what is breaking. Will the new syntax not work for v4 users? Can we have support for both v4 and v5, or do you think we should release a major and bump the peer dependency instead of adding v5 next to v4?

pedsm:
@balazsorban44 you do have a good point, this is probably backwards compatible with v4 but I didn't test it, will give it a spin at some point this week.

balazsorban44:
@pedsm were you able to figure out if this is backward compatible?

balazsorban44:
Closing in favor of #7392

PR #6212, opened 2022-12-28
Update typescript.md

mongolyy:
Remove unused import statement.

> _NOTE_:
>

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

None

## 📌 Resources

### DIFF
--- a/docs/docs/getting-started/typescript.md
+++ b/docs/docs/getting-started/typescript.md
@@ -75,8 +75,6 @@ export default function IndexPage() {
 To extend/augment this type, create a `types/next-auth.d.ts` file in your project:
 
 ```ts title="types/next-auth.d.ts"
-import NextAuth from "next-auth"
-
 declare module "next-auth" {
   /**
    * Returned by `useSession`, `getSession` and received as a prop on the `SessionProvider` React Context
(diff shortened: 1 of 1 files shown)


balazsorban44 (commented):
Not sure if this was a bug in earlier TS versions, but when I wrote that documentation, the augmentation only worked with those imports. 🤔 Can you verify it's not the case anymore?

mongolyy:
@balazsorban44 
Thank you for your comment.

> Can you verify it's not the case anymore?

I'm trying to verify with [this example](https://github.com/nextauthjs/next-auth/tree/main/apps/example-nextjs).
It seems that the behavior of requiring/not requiring import is different from my other project, perhaps due to different transpile options in TypeScript.
I will look into this a bit more.

balazsorban44:
Closing as seems to be stale. Happy to re-open if it's justified.

mongolyy:
Thanks for the close.
I tried using the example and could not find a situation where import would not be needed.

TypeScript documentation also describes import, so I think we should leave the current description as it is.
https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation

PR #5055, opened 2022-07-31
fix(firebase): add ability to pass in firebase config

chanceaclark:
## ☕️ Reasoning

Adds the ability to pass in Firebase config into the adapter to initialize the app, rather than initializing by default. This is useful if you've initialized the app on the client and want to reuse the config on the server. Alongside that, adds some documentation on how to use it.

_Note: Ran `prettier --write packages/adapter-firebase` which cleaned up some files._

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: #4927

## 📌 Resources

- [Contributing guidelines](./CONTRIBUTING.md)
- [Code of conduct](./CODE_OF_CONDUCT.md)
- [Contributing to Open Source](https://kcd.im/pull-request)

PR #5628, opened 2022-10-24
Using Firebase Admin SDK

nhuethmayr:
New adapter that uses the Firebase Admin instead of the client-side SDK

## ☕️ Reasoning

The current implementation of firebase-adapter uses the client SDK. This has the severe drawback that the Firestore Rules need to be opened up. Instead of using the client-side SDK, we should be using the server-side one. That way we can keep the Firestore rules closed and still integrate nicely with FB

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

This PR solves an architectural problem with the firebase-adapter as descibed in discussion [5049](https://github.com/nextauthjs/next-auth/issues/5049)

## 📌 Resources

- [Contributing guidelines](./CONTRIBUTING.md)
- [Code of conduct](./CODE_OF_CONDUCT.md)
- [Contributing to Open Source](https://kcd.im/pull-request)

## ❓ Questions 

- Do we want to release this code as an independent adapter? `firebase-admin-adapter`
- Do we want to replace the current implementation altogether?

Issue #4075, opened 2022-02-25
    CredentialsProvider session token cookie not updated on getServerSideProps


killjoy2013:
### Environment

  System:
    OS: Windows 10 10.0.19042
    CPU: (8) x64 Intel(R) Xeon(R) CPU E3-1505M v6 @ 3.00GHz
    Memory: 5.44 GB / 15.86 GB
  Binaries:
    Node: 14.17.2 - C:\Program Files\nodejs\node.EXE       
    Yarn: 1.22.4 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.14.13 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0)
    Internet Explorer: 11.0.19041.1202
  npmPackages:
    next: ~12.1.0 => 12.1.0 
    next-auth: ~4.2.1 => 4.2.1 
    react: 17.0.2 => 17.0.2 

### Reproduction URL

https://github.com/killjoy2013/nextauth-credential-rotate

### Describe the issue

Hi,
Using CredentialsProvider and need to rotate the token.  My [...nextauth.ts] is below;

```js
import NextAuth from 'next-auth';
import { JWT, JWTEncodeParams, JWTDecodeParams } from 'next-auth/jwt';
import CredentialsProvider from 'next-auth/providers/credentials';
import jsonwebtoken from 'jsonwebtoken';

function createToken(username: string) {
  return {
    username,
    willExpire: Date.now() + parseInt(process.env.TOKEN_REFRESH_PERIOD) * 1000,
  };
}

function refreshToken(token) {
  const { iat, exp, ...others } = token;

  return {
    ...others,
    willExpire: Date.now() + parseInt(process.env.TOKEN_REFRESH_PERIOD) * 1000,
  };
}

export default NextAuth({
  secret: process.env.TOKEN_SECRET,
  jwt: {
    secret: process.env.TOKEN_SECRET,
    maxAge: parseInt(process.env.TOKEN_MAX_AGE),
    encode: async (params: JWTEncodeParams): Promise<string> => {
      const { secret, token } = params;
      let encodedToken = '';
      if (token) {
        const jwtClaims = {
          username: token.username,
          willExpire: token.willExpire,
        };

        encodedToken = jsonwebtoken.sign(jwtClaims, secret, {
          expiresIn: parseInt(process.env.TOKEN_REFRESH_PERIOD),
          algorithm: 'HS512',
[...]

PR #6225, opened 2022-12-30
feat(adapters): move to `firebase-admin` in Firebase Adapter

### LINKED ISSUE #4927: Firebase Adapter Fails with App Name Already exist error

### LINKED ISSUE #5049: [firebase adapter] - Firebase adapter requires open security rules we cannot allow in production

wyattades:
## ☕️ Reasoning

This PR adds a new adapter for the Firebase Admin SDK. The current Firebase `next-auth` adapter (`@next-auth/firebase-adapter`) uses the *frontend* Firebase SDK, and should not be used because it requires developers to loosen their Firebase security rules to accommodate the adapter. (For that reason, I would recommend deprecating `@next-auth/firebase-adapter` if/when this new adapter is stable).

This adapter is inspired by @NorbertHuethmay's PR https://github.com/nextauthjs/next-auth/pull/5628 (which for some reason was never merged?) but also adds: tests, more use of firestore transactions, and a `preferSnakeCase` option.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes:
fixes https://github.com/nextauthjs/next-auth/issues/5049, closes #6230, closes #5449, closes #5055, fixes #5049, fixes #4927


balazsorban44 (commented):
Thanks for the PR! I am going through it right now!

I think we can just release it as a major bump of the existing adapter, so I made some changes. I'll test it, fix up the docs, and it will probably be good to merge! Nice work!

balazsorban44:
@wyattades So I also incorporated the changes from #6230 here.

Regardless, trying to initialize this adapter in the dev app, I am getting:

> '7 PERMISSION_DENIED: Permission denied on resource project authjs-firebase-adapter.'

I used the https://console.firebase.google.com/u/2/project/project-id/settings/serviceaccounts/adminsdk page to generate a service account, and then point to the JSON file via `GOOGLE_APPLICATION_CREDENTIALS=./firebase-admin-credentials.json`.

Do you have any idea what's happening?

packages/adapter-firebase/src/index.ts
```diff
@@ -1,282 +1,236 @@
...
+ *
+ * ##### Example
+ *
+ * ```ts title="pages/api/auth/[...nextauth].ts"
+ * import NextAuth from "next-auth"
+ * import GoogleProvider from "next-auth/providers/google"
+ * import { FirestoreAdapter } from "@next-auth/firebase-adapter"
+ *
+ * export default NextAuth({
+ *   adapter: FirestoreAdapter(),
```
wyattades, line 67: Even if it's a bit more verbose, I still prefer forcing the user to pass a firestore instance themselves. I can only speak from personal experience, but I usually end up using the instance somewhere else on the backend. Also, I think it removes some guessing on how the credentials and appconfig are used to initialize the app. What do you think?

packages/adapter-firebase/src/utils.ts
```diff
@@ -0,0 +1,147 @@
+import { type AppOptions, getApps, initializeApp } from "firebase-admin/app"
+import {
+  getFirestore,
+  initializeFirestore,
+  Timestamp,
+} from "firebase-admin/firestore"
```
wyattades, line 6: I think this import pattern breaks compatibility with firebase-admin v10. Not a big deal since we label the peer dependency as v11, but it might be a nice-to-have to support v10 as well?

wyattades (commented):
Thanks for cleaning this up! I left a few final suggestions

packages/adapter-firebase/src/utils.ts
balazsorban44: Since this is going to be a major bump, I would prefer we start only supporting the latest of the package. It also helps a lot with tree-shaking to decrease the size.

Based on https://firebase.google.com/support/release-notes/admin/node#version_1100_-_16_june_2022 it's also better to just drop Node 12 support.

packages/adapter-firebase/src/index.ts
balazsorban44: Could support both, but it was something many people wanted. I think it depends - as you say - if you want to use the instance somewhere else. Will follow up.

mariselli:
Looking forward this PR! 🙏🏻

Furkan-rgb:
If this gets merged, does that mean the next auth Firebase adapter uses the admin SDK, meaning we can implement Firestore security rules?

smashah:
@Furkan-rgb yes since admin sdk bypasses security rules.

Looking forward to seeing this merged.

This opens up the path of using the admin SDK to optionally [`createUser`](https://firebase.google.com/docs/reference/admin/node/firebase-admin.auth.baseauth.md#baseauthcreateuser) on the firebase Auth level instead of just in the firestore DB. This will allow for greater flexibility.

packages/adapter-firebase/src/index.ts
wyattades: Thanks for handling both :)

balazsorban44:
For anyone following, what's left here is the docs. We would like to create the API reference from the source code and I'm trying to work out the preferred format.

packages/adapter-firebase/package.json
```diff
@@ -12,35 +12,43 @@
...
+  "exports": {
+    ".": {
+      "types": "./index.d.ts",
+      "import": "./index.js"
+    }
+  },
   "files": [
-    "dist",
-    "index.d.ts"
+    "*.js",
```
ThangHuuVu, line 23: This will include the config files like `jest.config.js` so I think it's best to be specific about what files
```suggestion
    "index.js",
```
Could you add the readme.md file here as well? 🙌

ThangHuuVu (changes requested):
Nice PR, I suggest one change regarding the NPM publish files 🙌

packages/adapter-firebase/package.json
balazsorban44: :+1: FWIW, the readme is always included: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files:~:text=Certain%20files%20are%20always%20included

Tweaked a bit, `npm pack` now shows:
image

ThangHuuVu (approved):
Let's ship it!

balazsorban44:
This is now available under [`@next-auth/firebase-adapter@2.0.0`](https://github.com/nextauthjs/next-auth/releases/tag/%40next-auth%2Ffirebase-adapter%402.0.0)

Docs at https://authjs.dev/reference/adapter/firebase

Thanks for everyone following along! :tada: Check it out, and give us feedback/reports bugs (in issues) if you find any. :pray:

Furkan-rgb:
Can't install it with yarn yet!

NorbertHuethmayr:
> If this gets merged, does that mean the next auth Firebase adapter uses the admin SDK, meaning we can implement Firestore security rules?

This means that you won't need security rules cause the firebase adapter runs on the server and has full access to the DB.

Matgsan:
Not in npm repo yet?

dbeaudway:
Migrating to the new adapter is giving me this error:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../node_modules/@next-auth/firebase-adapter/utils' imported from .../node_modules/@next-auth/firebase-adapter/index.js
```

Firebase-admin: 11.5.0
Next-auth: 4.19.2
@next-auth/firebase-adapter: 2.0.0

I followed the documentation instructions and used the provided `initFirestore` function. Anyone else experiencing this?

darrencarlin:
@dbeaudway Yes, I'm facing the same issue. If you import it from node modules using a relative path it should work for now. 

Adapter works great other than that.

RisKakaN:
Have anyone tested this with Nextjs 13? I followed the instructions from https://authjs.dev/reference/adapter/firebase, using the credentials cert method.
But I am getting Invalid Compact JWE error. See logs.

So when I sign in, it does create sessions, users, accounts, in firestore. but i get this error, and when I call getServerSession(), it is null.

`[next-auth][error][JWT_SESSION_ERROR]
https://next-auth.js.org/errors#jwt_session_error Invalid Compact JWE {
  message: 'Invalid Compact JWE',
  stack: 'JWEInvalid: Invalid Compact JWE\n' +
    '    at compactDecrypt (webpack-internal:///(sc_server)/./node_modules/jose/dist/node/esm/jwe/compact/decrypt.js:20:15)\n' +
    '    at jwtDecrypt (webpack-internal:///(sc_server)/./node_modules/jose/dist/node/esm/jwt/decrypt.js:12:100)\n' +
    '    at Object.decode (webpack-internal:///(sc_server)/./node_modules/next-auth/jwt/index.js:44:53)\n' +
    '    at async Object.session (webpack-internal:///(sc_server)/./node_modules/next-auth/core/routes/session.js:25:34)\n' +
    '    at async AuthHandler (webpack-internal:///(sc_server)/./node_modules/next-auth/core/index.js:157:37)\n' +
    '    at async getServerSession (webpack-internal:///(sc_server)/./node_modules/next-auth/next/index.js:86:21)\n' +
    '    at async RootLayout (webpack-internal:///(sc_server)/./src/app/layout.tsx:16:21)',
  name: 'JWEInvalid'
}`

**Specifics:**
next: 13.1.6
next-auth: 4.19.2
firebase-admin: 11.5.0
@next-auth/firebase-adapter: 2.0.0
Provider: Google

darrencarlin:
@RisKakaN Have you added a JWT env variable?

RisKakaN:
@darrencarlin no i have not. is that required, if so, how do i add one? 
I have only added the config from the instructions.
then i also have NEXTAUTH_SECRET, generated with: `openssl rand -base64 32`.

if i look in the browser cookies, the session token is a uuid. not sure if that is an issue.

do you have a better example/guide i can follow other than what is in authjs. in case i am missing some basic steps.

darrencarlin:
Sorry that's what I was referring to, the `NEXTAUTH_SECRET`

Here is my simple repo with a working example using google login https://github.com/darrencarlin/auth 

If that doesn't help feel free to upload your code and I can take a look

RisKakaN:
thanks for the example, helps a lot!

I tried your repo with my firebase config and app, and it works fine! so then there should be no issues with my google provider or firebase.

the only difference i see is that i am using the new nextjs 13 app dir, and you are not.
i am also not getting the ERR_MODULE_NOT_FOUND mentioned above, even though i am on exact same versions.

so maybe there is an issue with using the new app dir.

i am also using middleware, but i dont see how that would be an issue. still getting error when i remove it.

feel free to check out my test branch here. just setup your own env.local file with the necessary variables.
my expected scenario is to not get the jwe error, as well as being able to be directed to a logged in state in my app. right now, im just stuck in the login page.

https://github.com/RisKakaN/food-recipe-manager/tree/test-1

darrencarlin:
@RisKakaN I was able to solve the issue by following this comment (I didn't add ctx) https://github.com/nextauthjs/next-auth/issues/4075#issuecomment-1186252347 

Gives a console warning leading to this page https://next-auth.js.org/configuration/nextjs#getServerSession

```
[next-auth][warn][EXPERIMENTAL_API] 
https://next-auth.js.org/configuration/nextjs#getServerSession} 
https://next-auth.js.org/warnings#EXPERIMENTAL_API
```

But I am able to login without the error previous error.

I think it's important to note this issue isn't related to the firebase adapter.

RisKakaN:
Huge thanks @darrencarlin !
I totally missed that you had to pass in the auth config to getServerSession. Because it was not needed when not using the adapter.

however, now i am running into the next issue :D
have you tried this with a middleware, which blocks pages if you are not authenticated?
the issue is, even if i am logged in, and have a session, the middleware says that i am not authenticated. so i cant navigate to my protected pages.

in my middleware, i am using withAuth, so that i can print out what is happening in the authorized callback. the token there is null, therefore not allowing me to go to any protected pages.

i think i read somewhere that this middleware currently does not support anything other that jwt session strategy, whereas the adapter uses a database session strategy.

would you know any workaround/solution to this? 

otherwise i would have to do manual checks if session exists. the neat thing with middleware is that it handles it all for you.

RisKakaN:
i ended up creating my own middleware logic. in the middleware withAuth authorized callback, i call an internal api endpoint, which in turn calls getSession to verify whether we have a session or not. 
and then in my middleware, i define some protected routes.

the downside is that there is an extra step with calling an endpoint which calls getSession.
but the use of getSession here should be alright. and i am not aware of any other downsides or risks.

gopherine:
@dbeaudway @darrencarlin  how did you resolve cannot find module error i am still getting it 

import { FirestoreAdapter } from "../../../node_modules/@next-auth/firebase-adapter";

tried to import it like this ???

PR #6251, opened 2023-01-01
feat(adapters): add SurrealDB adapter

martinschaer:
## ☕️ Reasoning

Adding a SurrealDB adapter.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: INSERT_ISSUE_LINK_HERE


Sacramentix:
I'm gonna test this today, thanks for the contribution. :)

martinschaer:
@Sacramentix thanks! I added it to a side-project hosted on vercel (nextjs) with SurrealDB on AWS Fargate + EFS. It works well for a 1 minute, then I get logged out… If I try to log in again I get this error "Try signing in with a different account". If I try after a while it works back again. Running it locally works fine. I haven't had the time to debug this, hopefully today or tomorrow.

This PR is in draft since I haven't added the documentation in the README.

packages/adapter-surrealdb/src/index.ts
```diff
@@ -0,0 +1,336 @@
...
+    async createUser(data) {
+      const surreal = await client
+      const doc = userToDoc(data)
+      const user = (await surreal.create("user", doc)) as UserDoc
+      return docToUser(user)
+    },
+    async getUser(id: string) {
+      const surreal = await client
+      try {
+        const users = (await surreal.select(`user:${id}`)) as UserDoc[]
```
cameronbraid, line 101: use query variables

hgoona:
Hey team! I don't know to formally contribute to these but I just managed to get the above adaptor working with a the new AuthJS for Sveltekit - with some minor tweaks, setup in the "Sveltekit" way.

Took a long time to figure it out so I documented it here (along with my noob struggles/errors): https://gist.github.com/hgoona/ad5d6fcfa446a7a56b29a977861747eb

Hope the code structure can be useful to other Sveltekit / AuthJs / SurrealDB users! 💪🏾

^consider my above as a "test" to vouch for @martinschaer work👏🏾👏🏾👏🏾

packages/adapter-surrealdb/src/index.ts
alexgorbatchev: `surreal.select` is a syntactic sugar for `surreal.query`.

https://surrealdb.com/docs/integration/libraries/nodejs#select

leolorenzoluis:
@martinschaer Is there no way for it not to be dependent on your other lib? Why use your own lib instead of the official SDK?

martinschaer:
> @martinschaer Is there no way for it not to be dependent on your other lib? Why use your own lib instead of the official SDK?

@leolorenzoluis the current SDK uses an RPC connection, so it is not compatible with a stateless server (live vercel)

I haven't had the chance to either find another solution or try to make my package part of the official SDK. Any recommendation? I'm planning to continue working on this soon

leolorenzoluis:
@martinschaer That’s a fair point. However, not everyone wants to deploy a stateless server or use REST. Also, surrealdb has a rest endpoint https://surrealdb.com/docs/integration/http

I also tested your lib but it doesn’t look like it’s properly passing the session token when `session` is called? It is persisting the session token and creating user in surrealdb. Do you have a complete configuration in the docs compared to your test files?

martinschaer:
> @martinschaer That’s a fair point. However, not everyone wants to deploy a stateless server or use REST. Also, surrealdb has a rest endpoint https://surrealdb.com/docs/integration/http
> 
> I also tested your lib but it doesn’t look like it’s properly passing the session token when `session` is called? It is persisting the session token and creating user in surrealdb. Do you have a complete configuration in the docs compared to your test files?
> 
> Update: Not sure if you'll ever get a chance since it looks like it's hard coded here
> 
> https://github.com/nextauthjs/next-auth/blob/36286b1fae4f07fa8c204e88aefb0876479e8b03/packages/core/src/lib/routes/session.ts#L127
> I'm sure there's a way to bypass this.

hi @leolorenzoluis , I created 2 separate tests, one using the official node.js library from surreal (`surrealdb.js`), and the other with my rest library (`surrealdb-rest-ts`). So people can choose either option.

My library is based on the rest endpoint you mentioned. I think `surrealdb.js` should include the option to use the rest endpoint instead of RPC, that way my library could be removed.

There is an issue related to that on their repo https://github.com/surrealdb/surrealdb.js/issues/61 , but it's closed :(

packages/adapter-surrealdb/src/index.ts
```diff
@@ -0,0 +1,364 @@
...
+  refresh_token?: string
+  access_token?: string
+  type: ProviderType
+  provider: string
+  providerAccountId: string
+  expires_at?: number
+}
+export type SessionDoc<T = string> = Document & { userId: T }
+
+const extractId = (surrealId: string) => surrealId.split(":")[1] ?? surrealId
```
leolorenzoluis, line 25: @martinschaer Any reason why you prefer to extract the id? This makes the query on the server side having to append the pre-split values.

martinschaer: @leolorenzoluis just to hide implementation details for security, so that if a user sees an ID its not that obvious what kind of DB its running below

jer-0:
any update on this?

rvdende:
Hi all, I'm wondering if there is a possibility of adding surrealdb scope authentication to this? The idea being that an authenticated user can then access the database from the frontend directly.

ThangHuuVu (approved):

ThangHuuVu:
@rvdende can you open a separate discussion for your question? 🙏

var-che:
Is the installation/setup page for surrealDB going to be updated?

PR #6257, opened 2023-01-03
Updated installation page for SvelteKit

oscarspalk:
I have explained the load session into layout and type errors note, which I felt were missing, when I had trouble using the library.


ThangHuuVu:
Thanks for sending the PR 🙌 However I think we've already covered this in one recent PR: https://github.com/nextauthjs/next-auth/pull/6184

PR #6259, opened 2023-01-03
fix(core): improve stack traces

balazsorban44:
When re-throwing errors, we should preserve the original stack-trace so the error can be easily identified. Also improved formatting.

Before:
image

After:
image


ThangHuuVu (approved):
One nit comment and that's all! I've tested the output, looking good 👍

PR #6270, opened 2023-01-03
@auth/remix framework for using @auth in remix

acoreyj:
MOVED TO https://github.com/nextauthjs/next-auth/pull/6767

### DIFF
--- a/packages/adapter-dynamodb/package.json
+++ b/packages/adapter-dynamodb/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@next-auth/dynamodb-adapter",
   "repository": "https://github.com/nextauthjs/next-auth",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "description": "AWS DynamoDB adapter for next-auth.",
   "keywords": [
     "next-auth",
--- a/packages/adapter-xata/package.json
+++ b/packages/adapter-xata/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@next-auth/xata-adapter",
-  "version": "0.2.1",
+  "version": "0.2.2",
   "description": "Xata adapter for next-auth.",
   "homepage": "https://authjs.dev",
   "repository": "https://github.com/nextauthjs/next-auth",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@auth/core",
-  "version": "0.2.5",
+  "version": "0.3.0",
   "description": "Authentication for the Web.",
   "keywords": [
     "authentication",
--- a/packages/frameworks-sveltekit/package.json
+++ b/packages/frameworks-sveltekit/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@auth/sveltekit",
-  "version": "0.1.12",
+  "version": "0.2.0",
   "description": "Authentication for SvelteKit.",
   "keywords": [
     "authentication",


balazsorban44 (commented):
Thanks for the PR! As a preliminary review to set some expectations, I would like to point out that we should aim to embrace the framework as much as possible, rather than reusing code from other framework implementations. I'll have to dig deeper into Remix, but I have a feeling that the current client module is not the "Remix-way". We could merge it for now, but will have to improve before hitting a stable release version.

packages/frameworks-remix/scripts/postbuild.js
```diff
@@ -0,0 +1,16 @@
+import path from "path";
```
balazsorban44, line 1: I don't think, this file should be needed.

packages/frameworks-remix/package.json
```diff
@@ -0,0 +1,53 @@
...
+  },
+  "devDependencies": {
+    "@auth/core": "workspace:*",
+    "@types/node": "^18.7.14",
+    "next-auth": "workspace:*",
+    "tsup": "^6.5.0",
+    "typescript": "^4.8.2"
+  },
+  "peerDependencies": {
+    "@auth/core": "~0.2.2 || ^0.2.2",
```
balazsorban44, line 38:
```suggestion
```

not necessary, since it's added as a dependency

packages/frameworks-remix/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "clean": "rm -rf client.* index.*"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "@auth/core": "workspace:*",
+    "@types/node": "^18.7.14",
+    "next-auth": "workspace:*",
+    "tsup": "^6.5.0",
```
balazsorban44, line 34: let's not add another tool. we already have `tsc`, let's remove this

packages/frameworks-remix/tsup.config.js
```diff
@@ -0,0 +1,14 @@
+import { defineConfig } from "tsup";
```
balazsorban44, line 1: let's remove

packages/frameworks-remix/tsconfig.eslint.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "./tsconfig.json",
```
balazsorban44, line 2: should only have single tsconfig, let's remove this. the eslint config is in the root directory

packages/frameworks-remix/package.json
acoreyj: @balazsorban44 

Hmm I'll make the change but just fyi I based it off

https://github.com/nextauthjs/next-auth/tree/main/packages/frameworks-solid-start

acoreyj:
> Thanks for the PR! As a preliminary review to set some expectations, I would like to point out that we should aim to embrace the framework as much as possible, rather than reusing code from other framework implementations. I'll have to dig deeper into Remix, but I have a feeling that the current client module is not the "Remix-way". We could merge it for now, but will have to improve before hitting a stable release version.

Hmm good call I think I can adapt to using the createCookieSessionStorage  function from remix

packages/frameworks-remix/package.json
balazsorban44: thanks for the heads up, we should do the same there too. 👍

OrJDev: solid start is now using tsc instead of tsup aswell

https://github.com/nextauthjs/next-auth/pull/6276/commits/cad4b59d3cc3c753956bee30f76a75ad45f9e7fa

acoreyj:
This is pretty much working but I'm hoping on seeing some bug fixes in upcoming remix releases that this needs so patches aren't needed.

Specifically

https://github.com/remix-run/react-router/pull/9913
https://github.com/remix-run/remix/pull/3330
https://github.com/remix-run/remix/pull/5095

cliffordfajardo:
@acoreyj - is the auth.js adapter for remix still something you'd want to tackle after those bug fixes you mentioned are resolved? Currently using Remix Auth to get Azure Authentication working, but would love to get it working with Auth.js Remix adapter 😄

acoreyj:
@cliffordfajardo 

Yea I have actually been using it myself by using patch-package to implement the change from the remix PR if that's something you want to do.

See the new MR at https://github.com/nextauthjs/next-auth/pull/6767

PR #6290, opened 2023-01-04
feat(providers): added a mattermost provider

### LINKED ISSUE #6271: New Provider: Mattermost
### Description 📓

[Mattermost is an opensource chat app](https://github.com/mattermost/mattermost-server) that has support for OAuth providers see [here](https://developers.mattermost.com/integrate/apps/authentication/oauth2/). I think adding a Mattermost provider would be a good idea as it was initially a pain for me to get working so others might appreciate a flawless experience.

### How to reproduce ☕️

[Here](https://github.com/FinnDore/t3-mattermost/blob/main/src/auth/mattermost-provider.ts#L3) is the provider I have created. I think it's pretty much good to go so I could create the PR asap providing it's wanted.

### Contributing 🙌🏽

[my current branch](https://github.com/FinnDore/next-auth/blob/feat/mattermost-provider/packages/core/src/providers/mattermost.ts)

FinnDore:
## ☕️ Reasoning

I have added a new provider for [Mattermost](https://github.com/mattermost/mattermost-server) 

## 🧢 Checklist

- [ X ] Documentation
- [ ] Tests
- [ X ] Ready to be merged

## 🎫 Affected issues

Resolves #6271


FinnDore:
I see `packages/next-auth/src/providers` also exists do I need to ctrl c + v the new provider in there as well to include it in next-auth?

docs/docs/reference/05-oauth-providers/mattermost.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Configuration
+
+http://my-cool-server.cloud.mattermost.com/mycoolteam/integrations/oauth2-apps
+
+## Options
+
+The **Mattermost provider** comes with a set of default options:
+
+- [Mattermost Provider options](https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/providers/mattermost.ts)
```
FinnDore, line 18: the `/packages/next-auth/src/providers/mattermost.ts` part would make it a dead link, so I guess ill have to also place the new provider there aswell? ill wait for clarification :)

packages/core/src/providers/mattermost.ts
```diff
@@ -0,0 +1,73 @@
...
+export default function Mattermost<P extends MattermostProfile>(
+  config: OAuthUserConfig<P> & { issuer: string }
+): OAuthConfig<P> {
+  const { issuer, ...rest } = config
+
+  return {
+    id: "mattermost",
+    name: "Mattermost",
+    type: "oauth",
+    client: { token_endpoint_auth_method: "client_secret_post" },
```
balazsorban44, line 51: Passing client_id and client_secret as query params were non-standard, so I changed it. See: https://www.rfc-editor.org/rfc/rfc7591.html#section-2

packages/core/src/providers/mattermost.ts
```diff
@@ -0,0 +1,73 @@
...
+    manualTimezone: string
+    useAutomaticTimezone: string
+  }
+  disable_welcome_email: boolean
+}
+
+export default function Mattermost<P extends MattermostProfile>(
+  config: OAuthUserConfig<P> & { issuer: string }
+): OAuthConfig<P> {
+  const { issuer, ...rest } = config
```
balazsorban44, line 45: Using `issuer` as this is the conventional name for the Authorization Server.

balazsorban44 (approved):
Thanks! Ideally, the docs should be generated from the source code. I approved this PR for now, but if you want to help out and follow up in another PR, check out GitHub for example: https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/providers/github.ts

Using JSDoc, we are able to generate consistent API references like these:

- https://authjs.dev/reference/core/functions/providers_github.default
- https://authjs.dev/reference/core/interfaces/providers_github.GithubProfile

FinnDore:
Oh wow! thanks a tone. I'll get on the docs after work today hopefully :).

jonasjancarik:
I am a bit confused - why doesn't the Mattermost provider seem to be included in the providers in the current version (4.24.7)? It's not in the docs and it's not in `node_modules/next-auth/providers` either. What am I missing?

PR #6300, opened 2023-01-05
Remove callback urls warning for GitHub provider

iAverages:
## ☕️ Reasoning

GitHub has supported multiple callback URLs since 2020 
See: https://github.blog/changelog/2020-10-26-github-apps-api-updates-now-ga/#multiple-callback-urls

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

### DIFF
--- a/docs/docs/providers/github.md
+++ b/docs/docs/providers/github.md
@@ -37,10 +37,6 @@ providers: [
 ...
 ```
 
-:::warning
-Only allows one callback URL per Client ID / Client Secret.
-:::
-
 :::tip
 Email address is always returned, even if the user doesn't have a public email address on their profile.
 :::


balazsorban44:
Thanks, but this is incorrect. GitHub apps are not the same as OAuth apps.

You can read about the difference here https://docs.github.com/en/developers/apps/getting-started-with-apps/about-apps#about-github-apps 

I could not find direct mention of only single callback URL being allowed for OAuth apps, but check out https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#redirect-urls.

Also, the GitHub UI does not allow adding more than one callback URL for OAuth apps either.

PR #6307, opened 2023-01-06
fix: alter getURL to utilise NEXTAUTH_URL_INTERNAL if defined

### LINKED ISSUE #6306: utilise `NEXTAUTH_URL_INTERNAL` for `unstable_getServerSession` calls
### Description 📓

#1449 introduced a `NEXTAUTH_URL_INTERNAL` env var that as [stated by the docs should be used for all server side calls](https://next-auth.js.org/configuration/options#nextauth_url_internal). 

However, this is not the case as `unstable_getServerSession` is explicitly a server side call but still utilises the `NEXTAUTH_URL` env var to form the url for fetch calls.

[unstable_getServerSession ](https://github.dev/nextauthjs/next-auth/blob/48746c896a41abcf364f42d9d79d72080885a076/packages/next-auth/src/next/index.ts#L143) uses the [getURL](https://github.dev/nextauthjs/next-auth/blob/48746c896a41abcf364f42d9d79d72080885a076/packages/next-auth/src/utils/node.ts#L32-L62) function. 

```ts
export function getURL(url: string | undefined, headers: Headers): URL | Error {
  try {
    if (!url) throw new Error("Missing url")
    if (process.env.NEXTAUTH_URL) {
      const base = new URL(process.env.NEXTAUTH_URL)
      if (!["http:", "https:"].includes(base.protocol)) {
        throw new Error("Invalid protocol")
      }
      const hasCustomPath = base.pathname !== "/"

      if (hasCustomPath) {
        const apiAuthRe = /\/api\/auth\/?$/
        const basePathname = base.pathname.match(apiAuthRe)
          ? base.pathname.replace(apiAuthRe, "")
          : base.pathname
        return new URL(basePathname.replace(/\/$/, "") + url, base.origin)
      }
      return new URL(url, base)
    }
    const proto =
      headers.get("x-forwarded-proto") ??
      (process.env.NODE_ENV !== "production" ? "http" : "https")
    const host = headers.get("x-forwarded-host") ?? headers.get("host")
    if (!["http", "https"].includes(proto)) throw new Error("Invalid protocol")
    const origin = `${proto}://${host}`
    if (!host) throw new Error("Missing host")
    return new URL(url, origin)
  } catch (error) {
    return error as Error
  }
}
```

As seen above this explicitly uses the `process.env.NEXTAUTH_URL` value and ignores `NEXTAUTH_URL_INTERNAL`.

Having a quick look over where `getURL` is used, and looking at it's JSDoc it looks like it is only ever used to create a URL to call in a function, not passed back to the user as a callback url or anything. So, we should be able to just edit this function to use `NEXT_AUTH_INTERNAL` if it is defined! 

ie: 
```ts
   const base = new URL(process.env.NEXTAUTH_URL_INTERNAL ?? process.env.NEXTAUTH_URL)
```

If we wanted to be super safe we could also check for `typeof window === "undefined"` as is done [in this function](https://github.dev/nextauthjs/next-auth/blob/48746c896a41abcf364f42d9d79d72080885a076/packages/next-auth/src/client/_utils.ts#L53-L60). 

I can throw in this as a quick PR if the maintainers are happy and can't see an issue in this change? 

Also Just a little thanks for making such an awesome auth package available to all!! 👍🙌

### How to reproduce ☕️

This is quite hard to show in a simple sandbox etc, but it is doing this as I have encountered it [in a current issue when deploying on fly.io](https://community.fly.io/t/name-resolution-of-apps-current-url-failing-internally-ssr-happening-for-others/9861/4) where the full external canonical URL is not resolving when Server Side Rendering (Which was the issue that caused the creation of `NEXTAUTH_URL_INTERNAL` in the first place!) 

### Contributing 🙌🏽

JonParton:
## ☕️ Reasoning

This will close out #6306 where `unstable_getServerSession` was not utilising the `NEXTAUTH_URL_INTERNAL` env variable even though the docs say all server side calls will use it. 

In an environment where the full canonical URL of the application cannot be resolved when server side, this causes issues if you utilise `unstable_getServerSession` for session information when rendering server side or in an API route. 

## 🧢 Checklist

- [X] Documentation => Existing documentation implies this is already the case!
- [ ] Tests
- [x] Ready to be merged => Subject toa  maintainer confirming they are happy 🙂

## 🎫 Affected issues

Fixes: #6306

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/utils/node.ts
+++ b/packages/next-auth/src/utils/node.ts
@@ -32,8 +32,8 @@ export function getBody(
 export function getURL(url: string | undefined, headers: Headers): URL | Error {
   try {
     if (!url) throw new Error("Missing url")
-    if (process.env.NEXTAUTH_URL) {
-      const base = new URL(process.env.NEXTAUTH_URL)
+    if (process.env.NEXTAUTH_URL_INTERNAL ?? process.env.NEXTAUTH_URL) {
+      const base = new URL(process.env.NEXTAUTH_URL_INTERNAL ?? process.env.NEXTAUTH_URL)
       if (!["http:", "https:"].includes(base.protocol)) {
         throw new Error("Invalid protocol")
       }


balazsorban44:
See https://github.com/nextauthjs/next-auth/issues/6306#issuecomment-1373926023

PR #6336, opened 2023-01-09
feat: allow async `generateSessionToken`

HommeSauvage:
Some random string libraries are async (such as nanoid). It would be great if `generateSessionToken()` can be async. This is backwards compatible as await'ing a sync function will work without issues.


balazsorban44 (changes requested):
Can you do these changes against https://github.com/nextauthjs/next-auth/tree/main/packages/core/src? We are only accepting critical fixes against `next-auth` at this point.

balazsorban44:
Closing in favor of #6536

PR #6132, opened 2022-12-21
fix(next-auth): revert to 4.17 to fix host issues but keep other fixes

balazsorban44:
The last few `next-auth` releases have been an attempt to fix something that was only supposed to be part of the new (experimental) `@auth/core` package, but got merged accidentally and did not realize that it would cause issues later on.

Rather than reverting then, I wanted to just fix the issue, but albeit all my efforts, the issue kept coming up in different forms, which led to a few broken releases (`4.18.0` or later.)

In this PR, I decided to mostly revert the code base to the state it was on `4.17.0` - the last known version without such issues - and replay the fixes since that version.

I am planning to put more effort into `@auth/core`, but want to make sure that `next-auth` keeps being the awesome library that so many rely on. :tada:

**DISCLAIMER**: I will also note here, that with this change, `next-auth` enters its maintenance phase. We are not going to add more features to this package and will be focusing on making `@auth/core` on pair in terms of features with this package, using standard Web APIs. **The only exception is security fixes or major bugs.**

We are obviously **not** abandoning Next.js, but expect a new, improved library `@auth/nextjs` coming out in the following months instead. It's going to be fully Edge compatible (heads up, Vercel's [Edge Functions](https://vercel.com/blog/edge-functions-generally-available) has just entered general availability, which is a great place to test out the upcoming package), will be smaller in size, and will introduce new APIs and conventions to leverage the App directory too. (`pages` will be supported for the foreseeable future still).

Fixes #6102, Fixes #6045, Fixes #6113

PR #6337, opened 2023-01-09
feat(core): allow clearing cookies from `jwt()`

iMrDJAi:
## ☕️ Reasoning
Currently there is no way to invalidate session from the `jwt()` callback. This change allows clearing the user cookies when `jwt()` returns ~~a falsy value~~ null.

~~Also, a new provider option has been added, `decodeJWT`, which allows passing the existing token of a user (if any) instead of the default one to the `jwt()` callback after provider's `authorize()`. This allows creating providers for the purpose of checking and invalidating existing user sessions.~~

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues
Fixes: https://github.com/nextauthjs/next-auth/discussions/6325

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -215,7 +215,7 @@ export interface CallbacksOptions<P = Profile, A = Account> {
     account?: A | null
     profile?: P
     isNewUser?: boolean
-  }) => Awaitable<JWT>
+  }) => Awaitable<JWT|null>
 }
 
 /** [Documentation](https://authjs.dev/reference/configuration/auth-config#cookies) */
@@ -464,6 +464,7 @@ export interface InternalOptions<TProviderType = ProviderType> {
   theme: Theme
   debug: boolean
   logger: LoggerInstance
+  decodeJWT?: boolean
   session: Required<SessionOptions>
   pages: Partial<PagesOptions>
   jwt: JWTOptions
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -207,7 +207,7 @@ export interface CallbacksOptions<P = Profile, A = Account> {
    * Returning `null` will invalidate the JWT session by clearing
    * the user's cookies. You'll still have to monitor and invalidate
    * unexpired tokens from future requests yourself to prevent
-   * unauthoried access.
+   * unauthorized access.
    *
    * By default the JWT is encrypted.
    *
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -464,7 +464,6 @@ export interface InternalOptions<TProviderType = ProviderType> {
   theme: Theme
   debug: boolean
   logger: LoggerInstance
-  decodeJWT?: boolean
   session: Required<SessionOptions>
   pages: Partial<PagesOptions>
   jwt: JWTOptions
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -203,6 +203,11 @@ export interface CallbacksOptions<P = Profile, A = Account> {
    * Its content is forwarded to the `session` callback,
    * where you can control what should be returned to the client.
    * Anything else will be kept inaccessible from the client.
+   * 
+   * Returning `null` will invalidate the JWT session by clearing
+   * the user's cookies. You'll still have to monitor and invalidate
+   * unexpired tokens from future requests yourself to prevent
+   * unauthoried access.
    *
    * By default the JWT is encrypted.
    *
--- a/packages/next-auth/src/core/routes/callback.ts
+++ b/packages/next-auth/src/core/routes/callback.ts
@@ -29,6 +29,7 @@ export default async function callback(params: {
     jwt,
     events,
     callbacks,
+    decodeJWT,
     session: { strategy: sessionStrategy, maxAge: sessionMaxAge },
     logger,
   } = options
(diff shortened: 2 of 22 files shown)


balazsorban44 (changes requested):
Few comments:

- This should be done against https://github.com/nextauthjs/next-auth/tree/main/packages/core as `next-auth` only receives critical fixes at this point
- Let's allow returning `null` in `jwt` which indicates that we should call `sessionStore.clean()`.
- Looks like you did changes to the `session` callback too, but I don't understand the reason. Could we revert that?

packages/next-auth/src/core/routes/callback.ts
```diff
@@ -37,6 +38,13 @@ export default async function callback(params: {
 
   const useJwtSession = sessionStrategy === "jwt"
 
+  const decodedToken = useJwtSession && decodeJWT
```
balazsorban44, line 41: Why is this necessary?

packages/next-auth/src/core/types.ts
```diff
@@ -15,6 +15,7 @@ import type { CookieSerializeOptions } from "cookie"
 import type { NextApiRequest, NextApiResponse } from "next"
 
 export type Awaitable<T> = T | PromiseLike<T>
+export type Falsy = null|undefined|false
```
balazsorban44, line 18: Let's not introduce a new type for this, only check for `null`. The other values don't make sense anyway.

packages/next-auth/src/core/routes/callback.ts
```diff
@@ -29,6 +29,7 @@ export default async function callback(params: {
     jwt,
     events,
     callbacks,
+    decodeJWT,
```
balazsorban44, line 32: Unclear what this tries to accomplish. Could you clarify?

packages/next-auth/src/core/types.ts
```diff
@@ -325,7 +326,7 @@ export interface CallbacksOptions<P = Profile, A = Account> {
     session: Session
     user: User | AdapterUser
     token: JWT
-  }) => Awaitable<Session>
+  }) => Awaitable<Session|Falsy>
   /**
    * This callback is called whenever a JSON Web Token is created (i.e. at sign in)
```
balazsorban44, line 331: The comment above needs to be extended that explains, returning `null` will invalidate the JWT session.

packages/next-auth/src/core/types.ts
```diff
@@ -325,7 +326,7 @@ export interface CallbacksOptions<P = Profile, A = Account> {
     session: Session
     user: User | AdapterUser
     token: JWT
-  }) => Awaitable<Session>
+  }) => Awaitable<Session|Falsy>
```
balazsorban44, line 329: Not sure why this is necessary

iMrDJAi:
@balazsorban44 Hello, sorry for the delay, working on the changes that you have requested.

Regarding the session callback, clearing the cookies could be done when fetching the session, not sure what else to return from it when the token is ~~falsy~~ `null`:

https://github.com/iMrDJAi/next-auth/blob/27dc1b20c0fa6c0d0e3c15f099ef27d8a132159e/packages/next-auth/src/core/routes/session.ts#L63-L68

Does setting the `response.body` to `undefined` does the job? Or should we avoid calling it at all in this case? What do you think?

Regarding the new option `decodeJWT` (or it could be `decodeExistingJWT`, it certainly needs a clearer name), the point is, if the auth cookies exist while the user is attempting to call a provider, it allows to decode the JWT of that user:

https://github.com/iMrDJAi/next-auth/blob/27dc1b20c0fa6c0d0e3c15f099ef27d8a132159e/packages/next-auth/src/core/routes/callback.ts#L41-L46

Then passing it to `jwt()` instead of `defaultToken`:

https://github.com/iMrDJAi/next-auth/blob/27dc1b20c0fa6c0d0e3c15f099ef27d8a132159e/packages/next-auth/src/core/routes/callback.ts#L41-L46

In this way, we could for example make a provider to sign out users, as we have their pervious state.

balazsorban44:
So basically passing the previous token if I understand correctly. That should be a different PR. Could we split it please and simplify this PR?

iMrDJAi:
@balazsorban44 Alright! one thing left:

> The comment above needs to be extended that explains, returning null will invalidate the JWT session.

Working on it...

iMrDJAi:
@balazsorban44 Before creating a new PR, what should we call that option?

balazsorban44:
don't think it should be an option, just give it as `prevToken` or `currentToken`. 🤔

BTW this will resolve #3946

rolanday:
Just curious regarding ...

> "This should be done against https://github.com/nextauthjs/next-auth/tree/main/packages/core as next-auth only receives critical fixes at this point"

What exactly does "at this point" mean? Is next-auth end-of-life?

iMrDJAi:
@rolanday Auth.js will replace NextAuth.js.

@balazsorban44 Actually that's a very good idea, we can pass both pervious and new tokens! I'll play with it a little bit then I'll create a PR.

packages/core/src/types.ts
```diff
@@ -203,6 +203,11 @@ export interface CallbacksOptions<P = Profile, A = Account> {
    * Its content is forwarded to the `session` callback,
    * where you can control what should be returned to the client.
    * Anything else will be kept inaccessible from the client.
+   * 
+   * Returning `null` will invalidate the JWT session by clearing
+   * the user's cookies. You'll still have to monitor and invalidate
+   * unexpired tokens from future requests yourself to prevent
+   * unauthoried access.
```
balazsorban44, line 210:
```suggestion
   * unauthorized access.
```

balazsorban44 (approved):
Thanks!

balazsorban44:
@rolanday https://github.com/nextauthjs/next-auth/pull/6132#issue-1505602450

PR #6357, opened 2023-01-10
feat: pass previous token to jwt() callback

iMrDJAi:
## ☕️ Reasoning
This change allows passing the existing JWT of a user (if any) to the `jwt()` callback after calling a provider's `authorize()`. This allows using multiple providers while being already authorized for the purpose of merging sessions or creating providers to check and invalidate existing user sessions (sign out users).

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues
Fixes: https://github.com/nextauthjs/next-auth/discussions/3946


iMrDJAi:
Not sure how is it practically possible to pass the previous session to the `session()` callback. Is that even needed?

iMrDJAi:
@balazsorban44 We should discuss whether or not we should pass it as `prevToken` or add an option to override `token` as I suggested earlier. Both options seem fine to me though.

iMrDJAi:
@balazsorban44 Could you please review this?

ThangHuuVu (commented):
I went through the discussion - returning the old session in the session callback could be possible if we call `callbacks.session` in the `/callback` route where we still have both old and new tokens. I think it's fine to do so, what do you think @balazsorban44 ? 🤔

lavcraft:
> Not sure how is it practically possible to pass the previous session to the `session()` callback. Is that even needed?
> 
> **Edit:** @balazsorban44 what was @lavcraft suggesting regarding the `session()` callback in #3946 is impossible to implement, since neither the client nor the server holds a copy of the previous session object. I believe this is something that devs should handle by themselves with some custom logic. Thoughts?

Sounds reasonable, but can you explain how dev do this? Now i don't understand in which kind of hooks i should implement this.

PS: Thanks for your improvements, it's really good :)

iMrDJAi:
> I went through the discussion - returning the old session in the session callback could be possible if we call `callbacks.session` in the `/callback` route where we still have both old and new tokens. I think it's fine to do so, what do you think @balazsorban44 ? 🤔

@ThangHuuVu The problem is, the previous token isn't available in the session callback. If we somehow could keep the previous token available at that point, isn't it better to just pass the token itself rather than the value `callbacks.session({ token, ...params })`? I mean, it's the same callback, why calling it twice?

> Sounds reasonable, but can you explain how dev do this? Now i don't understand in which kind of hooks i should implement this.
>
> PS: Thanks for your improvements, it's really good :)

@lavcraft In the most of cases, isn't the session object built based on the JWT object itself? in the JWT callback we could merge both tokens anyway, so in practice we could pass information from the old token to the session object.

ThangHuuVu:
> @ThangHuuVu The problem is, the previous token isn't available in the session callback. If we somehow could keep the previous token available at that point, isn't it better to just pass the token itself rather than the value callbacks.session({ token, ...params })? I mean, it's the same callback, why calling it twice?

@iMrDJAi I meant we could call `callbacks.session` in the `/callback` route, where we still have the old session - that way we can trigger the `session` callback once with the old session object. Does that make sense? 👀 I'm not talking about the `callbacks.session` call in the `/session` route

iMrDJAi:
@ThangHuuVu Yes, actually this does make sense. I assumed that a user would only need to call the session callback to return the session object, but I ignored that fact that in some scenarios, he could use it for something else (depending on the logic he implements, maybe to trigger something?), it's a callback after all.

leog:
Interested in this, to be able to link users through the login page, detecting an existing session different from new session would allow to deduce the situation and persist the linking in some custom way, as JWT sessions are not supporting linking users beyond accounts through different providers.

ThangHuuVu:
@iMrDJAi I think the PR is good, the only question left from last review is whether we should call callbacks.session in /callback route. I think we can leave it and merge the PR first 👍 cc @balazsorban44

iMrDJAi:
@ThangHuuVu I think that we should move forward with this and leave the session callback thing for another PR.

One more thing, can you by any chance allow backporting this change along with https://github.com/nextauthjs/next-auth/pull/6337 to `next-auth` package as well? Currently I cannot upgrade to `@auth/core` since it's not yet production ready and it's incompatible with express. Also, it would require tons of rewriting as I'm dealing with a complex setup (I don't actually use next.js, I use vite-plugin-ssr). Currently the only way to bring my changes to my projects is by manually patching the package or maintaining my own fork, with is inconvenient. Thank you in advance.

leog:
Yes, please consider porting it to `next-auth` as that's what we are using at @calcom for the time being

iMrDJAi:
@leog I think they still accept pull requests for `next-auth` but only under the `v4` branch. @ThangHuuVu Can you please confirm that?

leog:
> @leog I think they still accept pull requests for `next-auth` but only under the `v4` branch. @ThangHuuVu Can you please confirm that?

@iMrDJAi thsnks for the heads up, luckily we are using v4

leog:
I’m available to contribute to move this forward, JLMK @iMrDJAi @ThangHuuVu

leog (approved):
LGTM

leog:
Please consider to merge this @iMrDJAi @ThangHuuVu @lluia

ThangHuuVu:
@iMrDJAi @leog Sorry for the delay, I just got back from a discussion with @balazsorban44 - we're a bit hesitant to merge this change because we're rethinking the JWT callback (and other callbacks) in v5. Therefore, adding new features to these callbacks will introduce more resistant and make it harder to migrate. For now, please use `patch-package` if you want this change in your app. Thank you for your understanding. 🙏

leog:
> @iMrDJAi @leog Sorry for the delay, I just got back from a discussion with @balazsorban44 - we're a bit hesitant to merge this change because we're rethinking the JWT callback (and other callbacks) in v5. Therefore, adding new features to these callbacks will introduce more resistant and make it harder to migrate. For now, please use `patch-package` if you want this change in your app. Thank you for your understanding. 🙏

I completely understand, thanks for the explanation @ThangHuuVu, perhaps this PR should be closed on that ground. About v5, will the JWT callback support having the previous token available? Or perhaps officially support linking account in some way? Let me know if I can help testing it if so. Is there a public timeline for releasing v5? Just to adjust to it from our side. Thanks again.

iMrDJAi:
@ThangHuuVu @balazsorban44 Hello, thanks for your response. I understand the concerns about introducing more resistance to the callback changes in v5. However, I'd like to point out that the PR has been open for nearly 9 months, even before the discussion of v5 started. This feature would greatly benefit the current version (v4) and the community using it, it's very minimal and I doubt it would cause any migration issues in the future. Passing a property wouldn't cause too much harm, and users could always adapt to any design that the next release would introduce, providing that it will have major breaking changes anyway, regardless of this PR getting merged or not. Also, v5 is nowhere close to being ready, neither is Auth.js. This is why I've recreated this PR (https://github.com/nextauthjs/next-auth/pull/8260) and another one I had (https://github.com/nextauthjs/next-auth/pull/8256) to target the v4 branch instead. Thank you.

luisalrp:
Hello, I really need this functionality and I patched the package with PR (#8260), but I see the previous token as undefined the second time I sign in with a different oauth provider. Should I change [...nextauth].ts in any way besides including prevToken in the async jwt callback?

My use case is easy. Being signed with one provider and then sign in again to link the account, even if when the user signs again with another providers, it does not have the same email as the previous one.

Thanks!

iMrDJAi:
Hello @luisalrp, my apologies for the delay. Can you please share a minimal reproduction? It should work fine with any OAuth provider as we're passing `pervToken` to JWT callback here: https://github.com/nextauthjs/next-auth/pull/6357/files#diff-34764726f8865be5a9cc1e76b1c3b7407f024a45678781abbbac0e4901a0c34fR114.

Keep in mind this will only work if `session.strategy` is set to `jwt`.

iMrDJAi:
@leog @ThangHuuVu Hello, sorry to bother you, but since this PR, https://github.com/nextauthjs/next-auth/pull/8260 and https://github.com/nextauthjs/next-auth/pull/8256 have been open for a while without a clear conclusion, and as I see new releases of NextAuth.js (v4) popping up with no indication that these are going to make it to the package anytime soon, I have decided that I'm giving a deadline of one more week before closing them permanently. During this week feel free to review or merge or give more insight. I'm just sick and I have a total burnout and it's not worth the hustle anymore. I'll have to find and alternative as well since I neither can wait for Auth.js nor use patch-package for a library that is so critical and exposed to security vulnerabilities and attacks. Thank you for understanding.

feliciavonbraun:
Hi, I'm in need of the previous token feature as well and from what I can see in the beta 5.0.0-beta.22 this feature is not yet implemented. Does anyone have any updated either how you solved this (without using patched version) or if (and when) this will be implemented in v5?

iMrDJAi:
> Hi, I'm in need of the previous token feature as well and from what I can see in the beta 5.0.0-beta.22 this feature is not yet implemented. Does anyone have any updated either how you solved this (without using patched version) or if (and when) this will be implemented in v5?

@feliciavonbraun I guess you don't have much to do in this case but patching the package. You might use [patch-package](https://www.npmjs.com/package/patch-package) for that if you don't want to host your own patched version.

Issue #6363, opened 2023-01-11
Twitch Provider missing client id


brazelja:
### Provider type

Twitch

### Environment

System:
  

> OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
>   CPU: (16) x64 AMD Ryzen 9 5900HX with Radeon Graphics
>   Memory: 10.68 GB / 15.34 GB
>   Container: Yes
>   Shell: 5.8.1 - /usr/bin/zsh

Binaries:
  

> Node: 18.12.1 - ~/.volta/tools/image/node/18.12.1/bin/node
>   Yarn: 1.22.19 - ~/.volta/tools/image/yarn/1.22.19/bin/yarn
>   npm: 8.19.3 - ~/.volta/tools/image/npm/8.19.3/bin/npm
>   pnpm: 6.11.0 - ~/.volta/bin/pnpm

Packages:

> @auth/core: 0.2.5
> @auth/sveltekit: 0.1.12

### Reproduction URL

https://authjs.dev/reference/oauth-providers/twitch

### Describe the issue

I am using `@auth/sveltekit` and attempting to add Twitch and Discord login options. I followed the documentation pages for each and have them set up like so:

```ts
import { SvelteKitAuth } from '@auth/sveltekit';
import Discord from '@auth/core/providers/discord';
import Twitch from '@auth/core/providers/twitch';
import {
  DISCORD_CLIENT_ID,
  DISCORD_CLIENT_SECRET,
  TWITCH_CLIENT_ID,
  TWITCH_CLIENT_SECRET
} from '$env/static/private';

export const handle = SvelteKitAuth({
  providers: [
    Discord({
      clientId: DISCORD_CLIENT_ID,
      clientSecret: DISCORD_CLIENT_SECRET
    }),
    Twitch({
      clientId: TWITCH_CLIENT_ID,
      clientSecret: TWITCH_CLIENT_SECRET
    })
  ]
}
```

Using the demo project setup described [here[(https://authjs.dev/reference/sveltekit/modules/main) but replacing GitHub with Twitch/Discord I was able to get the project up and running. Logging in with Discord works perfectly fine, but when I log attempt to log in with Twitch I get:

```bash
CallbackRouteError: "response" is not a conform Token Endpoint response
[...]

PR #6365, opened 2023-01-11
fix(providers): conform Twitch provider to spec with escape hatch

balazsorban44:
Fixes #6363

This PR introduces a new `conform()` callback for the `token` OAuth config option. It's marked as internal for now, and not expected to be used in user-land. This is an escape hatch to work around non-conformities in our built-in providers in a way that does not require a full `request` override, but also does not add special handling to providers into the core.

This is a last resort, and ideally, such issues should be fixed on the Provider side. If you see this method and have the power, we encourage you to put pressure on the given provider so they fix the issue on their side.

Once a non-conformity is fixed, the `conform()` method will be pass-through and will show a warning. In this case, you can file an issue on this repo so we can remove the workaround.

### DIFF
--- a/packages/core/src/providers/twitch.ts
+++ b/packages/core/src/providers/twitch.ts
@@ -1,4 +1,4 @@
-import type { OAuthConfig, OAuthUserConfig } from "./index.js"
+import type { OIDCConfig, OIDCUserConfig } from "./index.js"
 
 export interface TwitchProfile extends Record<string, any> {
   sub: string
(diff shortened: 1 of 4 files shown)


packages/core/src/providers/twitch.ts
```diff
@@ -7,26 +7,52 @@ export interface TwitchProfile extends Record<string, any> {
...
+          id_token: { email: null, picture: null, preferred_username: null },
         },
       },
     },
+    token: {
+      async conform(response) {
+        const body = await response.json()
+        if (response.ok) {
+          if (typeof body.scope === "string") {
+            console.warn(
```
ThangHuuVu, line 32: I think we should warn in dev only to avoid cluttering the console output, WDYT? it seems tricky to determine the environment from within the core right?

balazsorban44: Yeah, it's practically impossible because of the lack of standards for environment variables, but we could expose a `dev: boolean` config option in `AuthConfig`, I guess.

balazsorban44: We could follow up in a different PR

ThangHuuVu: We could also use `config.debug` 👍

ThangHuuVu (approved):
LGTM tested with Twitch login before and after this PR! 🙌

PR #6379, opened 2023-01-12
feat(core): option to opt out of CSRF checks

balazsorban44:
**NOTE**: This is only intended for framework authors.

Some frameworks (like [SvelteKit](https://kit.svelte.dev/docs/configuration#csrf)) come with built-in CSRF protection. In that case, our CSRF check adds some overhead and makes it difficult to integrate in a way that wouldn't require client-side JS.

This new option makes it possible to opt out. Be careful!

Example:

```ts
// Some framework code
import { Auth, skipCSRFCheck } from "@auth/core"

const response = await Auth(new Request("https://example.com"), {
  skipCSRFCheck
})
```

### DIFF
--- a/docs/docs/reference/09-warnings.md
+++ b/docs/docs/reference/09-warnings.md
@@ -3,57 +3,13 @@ id: warnings
 title: Warnings
 ---
 
-This is a list of warning output from Auth.js.
+A list of warnings from Auth.js that need your attention.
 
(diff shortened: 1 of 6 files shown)


manucorporat:
So happy about this PR!

aaronteamAAR:
This would be nice to have in astro

PR #6390, opened 2023-01-14
support baseUrl from client SessionProvider

Jokcy:
## ☕️ Reasoning

Support the baseUrl prop of SessionProvider (it's already being provided but not working currently), so that we can depoly our auth service standalone using a differnt subdomain from our other service, which mean we can share one auth server with multi sites.

I also add a `credentials` option which represents the fetch option `credentials`, we need to pass it as `include` when making cross-site request otherwise fetch will not take the cookies.

## 🧢 Checklist

**This is an early preview PR for maintainers to give some feedback if we should support this feature. I tried to add unit tests but the old test can not run on my computer either so I may need more time to figure out the reason. So I will spend time for tests and documentation after we confirm that we will support this feature.**

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: Only a dicussion right now (and no feedback) https://github.com/nextauthjs/next-auth/discussions/6385

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/client/_utils.ts
+++ b/packages/next-auth/src/client/_utils.ts
@@ -15,6 +15,8 @@ export interface AuthClientConfig {
    * trigger session updates from places like `signIn` or `signOut`
    */
   _getSession: (...args: any[]) => any
+
+  credentials?: RequestCredentials
 }
 
 export interface CtxOrReq {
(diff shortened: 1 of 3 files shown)


ThangHuuVu:
Hi @Jokcy thanks for opening the PR - we decided to close this temporarily and will revisit it later if we gain enough interest. See: https://github.com/nextauthjs/next-auth/discussions/6385#discussioncomment-5624179

PR #6424, opened 2023-01-17
feat(providers): Add Vipps provider

endrekrohn:
Added Vipps provider.

> _NOTE_:
>

## ☕️ Reasoning

This pull request adds the OAuth-provider [Vipps](https://en.wikipedia.org/wiki/Vipps). Vipps is the largest payment application in Norway. Vipps also provides a "Log-In"-product, which this pull request adds support for.

## 🧢 Checklist

- [X] Documentation
- [X] Tests
- [X] Ready to be merged

## 🎫 Affected issues

Fixes: N/A.

## 📌 Resources


balazsorban44 (changes requested):
Thanks, could you move the code under: https://github.com/nextauthjs/next-auth/tree/main/packages/core/src/providers

And the docs do not need to be written manually anymore, just make sure to add proper JSDoc comments, example: https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/providers/auth0.ts

itschip:
Would it be possible to add `nin` as well @endrekrohn ?

balazsorban44:
@itschip we would like if all providers have the same default but you can access all values via the jwt callback's profile parameter, or in case you use a database, the profile callback of the provider can be overridden to save `nin` on your user

itschip:
> @itschip we would like if all providers have the same default but you can access all values via the jwt callback's profile parameter, or in case you use a database, the profile callback of the provider can be overridden to save `nin` on your user

Gotcha, I might continue on OPs PR if we don't see any update in the near future :)

endrekrohn:
> > @itschip we would like if all providers have the same default but you can access all values via the jwt callback's profile parameter, or in case you use a database, the profile callback of the provider can be overridden to save `nin` on your user
> 
> Gotcha, I might continue on OPs PR if we don't see any update in the near future :)

Please do so! I am quite busy at the moment 🙏

PR #6430, opened 2023-01-18
Fix token expiry comparision in database strategy

### LINKED ISSUE #6407: Incorrect condition for token expiry check for Database Strategy

In the page, 
https://authjs.dev/guides/basics/refresh-token-rotation#server-side

In Database Strategy, it looks like token expiry check condition is mistaken.

**What was Wrong?**

In session callbacks in Database Strategy,
```js
 if (google.expires_at >= Date.now()) {
        // If the access token has expired, try to refresh it
```
here, when `google.expired_at` is greater than `Date.now()` it would mean that
the token hasn't expired yet, but this condition has been used to implement the token refreshing.

**Solution** 

Just fliping the comparision, would solve it.

```js
 if (google.expires_at <= Date.now()) {
        // If the access token has expired, try to refresh it
```

When the `expires_at` is greater than `Date.now()` it would  mean that Token Expiry is in the future,
which is incorrect condition to implement token refresh as we do not need to refresh it.

So, the correct condition would be `expires_at` being less than or equal to `Date.now()`.

https://authjs.dev/guides/basics/refresh-token-rotation#server-side

rawbinary:
## ☕️ Reasoning

Incorrect condition for token expiry check for Database Strategy in "Refresh Token Rotation" Docs. (#6407)

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: #6407 

## 📌 Resources

### DIFF
--- a/docs/docs/guides/03-basics/refresh-token-rotation.md
+++ b/docs/docs/guides/03-basics/refresh-token-rotation.md
@@ -136,7 +136,7 @@ export default Auth(new Request("https://example.com"), {
       const [google] = await prisma.account.findMany({
         where: { userId: user.id, provider: "google" },
       })
-      if (google.expires_at >= Date.now()) {
+      if (google.expires_at <= Date.now()) {
         // If the access token has expired, try to refresh it
         try {
           // https://accounts.google.com/.well-known/openid-configuration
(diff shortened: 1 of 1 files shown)


docs/docs/guides/03-basics/refresh-token-rotation.md
```diff
@@ -136,7 +136,7 @@ export default Auth(new Request("https://example.com"), {
       const [google] = await prisma.account.findMany({
         where: { userId: user.id, provider: "google" },
       })
-      if (google.expires_at >= Date.now()) {
+      if (google.expires_at <= Date.now()) {
```
balazsorban44, line 139:
```suggestion
      if (google.expires_at < Date.now()) {
```

PR #6440, opened 2023-01-19
docs(xata-adapter): adjust json schema

atilafassina:
## ☕️ Reasoning

Adjusting the JSON schema in the Xata-Adapter to match the new format.

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged


balazsorban44 (approved):
I remembered something similar, turns out #5772 only updated this in the README.md :+1: I plan to add only links to the readme files, and also generate that from JSDoc comments, so we will have a single source for docs. But this is good for now.

PR #6448, opened 2023-01-20
Pass oauth_token_secret in OAuth 1.0 calls

StachowiakDawid:
## ☕️ Reasoning
Right now oauth_token_secret is being ignored in OAuth 1.0 calls made by Next Auth despite that this secret token is needed by some APIs, e.g. USOS API (University Study-Oriented System is an app used by Polish universities).

Our Computerization Comittee of Students' Council is moving from Google OAuth towards USOS OAuth and the lack of proper OAuth 1.0 support is a reason why we had to fork this repository.

### DIFF
--- a/packages/next-auth/src/core/lib/oauth/authorization-url.ts
+++ b/packages/next-auth/src/core/lib/oauth/authorization-url.ts
@@ -8,6 +8,7 @@ import type { AuthorizationParameters } from "openid-client"
 import type { InternalOptions } from "../../types"
 import type { RequestInternal } from "../.."
 import type { Cookie } from "../cookie"
+import oAuth1TokenStore from "./oauth1-token-store"
 
 /**
  *
@@ -44,7 +45,7 @@ export default async function getAuthorizationUrl({
       oauth_token_secret: tokens.oauth_token_secret,
       ...tokens.params,
     })}`
-
+    oAuth1TokenStore.setTokenSecret(tokens.oauth_token, tokens.oauth_token_secret);
     logger.debug("GET_AUTHORIZATION_URL", { url, provider })
     return { redirect: url }
   }
--- a/packages/next-auth/src/core/lib/oauth/callback.ts
+++ b/packages/next-auth/src/core/lib/oauth/callback.ts
@@ -12,6 +12,7 @@ import type { OAuthChecks, OAuthConfig } from "../../../providers"
 import type { InternalOptions } from "../../types"
 import type { RequestInternal } from "../.."
 import type { Cookie } from "../cookie"
+import oAuth1TokenStore from "./oauth1-token-store"
 
 export default async function oAuthCallback(params: {
   options: InternalOptions<"oauth">
@@ -40,9 +41,10 @@ export default async function oAuthCallback(params: {
       const client = await oAuth1Client(options)
       // Handle OAuth v1.x
       const { oauth_token, oauth_verifier } = query ?? {}
+      const oauth_token_secret = oAuth1TokenStore.getTokenSecret(oauth_token);
       const tokens = (await (client as any).getOAuthAccessToken(
         oauth_token,
-        null,
+        oauth_token_secret,
         oauth_verifier
       )) as TokenSet
       let profile: Profile = await (client as any).get(
@@ -62,6 +64,8 @@ export default async function oAuthCallback(params: {
       throw error
     }
   }
+
+  if (query?.oauth_token) oAuth1TokenStore.removeTokenSecret(query.oauth_token);
 
   try {
     const client = await openidClient(options)
--- a/packages/next-auth/src/core/lib/oauth/oauth1-token-store.ts
+++ b/packages/next-auth/src/core/lib/oauth/oauth1-token-store.ts
@@ -1,0 +1,11 @@
+const _storage = new Map();
+
+const oAuth1TokenStore = {
+  setTokenSecret: (token, tokenSecret) => _storage.set(token, tokenSecret),
+  getTokenSecret: function (token) { 
+    return _storage.get(token); 
+  },
+  removeTokenSecret: (token) => _storage.delete(token),
+}
+
+export default oAuth1TokenStore;


balazsorban44 (changes requested):
Thanks, I believe the choice was made as Twitter OAuth 1.0 worked that way, but I haven't checked in a while.

A few notes:

The `next-auth` library is in maintenance mode so we try to keep changes minimal, and those should be opened against the `v4` branch, not `main`. We can verify if this does not break current providers (Twitter is the only OAuth 1 provider) and could merge the PR against `v4` for backwards compatibility. cc @ThangHuuVu 

On the other hand, OAuth 1 is an outdated standard at this point, most providers are migrating away from it, and I would not really recommend migrating to it in new projects. 🤔 In our new `@auth/core` library, OAuth 1 won't even be supported, so that you are aware.

StachowiakDawid:
Thanks for your answer.

> and those should be opened against the v4 branch, not main
Ok, I overwrote my main branch with next-auth/v4 branch and changed the pull request base. I will update my branch when needed.
> On the other hand, OAuth 1 is an outdated standard at this point, most providers are migrating away from it, and I would not really recommend migrating to it in new projects.
I am aware of this, but OAuth 2 support in USOS is not relesead yet, they are only looking for beta testers. And USOS is more suitable for us than Google.

ThangHuuVu:
hi @StachowiakDawid, I'm looking a way to test this change, could you give more information about which provider are you using, so I can reproduce the issue and verify the fix? 🙌

StachowiakDawid:
> hi @StachowiakDawid, I'm looking a way to test this change, could you give more information about which provider are you using, so I can reproduce the issue and verify the fix? 🙌

You can't verify it yourself, because to do so you need an account in University Study-Oriented System and to have one you need to be a student/lecturer in any university from this list: https://apps.usos.edu.pl/developers/api/definitions/installations/
But I can assure you this code works with USOS and is already used on production in our app which uses Strapi and Next.js (with own patched Next Auth build).
Here is a snippet of PHP code for auth via USOS: https://github.com/ITE-2021/usos-oauth/blob/main/index.php
(later I'm going to share snippets for Strapi, Next-Auth and node-oauth too).
Here is an official documention, but it is not correct: https://apps.usos.pwr.edu.pl/developers/api/services/oauth/

You can check if my changes don't break Legacy Twitter support.

StachowiakDawid:
```
import { OAuthUserConfig, OAuthConfig } from "next-auth/providers";

export interface UsosProfile extends Record<string, any> {
  first_name: string;
  last_name: string;
  email: string;
  student_number: string;
}

export default function UsosProvider<P extends UsosProfile>(
  options: OAuthUserConfig<P>
): OAuthConfig<P> {
  return {
    id: "usos",
    name: "Usos",
    type: "oauth",
    version: "1.0",
    authorization: {
      url: "https://apps.usos.pwr.edu.pl/services/oauth/authorize",
    },
    accessTokenUrl: "https://apps.usos.pwr.edu.pl/services/oauth/access_token",
    requestTokenUrl:
      "https://apps.usos.pwr.edu.pl/services/oauth/request_token?scopes=studies|email",
    profileUrl:
      "https://apps.usos.pwr.edu.pl/services/users/user?fields=first_name|last_name|sex|student_number|email",
    userinfo: {
      url: "https://apps.usos.pwr.edu.pl/services/users/user?fields=first_name|last_name|sex|student_number|email",
    },
    profile(profile) {
      return {
        id: profile.student_number,
        name: profile.first_name + " " + profile.last_name,
        email: profile.email,
      };
    },
    options,
  };
}
```
This is how UsosProvider looks like in our app.

StachowiakDawid:
```
  async usos({ accessToken, query }) {
    const OAuth = require('oauth').OAuth;
    const client = new OAuth("",
      "",
      process.env['USOS_CLIENT_ID'],
      process.env['USOS_CLIENT_SECRET'],
      "1.0",
      null,
      "HMAC-SHA1");
    return new Promise((resolve, reject) => {
      client.getProtectedResource("https://apps.usos.pwr.edu.pl/services/users/user?fields=student_number|email", "GET",
        query.oauth_token, query.oauth_token_secret, (error, data) => {
          if (error) {
            reject(error);
          } else {
            data = JSON.parse(data);
            resolve({
              username: data.student_number,
              email: data.email,
            })
          }
        });
    });
```
Here is code used in Strapi

ndom91:
Some quick googling shows that USOS is testing OAuth 2.0 in beta as well - https://apps.usos.uj.edu.pl/developers/api/services/oauth2/. Is that an option for you?

StachowiakDawid:
> Some quick googling shows that USOS is testing OAuth 2.0 in beta as well - https://apps.usos.uj.edu.pl/developers/api/services/oauth2/. Is that an option for you?

We're not going to use it in a beta stage. It may break compatibility and it requires asking us our Uni (and Uni asking USOS admins) for permission, which will take time.

ThangHuuVu:
I can confirm that the current Twitter provider doesn't break - just tested it locally
However since Twitter - our only official provider doesn't need this, I'm a bit hesitant to merge this extra change for your custom OAuth 1.0 provider. Could you try the link @ndom91 suggested?

StachowiakDawid:
I've just improved it to pass the pipeline.

> I can confirm that the current Twitter provider doesn't break - just tested it locally However since Twitter - our only official provider doesn't need this, I'm a bit hesitant to merge this extra change for your custom OAuth 1.0 provider.

I understand hesitancy to merge PR from random dev, but I believe that passing `null` instead of `oauth_token_secret` is an unneeded obstacle to the full https://github.com/ciaranj/node-oauth functionality and is partially incompatible with what NextAuth docs claim:

> NextAuth.js is designed to work with any OAuth service, it supports OAuth 1.0, 1.0A

I know that USOS is not a popular provider, but both node-oauth and official PHP OAuth library are compatible with it by default, so it isn't some totally deranged OAuth implementation.

If you aren't going to merge it, please close this PR and we will make own build system for the fork and host it in an own repository manager to have latest updates included.

ThangHuuVu:
After looking at the [RFC](https://www.rfc-editor.org/rfc/rfc5849#section-2.1) and source code of `node-oauth`, passing `oauth_token_secret` is a reasonable enhancement for our OAuth 1.0 to me 👍 @balazsorban44 if you agree, let's merge this PR
Basically it is used to sign the subsequent requests to the server by `node-oauth`

StachowiakDawid:
Thank you for the merge.

PR #6484, opened 2023-01-23
Include an example for accessing the MongoDB _id in the session object when using the MongoDB Adapter

doemser:
We teach next-auth with the MongoDB adapter as part of our bootcamps. 
We usually need the _id. 
Here's how to get it.


balazsorban44:
It's generally a bad idea or at the least unnecessary to expose the entire user to the client.

Exposing extra fields (like the id) should already be explained in the callbacks documentation: https://next-auth.js.org/configuration/callbacks#session-callback

PR #6524, opened 2023-01-27
Make next 12.1.X compatible

alxhotel:
## ☕️ Reasoning

I'm using `next@12.1.x` and I would like to use the latest `next-auth` features. The current version of `next-auth` is compatible with it, so why not include it.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [X] Ready to be merged

## 🎫 Affected issues

When installing with `next@12.1.x` throws an error because of the `peerDependencies`.

### DIFF
--- a/packages/next-auth/package.json
+++ b/packages/next-auth/package.json
@@ -77,7 +77,7 @@
     "uuid": "^8.3.2"
   },
   "peerDependencies": {
-    "next": "^12.2.5 || ^13",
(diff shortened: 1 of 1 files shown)


balazsorban44:
Hi, we wouldn't like to add support back to older versions, since the Middleware signature has changed, and that would introduce more code. You can just upgrade Next.js to a newer version.

PR #6535, opened 2023-01-28
feat: remove `unstable_` prefix `getServerSession`

balazsorban44:
The API has been working well in production for a while. This PR removes the warning. We do not remove the `unstable_gerServerSession` yet for backwards compatibility, but it is now shown as deprecated.


kirtan-desai:
What is the reasoning behind this warning?

Issue #6231, opened 2022-12-30
Use AUTH_SECRET as a dynamic variable in the sveltekit package


ptrap:
### Description 📓

In the sveltekit package, AUTH_SECRET is required to be present on buildtime. When you want to bundle your package in a docker container, you need to compile this variable into your container. 

https://github.com/nextauthjs/next-auth/blob/39c78f27b590b0f33b50aa08a04138bfddd5290c/packages/frameworks-sveltekit/src/lib/index.ts#L86

Changing this to a Private dynamic environment variable will allow to pass the AUTH_SECRET to be injected in the docker container environment and will avoid shipping the actual AUTH_SECRET in the docker image.

### How to reproduce ☕️

https://github.com/nextauthjs/next-auth/blob/39c78f27b590b0f33b50aa08a04138bfddd5290c/packages/frameworks-sveltekit/src/lib/index.ts#L86

Changing this to a dynamic import

### Contributing 🙌🏽

No, I am afraid I cannot help regarding this

PR #6548, opened 2023-01-29
fix(sveltekit): use `$env/static/private` in `framework-sveltekit`

### LINKED ISSUE #6439: Auth.js for SvelteKit doesn't work with Vercel Edge Functions
### Environment

npx envinfo --system --binaries --browsers --npmPackages "@sveltejs/kit,@sveltejs/adapter-vercel,@auth/core,@auth/sveltekit" 

  System:
    OS: macOS 13.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 72.55 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 19.3.0 - /opt/homebrew/bin/node
    Yarn: 1.22.18 - /opt/homebrew/bin/yarn
    npm: 9.2.0 - /opt/homebrew/bin/npm
  Browsers:
    Firefox Developer Edition: 110.0
    Safari: 16.2
  npmPackages:
    @auth/core: ^0.2.5 => 0.2.5 
    @auth/sveltekit: ^0.1.12 => 0.1.12 
    @sveltejs/adapter-vercel: ^1.0.5 => 1.0.5 
    @sveltejs/kit: ^1.0.0 => 1.1.1 

### Reproduction URL

https://github.com/evangelos-ch/sveltekit-auth-example

### Describe the issue

I'd like to use Vercel's Edge Functions, which SvelteKit does support as demonstrated in this example project: <https://vercel.com/templates/svelte/sveltekit-edge-functions>.

When I set up my project that way though and I also use Auth.js for Discord authentication, deployment breaks on Vercel, whereas it works just fine before and also just fine if I use `@sveltejs/adapter-auto` and do not set the edge flag.

The error from Auth.js is `Host must be trusted`, which breaks the entire site and not just any auth flow stuff. I tried to resolve this error with `AUTH_TRUST_HOST` or `NEXTAUTH_HOST` env variables to no avail. But after all I shouldn't need them on Vercel, right?

Then I tried deploying your SvelteKit example to Vercel and modifying it for Discord auth, and voila it worked just fine with the same variables as my other project. Then when I changed it over to Edge Functions, it broke with the same behaviour: Host must be trusted errors.

### How to reproduce

1. Deploy the project from my Reproduction URL to Vercel
2. Set the env variables `DISCORD_CLIENT_SECRET`, `DISCORD_CLIENT_ID` and `AUTH_SECRET` (or change them around to e.g. GitHub and then modify the appropriate files in the repo for another provider to test)
3. Redeploy since the build would've failed the first time due to the missing variables

### Expected behavior

The deployment should work, unless Edge incompatibility is some intended or unavoidable limitation of Auth.js, which should then be documented somewhere.

ThangHuuVu:
> _NOTE_:
>

## ☕️ Reasoning

Make our environment variables available in the build step as static so they can be used in the edge functions. 

## 🧢 Checklist

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: #6439 

## 📌 Resources

### DIFF
--- a/packages/frameworks-sveltekit/.env
+++ b/packages/frameworks-sveltekit/.env
@@ -1,0 +1,4 @@
+# These environment variables are used by Sveltekit to inject into `$env/static/private`
+AUTH_SECRET=
+VERCEL=
+AUTH_TRUST_HOST=
--- a/packages/frameworks-sveltekit/.env
+++ b/packages/frameworks-sveltekit/.env
@@ -1,4 +1,0 @@
-# These environment variables are used by Sveltekit to inject into `$env/static/private`
-AUTH_SECRET=
-VERCEL=
-AUTH_TRUST_HOST=
--- a/.gitignore
+++ b/.gitignore
@@ -96,6 +96,5 @@ packages/frameworks-sveltekit/index.*
 packages/frameworks-sveltekit/client.*
 packages/frameworks-sveltekit/.svelte-kit
 packages/frameworks-sveltekit/package
-!packages/frameworks-sveltekit/.env
 packages/frameworks-sveltekit/vite.config.js.timestamp-*
 packages/frameworks-sveltekit/vite.config.ts.timestamp-*
--- a/.gitignore
+++ b/.gitignore
@@ -96,5 +96,6 @@ packages/frameworks-sveltekit/index.*
 packages/frameworks-sveltekit/client.*
 packages/frameworks-sveltekit/.svelte-kit
 packages/frameworks-sveltekit/package
+!packages/frameworks-sveltekit/.env
 packages/frameworks-sveltekit/vite.config.js.timestamp-*
 packages/frameworks-sveltekit/vite.config.ts.timestamp-*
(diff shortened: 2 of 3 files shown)


packages/frameworks-sveltekit/.env
```diff
@@ -0,0 +1,4 @@
+# These environment variables are used by Sveltekit to inject into `$env/static/private`
```
balazsorban44, line 1: Why is this file added to the package itself?

balazsorban44 (changes requested):
Hmm. 😬 This would contradict #6231.

The only way I could see this get solved is if we add an `env` config option to `SvelteKitAuthConfig` and there you can pass all the necessary variables. This would require an extra step from the user though.

packages/frameworks-sveltekit/.env
ThangHuuVu: Sorry, I used TS `declare` instead 😅

ThangHuuVu:
Closing in favor of the upstream fix linked in the issue

PR #6550, opened 2023-01-29
fix: inaccurate discord types + misc

Zamiell:
## ☕️ Reasoning

- Redid the types in the Discord object + added JSDoc comments to match the documentation.
- Fixed a few typos in code comments.
- Fixed the build on Windows computers.

Open questions:

1) Can someone explain how I can actually access the new fields provided in this PR?
2) Why is the discord.ts file completely duplicated in two separate places? Should I duplicate everything in the PR to both places? Seems like a mistake in the codebase.

### DIFF
--- a/packages/core/src/lib/oauth/callback.ts
+++ b/packages/core/src/lib/oauth/callback.ts
@@ -203,7 +203,7 @@ async function getProfile(
     // If we didn't get a response either there was a problem with the provider
     // response *or* the user cancelled the action with the provider.
     //
-    // Unfortuately, we can't tell which - at least not in a way that works for
+    // Unfortunately, we can't tell which - at least not in a way that works for
     // all providers, so we return an empty object; the user should then be
     // redirected back to the sign up page. We log the error to help developers
     // who might be trying to debug this when configuring a new provider.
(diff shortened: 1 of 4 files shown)


balazsorban44 (approved):
Thanks! I applied some suggestions, otherwise looks good!
To answer your questions:
1. https://github.com/nextauthjs/next-auth/blob/1bde7cc8df8d56a37de5f575efa4b11a7c7de488/packages/core/src/providers/oauth.ts#L134-L141 Therefore, we should not return anything else than the defaults in the built-in provider. I reverted those changes. If you want to save extra fields in your database, just define your own `profile()` callback in your `Discord({})` config object.
2. If you refer to https://github.com/nextauthjs/next-auth/tree/main/packages/next-auth/src/providers, the development of `next-auth` has been moved to a diferrent ([`v4`](https://github.com/nextauthjs/next-auth/tree/v4/packages/next-auth)) branch, and we were thinking about removing it on `main` to avoid confusion, but we still have a few references to that location. New changes (like in this PR) are welcome to be done against the `core` package as you did! :wink:

Zamiell:
Ok, thank you balazsorban44.

> If you want to save extra fields in your database, just define your own profile() callback in your Discord({}) config object.

Can we please document that somewhere with a nice example? I'm a brand-new user to Auth.js and I'm finding that extremely non-obvious, even after spending hours both pouring through the documentation + the source code.

PR #6571, opened 2023-01-31
Fix a recursive JSON.stringify

ottomated:
> _NOTE_:
>

## ☕️ Reasoning

Because this modified e.params in place, every time `normalizeEndpoint` is called, it recursively calls JSON.stringify and the url gets exponentially bigger, eventually causing an OOM.

## 🧢 Checklist

- [ ] Documentation (N/A)
- [ ] Tests (N/A)
- [x] Ready to be merged

## 🎫 Affected issues

None

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/providers.ts
+++ b/packages/core/src/lib/providers.ts
@@ -96,9 +96,14 @@ function normalizeEndpoint(
   // NOTE: This need to be checked when constructing the URL
   // for the authorization, token and userinfo endpoints.
   const url = new URL(e?.url ?? "https://authjs.dev")
-  for (const k in e?.params) {
-    if (e?.params && k === "claims") e.params[k] = JSON.stringify(e.params[k])
-    url.searchParams.set(k, e?.params[k])
+  if (e?.params) {
+    for (const k in e.params) {
+      if (k === "claims") {
+        url.searchParams.set(k, JSON.stringify(e.params[k]))
+      } else {
+        url.searchParams.set(k, e.params[k])
+      }
+    }
   }
   return { url, request: e?.request, conform: e?.conform }
 }


packages/core/src/lib/providers.ts
```diff
@@ -96,9 +96,14 @@ function normalizeEndpoint(
   // NOTE: This need to be checked when constructing the URL
   // for the authorization, token and userinfo endpoints.
   const url = new URL(e?.url ?? "https://authjs.dev")
-  for (const k in e?.params) {
-    if (e?.params && k === "claims") e.params[k] = JSON.stringify(e.params[k])
-    url.searchParams.set(k, e?.params[k])
+  if (e?.params) {
+    for (const k in e.params) {
```
balazsorban44, line 100: I don't think this change is necessary, `for..of` will only iterate if `e?.params` is an iterable object, which `undefined` isn't.

balazsorban44 (commented):
Thanks, could you give an example/reproduction where you have seen this issue?

packages/core/src/lib/providers.ts
ottomated: This check tells typescript that `e` is defined on lines 102 and 104.

balazsorban44: It's unnecessary as per my comment. You cannot enter the for loop where `e.params` is not defined.

balazsorban44:
Let's open an issue for this first. 👍

packages/core/src/lib/providers.ts
balazsorban44: hmm. current source code might suggest otherwise https://github.com/nextauthjs/next-auth/blob/326eadf0edad2a49a92545df44623304cc521958/packages/core/src/lib/providers.ts#L100

Will double check when on a computer (I believe TS should be smart enough here), but still, I would like to take a closer look at where this is causing an OOM issue. 🤔

PR #6589, opened 2023-02-01
Add success handler to getServerSideProps

ChinonsoIg:
## ☕️ Reasoning

When implementing OAuth sign in with multiple providers. The documentation did show how to map all the providers and render it in the signin page, which is great.
After the user clicks on a provider to sign in, it did not state how the user will know if the sign in was successful. Hence, I added a code that checks for that in the getServerSideProps. It gives the user (especially a new user) a better sense of direction.

## 🧢 Checklist

- [✔️ ] Documentation

## 🎫 Affected issues

None.

### DIFF
--- a/docs/docs/configuration/pages.md
+++ b/docs/docs/configuration/pages.md
@@ -79,6 +79,8 @@ In order to get the available authentication providers and the URLs to use for t
 
 ```jsx title="pages/auth/signin.js"
 import { getProviders, signIn } from "next-auth/react"
+import { getServerSession } from "next-auth/next"
+import { authOptions } from "../../pages/api/auth/[...nextauth]";
 
 export default function SignIn({ providers }) {
   return (
@@ -95,8 +97,8 @@ export default function SignIn({ providers }) {
 }
 
 export async function getServerSideProps(context) {
-  const { req } = context;
-  const session = await getSession({ req });
+  const session = await getServerSession(context.req, context.res, authOptions);
+  
   // If sign in is successful, redirect to homepage
   if (session) {
     return {
--- a/docs/docs/configuration/pages.md
+++ b/docs/docs/configuration/pages.md
@@ -95,7 +95,17 @@ export default function SignIn({ providers }) {
 }
 
 export async function getServerSideProps(context) {
-  const providers = await getProviders()
+  const { req } = context;
+  const session = await getSession({ req });
+  const providers = await getProviders(context);
+
+  // If sign in is successful, redirect to homepage
+  if (session) {
+    return {
+      redirect: { destination: "/" },
+    };
+  }
+  
   return {
     props: { providers },
   }
--- a/docs/docs/configuration/pages.md
+++ b/docs/docs/configuration/pages.md
@@ -99,13 +99,11 @@ export default function SignIn({ providers }) {
 export async function getServerSideProps(context) {
   const session = await getServerSession(context.req, context.res, authOptions);
   
-  // If the user is already logged in, redirect,
-  // unless the destination is on the same page,
-  // to prevent infinite redirects
-  const destination = "/"
-  // Note, this does not check hash (#) or search params (?)
-  if (session && destination !== authOptions.pages?.signIn) {
-    return { redirect: { destination } };
+  // If the user is already logged in, redirect.
+  // Note: Make sure not to redirect to the same page
+  // To avoid an infinite loop!
+  if (session) {
+    return { redirect: { destination: "/" } };
   }
 
   const providers = await getProviders(context);
--- a/docs/docs/configuration/pages.md
+++ b/docs/docs/configuration/pages.md
@@ -97,14 +97,14 @@ export default function SignIn({ providers }) {
 export async function getServerSideProps(context) {
   const { req } = context;
   const session = await getSession({ req });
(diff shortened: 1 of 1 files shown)


docs/docs/configuration/pages.md
```diff
@@ -95,7 +95,17 @@ export default function SignIn({ providers }) {
 }
 
 export async function getServerSideProps(context) {
-  const providers = await getProviders()
+  const { req } = context;
+  const session = await getSession({ req });
```
balazsorban44, line 99: Let's use `getServerSession`. Will also need the import as it seems to be missing: 
https://github.com/ChinonsoIg/next-auth/blob/0cef6f0eac11259eb8946c8f40ff68ea61262729/docs/docs/configuration/pages.md?plain=1#L81

https://next-auth.js.org/configuration/nextjs#getserversession

docs/docs/configuration/pages.md
```diff
@@ -95,7 +95,17 @@ export default function SignIn({ providers }) {
...
+  const providers = await getProviders(context);
+
+  // If sign in is successful, redirect to homepage
+  if (session) {
+    return {
+      redirect: { destination: "/" },
+    };
+  }
```
balazsorban44, line 107:
```suggestion
  // If sign in is successful, redirect to homepage
  if (session) {
    return {
      redirect: { destination: "/" },
    };
  }

  const providers = await getProviders(context);
```

Only get providers if there was no session

docs/docs/configuration/pages.md
ChinonsoIg: Is `getServerSession` stable enough.
I tried it on my local environment and got this error message, "Module not found: Can't resolve 'v8'".
I tried resolving it, but no headway yet.

balazsorban44: Yes, it has been for months, we recently removed the `unstable_` flag. The way it worked was always stable, I just wasn't sure about the API, so we kept the rights to change it. But since it has worked well in production for a long time, we could just remove the prefix.

You got that `v8` error because you were using it in the wrong context. `getServerSession` - as the name suggests - is meant for the server, and you were trying to use it somewhere in the browser. You could open an issue for that, but still, in this PR, we should use `getServerSession`.

ChinonsoIg: Ooh, I get it now.

docs/docs/configuration/pages.md
```diff
@@ -79,6 +79,8 @@ In order to get the available authentication providers and the URLs to use for t
 
 ```jsx title="pages/auth/signin.js"
 import { getProviders, signIn } from "next-auth/react"
+import { getServerSession } from "next-auth/next"
+import { authOptions } from "../../pages/api/auth/[...nextauth]";
```
balazsorban44, line 83:
```suggestion
import { authOptions } from "../api/auth/[...nextauth]";
```

balazsorban44 (approved):

docs/docs/configuration/pages.md
```diff
@@ -95,7 +97,17 @@ export default function SignIn({ providers }) {
 }
 
 export async function getServerSideProps(context) {
-  const providers = await getProviders()
+  const session = await getServerSession(context.req, context.res, authOptions);
+  
+  // If sign in is successful, redirect to homepage
+  if (session) {
```
balazsorban44, line 103: Actually, we should probably also check that the destination is not the same as the sign-in page, otherwise, you might end up in an infinite loop.

docs/docs/configuration/pages.md
```diff
@@ -95,7 +97,17 @@ export default function SignIn({ providers }) {
...
+    return {
+      redirect: { destination: "/" },
+    };
+  }
```
balazsorban44, line 107:
```suggestion
  // If the user is already logged in, redirect,
  // unless the destination is on the same page,
  // to prevent infinite redirects
  const destination = "/"
  if (session && destination !== authOptions.pages?.signIn) {
    return { redirect: { destination } };
  }
```

balazsorban44 (approved):
~Thanks, approved with a small addition. We need to check that the destination is not on the same page, to avoid an infinite redirect loop. The example only checks if the path matches. If [`authOptions.pages.signIn`](https://next-auth.js.org/configuration/pages) has a hash or query param, that should still not redirect, but I did not add that logic here since I assume it's less common.~

Added a comment rather, to just simply not redirect to the sign-in page by accident.

docs/docs/configuration/pages.md
```diff
@@ -95,7 +97,19 @@ export default function SignIn({ providers }) {
...
+  const session = await getServerSession(context.req, context.res, authOptions);
+  
+  // If the user is already logged in, redirect,
+  // unless the destination is on the same page,
+  // to prevent infinite redirects
+  const destination = "/"
+  // Note, this does not check hash (#) or search params (?)
+  if (session && destination !== authOptions.pages?.signIn) {
+    return { redirect: { destination } };
+  }
```
balazsorban44, line 109:
```suggestion
  // If the user is already logged in, redirect.
  // Note: Make sure not to redirect to the same page
  // To avoid an infinite loop!
  if (session) {
    return { redirect: { destination: "/" } };
  }
```

PR #6595, opened 2023-02-02
fix(ts): stop using typeof + generic together

balazsorban44:
#6560 introduced a TS regression. Apparently using `typeof`+ passing a generic is invalid TS:

https://stackoverflow.com/questions/50005595/getting-the-return-type-of-a-function-which-uses-generics


CSFlorin:
If it compiled, it's valid no?

PR #6608, opened 2023-02-03
docs: api reference restructure

balazsorban44:
Clean up the structure under /reference. Each module is its own section, listing the main and all submodules, in each module, it lists all functions, interfaces, types, etc.

This will greatly improve discoverability. Huge thanks to @tgreyuk for taking the feedback to figure out the best format and implementing it in `typedoc-plugin-markdown`. :pray:

Disclaimer: Even though we are still figuring out some minor formatting changes, I think the new format is so much better than the current one, that I want to go with it as soon as possible.


ThangHuuVu (approved):
LGTM, I'm not 100% sure about the configs changes though - I'm going to need sometime playing with them by myself later.

PR #6615, opened 2023-02-04
fix: export default the email provider

### LINKED ISSUE #6497: SvelteKitAuth: No Call signature for Email in @auth/core/providers/email
### Environment

  System:
    OS: Linux 6.0 Pop!_OS 22.04 LTS
    CPU: (8) x64 Intel(R) Core(TM) i5-10210U CPU @ 1.60GHz
    Memory: 2.70 GB / 7.60 GB
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 18.12.1 - /usr/bin/node
    npm: 9.2.0 - /usr/bin/npm
  Browsers:
    Firefox: 108.0

### Reproduction URL

https://github.com/nextauthjs/sveltekit-auth-example

### Describe the issue

his expression is not callable.
  Type 'typeof import(".../node_modules/@auth/core/providers/email")' has no call signatures.ts(2349)

--> build error:
TypeError: __vite_ssr_import_4__.default.default is not a function at /home/.../src/hooks.server.ts:47:17 at async instantiateModule (file:///home/.../node_modules/vite/dist/node/chunks/dep-5e7f419b.js:52224:9)

### How to reproduce

   1. Clone the repo
   2. Install the repo
   3.  npm run dev

### Expected behavior

expected build to complete without errors.

OrJDev:
> _NOTE_:
>

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources

Closes https://github.com/nextauthjs/next-auth/issues/6497

### DIFF
--- a/packages/core/src/providers/email.ts
+++ b/packages/core/src/providers/email.ts
@@ -111,6 +111,8 @@ export function Email(config: EmailConfig): EmailConfig {
   }
 }
 
+export default Email
+
 /**
  * Email HTML body
  * Insert invisible space into domains from being turned into a hyperlink by email


packages/core/src/providers/email.ts
```diff
@@ -111,6 +111,8 @@ export function Email(config: EmailConfig): EmailConfig {
   }
 }
 
+export default Email
```
binaryme, line 114: I tried hardcoding this into my node_modules/@auth/core/src/providers/email.ts but the error persist

OrJDev: ye because it should be email.js not .ts

binaryme: What else could I be missing? 
I'm sure I added that change to 
`@auth/core/providers/email.js`

OrJDev: > What else could I be missing? I'm sure I added that change to `@auth/core/providers/email.js`

the issue is that i wasn't default exported so it throws this issue, i tried this using solid-start (vite as-well) and it seemed to throw the exact same issue, after adding this line to email.js this issue disappears

OrJDev: > 
make sure you add this to email.js and restart the dev server.

binaryme (approved):

balazsorban44 (commented):
Thanks. Although going from default to a named export only on this provider went a bit unnoticed and we should have done it more holistically (for all providers at the same time), my opinion is that we should use as few default exports as possible, and as such, we should rather change this in other providers, and not revert this one.

If for nothing else, but to get auto-completion in IDEs like VSCode.

I would definitely not like to have both a named and a default export though.

What do you think?

binaryme:
I can confirm I can use the login with email when using the named export Email in the providers array

OrJDev:
> Thanks. Although going from default to a named export only on this provider went a bit unnoticed and we should have done it more holistically (for all providers at the same time), my opinion is that we should use as few default exports as possible, and as such, we should rather change this in other providers, and not revert this one.
> 
> 
> 
> If for nothing else, but to get auto-completion in IDEs like VSCode.
> 
> 
> 
> I would definitely not like to have both a named and a default export though.
> 
> 
> 
> What do you think?

I also think there should only be 1 export, 
IDEs like vsCode give you auto completion for both named and default export, so it depends if you want to keep the old nextAuth api reference or move over, this ofc will confuse previous users and there seems to be nothing wrong with it being default exported so maybe we should just stick to default?

canastro:
Don't you have an issue with typescript requiring you to pass "id", "type", "name" and "sendVerificationRequest"? It doesnt seem like the export is only thing that is currently wrong.

Issue #5308, opened 2022-09-08
Help on callback url


geniusit:
### Question 💬

Using custom OAuth provider, does the callback url **must** respect this format : [origin]/api/auth/callback/[provider] ?

As described [here](https://next-auth.js.org/configuration/providers/oauth#how-to)

May I use the callbackurl of my choice like http://localhost:3000/foo ?

As described [here](https://next-auth.js.org/getting-started/client#specifying-a-callbackurl)

Here is my conf : 

```js
providers: [
    {
        id: 'customprovider',
        name: 'Customprovider',
        type: 'oauth',
        wellKnown: 'https://mydomain.fr/openam/oauth2/IDP/.well-known/openid-configuration',
        authorization: { params: { scope: "profile" }},
        clientId: "myclientid",
    }
  ],
```

I have this button : 

```js
<button onClick={() => signIn("sncf", { callbackUrl: 'http://localhost:3000/oidc-redirect_uri' })}>Sign in</button>
```

When I click on it, the redirect_uri query params is : &redirect_uri=http://localhost:3000/api/auth/callback/customprovider

why not &redirect_uri=http://localhost:3000/oidc-redirect_uri

For custom oauth provider, the callback url seems to be ignored ... why ? 

### How to reproduce ☕️

Create a sample project with custom oauth provider definition

### Contributing 🙌🏽

No, I am afraid I cannot help regarding this

balazsorban44:
Hi, the `callbackUrl` is the URl you want the user to be redirected to after login, and not the same as the callback URL for the Identity Provider used during the OAuth sign-in flow.

You can set `callbackUrl` to anything, by default it will return the user to the same page they started on.

See the docs: https://next-auth.js.org/getting-started/client#specifying-a-callbackurl


We might want to rename `callbackUrl` to something like `redirectTo` in the future to avoid this confusion.

Issue #5305, opened 2022-09-07
Custom OAuth Provider doesn't use the the well-known authorization_endpoint


geniusit:
### Provider type

Custom provider

### Environment

 System:
    OS: Windows 10 10.0.19044
    CPU: (8) x64 Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
    Memory: 3.13 GB / 15.86 GB
  Binaries:
    Node: 16.15.1 - C:\Program Files\nodejs\node.EXE
    npm: 8.11.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 104.0.5112.102
    Edge: Spartan (44.19041.1266.0), Chromium (105.0.1343.27)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    next: 12.2.5 => 12.2.5
    next-auth: ^4.10.3 => 4.10.3
    react: 18.2.0 => 18.2.0

### Reproduction URL

https://github.com/geniusit/nextauthissue

### Describe the issue

Using a custom OAuthProvider, the wellKnown URL value doesn't seem to be used.
When I click on the SignIn button, Im redirected to the default built in custom provider login page with this URL : 
http://localhost:3000/api/auth/signin?callbackUrl=http%3A%2F%2Flocalhost%3A8080%2Foidc-redirect_uri

![image](https://user-images.githubusercontent.com/3659037/188959801-d9d629fc-28d1-4bf5-9e22-106120f03c2f.png)

 

### How to reproduce

1) Create a nextjs application and add the nextauth dependency
2) Create the [...nextauth].js file
3) Add a custom provider : 

```js
import NextAuth from "next-auth"
export default NextAuth({
  providers: [
    {
        id: 'myprovider',
        name: 'MyProvider',
        type: 'oauth',
        version: "2.0",
        wellKnown: 'https://mycustomdomain.com/openam/oauth2/IDP/.well-known/openid-configuration',
        authorization: { params: { scope: "profile", response_type: "code", client_id: "myclientid", redirect_uri: "http://localhost:8080/oidc-redirect_uri" } }
    }
  ],
})
```

4) Go to localhost:3000 and click to the signin button

### Expected behavior

For SSO, I expect when I click on the SignIn button to be redirected to the "authorization_endpoint" of the well-known json object.
After I click to authorize, I need to be redirected to the redirect_uri

balazsorban44:
[...]

PR #6621, opened 2023-02-05
Respect callbackUrl if provided to signIn

feynmanliang:
> _NOTE_:
>

## ☕️ Reasoning

I am using auth.js with sveltekit. When I call `signIn(providerId, { callbackUrl })` (https://github.com/nextauthjs/next-auth/blob/main/packages/frameworks-sveltekit/src/lib/client.ts#L19-L30)with an oauth provider, I expect the user to be redirected to `callbackUrl`  after provider authn.

Currently, the actual behavior is that they always get redirected to `${request.url}/callback/${providerId}`: https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/lib/providers.ts#L32.

The `redirect_uri` controls where redirect after provider authn goes, and it is set equal to the `provider.callbackUrl` https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/lib/oauth/authorization-url.ts#L49. The root cause is that the `provider.callbackUrl`  is lost during `AuthInternal` handler initialization because `parseProviders` ignores `callbackUrl` when instantiating providers (https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/lib/init.ts#L55-L59).

One fix is suggested by this PR: when a `callbackUrl` is passed to `init()`, thread it through to `parseProviders` so that it is ultimately used for `redirect_uri`. When it is `undefined`, default back to previous behavior to maintain backwards compatibility.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/init.ts
+++ b/packages/core/src/lib/init.ts
@@ -55,6 +55,7 @@ export async function init({
   const { providers, provider } = parseProviders({
     providers: authOptions.providers,
     url,
+    callbackUrl: reqCallbackUrl,
     providerId,
   })
 
--- a/packages/core/src/lib/providers.ts
+++ b/packages/core/src/lib/providers.ts
@@ -16,12 +16,13 @@ import type {
 export default function parseProviders(params: {
   providers: Provider[]
   url: URL
+  callbackUrl: string | undefined
   providerId?: string
 }): {
   providers: InternalProvider[]
   provider?: InternalProvider
 } {
-  const { url, providerId } = params
+  const { url, callbackUrl, providerId } = params
 
   const providers = params.providers.map((provider) => {
     const { options: userOptions, ...defaults } = provider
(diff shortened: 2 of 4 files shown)


balazsorban44:
Thanks, but this is actually on purpose. We want to control the callback endpoint.

In this case, using the request's callback URL (from either search params or body) would not be safe, as it hasn't been verified at that point in the code. It's happening in https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/lib/callback-url.ts#L14

The client-side `callbackUrl` option should not control where the OAuth provider sends the response in the OAuth code/token exchange. It is intended to do a redirect at the end of it, which should already happen. The default is to redirect back to the same page the user was currently on when clicking the sign-in button.

If this behavior is incorrect in your experience, please open a bug report first with a reproduction, so we can verify.

FWIW, this confusion has been made before, and we are trying to make this clear by renaming the `callbackUrl` option. See #5305, #5308 and #6580

Small note: `next-auth`'s development branch is https://github.com/nextauthjs/next-auth/tree/v4/packages/next-auth, and at this point, we would only like to accept bug fixes anyway.

feynmanliang:
Hmm... how do we handle the case of a tunneled connection (e.g. GitHub codespaces)? In this situation, the `req.url` is `localhost` but it is exposed via reverse proxy and I need oauth to redirect to that external URL not `localhost`.

We already control where the OAuth provider redirects via the authorized callback URLs, so is safety actually an issue?

PR #6340, opened 2023-01-09
feat(providers): Add BankID Norge provider

balazsorban44:
image
image

PR #6713, opened 2023-02-13
feat(core/providers): Add Vipps Login Provider

itschip:
> _NOTE_:
>

## ☕️ Reasoning
This PR adds Vipps Login API to next-auth. One of the most popular login methods used in Norway.

Completes #6424 (Credits to @endrekrohn for the initial PR)

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues
None

## 📌 Resources


packages/core/src/lib/oauth/callback.ts
```diff
@@ -142,8 +142,14 @@ export async function handleOAuth(
       throw new Error("TODO: Handle OIDC response body error")
     }
 
-    profile = o.getValidatedIdTokenClaims(result)
+    // profile = o.getValidatedIdTokenClaims(result)
     tokens = result
+
+    if (provider.userinfo?.request) {
```
ThangHuuVu, line 148: I don't think changes in this file are required at all 🤔 Could you revert?

packages/core/src/providers/vipps.ts
```diff
@@ -0,0 +1,90 @@
...
+): OIDCConfig<P> {
+  options.issuer ??= "https://api.vipps.no/access-management-1.0/access"
+  return {
+    id: "vipps",
+    name: "Vipps",
+    type: "oidc",
+    client: {
+      token_endpoint_auth_method: "client_secret_post",
+    },
+    authorization: {
```
ThangHuuVu, line 52: Since this is an OIDC provider can we just use the discovery endpoint? https://vippsas.github.io/vipps-developer-docs/docs/APIs/login-api/vipps-login-api#manual-integration

packages/core/src/providers/vipps.ts
```diff
@@ -0,0 +1,90 @@
...
+  email_verified: boolean
+  family_name: string
+  given_name: string
+  name: string
+  phone_number: string
+  sub: string
+}
+
+interface AdditonalConfig {
+  redirectUri?: string
```
ThangHuuVu, line 17: This isn't used anywhere

ThangHuuVu (changes requested):
Thanks, I added some comments. Besides, could you add an option to the [issue template](https://github.com/nextauthjs/next-auth/blob/a5e0db4bb31cb1a1e10a86098c7c80560e289461/.github/ISSUE_TEMPLATE/2_bug_provider.yml)? 🙏

packages/core/src/lib/oauth/callback.ts
itschip: Yeah 😅 I looked at another PR implementing BankID provider, and added after having some trouble. Forgot to remove it :)

packages/core/src/providers/vipps.ts
itschip: Although, from what I can tell, we do need the `issuer` and `token_endpoint_auth_method` anyways.

packages/core/src/providers/vipps.ts
```diff
@@ -0,0 +1,90 @@
...
+    name: "Vipps",
+    type: "oidc",
+    client: {
+      token_endpoint_auth_method: "client_secret_post",
+    },
+    authorization: {
+      params: {
+        scope: "openid name email",
+        client_id: options.clientId,
+        response_type: "code",
```
ThangHuuVu, line 56: client_id and response_type is automatically added, so I think you can remove them here

packages/core/src/providers/vipps.ts
```diff
@@ -0,0 +1,90 @@
...
+      },
+    },
+    userinfo: {
```
ThangHuuVu, line 59: Sorry, I meant you could use the `wellKnown` option with `idToken: true`, described here: https://authjs.dev/reference/providers/oauth#userinfo-option
since this is an OIDC-compliant provider, we should be able to extract the user info without making a request to that endpoint

packages/core/src/providers/vipps.ts
```diff
@@ -0,0 +1,90 @@
...
+            headers: {
+              Authorization: `Bearer ${tokens.access_token}`,
+            },
+          }
+        )
+
+        return await response.json()
+      },
+    },
+    profile(profile) {
```
ThangHuuVu, line 73: profile isn't needed afaik

packages/core/src/providers/vipps.ts
itschip: Last time I checked idToken wasn't type on OIDCConfig. Should I add it?

itschip:
Ok, so after https://github.com/nextauthjs/next-auth/pull/6713/commits/ace052b5a31bb2bc0043ccfe20d9bc31bb17c5c7 I seem to get the decoded id_token as the OAuthProfile. Is that correct?

itschip:
However, if I bring back the changes from https://github.com/nextauthjs/next-auth/pull/6713/commits/ab8da6c32aedd2f8fbc50bb39484e1c23eee724c

as well as including a custom request for the userinfo and adding the profile function back, I get "actual" profile data with email, name..etc

@balazsorban44 If I may ask, what was the reason for https://github.com/nextauthjs/next-auth/pull/6340/commits/740a2db07f5a88b347fdc53f96098f350c09c122, as that is where I took it from :)

balazsorban44 (commented):
I would like to test this locally before merging. I'm currently struggling with Vipps Support to check why my redirect url is not accepted. 🙏

The Vipps provider is a standard OIDC provider, so `userinfo.request` should be unnecessary as it's reserved for non-compliant providers as an escape hatch. I'm looking into this. 👍

itschip:
> I would like to test this locally before merging. I'm currently struggling with Vipps Support to check why my redirect url is not accepted. 🙏
> 
> The Vipps provider is a standard OIDC provider, so `userinfo.request` should be unnecessary as it's reserved for non-compliant providers as an escape hatch. I'm looking into this. 👍

Thank you! Let me know if I can help you with anything :)

itschip:
@balazsorban44 did you resolve your issue with the redirect url? :)

itschip:
@balazsorban44 Do you think it is fine to just use 'oauth' for now. I am currently using that for my project now, and it works fine. If not I'll try once more with 'oidc', even though I have not had any luck with it yet.

Gregor23021 (approved):

Issue #6716, opened 2023-02-14
No sign-in options are passed as query string params at all?


khteh:
### Provider type

Identity Server 4

### Environment

  System:
    OS: Linux 5.19 Ubuntu 22.10 22.10 (Kinetic Kudu)
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12800H
    Memory: 49.72 GB / 62.48 GB
    Container: Yes
    Shell: 5.2.2 - /bin/bash
  Binaries:
    Node: 19.6.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 9.4.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 110.0.5481.96
    Firefox: 109.0.1

### Reproduction URL

http://localhost:3000

### Describe the issue

Manage to reach my auth server but the logs show there is no sign-in options passed as query string at all. Without the query string the auth server wouldn't know where to redirect to after login.

### How to reproduce

Run the example app

### Expected behavior

Send sign-in options as query string to auth server.

balazsorban44:
https://next-auth.js.org/getting-started/client#specifying-a-callbackurl

khteh:
> https://next-auth.js.org/getting-started/client#specifying-a-callbackurl

The callbackUrl specifies to which URL the user will be redirected after signing in. **Defaults to the page URL the sign-in is initiated from** but the authentication server does NOT receive this?

balazsorban44:
The AS does not need to receive it either. The expected URL you have to set in the AS is defined here:

https://next-auth.js.org/getting-started/rest-api#getpost-apiauthcallbackprovider

For example for GitHub, it would be `http://localhost:3000/api/auth/callback/github`

Then, when you are redirected back, NextAuth.js will make sure you land on the page you defined in `callbackUrl`.

If you want to pass other params to the authorization endpoint, you can pass them through via the third `signIn` argument https://next-auth.js.org/getting-started/client#additional-parameters.

PR #6726, opened 2023-02-15
fix: pass user-supplied provider.authorization.params.state to service provider and signIn callback (#5727)

shirish87:
## ☕️ Reasoning

* Value supplied in `[...nextauth].ts` `provider.authorization.params.state` is currently not supplied to the provider service
* PR based on suggestion in comment (https://github.com/nextauthjs/next-auth/issues/5727#issuecomment-1304666934)
* Allowing `state` parameter to be available in the `signIn` callback helps decide post-login steps (redirect to specific page based on state?)

## 🧢 Checklist

Documentation for the `signIn` callback will need to be updated for the additional `state` parameter. I'll update it if that change is acceptable, since this may need discussion.
Since the `state` parameter in that callback will only be available for OAuth providers, adding it there may be confusing?

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

- #5727 
- If I understand correctly, maybe #6716

Fixes: https://github.com/nextauthjs/next-auth/issues/5727

### DIFF
--- a/packages/next-auth/src/core/lib/oauth/authorization-url.ts
+++ b/packages/next-auth/src/core/lib/oauth/authorization-url.ts
@@ -54,7 +54,7 @@ export default async function getAuthorizationUrl({
   const authorizationParams: AuthorizationParameters = params
   const cookies: Cookie[] = []
 
-  const state = await createState(options)
+  const state = await createState(options, authorizationParams.state)
   if (state) {
     authorizationParams.state = state.value
     cookies.push(state.cookie)
--- a/packages/next-auth/src/core/lib/oauth/callback.ts
+++ b/packages/next-auth/src/core/lib/oauth/callback.ts
@@ -56,7 +56,7 @@ export default async function oAuthCallback(params: {
       }
 
       const newProfile = await getProfile({ profile, tokens, provider, logger })
-      return { ...newProfile, cookies: [] }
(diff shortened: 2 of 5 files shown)


septatrix:
I am not sure that passing state as a string is the best or most secure option. The state is supposed to be a nonce, when users can directly pass a string it makes it easy to miss this. In most cases it should be just as good to accept an object and merge it with a nonce generated by this library internally. This approach was also suggested by @jasonkuhrt (the author of the original issue) in https://github.com/nextauthjs/next-auth/issues/5727#issuecomment-1305612236.

This could then be implemented as a callback where one is able to return a state object (or maybe augment one generated by the library with an already generated nonce). If people really want to control the raw state we might allow returning a string instead of an object. If a string is returned it is used as-is, however, when an object is returned, we base64 encode the JSON and ideally also sign it.

Merging this PR as-is would likely still not solve all use-cases for which manual control over state is desired or at least make implementations way harder and more error-prone. Also the current design does not seem to lend itself to be easily backwards compatible to allow changes as required for the more user friendly API.

balazsorban44:
Closing since the code has significantly changed. v4 is now on a different branch https://github.com/nextauthjs/next-auth/tree/v4/packages/next-auth

PR #6739, opened 2023-02-16
fix: compatibility with edge runtimes

### LINKED ISSUE #6736: @auth/solid-start does not work in the Vercel edge runtime
### Environment

| Package           | Version |
|-------------------|---------|
| @auth/core        | 0.4.0   |
| @auth/solid-start | 0.1.0   |
| solid-start       | 0.2.21  |
| vite              | 4.1.1   |

### Reproduction URL

N/A

### Describe the issue

When building a solid-start project that imports something from `@auth/solid-start` on Vercel with the edge runtime enabled, Vercel will fail the build with the following error message.

```
Error: The Edge Function "render" is referencing unsupported modules:
--
12:12:32.820 | - vercel/path0/packages/app/dist/index.mjs: crypto

```

I tracked the issue down to the `@panva/hkdf` project, and if I modify your package to exclude that import the build succeeds.

Looking into the hkdf package I noticed that it seems to be doing things correctly. More specifically, they have multiple types of exports. I would expect the **browser** export to be selected here, but my application is actually built with the **import** export. This is the root cause of the breaking build.

```jsonc
{
  "exports": {
    ".": {
      "types": "./dist/types/index.d.ts",
      "browser": "./dist/web/index.js", // This one should be used...
      "import": "./dist/node/esm/index.js", // ... but this one is used instead
      "require": "./dist/node/cjs/index.js"
    },
    "./package.json": "./package.json"
  }
}
  ```

Now, I'm not experienced enough with publishing libraries to know if this is something you can fix on your end, or if it somehow needs to be fixed in the build setup of solid-start/vite, but I figured it would be nice to track it in your issues since it will seem like the issue comes from your package from a user perspective.

### How to reproduce

* Bootstrap a new solid-start project
* Enable SSR and set up the vercel adapter with edge runtime enabled
* Import `@auth/solid-start` somewhere in your application (I did in a middleware in the `entry-server.tsx` file).
* Push build to vercel
* Failed build  

### Expected behavior

The project should build successfully on Vercels edge runtime

jakst:
## ☕️ Reasoning

The underlying change in @panva/hkdf is adding a `worker` entry to `package.json#exports`, so it does not fall back to `imports` which points to a node module.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: #6736

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -61,7 +61,7 @@
   },
   "license": "ISC",
   "dependencies": {
-    "@panva/hkdf": "^1.0.2",
+    "@panva/hkdf": "^1.0.3",
     "cookie": "0.5.0",
     "jose": "^4.11.1",
     "oauth4webapi": "^2.0.6",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -61,7 +61,7 @@
   },
   "license": "ISC",
   "dependencies": {
-    "@panva/hkdf": "^1.0.3",
+    "@panva/hkdf": "^1.0.4",
     "cookie": "0.5.0",
     "jose": "^4.11.1",
     "oauth4webapi": "^2.0.6",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -136,7 +136,7 @@ importers:
       vercel: ^23.1.2
     dependencies:
       dotenv: 16.0.3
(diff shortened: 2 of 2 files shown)


balazsorban44:
Thank you for tracking the issue down, and fixing it both here and in the underlying package! :green_heart:

jakst:
My pleasure!

PR #6742, opened 2023-02-17
fix(providers): add default user agent for GitHub

### LINKED ISSUE #6741: GitHub provider broken on cloudflare, missing user-agent header
### Provider type

GitHub

### Environment

The important part of the environment in this case is that the code is deployed on cloudflare wrangler workers or pages (dev environment) or cloudflare worker or pages production.

  System:
    OS: macOS 13.1
    CPU: (8) arm64 Apple M1
    Memory: 70.20 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.13.0 - ~/.nvm/versions/node/v18.13.0/bin/node
    npm: 8.19.3 - ~/.nvm/versions/node/v18.13.0/bin/npm
  Browsers:
    Firefox: 105.0.1
    Safari: 16.2

### Reproduction URL

http://localhost:5173 and http://127.0.0.1:8788 or as noted in "How to reproduce" below.

### Describe the issue

Users will see errors during the callback processing phase of logging in when clicking on the GitHub button in the sveltekit demo.   If you turn logging levels to debug, it looks like an error in the undici library, however what is really happening is that GitHub is responding with an error page stating that no user-agent http header was provided in the request to get a token during the callback phase of processing.   The fetch implementation in cloudflare wrangler (undici) and the cloudflare worker platform (unknown) doesnt provide a default user-agent.  The fetch implementation used by vite / node (which is undici again) does provide a default header, so the user will not see this issue when running `npm run dev` in the example.

### How to reproduce

First, sorry this is lengthy, there are a lot of details to set up to make sure that the sveltekit demo is working with `npm run dev` and not working with cloudflare wrangler.   Please give this a read before starting.
1.  Make sure you are up to date with next_auth on the main branch.
2. Make a new branch just so that its easy to clean up when you are done.
3. in a terminal `cd apps/examples/sveltekit`
4. run `npm install`
5. configure your environment variables per the instructions in the demo, this requires an OAuth application setup on GitHub.  Make a copy of .env.example to .env and fill in all of the variables.   Remember that you will need to change the homepage url and the redirect url on the GitHub OAuth configuration page to whichever server you are starting below (npm run dev vs wrangler), I got http://localhost:5173 for running `npm run dev` and http://127.0.0.1:8788 for running with wrangler.  Some dummy values will work fine for basic setup.   The localhost vs 127.0.0.1 difference as well as the port difference is important , make sure that the host and callback urls all match.  [Here](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/creating-an-oauth-app) is a  page to help you set up a GitHub OAuth app.  Your callback url should be of the form http://host:port/auth/callback/github.  Note that the official next_auth docs will instruct you to put it in the form http://host:port/api/auth/callback/github, the api fragment is incorrect for the current codebase.  Keep the OAuth app page up in your browser, you will have to make changes.  The changes there have always been instantaneous for me during my debugging.
6. Make src/hooks.server.ts look like the following:
```
import { SvelteKitAuth } from "@auth/sveltekit"
import GitHub from "@auth/core/providers/github"
import { GITHUB_ID, GITHUB_SECRET, AUTH_SECRET } from "$env/static/private"

  debug:true,
  trustHost: true,
  secret: AUTH_SECRET,
  providers: [GitHub({ clientId: GITHUB_ID, clientSecret: GITHUB_SECRET })],
})
```

7. Start the vite dev server: `npm run dev`.  Make note of the hostname and port and that it corresponds to the settings in your GitHub OAuth app.
8. Use your browser to make sure that you can login and don't get any errors.  (everything should work fine if you follow the instructions in the example)
9. Log out on the example app in the browser and `ctrl-c` the terminal to quit the example app.
10. Install the cloudflare sveltekit adapter in the example dir, `npm i --save-dev @sveltejs/adapter-cloudflare`
11. Perform the following replacement in svelte.config.js (read like a diff)
```
- import adapter from '@sveltejs/adapter-auto;
+ import adapter from '@sveltejs/adapter-cloudflare';
```
12. If you don't have it, install cloudflare wrangler. https://developers.cloudflare.com/workers/wrangler/install-and-update/
13. Build the application for production `npm run build`.  This will tell vite to build the application and the use of adapter-cloudflare will build a production asset `.svelte-kit/cloudflare/_worker.js` which is suitable for testing with wrangler.
14. Launch the app with wrangler `wrangler pages dev --local .svelte-kit/cloudflare`. Press 'b' in your terminal to launch a browser. Don't forget to change the GitHub OAuth page homepage and redirect urls to match your browser.  If your browser launched to localhost, change your browser to 127.0.0.1:[same port] and reload and make the GitHub OAuth page match.  Im not sure why but 127.0.0.1 seems to work with wrangler always.
15. Run through the login process, it will error when you click on the GitHub button.  The browser will show you a generic 'Error' page.  It just says 'Error'.   Go look at your terminal and you will see some debug output.   It will look like some kind of premature TLS connection error.  That's a ruse.   I spent hours trying to debug it thinking it was something to do with lower level networking.   In reality, the fetch implementation in the GitHub provider has two bare fetch calls (not via oauth4webapi) and those calls aren't sending a user-agent header to GitHub.  Github is responding with a webpage telling the user that it needs a user-agent.   The callback processing code sees this as an error and tells undici to kill the network connection.
16. At this point if you would like to see it work, you can make 2 simple changes to the GitHub provider code inside of the node_modules directory of the example.  In node_modules/@auth/core/providers/github.js look for the 2 fetch calls and add a user-agent header to both `headers: { Authorization: `Bearer ${tokens.access_token}`, 'User-Agent': 'undici' },`.   Rerun `npm run build` to produce a new production asset and restart wrangler like step 14 above.  The demo app will now work.  I have tested this to make sure it works on the production cloudflare platform and it is broken without the headers and works with the headers.

NOTE: It is tempting to think you can setup in the dev version of the svelte demo in apps/dev/sveltekit with the appropriate environment variables and code changes to the app in hooks.server.js etc... and then make changes in the next_auth source of the GitHub provider and rebuild the whole project and test that this works.  Unfortunately there is a bug with running `pnpm run build` in the dev/sveltekit code.  pnpm is doing something that messes with the build process where it can't find the 'crypto' library.  I have filed a bug for this already.  Because it's not buildable with pnpm it's not testable in this way :(.   I am going to submit a PR for this and attach it to this issue but it won't be testable within that PR.

### Expected behavior

The user should not see an error page when clicking on the GitHub button in the svelte example, they should be redirected to the homepage and it should show them as logged in.

jschlesser:
> _NOTE_:
>

## ☕️ Reasoning

This PR adds user-agent headers for the GitHub provider because certain `fetch()` implementations don't provide a default and GitHub rest apis require it.   This specifically will fix the provider for running the svelte examples on the cloudflare platform.  It should fix it for other platforms with similar `fetch()` implementations.

## 🧢 Checklist

- [ x] Documentation (no doc changes needed)
- [ ] Tests (see test information in the linked [Issue](https://github.com/nextauthjs/next-auth/issues/6741)
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: [Issue](https://github.com/nextauthjs/next-auth/issues/6741)

## 📌 Resources

### DIFF
--- a/packages/core/src/providers/github.ts
+++ b/packages/core/src/providers/github.ts
@@ -129,14 +129,14 @@ export default function GitHub(
       url: "https://api.github.com/user",
       async request({ tokens, provider }) {
         const profile = await fetch(provider.userinfo?.url as URL, {
-          headers: { Authorization: `Bearer ${tokens.access_token}`, 'User-Agent': 'undici' },
+          headers: { Authorization: `Bearer ${tokens.access_token}`, 'User-Agent': 'authjs' },
         }).then(async (res) => await res.json())
 
         if (!profile.email) {
           // If the user does not have a public email, get another via the GitHub API
           // See https://docs.github.com/en/rest/users/emails#list-public-email-addresses-for-the-authenticated-user
           const res = await fetch("https://api.github.com/user/emails", {
-            headers: { Authorization: `Bearer ${tokens.access_token}`, 'User-Agent': 'undici' },
+            headers: { Authorization: `Bearer ${tokens.access_token}`, 'User-Agent': 'authjs' },
           })
 
           if (res.ok) {
--- a/packages/core/src/providers/github.ts
+++ b/packages/core/src/providers/github.ts
@@ -129,14 +129,14 @@ export default function GitHub(
       url: "https://api.github.com/user",
       async request({ tokens, provider }) {
         const profile = await fetch(provider.userinfo?.url as URL, {
-          headers: { Authorization: `Bearer ${tokens.access_token}` },
+          headers: { Authorization: `Bearer ${tokens.access_token}`, 'User-Agent': 'undici' },
         }).then(async (res) => await res.json())
 
         if (!profile.email) {
           // If the user does not have a public email, get another via the GitHub API
           // See https://docs.github.com/en/rest/users/emails#list-public-email-addresses-for-the-authenticated-user
           const res = await fetch("https://api.github.com/user/emails", {
-            headers: { Authorization: `Bearer ${tokens.access_token}` },
+            headers: { Authorization: `Bearer ${tokens.access_token}`, 'User-Agent': 'undici' },
           })
 
           if (res.ok) {


balazsorban44 (commented):
I'm a bit hesitant to add special handling for certain providers. <picture data-single-emoji=":thinking:" title=":thinking:">:thinking:</picture>, in this case even when the config is confined to the GitHub provider.

But I found the docs at least: https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#user-agent-required

and it does not look like we can work around it. :confused:

Even if we end up adding a user-agent, it should probably not be `"undici"` as we cannot guarantee that the runtime uses Undici for the request. something like `"authjs"` might be more appropriate, I believe.

jschlesser:
I changed the user-agent header from undici to authjs.

kpbaseten:
Any other thoughts, @balazsorban44 - keen to get github provider working with cloudflare.

balazsorban44 (approved):

PR #6744, opened 2023-02-17
added optional dynamic sveltekit options

### LINKED ISSUE #6743: Allow the @auth/sveltekit framework to use platform variables for config data
### Description 📓

The current @auth/sveltekit framework assumes that config variables will be available outside of the scope of the hooks handler.  Note that in the code below, the SvelteKitAuth function is returning a handler function that is generated anonymously in the framework.

```
import { GITHUB_ID, GITHUB_SECRET } from "$env/static/private"

  providers: [GitHub({ clientId: GITHUB_ID, clientSecret: GITHUB_SECRET })],
})
```

Some platforms, notably cloudflare for my case, provide environment variables via the platform properties which are only available in the hooks handler via the event parameter.  Note that the sveltekit hooks Handler has a signature of:
```
export interface Handle {
	(input: {
		event: RequestEvent;
		resolve(event: RequestEvent, opts?: ResolveOptions): MaybePromise<Response>;
	}): MaybePromise<Response>;
}
```

Here is a [link](https://kit.svelte.dev/docs/adapter-cloudflare) to the cloudflare sveltekit docs for environment variables as a reference.

Typically to get to platform variables a developer will use `event.platform.env.VARIABLE_NAME`.
Because of the method of generating the config data outside of the hooks handler, the event object isn't available.  I would like to add an optional dynamicOptions parameter that is a function that will generate a config and can have access to the event object.  See proposed changes in the linked PR.

### How to reproduce ☕️

Unfortunately this is a complicated feature to test because there is a bug with running `pnpm run build` in the apps/dev/sveltekit directory.  It causes vite build to fail to resolve the 'crypto' module.  See [issue](https://github.com/nextauthjs/next-auth/issues/6732)  `pnpm run build` is required to test with the cloudflare local development platform "wrangler".   I have 'hacked' the node_modules/@auth/sveltekit module outside of the next_auth development environment and confirmed that the changes do provide access to the platform variables locally on wrangler and also when deployed to cloudflare production.   I have tried a couple of ways to get a publishable test working that looks verifiable.

1.  I tried to import from my personal GitHub fork via a git url but workspaces aren't supported that way.
2. I tried to publish the changes to a local verdaccio npm server but I discovered that the release process is highly customized.  I can't simply run `pnpm run build` and then do regular publish commands.  The workspace declarations in package.json aren't getting converted.    

Perhaps a core contributor could publish a feature branch so that I could normally import @auth/sveltekit from the feature branch?   I could then publish an app to a public GitHub repo that should work just fine.  Note that getting the sveltekit demo working on cloudflare is also dependent on https://github.com/nextauthjs/next-auth/issues/6741 .  I opened that issue and I could consolidate these two if necessary for a feature branch to make testing easier.

### Contributing 🙌🏽

jschlesser:
> _NOTE_:
>

## ☕️ Reasoning

This is a feature enhancement to allow for using sveltekit's 'platform' properties in @auth/sveltekit configuration data.   The structure of the current plugin precludes getting access to this data and it is the preferred way to get environment variables and secrets on the cloudflare platform and perhaps other platforms.  The proposed changes are optional and shouldn't break any previous code.   Testing this is a challenge, please see notes in the linked issue.

Here are the docs from sveltekit and cloudflare, neither shows the whole picture.  You have to see both to get it to work.
[Cloudflare Docs](https://kit.svelte.dev/docs/adapter-cloudflare#environment-variables)
[SvelteKit Docs](https://kit.svelte.dev/docs/adapter-cloudflare#environment-variables)

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: [ISSUE](https://github.com/nextauthjs/next-auth/issues/6743)

## 📌 Resources


jschlesser:
Please review but note that I think it needs a touch more work.
As I noted in the attached issue, after playing around with this solution in a project, I think that the dynamicOptions param needs to optionally be async so that the hook can make some await calls and I need to convert the returned handler to an optional promise.  I'll do some more work to make sure that they types are all correct and it works in both async and non async cases.

jschlesser:
Ready for review now.

packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -237,11 +253,14 @@ const actions: AuthAction[] = [
   "error",
 ]
 
-function AuthHandle(prefix: string, authOptions: AuthConfig): Handle {
-  return function ({ event, resolve }) {
+type OptionsFunction = (event: RequestEvent, options: SvelteKitAuthConfig) => Promise<SvelteKitAuthConfig> | SvelteKitAuthConfig
```
ThangHuuVu, line 256:
```suggestion
type DynamicSvelteKitAuthConfig = (event: RequestEvent, options: SvelteKitAuthConfig) => PromiseLike<SvelteKitAuthConfig>
```

ThangHuuVu, line 256:
```suggestion
type DynamicSvelteKitAuthConfig = (event: RequestEvent, options: SvelteKitAuthConfig) => PromiseLike<SvelteKitAuthConfig>
```

packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -251,19 +270,19 @@ function AuthHandle(prefix: string, authOptions: AuthConfig): Handle {
...
  * The main entry point to `@auth/sveltekit`
  * @see https://sveltekit.authjs.dev
  */
-export function SvelteKitAuth(options: SvelteKitAuthConfig): Handle {
+export function SvelteKitAuth(options: SvelteKitAuthConfig, dynamicOptions?: OptionsFunction): Handle {
   const { prefix = "/auth", ...authOptions } = options
   authOptions.secret ??= env.AUTH_SECRET
   authOptions.trustHost ??= !!(env.AUTH_TRUST_HOST ?? env.VERCEL ?? dev)
-  return AuthHandle(prefix, authOptions)
+  return AuthHandle(prefix, authOptions, dynamicOptions)
```
ThangHuuVu, line 285: IMO we should only keep one parameter for `options`, so it could be typed as `SvelteKitAuthConfig | DynamicSvelteKitAuthConfig`

Let's do it like this:
```ts
export function SvelteKitAuth(
  options: SvelteKitAuthConfig | DynamicSvelteKitAuthConfig
): Handle {
  if (typeof options === "object") {
    options.secret ??= env.AUTH_SECRET
    options.trustHost ??= !!(env.AUTH_TRUST_HOST ?? env.VERCEL ?? dev)
  }
  return AuthHandle(options)
}
```

Since the assignment of `secret` and `trustHost` requires `env`, we have to do it in the function callback anyway.
To keep the `prefix` options working let's move its default value assignment part to inside `AuthHandle`

packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -237,11 +253,14 @@ const actions: AuthAction[] = [
   "error",
 ]
 
-function AuthHandle(prefix: string, authOptions: AuthConfig): Handle {
-  return function ({ event, resolve }) {
+type OptionsFunction = (event: RequestEvent, options: SvelteKitAuthConfig) => Promise<SvelteKitAuthConfig> | SvelteKitAuthConfig
+
+function AuthHandle(prefix: string, authOptions: AuthConfig, dynamicOptions?: OptionsFunction): Handle {
```
ThangHuuVu, line 258:
```suggestion
function AuthHandle(svelteKitAuthOptions: SvelteKitAuthConfig | DynamicSvelteKitAuthConfig): Handle {
```

packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -237,11 +253,14 @@ const actions: AuthAction[] = [
...
+  return async function ({ event, resolve }) {
+    const resolvedAuthOptions: SvelteKitAuthConfig = await dynamicOptions?.(event, authOptions) || authOptions
```
ThangHuuVu, line 260:
```suggestion
    const authOptions =
      typeof svelteKitAuthOptions === "object"
        ? svelteKitAuthOptions
        : await svelteKitAuthOptions(event)
     const { prefix = "/auth" } = authOptions
```

packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -26,6 +26,22 @@
...
+ * import GitHub from "@auth/core/providers/github"
+ * import type { Handle } from "@sveltejs/kit";
+ *
+ * export const handle = SvelteKitAuth({providers: []}, async (event, options) => {
+ *   // await somethingAsync()
+ *   const providers = [GitHub({ clientId: event.platform.env.GITHUB_ID, clientSecret: event.platform.env.GITHUB_SECRET })]
+ *   options.providers = providers
+ *   options.secret = event.platform.env.AUTH_SECRET
+ *   return options
+ * }) satisfies Handle;
```
ThangHuuVu, line 43: The example would be much simpler IMO:
```
 * export const handle = SvelteKitAuth(async (event) => {
 *   const authOptions = {
 *     providers: [GitHub({ clientId: event.platform.env.GITHUB_ID, clientSecret: event.platform.env.GITHUB_SECRET })]
 *     secret: event.platform.env.AUTH_SECRET,
 *     trustHost: true
 *   }
 *   return authOptions
 * }) satisfies Handle;
```

packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -237,11 +253,14 @@ const actions: AuthAction[] = [
...
-  return function ({ event, resolve }) {
+type OptionsFunction = (event: RequestEvent, options: SvelteKitAuthConfig) => Promise<SvelteKitAuthConfig> | SvelteKitAuthConfig
+
+function AuthHandle(prefix: string, authOptions: AuthConfig, dynamicOptions?: OptionsFunction): Handle {
+  return async function ({ event, resolve }) {
+    const resolvedAuthOptions: SvelteKitAuthConfig = await dynamicOptions?.(event, authOptions) || authOptions
     const { url, request } = event
 
-    event.locals.getSession ??= () => getSession(request, authOptions)
+    event.locals.getSession ??= () => getSession(request, resolvedAuthOptions)
```
ThangHuuVu, line 263:
```suggestion
    event.locals.getSession ??= () => getSession(request, authOptions)
```

packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -251,19 +270,19 @@ function AuthHandle(prefix: string, authOptions: AuthConfig): Handle {
       return resolve(event)
     }
 
-    return Auth(request, authOptions)
+    return Auth(request, resolvedAuthOptions)
```
ThangHuuVu, line 273:
```suggestion
    return Auth(request, authOptions)
```

ThangHuuVu (changes requested):
Thank you for sending the PR. IMO let's keep the configuration clean to avoid writing two `authOptions` in the parameters. Let me know what you think 🙌

packages/frameworks-sveltekit/src/lib/index.ts
jschlesser: Sounds good.  I like it, it's clean.  I pushed a commit that keeps the api signature, keeps the default /auth prefix for both static and dynamic options and keeps getting values for secret and TrustHost from env in the static configuration.   I have done some basic end to end testing locally and it works in my environment.   I think it achieves what you want.

packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -259,11 +283,14 @@ function AuthHandle(prefix: string, authOptions: AuthConfig): Handle {
...
-  authOptions.secret ??= env.AUTH_SECRET
-  authOptions.trustHost ??= !!(env.AUTH_TRUST_HOST ?? env.VERCEL ?? dev)
-  return AuthHandle(prefix, authOptions)
+export function SvelteKitAuth(options: SvelteKitAuthConfig | DynamicSvelteKitAuthConfig): Handle {
+  if(typeof options === "object") {
+    options.secret ??= env.AUTH_SECRET
+    options.trustHost ??= !!(env.AUTH_TRUST_HOST ?? env.VERCEL ?? dev)
+    return AuthHandle(options)
+  } else {
+    return AuthHandle(options)
```
ThangHuuVu, line 292: Can we have only one return statement here? 👀

jschlesser: Fixed

ThangHuuVu (approved):
LGTM. Thanks again @jschlesser 🙌

packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -259,11 +283,12 @@ function AuthHandle(prefix: string, authOptions: AuthConfig): Handle {
  * The main entry point to `@auth/sveltekit`
  * @see https://sveltekit.authjs.dev
  */
-export function SvelteKitAuth(options: SvelteKitAuthConfig): Handle {
-  const { prefix = "/auth", ...authOptions } = options
-  authOptions.secret ??= env.AUTH_SECRET
-  authOptions.trustHost ??= !!(env.AUTH_TRUST_HOST ?? env.VERCEL ?? dev)
-  return AuthHandle(prefix, authOptions)
+export function SvelteKitAuth(options: SvelteKitAuthConfig | DynamicSvelteKitAuthConfig): Handle {
+  if(typeof options === "object") {
```
ThangHuuVu, line 287:
```suggestion
  if (typeof options === "object") {
```

packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -259,11 +283,12 @@ function AuthHandle(prefix: string, authOptions: AuthConfig): Handle {
...
+    options.secret ??= env.AUTH_SECRET
+    options.trustHost ??= !!(env.AUTH_TRUST_HOST ?? env.VERCEL ?? dev)
+  }
+    return AuthHandle(options)
```
ThangHuuVu, line 291:
```suggestion
  return AuthHandle(options)
```

PR #6767, opened 2023-02-20
@auth/remix framework for using @auth in remix

acoreyj:
## ☕️ Reasoning

Integrate auth.js into a remix project

## 🧢 Checklist

- [X] Documentation
- [ ] Tests
- [ ] Ready to be merged

Note this is working in my testing but setup a good example project and merge I need at least the first and preferably both of the following remix issues fixed/merged

https://github.com/remix-run/remix/pull/4148
https://github.com/remix-run/remix/pull/3330

But I'm reopening this MR because some people have expressed interest in using the code

## 📌 Resources


balazsorban44:
Thank you! Small note, we are still very much interested, but our priority is on making the core more stable and finishing up the docs migration (https://next-auth.js.org and https://authjs.dev have a lot of shared content right now).

Once that's done, we will have more time to look into new frameworks again! :green_heart:

In the meantime, I was wondering about your opinion on this comment by @sergiodxa: https://github.com/nextauthjs/next-auth/discussions/6246#discussioncomment-4670939, considering that there is https://remix.run/docs/en/v1/utils/sessions

acoreyj:
> Thank you! Small note, we are still very much interested, but our priority is on making the core more stable and finishing up the docs migration (https://next-auth.js.org and https://authjs.dev have a lot of shared content right now).
> 
> Once that's done, we will have more time to look into new frameworks again! 💚
> 
> In the meantime, I was wondering about your opinion on this comment by @sergiodxa: [#6246 (reply in thread)](https://github.com/nextauthjs/next-auth/discussions/6246#discussioncomment-4670939), considering that there is https://remix.run/docs/en/v1/utils/sessions

No problem, it seems silly to merge it when that remix bug means you have to use a patch and for some reason they still haven't merged in the fix.

But yes I agree it would be easier to work with remix if it didn't return a response object and just the info you needed to create the response. Not even mainly to use the built in remix session storage methods, although that could be especially nice for things like cloudflare. But mainly after playing with it a lot and the best way to integrate it, is as I lay out in the README, is using a `fetcher.Form` and you are basically ignoring the html of the response but of course you need the set-cookie. 

This ends up working fine for most things as if it succeeds you can easily show they are signed in. But for errors you can tell it encountered an error via the response code but would need to parse the html response to see if it's something you want to tell the user like if the email already exists in the database.

Also it becomes a pain for strategies that succeed but don't log the user in like with email magic links.

Guria:
@acoreyj just a ping reminder. Is it possible to finish the work you have started here? Do you have an y plans for that?

acoreyj:
@voltuer @Guria 

I can make sure this is working with the latest Remix. Although a bit confused on what Remix is to become as it is merging with React Router so not clear if this makes sense in that context

ndom91:
> @voltuer @Guria
> 
> I can make sure this is working with the latest Remix. Although a bit confused on what Remix is to become as it is merging with React Router so not clear if this makes sense in that context

Yeah it'd be great if you could double check this against the latest versions of Remix!

The way I see it, I can't imagine the framework will disappear in any sense. Shopify bought them as replacement for their homegrown framework and as a competitor to Vercel/next, netlify/gatsby, etc.

ndom91:
You shouldn't touch the package.json version manually. Just set it to `0.0.1`, for example, for the first release. And then once released, subsequent updates are automatically bumped by our CI tooling

okalil:
Maybe the name should change to '@auth/react-router' because Remix will be merged into the upcoming React Router 7

gsaandy:
any plan to get this merged?

PR #6787, opened 2023-02-22
feat(adapters): grafbase

notrab:
## ☕️ Reasoning

This PR adds a new Adapter for [Grafbase](https://grafbase.com).

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## Todo

Since Grafbase uses the CLI when working locally, and the hosted Grafbase in prod, what is the best way to proceed in regards to adding tests?

The CLI runs a local development server by invoking `npx grafbase dev` in the root of this project. It will detect `grafbase/schema.graphql`.

The file `grafbase.schema.graphql` is also used by the `npx grafbase init` command to scaffold a new project (as outlined in the README).

There is one final `TODO` in the code to change in regards to scoped unique fields. This is expected to land in the CLI over the next few days.

Let me know what you think! Thank you.


JohnGrisham:
I would really love to have this as an option 👍

notrab:
Yeah this would have been great but I haven’t been able to get back in touch with @balazsorban44 about next steps.

JohnGrisham:
What's the hold up here @lluia @ndom91?

notrab:
@JohnGrisham it's worth noting that this uses the Grafbase Database which might not be around forever as we expand our DB connectors. This means that this will likely need updating to allow custom queries and mutations that map through to the database connected to the Grafbase API.

I'll give it some thought on how we can update this PR to include support for that as well as the unique constraint. 

I'm still not sure how we go about the tests with this custom adapter.

JohnGrisham:
> I'm still not sure how we go about the tests with this custom adapter.

@notrab I would just do what the other adapters are doing if possible. Such as the one for [dynamodb](https://github.com/nextauthjs/next-auth/blob/main/packages/adapter-dynamodb/tests/index.test.ts).

I guess if the plan is to allow devs to bring their own DB or change the default DB that does complicate things a bit. It might just be a concession that the adapter is only usable when using the default DB for now. I expect most users won't need this kind of integration so that might not be the worst solution.

notrab:
@JohnGrisham we could always propose a `graphql-adapter` that lets you provide your own query and mutations to satisfy the Auth.js adapter API. Wdyt?

JohnGrisham:
@notrab A graphql-adapter would be fine too and add support for not just Grafbase but any endpoint. As long as it would function in the same way as a DB adapter I'm fine with that as well. Another consideration might be performance, I'm not sure if a direct DB connection would be faster or not but I would assume so.

ThangHuuVu:
hi @notrab , sorry for the delay, I was trying to get the PR merged by pushing some commits to format it. May I know why did you close it? together, we can get this PR to the finish line this weekend!

notrab:
@ThangHuuVu thanks for the efforts to get this over the line. In its current state, it would be using the Grafbase Database but since we are moving away from that product offering, it might make more sense to create something more agnostic for GraphQL in general where users can pass their own queries and mutations from the config to do what's necessary.

I will give this some thought and open a new PR with the proposal. Thank you, and sorry for the time wasted this weekend.

ThangHuuVu:
@notrab no worries, feel free to reopen this PR or create a new one!

PR #6875, opened 2023-03-05
docs: Authentik trailing slash information.

thiccaxe:
Mention that trailing slashes in `issuer` may or may not be required.

## ☕️ Reasoning

From https://github.com/nextauthjs/next-auth/pull/4646, we saw that the author mentions that the Authentik Provider appends a trailing slash to `issuer`. While no proof of this has been found at this time (An up-to-date installation of authentik tested with a solid-start application requires a trailing slash), the author's findings may still have some merit. To this end, the documentation now states that the trailing slash may or may not be required.

## 🧢 Checklist

- [X] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/docs/docs/providers/authentik.md
+++ b/docs/docs/providers/authentik.md
@@ -31,5 +31,5 @@ providers: [
 ```
 
 :::note
-`issuer` has been known to sometimes require a trailing slash (e.g., `https://my-authentik-domain.com/application/o/my_slug/`) and others time require no trailing slash (e.g. `https://my-authentik-domain.com/application/o/my_slug`).
+`issuer` requires a trailing slash (e.g. `https://my-authentik-domain.com/application/o/my_slug/`).
 :::
--- a/docs/docs/providers/authentik.md
+++ b/docs/docs/providers/authentik.md
@@ -31,5 +31,5 @@ providers: [
 ```
 
 :::note
-`issuer` should include the slug without a trailing slash – e.g., `https://my-authentik-domain.com/application/o/My_Slug`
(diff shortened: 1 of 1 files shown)


docs/docs/providers/authentik.md
```diff
@@ -31,5 +31,5 @@ providers: [
 ```
 
 :::note
-`issuer` should include the slug without a trailing slash – e.g., `https://my-authentik-domain.com/application/o/My_Slug`
+`issuer` has been known to sometimes require a trailing slash (e.g., `https://my-authentik-domain.com/application/o/my_slug/`) and others time require no trailing slash (e.g. `https://my-authentik-domain.com/application/o/my_slug`).
```
balazsorban44, line 34: I feel that this is too vague and we should find the root cause of why this is happening. if this happens randomly, it sounds like a bug in Authentik. If it's a config option, we should mention that.

balazsorban44 (changes requested):
Let's find the reason and document, rather than giving a non-helpful comment

thiccaxe:
Based on the current state of https://github.com/goauthentik/authentik/blob/main/authentik/providers/oauth2/models.py#L226 and the history of the issuer url logic, Authentik has always mandated a `/`. It seems that further research shows that the author of 4646 may have referred to https://github.com/nextauthjs/next-auth/blob/a4d831d1b990b98db36d2d6ef2b45dcc15a65860/src/providers/authentik.ts#L30 when mentioning that provider adds a `/`. Since this is now obsolete, replaced by https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/providers/authentik.ts, I think the matter has been resolved, in that you require a trailing slash.

Thank you.

PR #6927, opened 2023-03-12
feat: token provider

ThangHuuVu:
## ☕️ Reasoning

This PR introduces the following changes:
- add `tokenId` and `tokenIdVerified` to `AdapterUser`. We set them similar to how we set `email` and `emailVerified`. `email` and `emailVerified` are kept around to keep migration effort low.
- `adapter.getUserByEmail` is changed to `adapter.getUserByTokenId`. This requires a lot of changes to follow in the adapters package. 
- add a `Token` provider. By default, it will throw an error for `sendVerificationToken`, signaling that if the developer use this provider, they should implement this method.
- `Email` provider is now `SmtpEmail` provider with specific settings for SMTP-based email servers. It works just like before, the name changing should not break anything, since this is a default `export`. This is to distinguish with the other HTTP-based email provider that we want to add in the future, like SendGrid.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Addressing #1465 

## 📌 Resources

### DIFF
--- a/apps/dev/nextjs/lib/db.ts
+++ b/apps/dev/nextjs/lib/db.ts
@@ -1,0 +1,237 @@
+import type {
+  Adapter,
+  AdapterUser,
+  AdapterAccount,
+  AdapterSession,
+} from "@auth/core/adapters"
+
+export interface LocalStorageAdapterOptions {
+  /**
+   * The base prefix for your keys
+   */
+  baseKeyPrefix?: string
+  /**
+   * The prefix for the `account` key
+   */
+  accountKeyPrefix?: string
+  /**
+   * The prefix for the `accountByUserId` key
+   */
+  accountByUserIdPrefix?: string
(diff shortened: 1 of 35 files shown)


balazsorban44 (commented):
I was thinking that we could refactor things in a way that the `email` provider essentially just implements the token provider, so we don't need another provider type, just renaming `"email"` to `"token"` internally.

Gregor23021 (approved):

ThangHuuVu:
@balazsorban44 do you think we should 🚢 this with v5? Let me know what you think of the current state of this PR. 👀

philbaker4:
Cross posting [my RFC comment](https://github.com/nextauthjs/next-auth/discussions/1465#discussioncomment-7203869) for reference

PR #7056, opened 2023-03-25
feat: make it possible to update the session

balazsorban44:
Implements https://github.com/nextauthjs/next-auth/discussions/3941#discussioncomment-5426142

- [x] Implementation
- [x] Documentation
- Port to `@auth/core` (can be done in a separate PR)
- [x] Tested locally, we can add one together with `@auth/core`. We don't currently have a good E2E setup. :grimacing:

## 🎫 Affected issues

#596, #2208, #2129


packages/next-auth/src/react/index.tsx
```diff
@@ -456,6 +462,21 @@ export function SessionProvider(props: SessionProviderProps) {
...
         : "unauthenticated",
+      async update(data) {
+        setLoading(true)
+        const session = await fetchData<Session>(
+          "session",
+          __NEXTAUTH,
+          logger,
+          { req: { body: { csrfToken: await getCsrfToken(), data } } }
+        )
+        setLoading(false)
```
remorses, line 473: should call `setLoading(false)` in a `try` `finally` block in case the fetch fails

balazsorban44: `fetchData` already handles fetch related issues.

https://github.com/nextauthjs/next-auth/blob/cd91c03211f23fd8dec270ae911de9facb11b704/packages/next-auth/src/client/_utils.ts#L54C11-L55

packages/next-auth/src/core/types.ts
```diff
@@ -313,30 +313,59 @@ export interface CallbacksOptions<P = Profile, A = Account> {
...
+          /**
+           * When using {@link SessionOptions.strategy} `"jwt"`, this is the data
+           * sent from the client via the [`useSession().update`](https://next-auth.js.org/getting-started/client#update-session) method.
+           *
+           * ⚠ Note, you should validate this data before using it.
+           */
+          session?: any
+          status: "update"
+        }
+      | { token: JWT; isNewUser?: boolean }
```
balazsorban44, line 367: Hitting TypeScript issues, any help is appreciated:

image

packages/next-auth/src/react/index.tsx
```diff
@@ -456,6 +462,21 @@ export function SessionProvider(props: SessionProviderProps) {
...
+        if (session) {
+          setSession(session)
+          broadcast.post({ event: "session", data: { trigger: "getSession" } })
```
ThangHuuVu, line 476: should it be `updateSession` trigger?

balazsorban44: We only have `signOut` and `getSession`:

https://github.com/nextauthjs/next-auth/blob/feaeda9e2a2d6768339951fd4daa8aa2cf20ecbf/packages/next-auth/src/client/_utils.ts#L69

Should not really matter, AFACT, we don't really use this value for anything really. From the other window's perspective, it should also not matter what triggered this, it just n eeds to get the updated session.

packages/next-auth/src/core/types.ts
remorses: You can remove the last `| { token: JWT; isNewUser?: boolean }` to make the type simpler and make typescript discriminated types work

To tell that `status` could be undefined use `status?: "update"`

remorses: I would also deprecate the `isNewUser` field and add another status option with `signUp`

balazsorban44: We don't want to deprecate anything yet, this should be a non-breaking update. But agree, we can add `signUp`. :+1:

remorses:
in my opinion a better design would be `getSession()` always retrigger the `params.jwt` function with the `user` argument (just like a normal signIn)

You don't pass the data to add to the jwt in the client, instead you mutate the `user` object on the database and mutate the jwt inside the callback

An example implementation that adds an `orgId` field to the jwt would be
```ts
// server
  callbacks: {
    jwt({ token, user }) {
      if (user) {
        token.orgId = user.orgId
      }
      return token
    },
  },
}


// client

import { getSession } from 'next-auth/react'

// add orgId to user in database
getSession()
```

benefits:
- no need to add a new `status` argument
- no need to validate data inside the `jwt` callback, data comes from the database
- no need to handle different cases inside the `jwt` callback, making everything simpler

packages/next-auth/src/core/routes/session.ts
```diff
@@ -123,19 +130,16 @@ export default async function session(
...
-            user: {
-              name: user.name,
-              email: user.email,
-              image: user.image,
-            },
+            user: { name: user.name, email: user.email, image: user.image },
             expires: session.expires.toISOString(),
           },
           user,
+          trigger: isUpdate ? "update" : "database",
```
ThangHuuVu, line 141: why do we need `trigger: "database"` or `trigger: "jwt"` ? As far as I concern, if the developers set the strategy, they already know about this info, so it's kinda redundant IMO

balazsorban44: For discriminated union types. trying to make TS happy. Our current typing is not safe as I've seen people trying it use either `user` or `token` when it's not available because of the strategy they were using.

balazsorban44:
@remorses we cannot know if it's the user you are trying to update, or the account, or just the session, or a combination of these.

So for every `getSession` invocation, we would have to query the user, the session and optionally the account (which?). 

By passing the whole object, it's up to the developer to update the correct things in the database.

If you already updated the DB in a different way, you don't have to pass anything to `update()`, just fetch the user yourself from the database and do what you showed.

In addition, I think `getSession` has been a bad decision by us in the past and I want to get away from it. It's not hooking into React correctly (triggers `setState` outside React) for example.

It's essentially just a fetch call anyway, we should not need a special API for that.

On the server, it's already replaced by `getServerSession` for faster reads as well.

IMO there should be no reason for it to exist at all at this point.

packages/next-auth/src/core/routes/session.ts
balazsorban44: Addressed in https://github.com/nextauthjs/next-auth/pull/7056/commits/3c9b49411c72a2afa05c97763cdf4fbbe434c74e

Maybe it's not worth breaking current types for this. We can do it in `@auth/core` instead.

ThangHuuVu (approved):

ajin-iqness:
@ThangHuuVu Could you please tell me when this feature will be available? I am currently using `4.20.1` and this feature not available.

packages/next-auth/src/core/types.ts
```diff
@@ -313,30 +316,87 @@ export interface CallbacksOptions<P = Profile, A = Account> {
...
    */
-  session: (params: {
-    session: Session
-    user: User | AdapterUser
-    token: JWT
-  }) => Awaitable<Session>
+  session: (
+    params:
+      | {
+          session: DefaultSession
```
johnslemmer, line 322: Was changing this to `DefaultSession` necessary? I was under the impression that the `Session` type exists so we could extend it in our own projects to add more data to it.  As documented here: https://next-auth.js.org/getting-started/typescript#extend-default-interface-properties

Like I currently have something like this in my project:

```
declare module "next-auth" {
  interface Session extends DefaultSession {
    user: {
      id: string;
      role: UserRole;
    } & DefaultSession["user"];
  }
}
```

And with the 4.21.0 this makes typescript angry about adding more data to the session.  Can we just switch this back to the `Session` type?  Or is there something I'm not understanding here.

ajin-iqness:
> @ThangHuuVu Could you please tell me when this feature will be available? I am currently using `4.20.1` and this feature not available.

I got this when updated to `4.21.0`  :100:

packages/next-auth/src/core/types.ts
balazsorban44: See my comment here: https://github.com/nextauthjs/next-auth/issues/7132#issuecomment-1493924161 open for constructive thoughts on the issue!

ttebify:
Thank you, @balazsorban44, for addressing this issue. I previously had to resort to some unconventional workarounds. For those interested in integrating this fix, please refer to the updated documentation at: https://next-auth.js.org/getting-started/client#updating-the-session.

rokob:
> * Port to `@auth/core` (can be done in a separate PR)

Is there an issue tracking this or a PR in progress?

marcosriver:
Would it be possible to create something like getSession(), but updateSession() that doesn't need to be imported from a hook and can be imported to update sessions?

It would be helpful to implement things like fetch interceptors for refresh token rotations.

PR #7088, opened 2023-03-27
fix : Type 'string | undefined' is not assignable to type 'string'.

jabedzaman:
## ☕️ Reasoning

fix typescript error of undefined not assignable to clientid and secret

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

N/A

## 📌 Resources


balazsorban44:
Thanks, but this is technically correct as-is. You cannot know for sure that those values have been set in the environment, thus adding `as string` would be a lie.

You would need to add a runtime check like the following instead:
```ts
if (!process.env.GITHUB_ID || !process.env.GITHUB_SECRET) {
  throw new Error("Missing environment variable(s)")
}
```

However I believe this would complicate the example code, but feel free to do this in your own code!

Issue #6920, opened 2023-03-11
upstash redis integration, store users as a json and store user id in the session


FrancescoSaverioZuppichini:
### Description 📓

Hi :wave: 

Upstash redis integration stores the user as a string, making it hard to work with the redis json package. 

![image](https://user-images.githubusercontent.com/15908060/224487047-c8162639-eaff-417e-accc-133b472ba531.png)

Would it be possible to store is as a json?

Moreover, the `user.id` that next auth stores in the session is not the id that upstash redis uses internally 

The id uses in next auth is the user id from GitHub (the provider I am using) 

![image](https://user-images.githubusercontent.com/15908060/224487104-f2ba5149-de1d-46d2-b3e5-149fc25f5d47.png)

Would it be possible to store the actual stored user's id instead?

Thanks a lot :heart: 

Cheers,

Fra

### How to reproduce ☕️

Use the following code and sing in once

```typescript

import NextAuth from "next-auth";
import GithubProvider from "next-auth/providers/github";
import type { NextAuthOptions } from "next-auth";
import { UpstashRedisAdapter } from "@next-auth/upstash-redis-adapter";

import logger from "@/lib/logger";
import { Redis } from "@upstash/redis";

const redis = new Redis({
  url: process.env.UPSTASH_REDIS_URL,
  token: process.env.UPSTASH_REDIS_TOKEN,
});

export const authOptions: NextAuthOptions = {
  adapter: UpstashRedisAdapter(redis),
  providers: [
    GithubProvider({
      clientId: process.env.GITHUB_ID,
      clientSecret: process.env.GITHUB_SECRET,
    }),
  ],
  ```

### Contributing 🙌🏽

Yes, I am willing to help implement this feature in a PR

balazsorban44:
[...]

PR #7124, opened 2023-04-01
feat(adapters): add RedisJSON adapter

akremer:
## ☕️ Reasoning

This adapter enables use of NextAuth with any Redis server that has the RedisJSON module installed. It mirrors the functionality of the existing upstash-redis adapter, but allows use of any Redis server without the need for Upstash.

## 🧢 Checklist

- [X] Documentation
- [X] Tests
- [X] Ready to be merged

## 🎫 Affected issues

Addresses: #6920, #4977

## 📌 Resources


akremer:
@balazsorban44 I know there is a bunch of stuff in flight due to the conversion to auth.js; can you provide an ETA on when this might be able to be merged? Are new adapters on hold?

ndom91:
@akremer are you still wanting to get this merged? We've been discussing [here](https://github.com/nextauthjs/next-auth/pull/8805) whether to support [unstorage](https://unstorage.unjs.io/) as a common KV storage API as it has drivers for all common KV storage adapters, what do you think?

akremer:
I think Unstorage is a great addition but it does not seem to use the RedisJSON module at its core, unless my quick skim of the documentation is wrong. It also seems fixed to using the ioredis adapter for connectivity, which is also not 100% common. 

There are a ton of benefits to using the RedisJSON module in a Redis cluster at scale, so while it seems at a high level a similar use case, it is actually different at the underlying level.

A few additional points/pros/cons:

- Unstorage has a ton of extensibility. This is a great thing,  and most use cases of NextAuth on top of Redis will be more than well covered by Unstorage's implementation. 

- But, for those of us who use Redis at scale, using Unstorage would be a regression compared to the RedisJSON module's much superior JSON handling within Redis. While it is very unlikely an authentication collection would in any way test Redis' capabilities without the use of the RedisJSON module, it seems kind of a waste not to use it if it's already there and being used elsewhere. 

- It would also enforce usage of ioredis which in some cases could make it not usable. 

We then come back to the maintainability question: Is it worth having a RedisJSON adapter and an Unstorage adapter and an Upstash adapter? That is a question only the NextAuth team can answer, but I would posit that they should coexist. That is, after all, the point of having a multitude of adapters and an extensible ecosystem. The JSON api in the RedisJSON module has been stable for many years and it really hasn't changed much, so I expect it might be rarely needing to be updated. But, I also get the broader maintainability aspect. 

For what it's worth, we have been using this code in production for the better part of 2023 and it's been flawless. We will continue to adapt and use it whether it's officially merged or not :-)

ndom91:
@akremer thanks for the detailed answer! I agree keeping a separate redisjson adapter then makes sense as well 👍

ndom91:
@akremer going forward with this PR then, do you mind cleaning up the merge conflicts and we can continue as normal 👍

ndom91:
Looks like it is possible to use a custom driver with unstorage so that RedisJSON could be leveraged, for example: https://github.com/nextauthjs/next-auth/pull/9501/files#diff-ee4275a235cdf5c7bddda8600e8cf1f68ecc043926a19039baa2eba79b84515a :thinking:

ndom91:
@akremer after talking with some of the other folks we've decided to close this PR as its an additional maintenance burden which doesn't add anything that is not already covered by unstorage adapter (see example I posted above).

PR #7145, opened 2023-04-04
chore: move `prettier` and `eslint` configs under `package.json`

SauravMaheshkar:
## ☕️ Reasoning

Changes proposed by this PR can be summarised as follows :-

* Move the content from `.prettierrc` under the `prettier` key in `package.json`. [Source(prettier docs)](https://prettier.io/docs/en/configuration.html)
* Move the content from `.eslintrc.js` under the `eslintConfig` key in `package.json`. [Source(eslint docs)](https://eslint.org/docs/latest/use/configure/configuration-files#using-configuration-files)
* Move the content from `.eslintignore` under the `eslintIgnore` key in `package.json`. [Source(eslint docs)](https://eslint.org/docs/latest/use/configure/ignore#using-eslintignore-in-packagejson)

Doing so, would lead to a more minimal project structure without any functional change.


balazsorban44 (approved):

balazsorban44:
Thanks, @SauravMaheshkar! Do you know if https://github.com/nextauthjs/next-auth/blob/main/.prettierignore could also be moved for completeness? :thinking: 

This is picked up correctly by VSCode, right? I did merge it in good faith. If it breaks though, we might have to revert...

SauravMaheshkar:
Sadly there isn't. I really would've liked that though. I think we could raise a issue over at Prettier. Eslint supports, why not prettier 🤷🏽‍♂️ . 

[Source: Docs](https://prettier.io/docs/en/ignore.html)

PR #7165, opened 2023-04-06
feat(adapters): Drizzle adapter

anthonyshew:
## ☕️ Reasoning

Introducing a `drizzle-orm` adapter!

With `drizzle`'s increasing popularity, it's no surprise that [there's a discussion](https://github.com/nextauthjs/next-auth/discussions/7005) for an adapter. This PR adds it.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources

# Notes
- There is currently an open issue with Drizzle's recent versions that breaks monorepo dependencies. Please use 0.24.x until this issue is resolved: https://github.com/drizzle-team/drizzle-orm/issues/283
- There is currently an open issue with Drizzle truncating millisecond precision. I've added an extra option to the `runBasicTests` utility to be able to provide some custom fixtures to workaround this. https://github.com/drizzle-team/drizzle-orm/pull/668


balazsorban44:
Good point in https://github.com/nextauthjs/next-auth/discussions/7005#discussioncomment-5542019, can we make sure to support multiple SQL flavors? :thinking:

anthonyshew:
@infamous55 Correct! That was a first cut at making things go. I meant to mark it as a draft but forgot to hit the button. 😅

I've got the different flavors on there now!

infamous55:
@anthonyshew Glad to see you added support for the other SQL flavors!

I would also like to suggest having only one Drizzle adapter function that returns the correct type of adapter depending on the database client it receives. Like this:

```ts
export function DrizzleAdapter(db: any): Adapter {
	if (isMySqlDatabase(db)) return MySqlAdapter(db)
	else if (isPgDatabase(db)) return PgAdapter(db)
	else if (isSQLiteDatabase(db)) return SQLiteAdapter(db)
	else throw new Error("Unsupported database")
}
```

```ts
import { MySqlDatabase } from "drizzle-orm/mysql-core"
import { PgDatabase } from "drizzle-orm/pg-core"
import { BaseSQLiteDatabase } from "drizzle-orm/sqlite-core"

export function isMySqlDatabase(db: any): db is AnyMySqlDatabase {
	return db instanceof MySqlDatabase<any, any>
}

export function isPgDatabase(db: any): db is AnyPgDatabase {
	return db instanceof PgDatabase<any, any>
}
	 
export function isSQLiteDatabase(db: any): db is AnySQLiteDatabase {
	return db instanceof BaseSQLiteDatabase<any, any>
}
```

In this case, the user would only have to import `DrizzleAdapter` from `@next-auth/adapter-drizzle`. I find this setup the cleanest, but this is my personal opinion. It's up to you to decide how you want the publishing done.

anthonyshew:
@infamous55 I had considered this as well!

I didn't know how much we cared about compiled size so I did them split up. API convenience vs. bundle sizing tradeoff, a decision I'll let the core maintainers make. 😄 

I've now added another adapter directory (`adapter-drizzle-all`) using your strategy for comparison.

I'll attach bundle sizes below:

| SQLite | Postgres | MySQL | All 3 |
|--------|--------|--------|--------|
| 28k | 28k | 36k | 120k |

Still incredibly small, of course, but don't know how strict about bundle sizes @balazsorban44 wants to be!

To summarize, where this PR currently stands:
- Four adapters currently written (one for each flavor, one with a compatibility later for all three)
- Tests in `adapter-drizzle-sqlite` work
- Documentation is not handled (waiting on API surface decision and then happy to write!)

jeanhdev:
Hey guys - I'd really like to use the adapter for Drizzle with my PSQL DB, any news on this ?

rikimbili:
> Hey guys - I'd really like to use the adapter for Drizzle with my PSQL DB, any news on this ?

As a temporary or permanent solution you can create your own adapter [using the code here from this PR](https://github.com/anthonyshew/next-auth/blob/main/packages/adapter-drizzle-pg/src/index.ts). This way you can also modify how user data is stored and queried from your db if you have something more custom. I tried this approach a week ago and the adapter has been working great.

packages/adapter-drizzle-all/src/sqlite/schema.ts
```diff
@@ -0,0 +1,52 @@
...
+import { migrate } from 'drizzle-orm/better-sqlite3/migrator';
+import { ProviderType } from 'next-auth/providers';
+
+const sqlite = new Database('db.sqlite');
+
+export const users = sqliteTable('users', {
+  id: text('id').notNull().primaryKey(),
+  name: text('name'),
+  email: text("email").notNull(),
+  emailVerified: integer("emailVerified", { mode: "timestamp_ms" }),
```
vilius, line 16: `emailVerifiedAt` would be more consistent

ferntheplant:
> As a temporary or permanent solution you can create your own adapter [using the code here from this PR](https://github.com/anthonyshew/next-auth/blob/main/packages/adapter-drizzle-pg/src/index.ts)

Whenever my `user` model extends the minimal `id, name, email, image` fields I get some pretty silly type errors. For example `adapter.createUser` **exclusively** accepts an object of shape 
```typescript
type Omit<AdapterUser, "id"> = { name?: string | null, email?: string | null, image?: string | null }```

I think it really should accept anything that _extends_ this type instead so I can extend my user schema in a typesafe way

File: packages/adapter-drizzle-all/src/pg/index.ts [hunk 2]
```diff
@@ -0,0 +1,264 @@
...
+ *  id_token: text("id_token"),
+ *  session_state: text("session_state"),
+ * }, (account) => ({
+ *   _: primaryKey(account.provider, account.providerAccountId)
+ * }))
+ *
+ * export const sessions = pgTable("sessions", {
+ *  userId: text("userId").notNull().references(() => users.id, { onDelete: "cascade" }),
+ *  sessionToken: text("sessionToken").notNull().primaryKey(),
+ *  expires: integer("expires").notNull(),
```
ferntheplant, line 90: this and the verificationTokens expires below should be timestamps not integers

packages/adapter-drizzle-all/src/mysql/index.ts
```diff
@@ -0,0 +1,293 @@
...
+ *
+ * ```
+ *
+ **/
+export function MySqlAdapter(client: typeof db): Adapter {
+  return {
+    createUser: async (data) => {
+      await client
+        .insert(users)
+        .values({ ...data, id: "123" })
```
pyrossh, line 118: Can these mocked id's be removed?

packages/adapter-drizzle-all/src/pg/schema.ts
```diff
@@ -0,0 +1,48 @@
+import { integer, timestamp, pgTable, text, primaryKey } from 'drizzle-orm/pg-core';
+import { drizzle } from 'drizzle-orm/postgres-js';
```
pyrossh, line 2: There are serverless drivers as well that needs to be taken into consideration. Ex: I use `drizzle-orm/neon-serverless` driver for postgres instances hosted by neon. More drivers here https://github.com/drizzle-team/drizzle-orm/blob/main/drizzle-orm/src/pg-core/README.md

packages/adapter-drizzle-all/src/mysql/index.ts
```diff
@@ -0,0 +1,293 @@
...
+    },
+    linkAccount: async (rawAccount) => {
+      await client
+        .insert(accounts)
+        .values(rawAccount)
+        .then(res => res[0])
+    },
+    getUserByAccount: async (account) => {
+      return client.select({
+        id: users.id,
```
vilius, line 214: Inconsistency: `getUser` and `getUserByEmail` return whole user record, where `getUserByAccount` returns only selected fields

packages/adapter-drizzle-all/src/mysql/index.ts
```diff
@@ -0,0 +1,293 @@
...
+        .then(res => res[0])
+    },
+    linkAccount: async (rawAccount) => {
+      await client
+        .insert(accounts)
+        .values(rawAccount)
+        .then(res => res[0])
+    },
+    getUserByAccount: async (account) => {
+      return client.select({
```
vilius, line 213: Inconsistency: `getUser` and `getUserByEmail` returns whole user model where `getUserByAccount` returns only partial.

```ts
const userResult =
  client
    .select({
      user: usersTable,
    })
    .from(usersTable)
    .innerJoin(
      accountsTable,
      and(
        eq(accountsTable.providerAccountId, providerAccountId),
        eq(accountsTable.provider, provider),
      ),
    )
    .get() ?? null;

return userResult.then((userResult) => userResult.user);
```

packages/adapter-drizzle-all/src/mysql/index.ts
```diff
@@ -0,0 +1,293 @@
...
+ * });
+ *
+ * export const db = drizzle(pool);
+ *
+ * migrate(db, { migrationsFolder: "./drizzle" })
+ *
+ * ```
+ *
+ **/
+export function MySqlAdapter(client: typeof db): Adapter {
```
ThangHuuVu, line 113: I think we need to have the schema objects passed in the Adapter, like this:
```suggestion
export function MySqlAdapter(client: typeof db, { users, accounts, sessions, verificationTokens  }: Schema): Adapter {
```

Then, we can get rid of the `schema.ts` file and `client.ts` file in the adapter code. WDYT? Normally, the userland `users` table is more complex than the standard `next-auth`'s `users` table

packages/adapter-drizzle-all/src/mysql/index.ts
```diff
@@ -0,0 +1,293 @@
...
+      return client
+        .select()
+        .from(verificationTokens)
+        .where(eq(verificationTokens.identifier, token.identifier))
+        .then(res => res[0])
+    },
+    useVerificationToken: async (token) => {
+      try {
+        await client
+          .delete(verificationTokens)
```
ThangHuuVu, line 251: Shouldn't we select the token for later return first before deleting it?

packages/adapter-drizzle-all/src/index.ts
```diff
@@ -0,0 +1,26 @@
+import { PgAdapter } from "./pg"
+import { MySqlAdapter } from "./mysql"
```
ThangHuuVu, line 2: should we get the other DB adapters from their packages?
```suggestion
import { MySqlAdapter } from "@next-auth/drizzle-adapter-mysql"
```

packages/adapter-drizzle-pg/src/index.ts
```diff
@@ -0,0 +1,264 @@
...
+ *
+ * ```
+ *
+ **/
+export function DrizzleAdapter(client: typeof db): Adapter {
+  return {
+    createUser: async (data) => {
+      return client
+        .insert(users)
+        .values({ ...data, id: "123" })
```
ThangHuuVu, line 117: We need to get rid of these mock IDs

packages/adapter-drizzle-pg/src/index.ts
```diff
@@ -0,0 +1,264 @@
...
+    linkAccount: async (rawAccount) => {
+      const updatedAccount = await client
+        .insert(accounts)
+        .values(rawAccount)
+        .returning()
+        .then(res => res[0])
+
+      const account: ReturnType<Adapter["linkAccount"]> = {
+        ...updatedAccount,
+        access_token: updatedAccount.access_token ?? undefined,
```
ThangHuuVu, line 184: Could you give this block of code some comments, it feels unnecessary to me 🤔

packages/adapter-drizzle-sqlite/src/index.ts
```diff
@@ -0,0 +1,261 @@
...
+ *
+ * ```
+ *
+ **/
+export function DrizzleAdapter(client: typeof db): Adapter {
+  return {
+    createUser: (data) => {
+      return client
+        .insert(users)
+        .values({ ...data, id: "123" })
```
ThangHuuVu, line 117: we need to get rid of these mock Ids

ThangHuuVu (changes requested):

packages/adapter-drizzle-all/src/pg/index.ts
anthonyshew: I'll make sure docs get handled after interfaces are settled! https://github.com/nextauthjs/next-auth/pull/7165#issuecomment-1500571782

packages/adapter-drizzle-pg/src/index.ts
anthonyshew: Added comment!
      // Drizzle will return `null` for fields that are not defined.
      // However, the return type is expecting `undefined`.

packages/adapter-drizzle-all/src/pg/schema.ts
anthonyshew: I'm under the impression that the drivers are handled in userland code.

Users will pass their final database client to `next-auth` which will use whichever SQL flavor the user's driver is using. As long as your SQL flavor matches the `next-auth` adapter you're using, you should be good to go.

That's at least my impression of things; hope I'm not being a fool. 😀

packages/adapter-drizzle-all/src/mysql/index.ts
anthonyshew: Good catch! Forgot that the query returns both tables. 😄 Thank you!

packages/adapter-drizzle-all/src/mysql/index.ts
anthonyshew: Oops, yes!

anthonyshew:
After speaking with the Drizzle team, it sounds like their monorepo support is currently slightly unstable. I've opted to remove the "all"-style adapter because of this at this time. I'm figuring we can revisit this in the future if it turns out there is a need for such an adapter!

packages/adapter-drizzle-all/src/mysql/index.ts
anthonyshew: See https://github.com/nextauthjs/next-auth/pull/7165#issuecomment-1537254219

This has been updated for the other three flavored packages!

anthonyshew:
@ThangHuuVu Did a clean up of your and others' reviews!

I'm having trouble getting tests to run at all this time around. Would you be able to help me with that?

oscartbeaumont:
To get the mysql adapter in this PR working with [Planetscale](https://planetscale.com) I had to change a couple of things.

First I had to remove `.references` as Planetscale doesn't support foreign key constraints. This can be done in userspace but does cause a Typescript mismatch.

I was also using the `@planetscale/database` serverless driver and this has a Typescript mismatch with the first argument to `DrizzleAdapterMySQL` as the DB type within this adapter is derived from a client using the `mysql` package instead.

Planetscale also doesn't seem to like the usage of MySQL `text` so I had to add the following alias into my Drizzle schema. This can be done in userspace but obviously causes a Typescript type error when passed to `DrizzleAdapterMySQL`.

```
const text = (name: string) =>
  varchar(name, {
    length: 255,
  });
```

I also had an issue with it crashing and have left that as a comment [here](https://github.com/nextauthjs/next-auth/pull/7165/files#diff-adab6ee4e101026d4b0131c568ceb9a01749b928bde501acef5d8acdf28bcb30R227).

Just wanted to document this as I think it will be fairly common for the MySql adapter to be used with Planetscale and the Planetscale serverless driver.

anthonyshew:
Thanks for those notes, @oscartbeaumont! I had been wondering about compatibility up the dependency tree.

I'll yield to @ThangHuuVu and the core team here. How do y'all prefer to handle compatibility situations like these for adapters?

packages/adapter-drizzle-mysql/src/index.ts
```diff
@@ -0,0 +1,293 @@
...
+      return client
+        .select()
+        .from(sessions)
+        .where(eq(sessions.sessionToken, data.sessionToken))
+        .then(res => res[0])
+    },
+    linkAccount: async (rawAccount) => {
+      await client
+        .insert(accounts)
+        .values(rawAccount)
```
khuezy, line 212: In the `accounts` schema: `expires_at`. But the rawAccount provided by the Google provider has `expires_in`. This will cause the insert to fail.
You'll probably want to pick out `expires_in` and add it to `expires_at: (Date.now()/1000 + expires_in)>>0`

anthonyshew: Interesting, I was under the impression that the Github provider would handle this at its layer. 🤔  I'll be sure to be cognizant of this while I go around and test!

anthonyshew:
For visibility:

- The core team has instructed me to opt for one adapter that handles all the SQL flavors.
- You'll also find support specifically for PlanetScale, paving a path for any other database providers that have quirks.
- I believe I've corrected TypeScript (in a messy way) such that users can now extend their schemas while still enforcing the minimum requirements of `next-auth`.
- I hope to take the `dist` out for a spin myself later this week and appreciate anyone else who goes and tests out stability!
- At this point, I still expect issues. 😀
- I have not handled documentation today as I still want to make sure that design is handled correctly.

pyrexfm:
The varchar(255) will lead to problems in the id_token field (happens with Google provider). Looking into [Prisma adapter for AuthJS](https://authjs.dev/reference/adapter/prisma), it seems that access_token, refresh_token and id_token should all be Text, so I changed and it works now.
However, I can't get it to work with Github provider.

`https://next-auth.js.org/errors#oauth_callback_handler_error Cannot read properties of undefined (reading 'name') TypeError: Cannot read properties of undefined (reading 'name')
    at eval (webpack-internal:///(sc_server)/./node_modules/drizzle-orm/session-6deff015.mjs:610:124)
    at Array.map (<anonymous>)
    at MySqlDialect.buildInsertQuery (webpack-internal:///(sc_server)/./node_modules/drizzle-orm/session-6deff015.mjs:610:40)
    at QueryPromise.getSQL (webpack-internal:///(sc_server)/./node_modules/drizzle-orm/session-6deff015.mjs:285:29)
    at QueryPromise.prepare (webpack-internal:///(sc_server)/./node_modules/drizzle-orm/session-6deff015.mjs:292:71)
    at MySqlInsert.execute (webpack-internal:///(sc_server)/./node_modules/drizzle-orm/session-6deff015.mjs:264:25)
    at QueryPromise.then (webpack-internal:///(sc_server)/./node_modules/drizzle-orm/relations-7adf6be2.mjs:1091:21)
    at linkAccount (webpack-internal:///(sc_server)/./src/lib/auth/drizzle-adapter.ts:58:108)
    at _callee2$ (webpack-internal:///(sc_server)/./node_modules/next-auth/core/errors.js:301:43)
    at tryCatch (webpack-internal:///(sc_server)/./node_modules/@babel/runtime/helpers/regeneratorRuntime.js:36:25)
    at Generator.eval (webpack-internal:///(sc_server)/./node_modules/@babel/runtime/helpers/regeneratorRuntime.js:120:30)
    at Generator.eval [as next] (webpack-internal:///(sc_server)/./node_modules/@babel/runtime/helpers/regeneratorRuntime.js:64:29)
    at asyncGeneratorStep (webpack-internal:///(sc_server)/./node_modules/@babel/runtime/helpers/asyncToGenerator.js:4:28)
    at _next (webpack-internal:///(sc_server)/./node_modules/@babel/runtime/helpers/asyncToGenerator.js:22:17)
    at eval (webpack-internal:///(sc_server)/./node_modules/@babel/runtime/helpers/asyncToGenerator.js:27:13)
    at new Promise (<anonymous>)
    at eval (webpack-internal:///(sc_server)/./node_modules/@babel/runtime/helpers/asyncToGenerator.js:19:16)
    at callbackHandler (webpack-internal:///(sc_server)/./node_modules/next-auth/core/lib/callback-handler.js:146:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.callback (webpack-internal:///(sc_server)/./node_modules/next-auth/core/routes/callback.js:71:57)
    at async AuthHandler (webpack-internal:///(sc_server)/./node_modules/next-auth/core/index.js:202:38)
    at async NextAuthRouteHandler (webpack-internal:///(sc_server)/./node_modules/next-auth/next/index.js:49:30)
    at async NextAuth._args$ (webpack-internal:///(sc_server)/./node_modules/next-auth/next/index.js:83:24)
    at async eval (webpack-internal:///(sc_server)/./node_modules/next/dist/server/future/route-modules/app-route/module.js:242:37) {
  name: 'LinkAccountError',
  code: undefined`

howarddo2208:
might be a dumb question, but how can I test out this adapter now before it's even merged? Thank u.

pyrexfm:
> might be a dumb question, but how can I test out this adapter now before it's even merged? Thank u.

Go to the src folder you need (pg, planetscale, mysql) and copy the schema and the index file into your project and create an adapter out of the index file. This allows you to experiment more easily.

packages/adapter-drizzle-mysql/src/index.ts
anthonyshew: Going through testing now and it does look like this is working! Let me know if it isn't.

packages/adapter-drizzle/src/planetscale/index.ts
```diff
@@ -0,0 +1,178 @@
...
+            )
+          )
+
+        return deletedToken
+      } catch (err) {
+        throw new Error("No verification token found.")
+      }
+    },
+    deleteUser: async (id) => {
+      console.log({ id })
```
khuezy, line 156: please remove this console.log

anthonyshew: Oops, thank you! 😄

packages/adapter-drizzle/src/mysql/index.ts
```diff
@@ -0,0 +1,174 @@
...
+        .where(eq(sessions.sessionToken, data.sessionToken))
+        .then((res) => res[0])
+    },
+    linkAccount: async (rawAccount) => {
+      await client
+        .insert(accounts)
+        .values(rawAccount)
+        .then((res) => res[0])
+    },
+    getUserByAccount: async (account) => {
```
c-wide, line 93: Hello! Thank you for your hard work on this. 

I've been testing this out in a project I'm using next auth and drizzle for. I could be wrong here but I think this method will return the wrong user sometimes.

It appears to be selecting every user, performing a join, then returning the first user which will always be the first user registered in the database.

For example, I register an email account under 'test@example.com', log out and then register through an oauth provider that has a different email as the primary email (so a separate account is made), then log out and log back in through that same oauth provider you should be logged in as the first user in the database, not your oauth user. 

It's a bit of a convoluted example but it's the only one I've tested lol

Again thank you for your hard work so far!

khuezy: Good point. This should:
1) query `accounts` table w/ `provider` and `providerAccountId`
2) If exists, query `users` by `account.id` to get user

mcnaveen:
Wow, Thanks guys for working on it. 🎉  Looking forward to it!

packages/adapter-drizzle/src/mysql/index.ts
anthonyshew: Good catch! Thank you!

anthonyshew:
Another update!:
- Manually tested the Postgres and Planetscale adapters. They appear to be working correctly for me. They still worked as expected when I added additional fields to the schema.
- Updated the API interface so that you don't have to provide a magic string for your SQL flavor.
- Wrote documentation.

Outstanding issues (that I'd probably need help resolving):
- Still can't get the tests to run! Drizzle resolves fine everywhere but in the Jest tests. 🤔 I've scaffolded out the tests for Postgres if anyone wants to try their hand finding the fix.
- The way I have TypeScript set up, it doesn't enforce the minimum schema requirements. This does leave some room for error for users - but I'm not certain about how to describe the types more succinctly. I'm not sure about the strictness of the other adapters so I'm not sure if this is to be considered blocking or not.

Additionally, if anyone wants to keep trying it out, feel free to report bugs! 😄

We're close!

packages/adapter-drizzle/src/mysql/index.ts
anthonyshew: Yes, updated again!

balazsorban44:
Fixed merge conflicts and the tests. `drizzle-orm` only publishes `.mjs` and `.cjs` files and our jest config ignored those.

Lermatroid:
Hi, I'm attempting to use this at the moment but I seem to be running into a type error when setting the adapter.

Here is my next auth config:

```ts
  callbacks: {
    session: ({ session, user }) => ({
      ...session,
      user: {
        ...session.user,
        id: user.id,
      },
    }),
  },
  adapter:  DrizzleAdapter(db, { users, accounts, sessions, verificationTokens }),
  providers: [
    DiscordProvider({
      clientId: env.DISCORD_CLIENT_ID,
      clientSecret: env.DISCORD_CLIENT_SECRET,
    }),
    /**
     * ...add more providers here.
     *
     * Most other providers require a bit more work than the Discord provider. For example, the
     * GitHub provider requires you to add the `refresh_token_expires_in` field to the Account
     * model. Refer to the NextAuth.js docs for the provider you want to use. Example:
     *
     * @see https://next-auth.js.org/providers/github
     */
  ],
};
```

The error is as follows:

```
Type 'PgTableWithColumns<{ name: "verificationToken"; schema: undefined; columns: { identifier: PgText<{ tableName: "verificationToken"; data: string; enumValues: [string, ...string[]]; name: "identifier"; driverParam: string; hasDefault: false; notNull: true; }>; token: PgText<...>; expires: PgInteger<...>; }; }>' is not assignable to type 'PgTableWithColumns<{ name: "verificationToken"; schema: undefined; columns: { identifier: PgText<{ tableName: "verificationToken"; data: string; enumValues: [string, ...string[]]; name: "identifier"; driverParam: string; hasDefault: false; notNull: true; }>; token: PgText<...>; expires: PgTimestamp<...>; }; }>'.
  Type 'PgTableWithColumns<{ name: "verificationToken"; schema: undefined; columns: { identifier: PgText<{ tableName: "verificationToken"; data: string; enumValues: [string, ...string[]]; name: "identifier"; driverParam: string; hasDefault: false; notNull: true; }>; token: PgText<...>; expires: PgInteger<...>; }; }>' is not assignable to type 'PgTable<{ name: "verificationToken"; schema: undefined; columns: { identifier: PgText<{ tableName: "verificationToken"; data: string; enumValues: [string, ...string[]]; name: "identifier"; driverParam: string; hasDefault: false; notNull: true; }>; token: PgText<...>; expires: PgTimestamp<...>; }; }>'.
    The types of '_.config.columns.expires' are incompatible between these types.
      Type 'PgInteger<{ tableName: "verificationToken"; data: number; name: "expires"; driverParam: string | number; hasDefault: false; notNull: true; }>' is missing the following properties from type 'PgTimestamp<{ tableName: "verificationToken"; data: Date; name: "expires"; driverParam: string; hasDefault: false; notNull: true; }>': withTimezone, precisionts(2322)
schema.ts(73, 3): The expected type comes from property 'verificationTokens' which is declared here on type 'Schema & Record<string, AnyPgTable>'
```

Any idea what could be causing this?

packages/adapter-drizzle/src/mysql/index.ts
```diff
@@ -0,0 +1,252 @@
...
+  (vt) => ({
+    compoundKey: primaryKey(vt.identifier, vt.token),
+  })
+)
+
+const poolConnection = process.env.AUTH_DRIZZLE_MYSQL_URL ? createPool(
+  process.env.AUTH_DRIZZLE_MYSQL_URL as string,
+) : createPool({
+  host: "localhost",
+  user: "root",
```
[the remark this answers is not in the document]
mcnaveen: Is this supposed to be coming from the env right?

XMadFoX:
Hi @anthonyshew, thank you for working on this adapter.

I've been testing it and faced a problem with implementing [credentials provider](https://next-auth.js.org/providers/credentials) (username/password example is simple reproduction), (email provider works fine).
As I understand, after calling `authorize` and returning user object from there, a session should be created in database.

But debugging network & adapter shows:
After authorize called, 302 /credentials returns next-auth.session-token (jwt)
Then comes request to /session (200) which deletes token from cookie (returns value `""`)
as `getSessionAndUser(token)` returns null, as it cannot find session (it's not being created as mentioned above)

Database is sqlite/turso, commenting out adapter in authOptions => credentials provider works as expected

Could you please look into that?

packages/adapter-drizzle/src/planetscale/index.ts
```diff
@@ -0,0 +1,248 @@
...
+          .where(
+            and(
+              eq(accounts.providerAccountId, account.providerAccountId),
+              eq(accounts.provider, account.provider)
+            )
+          )
+          .leftJoin(users, eq(accounts.userId, users.id))
+          .then(res => res[0])
+
+      return dbAccount.users
```
matheins, line 182: Returning the query result is causing an Callback error, in one edge case:
- If user already exists but has no linked oauth account and tries to sign in

This check before returning solves it: 
`
if (!dbAccount) return null
`

packages/adapter-drizzle/src/planetscale/index.ts
```diff
@@ -0,0 +1,248 @@
...
+    type: varchar("type", { length: 255 }).$type<AdapterAccount["type"]>().notNull(),
+    provider: varchar("provider", { length: 255 }).notNull(),
+    providerAccountId: varchar("providerAccountId", { length: 255 }).notNull(),
+    refresh_token: varchar("refresh_token", { length: 255 }),
+    access_token: varchar("access_token", { length: 255 }),
+    expires_at: int("expires_at"),
+    token_type: varchar("token_type", { length: 255 }),
+    scope: varchar("scope", { length: 255 }),
+    id_token: varchar("id_token", { length: 255 }),
+    session_state: varchar("session_state", { length: 255 }),
```
matheins, line 35: `linkAccount` function fails due to missing prop in the schema, when trying to register via github:
`refresh_token_expires_in: int("refresh_token_expires_in"),`

matheins (changes requested):

packages/adapter-drizzle/src/planetscale/index.ts
PatrickJoannisse: I got an issue related to linkAccount as well but the issue was that id_token's length was not long enough for some providers. Notably Google and Twitch. I switched it to a Text type which is probably overkill but it works.

nolanholden:
@anthonyshew -- Do you forsee instability using this branch with Planetscale as-is? Considering using it as a holdover until your merge goes through.

anthonyshew:
@nolanholden Yes, I was working on this without being able to run tests for awhile so I've got a lot to circle back on fixing. 😄

packages/adapter-drizzle/src/pg/index.ts
```diff
@@ -0,0 +1,238 @@
...
+      } catch (err) {
+        throw new Error("No verification token found.")
+      }
+    },
+    deleteUser: async (id) => {
+      await client
+        .delete(users)
+        .where(eq(users.id, id))
+        .returning()
+        .then((res) => res[0])
```
andresgutgon, line 223: I see this when I copy this file in my code
image


Shouldn't be `res.rows[0]`?

packages/adapter-drizzle/src/pg/index.ts
```diff
@@ -0,0 +1,238 @@
...
+      schema?.verificationTokens ?? defaultSchema.verificationTokens,
+  }
+
+  return {
+    createUser: async (data) => {
+      return client
+        .insert(users)
+        .values({ ...data, id: crypto.randomUUID() })
+        .returning()
+        .then((res) => res[0])
```
andresgutgon, line 90: This is giving me a typescript error when trying to pass the adapter to `nextauth` as Adapter

```
Type 'Adapter' is not assignable to type 'Adapter | undefined'.
  Type 'Adapter' is not assignable to type 'DefaultAdapter & { createVerificationToken: (verificationToken: VerificationToken) => Awaitable<VerificationToken | null | undefined>; useVerificationToken: (params: { ...; }) => Awaitable<...>; }'.
    Type 'Adapter' is not assignable to type 'DefaultAdapter'.
      Types of property 'createUser' are incompatible.
        Type '((user: Omit<AdapterUser, "id">) => Awaitable<AdapterUser>) | undefined' is not assignable to type '(user: Omit<AdapterUser, "id">) => Awaitable<AdapterUser>'.
          Type 'undefined' is not assignable to type '(user: Omit<AdapterUser, "id">) => Awaitable<AdapterUser>'.

```

andresgutgon, line 90: This is giving me a typescript error when trying to pass the adapter to `nextauth` as Adapter

```
Type 'Adapter' is not assignable to type 'Adapter | undefined'.
  Type 'Adapter' is not assignable to type 'DefaultAdapter & { createVerificationToken: (verificationToken: VerificationToken) => Awaitable<VerificationToken | null | undefined>; useVerificationToken: (params: { ...; }) => Awaitable<...>; }'.
    Type 'Adapter' is not assignable to type 'DefaultAdapter'.
      Types of property 'createUser' are incompatible.
        Type '((user: Omit<AdapterUser, "id">) => Awaitable<AdapterUser>) | undefined' is not assignable to type '(user: Omit<AdapterUser, "id">) => Awaitable<AdapterUser>'.
          Type 'undefined' is not assignable to type '(user: Omit<AdapterUser, "id">) => Awaitable<AdapterUser>'.

```

packages/adapter-drizzle/src/mysql/index.ts
```diff
@@ -0,0 +1,254 @@
...
+    createUser: async (data) => {
+      const id = crypto.randomUUID()
+
+      await client.insert(users).values({ ...data, id })
+
+      return client
+        .select()
+        .from(users)
+        .where(eq(users.id, id))
+        .then((res) => res[0])
```
andresgutgon, line 103: This is giving me a typescript error when trying to pass the adapter to `nextauth` as Adapter

```
Type 'Adapter' is not assignable to type 'Adapter | undefined'.
  Type 'Adapter' is not assignable to type 'DefaultAdapter & { createVerificationToken: (verificationToken: VerificationToken) => Awaitable<VerificationToken | null | undefined>; useVerificationToken: (params: { ...; }) => Awaitable<...>; }'.
    Type 'Adapter' is not assignable to type 'DefaultAdapter'.
      Types of property 'createUser' are incompatible.
        Type '((user: Omit<AdapterUser, "id">) => Awaitable<AdapterUser>) | undefined' is not assignable to type '(user: Omit<AdapterUser, "id">) => Awaitable<AdapterUser>'.
          Type 'undefined' is not assignable to type '(user: Omit<AdapterUser, "id">) => Awaitable<AdapterUser>'.

```

andresgutgon: I fixed the error by lying Typescript, not even tried yet. But I would like to understand why `.where` is giving `undefined` in Drizzel. Please comment if you know why : )

image

packages/adapter-drizzle/src/mysql/index.ts
```diff
@@ -0,0 +1,254 @@
...
+  emailVerified: timestamp("emailVerified", { mode: "date" }),
+  image: varchar("image", { length: 255 }),
+})
+
+export const accounts = mysqlTable(
+  "accounts",
+  {
+    userId: varchar("userId", { length: 255 })
+      .notNull()
+      .references(() => users.id, { onDelete: "cascade" }),
```
andresgutgon, line 28: Planet scale does not support foreign keys. So I had to remove `.references` to be able to push the migrations to their servers.

image

PatrickJoannisse: I'm also using Planetscale but have you seen this `packages/adapter-drizzle/src/planetscale/index.ts`?

andresgutgon: It's what I'm using. The problem is I think that Planet Scale use a flavor of MySQL called Vitess which is mean to be scalable and they don't allow foreign keys. That's my theory. 

Any way this issue goes away not using `.references`. You don't have foreign keys but that's the price for using something at planet scale (pum intended)

VaniaPopovic: There are two seperate adapters, one for mysql and one for planetscale in the PR.

andresgutgon: Oh you're right my bad. Ignore my comment here

packages/adapter-drizzle/src/sqlite/index.ts
```diff
@@ -0,0 +1,223 @@
...
+        refresh_token: updatedAccount.refresh_token ?? undefined,
+        scope: updatedAccount.scope ?? undefined,
+        expires_at: updatedAccount.expires_at ?? undefined,
+        session_state: updatedAccount.session_state ?? undefined,
+      }
+
+      return account
+    },
+    getUserByAccount: (account) => {
+      const dbAccount =
```
khuezy, line 155:
```suggestion
      const results = await client
       .select().from(accounts)
       .leftJoin(users, eq(users.id, accounts.id))
       .where(
         and(eq(accounts.provider, account.provider), eq(accounts.providerAccountId, account.providerAccountId))
       ).get()
       
       return results.users ?? null;
      
```

nolanholden:
Excited for this one! @anthonyshew

packages/adapter-drizzle/src/planetscale/index.ts
```diff
@@ -0,0 +1,248 @@
...
+        .select()
+        .from(sessions)
+        .where(eq(sessions.sessionToken, data.sessionToken))
+        .then((res) => res[0])
+    },
+    linkAccount: async (rawAccount) => {
+      await client
+        .insert(accounts)
+        .values(rawAccount)
+        .then((res) => res.rows[0])
```
andresgutgon, line 166: This is failing for me trying Google Auth. The `rawAccount.id_token` is bigger than 255 chars. Is 1185 chars. So maybe this column `id_token` should be a text column.

packages/adapter-drizzle/src/sqlite/index.ts
michaelangrivera: This is likely more performant. I don't see any indexes made in this PR. Is that something that should be done as well?

packages/adapter-drizzle/src/sqlite/index.ts
```diff
@@ -0,0 +1,223 @@
...
+    access_token: text("access_token"),
+    expires_at: integer("expires_at"),
+    token_type: text("token_type"),
+    scope: text("scope"),
+    id_token: text("id_token"),
+    session_state: text("session_state"),
+  },
+  (account) => ({
+    compoundKey: primaryKey(account.provider, account.providerAccountId),
+  })
```
michaelangrivera, line 37: Indexes should be made for this table, likely on the following columns:
- `userId` -> uniqueIndex
- `type` -> index
- `provider` -> index
- `expires_at` -> index (descending?) 
- `session_state` -> index

michaelangrivera: https://orm.drizzle.team/docs/indexes-constraints#indexes

packages/adapter-drizzle/src/pg/index.ts
```diff
@@ -0,0 +1,244 @@
...
+  {
+    userId: text("userId")
+      .notNull()
+      .references(() => users.id, { onDelete: "cascade" }),
+    type: text("type").$type<AdapterAccount["type"]>().notNull(),
+    provider: text("provider").notNull(),
+    providerAccountId: text("providerAccountId").notNull(),
+    refresh_token: text("refresh_token"),
+    access_token: text("access_token"),
+    expires_at: integer("expires_at"),
```
michaelangrivera, line 34: why not use the native PG date column? Let the database do the work, not the consumer. 

https://orm.drizzle.team/docs/column-types/pg#timestamp

packages/adapter-drizzle/src/sqlite/index.ts
michaelangrivera: https://github.com/nextauthjs/next-auth/pull/7165/files#r1253130773 index suggestions here. There should be indexes for each type (pg, MySQL, sqlite) IMO

packages/adapter-drizzle/src/index.ts
```diff
@@ -0,0 +1,355 @@
...
+ * import postgres from "postgres";
+ *
+ * const sql = postgres("...", { max: 1 })
+ * const db = drizzle(sql);
+ *
+ * await migrate(db, { migrationsFolder: "drizzle" });
+ * ```
+ *
+ **/
+export function DrizzleAdapter<SqlFlavor extends SqlFlavorOptions>(
```
michaelangrivera, line 329: What is the plan for migrations?

packages/adapter-drizzle/src/sqlite/index.ts
khuezy: The adapter's access pattern is only user.id and user.email. For `accounts` it's only by `provider and providerAccountId`, so we don't need indexes for those fields unless you're doing something specific out of scope of the Adapter.

michaelangrivera: Perhaps it would be nice to give the option of accessing these?

ImBIOS (approved):

packages/adapter-drizzle/src/sqlite/index.ts
```diff
@@ -0,0 +1,223 @@
...
+    type: text("type").$type<AdapterAccount["type"]>().notNull(),
+    provider: text("provider").notNull(),
+    providerAccountId: text("providerAccountId").notNull(),
+    refresh_token: text("refresh_token"),
+    access_token: text("access_token"),
+    expires_at: integer("expires_at"),
+    token_type: text("token_type"),
+    scope: text("scope"),
+    id_token: text("id_token"),
+    session_state: text("session_state"),
```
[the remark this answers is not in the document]
anthonyshew: We need to follow the spec: https://authjs.dev/reference/adapters#models

Users will be able to customize the tables if they want.

packages/adapter-drizzle/src/pg/index.ts
anthonyshew: Need to follow the spec: https://authjs.dev/reference/adapters#models

packages/adapter-drizzle/src/planetscale/index.ts
anthonyshew: Will be dropping PlanetScale to cut scope so we can ship. We should be able to add PlanetScale back in the future!

packages/adapter-drizzle/src/sqlite/index.ts
anthonyshew: Users will be able to provide custom tables; you'll be able to add your own custom indices.

Do we think any of these suggestions should be there by default? Going to mark as resolved because it doesn't look like like there should be at a glance?

packages/adapter-drizzle/src/sqlite/index.ts
anthonyshew: Thanks!

packages/adapter-drizzle/src/sqlite/index.ts
arivera-xealth: @anthonyshew I think at the very least the indexes should be mentioned in the doc. When you scale to millions of users, this is a big deal.

nolanholden:
hey @anthonyshew, not attacking the PR here, but I am curious. Why are there 189 commits? Is this because you are merging main back into this branch without ever rebasing? Could some commits be squashed? I know if I was a maintainer I would be concerned about this.

But there may be something I don't know! Interested about your process.

packages/adapter-drizzle/src/sqlite/index.ts
anthonyshew: @arivera-xealth We need to ensure we're careful about scope here. The docs need to describe what makes sense for the default configuration. If users are creating extra access patterns such that indices make sense for their application, that doesn't seem like something we need to document in next-auth.

For the default usage of next-auth, do we need to add any extra indices? From what I can tell looking at prior art (namely, the Prisma adapter), it doesn't look like we need to.

andresgutgon: My 2 cents. I would follow previous implementations on other adapters. Add just the indices those adapters added

anthonyshew:
@nolanholden I'm going to do some rebasing and squashing once I'm done; just haven't cared to do it yet. 😁 We've changed the API we were going for like four times and keeping the prior art around has helped a few times.

anthonyshew:
Okay, folks, I think we've got it! I've written up tests and manually tested all three SQL flavors with Docker.

One thing of note: I would recommend marking this adapter as experimental. 
- Through the course of working on this, Drizzle shipped several breaking changes that affected both the correctness and development velocity of this PR. Don't read this as a critique; rather, this would be an acknowledgement that library is still quite young (despite its popularity). The Drizzle core team is *excellent* and their ORM will be stable and awesome in the future - but it's important to note the downstream dependency, in my opinion.

packages/adapter-drizzle/src/mysql/index.ts
```diff
@@ -0,0 +1,261 @@
...
+      .notNull(),
+    provider: varchar("provider", { length: 255 }).notNull(),
+    providerAccountId: varchar("providerAccountId", { length: 255 }).notNull(),
+    refresh_token: varchar("refresh_token", { length: 255 }),
+    access_token: varchar("access_token", { length: 255 }),
+    expires_at: int("expires_at"),
+    token_type: varchar("token_type", { length: 255 }),
+    scope: varchar("scope", { length: 255 }),
+    id_token: varchar("id_token", { length: 255 }),
+    session_state: varchar("session_state", { length: 255 }),
```
andresgutgon, line 43: I copied this code in my project because I wanted to start using Planetscale + Next-Auth + Drizzel. So thank you!. 

Is this mysql adapter the one we should use if we want to use Planetscale? If yes, in my case I had to use for `id_token` and `session_state` a `text` field not a `varchar` because I got an error when using Google Auth.

andresgutgon: Sorry, I didn't see your previous comment here https://github.com/nextauthjs/next-auth/pull/7165#discussion_r1263313311

Ignore this. I guess I'll use my custom adapter for planet scale for now. Thanks for the work!

packages/adapter-drizzle/src/pg/index.ts
```diff
@@ -0,0 +1,231 @@
...
+    },
+    updateSession: async (data) => {
+      return await client
+        .update(sessions)
+        .set(data)
+        .where(eq(sessions.sessionToken, data.sessionToken))
+        .returning()
+        .then((res) => res[0])
+    },
+    linkAccount: async (rawAccount) => {
```
michaelangelo-io, line 135: If you get the following error with Github provider:
```
      "message": "Cannot read properties of undefined (reading 'name')",
      "stack": "TypeError: Cannot read properties of undefined (reading 'name')\n
```
It's because GitHub adds an extra property `refresh_token_expires_in`: 
```typescript
{
  provider: 'github',
  type: 'oauth',
  providerAccountId: '',
  access_token: '',
  expires_at: 1690073663,
  refresh_token: '',
  refresh_token_expires_in: 15897600, // <----- 
  token_type: 'bearer',
  scope: '',
  userId: '53c777ed-353a-438b-a57e-wqweqwe'
}
```
I think maybe it would be best (for compatibility) to add this column to the provider? Or I guess maybe point that out in the docs so people can add that column to the schema? 

Also, it may be better to use bigInt (with mode `number`) instead of integer for the expires column(s) for flexibility. https://orm.drizzle.team/docs/column-types/pg#bigint

mschieller:
Is there a reason to why this adapter is not using prepared statements? Hence prepared statements has a much quicker build time that standard typings, and that these auth request would be called millions of millions of time, prepared statements should be really beneficial?
And if decided not to use prepared as default, perhaps make it optional in the adapter, or would this change the decires outcome of the generality of all the adapters?

packages/adapter-drizzle/src/mysql/index.ts
marvkr: Would you mind sharing your adapter? I'm also using Planetscale so this bugs me as well ... 😕

demarchenac: Hi @andresgutgon! It would be nice, maybe you have a public repo in which we can read your twist of the adapter?

iukea1: > I copied this code in my project because I wanted to start using Planetscale + Next-Auth + Drizzel. So thank you!. 
> 
> Is this mysql adapter the one we should use if we want to use Planetscale? If yes, in my case I had to use for `id_token` and `session_state` a `text` field not a `varchar` because I got an error when using Google Auth.
> 
> 
Sharing is caring

andresgutgon: Not much to it. I just copy & paste from this PR when there was the planetscale adapter and fixed the errors for Google Auth provider.
This is working for me: Google Auth + Planetscale
https://gist.github.com/andresgutgon/dcbf135f890450f0a03eeb9b7891b268

demarchenac: I was just wondering since I was looking at the schema generated for Primsa, but shouldn't the `refresh_token` & `access_token` columns have a `text` type as well?

andresgutgon: In my case with Google Auth was not necessary. The problem with using text instead of varchar is that it takes more space in the db to store the column afaik. So i use text only when strictly necessary

iukea1: @andresgutgon  did you get prepared queries working?

andresgutgon: I don't even know what's that lol. So I guess no xD. What you see in the gist is what I did. Nothing else related with Drizzle + NextAuth + Planetscale

andresgutgon:
> I created a fully functional NextAuth adapter for Drizzle+PlanetScale, which utilises the new query methodologies, optional prepared statements to be used in adapter. If any would want it, I can share my findings and code :-)

Yes please, I [shared mine here](https://github.com/nextauthjs/next-auth/pull/7165#discussion_r1278520458) but not with prepare statements. I'm curious

mschieller:
> > I created a fully functional NextAuth adapter for Drizzle+PlanetScale, which utilises the new query methodologies, optional prepared statements to be used in adapter. If any would want it, I can share my findings and code :-)
> 
> 
> 
> Yes please, I [shared mine here](https://github.com/nextauthjs/next-auth/pull/7165#discussion_r1278520458) but not with prepare statements. I'm curious

Currently flying home from vacation, but I'll make a test-next repo tonight and post it here, so all of it can be seen 🫡

packages/adapter-drizzle/src/mysql/index.ts
iukea1: @andresgutgon  prepared queries with your build are not erroring out locally. Pushing to Versel now

andresgutgon: Can you explain here quickly what's a prepared query and what's the benefit?

mschieller: A prepared statement is kinda what the name implies. Instead of generate each sql query on request which is the normal procedure usually when using ORMs, Drizzle can make prepared statements, where the SQL is made on build, and when requested by code it it can use it directly which improves performance a lot because it can reuse the binary compilation. It can also take in params if doing joins and what not, which still is much faster. See here for more https://orm.drizzle.team/docs/performance#

iukea1: so Prepared statement

> When you run query to the database there're several things that happens:
> 
> - all the configurations of the query builder got concatenated to the SQL string
>-  that string and params are sent to the database driver
>-  driver compiles SQL query to the binary SQL executable format and sends it to the database
>-  With prepared statements you do SQL concatenation once on the Drizzle ORM side and then database driver is able to reuse precompiled binary SQL instead of parsing query all the time. It has extreme performance benefits on large SQL queries.

https://orm.drizzle.team/docs/performance

can also watch the first half of this video
- [](https://www.youtube.com/watch?v=O9zxsSl9zgE&t=430s)

mschieller:
> > > I created a fully functional NextAuth adapter for Drizzle+PlanetScale, which utilises the new query methodologies, optional prepared statements to be used in adapter. If any would want it, I can share my findings and code :-)
> > 
> > 
> > Yes please, I [shared mine here](https://github.com/nextauthjs/next-auth/pull/7165#discussion_r1278520458) but not with prepare statements. I'm curious
> 
> Currently flying home from vacation, but I'll make a test-next repo tonight and post it here, so all of it can be seen 🫡

**As promised** https://github.com/mschieller/drizzle-nextauth-planetscale
Feel free to ask why, and if you see something wrong, please let me know.

be aware that typing was a bit tricky so the adapter is not typesafe yet

zivtamary:
> **As promised** https://github.com/mschieller/drizzle-nextauth-planetscale Feel free to ask why, and if you see something wrong, please let me know.
> 
> be aware that typing was a bit tricky so the adapter is not typesafe yet

Hey, thanks for that :) would it also work for database session strategy?

mschieller:
> > **As promised** https://github.com/mschieller/drizzle-nextauth-planetscale Feel free to ask why, and if you see something wrong, please let me know.
> 
> > 
> 
> > be aware that typing was a bit tricky so the adapter is not typesafe yet
> 
> 
> 
> Hey, thanks for that :) would it also work for database session strategy?

Pretty confident it does yes, haven't tried though. Please give feedback if/when you have tried 😀

zivtamary:
> > > **As promised** https://github.com/mschieller/drizzle-nextauth-planetscale Feel free to ask why, and if you see something wrong, please let me know.
> > 
> > 
> > > 
> > 
> > 
> > > be aware that typing was a bit tricky so the adapter is not typesafe yet
> > 
> > 
> > Hey, thanks for that :) would it also work for database session strategy?
> 
> Pretty confident it does yes, haven't tried though. Please give feedback if/when you have tried 😀
Unfortunately getting some errors
`[next-auth][warn][EXPERIMENTAL_API] 
https://next-auth.js.org/configuration/nextjs#getServerSession}
https://next-auth.js.org/warnings#EXPERIMENTAL_API
[next-auth][error][adapter_error_linkAccount] 
https://next-auth.js.org/errors#adapter_error_linkaccount Cannot read properties of undefined (reading 'name') {
  message: "Cannot read properties of undefined (reading 'name')",
  stack: "TypeError: Cannot read properties of undefined (reading 'name')\n" +
    '    at eval (webpack-internal:///(rsc)/./node_modules/drizzle-orm/session-9628aea0.mjs:727:127)\n' +
    '    at Array.map (<anonymous>)\n' +
    '    at MySqlDialect.buildInsertQuery (webpack-internal:///(rsc)/./node_modules/drizzle-orm/session-9628aea0.mjs:727:40)\n' +
    '    at QueryPromise.getSQL (webpack-internal:///(rsc)/./node_modules/drizzle-orm/session-9628aea0.mjs:383:29)\n' +
    '    at QueryPromise.prepare (webpack-internal:///(rsc)/./node_modules/drizzle-orm/session-9628aea0.mjs:390:71)\n' +
    '    at MySqlInsert.execute (webpack-internal:///(rsc)/./node_modules/drizzle-orm/session-9628aea0.mjs:360:25)\n' +
    '    at QueryPromise.then (webpack-internal:///(rsc)/./node_modules/drizzle-orm/alias-3e926a50.mjs:766:21)\n' +
    '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)',
  name: 'TypeError'
}
[next-auth][error][OAUTH_CALLBACK_HANDLER_ERROR] 
https://next-auth.js.org/errors#oauth_callback_handler_error Cannot read properties of undefined (reading 'name') TypeError: Cannot read properties of undefined (reading 'name')
    at eval (webpack-internal:///(rsc)/./node_modules/drizzle-orm/session-9628aea0.mjs:727:127)
    at Array.map (<anonymous>)
    at MySqlDialect.buildInsertQuery (webpack-internal:///(rsc)/./node_modules/drizzle-orm/session-9628aea0.mjs:727:40)
    at QueryPromise.getSQL (webpack-internal:///(rsc)/./node_modules/drizzle-orm/session-9628aea0.mjs:383:29)
    at QueryPromise.prepare (webpack-internal:///(rsc)/./node_modules/drizzle-orm/session-9628aea0.mjs:390:71)
    at MySqlInsert.execute (webpack-internal:///(rsc)/./node_modules/drizzle-orm/session-9628aea0.mjs:360:25)
    at QueryPromise.then (webpack-internal:///(rsc)/./node_modules/drizzle-orm/alias-3e926a50.mjs:766:21)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  name: 'LinkAccountError',
  code: undefined
}`

zivtamary:
> > > **As promised** https://github.com/mschieller/drizzle-nextauth-planetscale Feel free to ask why, and if you see something wrong, please let me know.
> > 
> > 
> > > 
> > 
> > 
> > > be aware that typing was a bit tricky so the adapter is not typesafe yet
> > 
> > 
> > Hey, thanks for that :) would it also work for database session strategy?
> 
> Pretty confident it does yes, haven't tried though. Please give feedback if/when you have tried 😀

Unfortunately getting some errors... btw I'm using Google adapter
```
[next-auth][error][adapter_error_linkAccount] 
https://next-auth.js.org/errors#adapter_error_linkaccount Cannot read properties of undefined (reading 'name') {
  message: "Cannot read properties of undefined (reading 'name')",
  stack: "TypeError: Cannot read properties of undefined (reading 'name')\n" +
```

mschieller:
> > > > **As promised** https://github.com/mschieller/drizzle-nextauth-planetscale Feel free to ask why, and if you see something wrong, please let me know.
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > be aware that typing was a bit tricky so the adapter is not typesafe yet
> 
> > > 
> 
> > > 
> 
> > > Hey, thanks for that :) would it also work for database session strategy?
> 
> > 
> 
> > Pretty confident it does yes, haven't tried though. Please give feedback if/when you have tried 😀
> 
> 
> 
> Unfortunately getting some errors... btw I'm using Google adapter
> 
> ```
> 
> [next-auth][error][adapter_error_linkAccount] 
> 
> https://next-auth.js.org/errors#adapter_error_linkaccount Cannot read properties of undefined (reading 'name') {
> 
>   message: "Cannot read properties of undefined (reading 'name')",
> 
>   stack: "TypeError: Cannot read properties of undefined (reading 'name')\n" +
> 
> ```

I got the same errors my self, forgot to redo a change I had made as a test prior. Check the new update I have made to the auth schema :-)

zivtamary:
> > > > > **As promised** https://github.com/mschieller/drizzle-nextauth-planetscale Feel free to ask why, and if you see something wrong, please let me know.
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > >
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > > be aware that typing was a bit tricky so the adapter is not typesafe yet
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > Hey, thanks for that :) would it also work for database session strategy?
> > 
> > 
> > > 
> > 
> > 
> > > Pretty confident it does yes, haven't tried though. Please give feedback if/when you have tried 😀
> > 
> > 
> > Unfortunately getting some errors... btw I'm using Google adapter
> > ```
> > 
> > [next-auth][error][adapter_error_linkAccount] 
> > 
> > https://next-auth.js.org/errors#adapter_error_linkaccount Cannot read properties of undefined (reading 'name') {
> > 
> >   message: "Cannot read properties of undefined (reading 'name')",
> > 
> >   stack: "TypeError: Cannot read properties of undefined (reading 'name')\n" +
> > ```
> 
> I got the same errors my self, forgot to redo a change I had made as a test prior. Check the new update I have made to the auth schema :-)

It works like a charm! Thank you! 😀

balazsorban44:
Hi folks, this is now released under `@auth/drizzle-adapter`! Please try it out. 🙏

Docs: https://authjs.dev/reference/adapter/drizzle

Note, it should work with both NextAuth.js and Auth.js libs.

Thanks for everyone chipping in and your patience! Let us know!

PR #7220, opened 2023-04-11
fix: next route handler breaks host detection with headers

hellodamien:
## ☕️ Reasoning

I upgraded to use Route Handler with Next 13.3.
It works fine in dev, but when I deploy to Vercel, the url callback does not take `VERCEL_URL` or `NEXTAUTH_URL`.
The `callbackUrl` contains `https://localhost:3000` and not the right url.
I use the same configuration as I had with the api in `pages/`.
I think the host must be retrieved via the new `headers()` method provided by Next and not with req.headers.

### DIFF
--- a/packages/next-auth/src/core/index.ts
+++ b/packages/next-auth/src/core/index.ts
@@ -1,5 +1,5 @@
 import logger, { setLogger } from "../utils/logger"
-import { detectHost } from "../utils/detect-host"
+import { detectOrigin } from "../utils/detect-origin"
 import * as routes from "./routes"
 import renderPage from "./pages"
 import { init } from "./init"
@@ -13,7 +13,7 @@ import { parse as parseCookie } from "cookie"
 
 export interface RequestInternal {
   /** @default "http://localhost:3000" */
-  host?: string
+  origin?: string
   method?: string
   cookies?: Partial<Record<string, string>>
   headers?: Record<string, any>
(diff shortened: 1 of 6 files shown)


bugplanter:
Experiencing the same issue.

hellodamien:
I have also tried `next-auth@0.0.0-pr.7214.4142955d` on Vercel before this PR and same, it doesn't fix the issue.
I deployed this PR with headers fix on Vercel and it's working.
Just waiting review to get an experimental release.

hellodamien:
Hi @balazsorban44, yes it's because you have changed the way host/origin is detected since #7214 (host attribute removed from req object).

I have retried you experimental release `next-auth@0.0.0-pr.7214.4142955d` to be sure, and I confirmed it's not working on Vercel.

The issue is coming from this line https://github.com/nextauthjs/next-auth/blob/7a8c0068c4c42761b54eff496f6cb8820b54b5d1/packages/next-auth/src/core/index.ts#L57

When using Route Handler, req is not an instance of Request and setting origin attribute is skipped.
So default url is used and we are redirected to localhost:3000 in production...

uday8373:
Still facing this issue in version 4.22.1.

haseakito:
simple fix it to drop NextAuth and use [Clerk](https://clerk.com/) which is way more supportive and user friendly

simkessy:
I'm experiencing this currently, is there actually not a fix for this yet?

FrancescoSaverioZuppichini:
I am also redirected to localhost 3000 even if I've set the `NEXTAUTH_URL` var to the correct public url on vercel

jjalonso:
im experimenting something similar, on dev works, on previews (Despite there i dont use NEXTAUTH_URL as doc explain, just in dev) but I still get some references un request cookies  'next-auth.callback-url=http%3A%2F%2Flocalhost%3A3000'

this end up in error 504

PR #7225, opened 2023-04-11
feat(sveltekit): allow passing custom handles

intrnl:
## ☕️ Reasoning

This change adds a configurable `handle` property allowing users to handle requests not relating to SvelteKit Auth.

This is necessary because we set up the `event.locals.getSession` function so that it can be called by routes, and while the user could feasibly set it up themselves, it would be nice if SvelteKit Auth does so instead.

Helps answer this question: https://github.com/nextauthjs/next-auth/discussions/7187


intrnl:
haven't looked at this pull request for quite some time, think I'll close this, anyone interested can pick it up

PR #7234, opened 2023-04-13
Added `signinInfo` when signing in

### LINKED ISSUE #1069: Collect profile info in passwordless email signIn() flow

AlexErrant:
## ☕️ Reasoning

Since [this PR](https://github.com/nextauthjs/next-auth/pull/4747) seems abandoned, I took the initiative of moving it to `packages/core` as suggested by Thang.

I made some changes to the naming and semantics. Instead of being specific to new users, the name `signinInfo` refers to info passed in during the signin process. After all, without an adapter, isn't _every_ user "new"?

## 🧢 Checklist

- [x] Documentation
- [ ] Tests (Not sure how to test, tbh...)
- [x] Ready to be merged

## 🎫 Affected issues

Supersedes https://github.com/nextauthjs/next-auth/pull/4747
Fixes https://github.com/nextauthjs/next-auth/issues/1069

### DIFF
--- a/packages/core/src/adapters.ts
+++ b/packages/core/src/adapters.ts
@@ -116,7 +116,7 @@
  */
 
 import { ProviderType } from "./providers/index.js"
-import type { Account, Awaitable, User } from "./types.js"
+import type { Account, Awaitable, SigninInfo, User } from "./types.js"
 // TODO: Discuss if we should expose methods to serialize and deserialize
 // the data? Many adapters share this logic, so it could be useful to
 // have a common implementation.
@@ -216,7 +216,7 @@ export interface VerificationToken {
  * :::
  */
 export interface Adapter {
-  createUser?(user: Omit<AdapterUser, "id">): Awaitable<AdapterUser>
+  createUser?(user: Omit<AdapterUser, "id">, info?: { signinInfo?: SigninInfo }): Awaitable<AdapterUser>
   getUser?(id: string): Awaitable<AdapterUser | null>
   getUserByEmail?(email: string): Awaitable<AdapterUser | null>
   /** Using the provider id and the id of the user for a specific account, get the user. */
--- a/packages/core/src/lib/index.ts
+++ b/packages/core/src/lib/index.ts
@@ -11,6 +11,8 @@ import type {
   ResponseInternal,
 } from "../types.js"
 
+const SIGNIN_INFO_MAX_AGE = 60 * 15 // 15 minutes in seconds
+
 /** @internal */
 export async function AuthInternal<
   Body extends string | Record<string, any> | any[]
@@ -139,6 +141,18 @@ export async function AuthInternal<
   } else {
     switch (action) {
       case "signin":
+        if (request.body?.signinInfo !== undefined) {
+          const expires = new Date()
+          expires.setTime(expires.getTime() + SIGNIN_INFO_MAX_AGE * 1000)
+          cookies.push({
+            name: options.cookies.signinInfo.name,
+            value: request.body.signinInfo,
+            options: {
+              ...options.cookies.signinInfo.options,
+              expires,
+            },
+          })
+        }
         if ((csrfDisabled || options.csrfTokenVerified) && options.provider) {
           const signin = await routes.signin(
             request.query,
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -163,6 +163,8 @@ export interface CallbacksOptions<P = Profile, A = Account> {
     }
     /** If Credentials provider is used, it contains the user credentials */
     credentials?: Record<string, CredentialInput>
+    /** Info from the client's `signIn()` function. */
+    signinInfo?: SigninInfo
   }) => Awaitable<boolean>
   /**
    * This callback is called anytime the user is redirected to a callback URL (e.g. on signin or signout).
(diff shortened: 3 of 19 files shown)


packages/core/src/adapters.ts
```diff
@@ -216,7 +216,7 @@ export interface VerificationToken {
  * :::
  */
 export interface Adapter {
-  createUser?(user: Omit<AdapterUser, "id">): Awaitable<AdapterUser>
+  createUser?(user: Omit<AdapterUser, "id">, info?: { signinInfo?: SigninInfo }): Awaitable<AdapterUser>
```
AlexErrant, line 219: I add it as an optional param so it may be extended in the future with other data. Also shouldn't break exiting adapters.

packages/core/src/lib/index.ts
```diff
@@ -139,6 +141,18 @@ export async function AuthInternal<
...
+          expires.setTime(expires.getTime() + SIGNIN_INFO_MAX_AGE * 1000)
+          cookies.push({
+            name: options.cookies.signinInfo.name,
+            value: request.body.signinInfo,
+            options: {
+              ...options.cookies.signinInfo.options,
+              expires,
+            },
+          })
+        }
```
AlexErrant, line 155: Absolutely no idea if this is the right place to put it... but it works!

packages/core/src/types.ts
```diff
@@ -237,9 +239,12 @@ export interface CookiesOptions {
   csrfToken: CookieOption
   pkceCodeVerifier: CookieOption
   state: CookieOption
+  signinInfo: CookieOption
   nonce: CookieOption
 }
 
+export type SigninInfo = string
```
AlexErrant, line 246: Is it `SigninInfo` or `SignInInfo` 🤔🤔 Choosing the former arbitrarily because less stuttering.

AlexErrant (commented):
I'm not confident about the flow of `signinInfo` from query param => cookie; open to suggestions!

ThangHuuVu (commented):
Hi @AlexErrant thanks for sending the PR! After discussed with @balazsorban44 we agreed that we don't want to have this built in at the moment - We will revisit this PR once we see enough interests for it 🫶 Thanks again for your understanding

I suppose the workaround would be intercepting the request using [Advanced Initialization](https://next-auth.js.org/configuration/initialization#advanced-initialization) and handle the extra query params in the application code.

ThangHuuVu:
Hi @AlexErrant ,

AlexErrant:
@ThangHuuVu I don't tag you lightly, but may I encourage you to reconsider this? [The issue this fixes](https://github.com/nextauthjs/next-auth/issues/1069) has +9, which would put it in the **_top three_** [most "reactions" issues](https://github.com/nextauthjs/next-auth/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-desc+)... if it weren't closed by StaleBot. If this doesn't qualify as enough interest, I would inquire as to what _would_ qualify as enough interest. Note too that the people who upvote that issue are probably people who aren't using the NextAuth because it doesn't meet their business requirements; it would naturally have low engagement because people don't often look for a new auth library.

As to your workaround of intercepting the request, that's impossible because in the Authorization Code flow, [the code must be only used once](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2).

> The client MUST NOT use the authorization code more than once.

Here's the series of interceptable requests as seen by the constructor when I try what's in Advanced Initialization:

```
GET https://localhost:3014/
GET https://localhost:3014/api/auth/csrf
POST https://localhost:3014/api/auth/signin/github
GET https://localhost:3014/api/auth/callback/github?code=afa303a221a04d6d3a9f
GET https://localhost:3014/
```

That final GET is the final redirect to the homepage, and the penultimate GET is the code request which application code cannot use without violating the OAuth spec. Devs cannot use the code to get the profile to add the user to the database.

I don't mean to be argumentative. Please let me know if I'm missing something in my attempted workaround. I understand that you're trying to reduce the maintenance burden, but far as I can tell this feature is impossible to implement in application code, at least via what's suggested in Advanced Initialization.

ThangHuuVu:
hello, @AlexErrant thanks for the detailed comment, I apologize for giving you the incorrect workaround. If I understand correctly, this feature is only helpful for the Email login flow. For the OAuth flow, you can add `authorizationParams` as the third parameter in the `signIn` method:
```ts
signIn('github', undefined, { foo: 'bar'})
```

and you will receive it in the 4th request
```
GET https://localhost:3014/api/auth/callback/github?code=123&foo=bar
``` 

Right now we are focusing on shipping v5 and will re-visit the PR once we reach a more stable after the release. I apologize again for the inconvenience. 🫶

PR #7331, opened 2023-04-21
feat(solidstart): `createSession` hook & hydration fixes

OrJDev:
then you can use the `createSession` hook:

```ts
import { createSession } from "@auth/solid-start/client"
const session = createSession()

const user = () => session()?.user
```

This will return a resource, in Solid a resource can be used to trigger suspense so this could be used like:

```tsx
import { Show, type VoidComponent } from "solid-js"
import { A } from "solid-start"
import { createSession, signIn, signOut } from "~/auth/client"

const AuthShowcase: VoidComponent = () => {
  const session = createSession()
  return (
    <div class="flex flex-col items-center justify-center gap-4">
      <Show
        when={session()}
        fallback={
          <button
            onClick={() => signIn("discord", { redirectTo: "/" })}
            class="rounded-full bg-white/10 px-10 py-3 font-semibold text-white no-underline transition hover:bg-white/20"
          >
            Sign in
          </button>
        }
      >
        <span class="text-xl text-white">Welcome {session()?.user?.name}</span>
        <button
          onClick={() => signOut({ redirectTo: "/" })}
          class="rounded-full bg-white/10 px-10 py-3 font-semibold text-white no-underline transition hover:bg-white/20"
        >
          Sign out
        </button>
      </Show>
    </div>
  )
}
```

This also triggers suspense so the page will not be loaded until the session was passed.


OrJDev:
failing test has nothing to do with my changes

packages/frameworks-solid-start/package.json
```diff
@@ -15,39 +17,47 @@
...
-    "solid-start": "^0.2.14",
-    "tsup": "^6.5.0",
-    "typescript": "^4.8.2"
+    "solid-start": "^0.2.15",
+    "typescript": "^4.8.2",
+    "vite": "^4.3.1"
   },
   "peerDependencies": {
-    "@auth/core": "~0.2.2 || ^0.2.2",
+    "@auth/core": "0.5.1",
```
AlexErrant, line 51: [The Svelte package uses](https://github.com/nextauthjs/next-auth/blob/main/packages/frameworks-sveltekit/package.json#L47-L49)

```ts
  "dependencies": {
    "@auth/core": "workspace:*"
  },
```

instead of being a peerDep with a hardcoded version. Should we follow suit?

OrJDev: The sveltekit version is wrong, ad the user also needs to install @auth/core on his project in order to get access for providers, @balazsorban44 said that in the future we will have an @auth/providers package

OrJDev: https://github.com/OrJDev/next-auth/commit/f2b7afa65385b2c82365c22c805007d37393d14c

no longer hardcoding the version tho

AlexErrant: How about `"@auth/core": "*"`? Not sure why the `~ || ^` aren't working, but I currently get this in my own project:

image

PR #6939, opened 2023-03-13
chore(ci): disable node cache to speed up ci

boredland:
## ☕️ Reasoning

in my experience restoring and storing cache costs so much time, that the (tiny if even) gain during install is gone. this tries to prove that for next-auth.

[before](https://github.com/nextauthjs/next-auth/actions/runs/4409775149/jobs/7726475689): 

install: 1m 16s
post-setup: 1m 44s

[after](https://github.com/nextauthjs/next-auth/actions/runs/4409880568/jobs/7726722542):

install: 1m 11s
post-setup: 0s

... also yamllint complaint about redundant string indicators.

## 🧢 Checklist

- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged

PR #7332, opened 2023-04-21
feat(ci): enable `pnpm` caching in CI

SauravMaheshkar:
Changes made by this PR can be summarized as follows:

- Set the `cache` and `cache-dependency-path` argument to enable caching to speed up CI times.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,6 +24,8 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 18
+          cache: "pnpm"
+          cache-dependency-path: "pnpm-lock.yaml"
       - name: Install dependencies
         run: pnpm install
       - name: Run tests
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,7 +25,6 @@ jobs:
         with:
           node-version: 18
           cache: "pnpm"
-          cache-dependency-path: "pnpm-lock.yaml"
       - name: Install dependencies
         run: pnpm install
       - name: Run tests


.github/workflows/release.yml
```diff
@@ -24,6 +24,8 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 18
+          cache: "pnpm"
+          cache-dependency-path: "pnpm-lock.yaml"
```
AlexErrant, line 28:
```suggestion
```

I don't think you need this since its defined in the action [here](https://github.com/actions/setup-node/blob/main/src/cache-utils.ts#L20). I didn't need `cache-dependency-path` [when I got it working](https://github.com/AlexErrant/Pentive/actions/runs/4758458558/workflow) in my own repo:

chrome_auuC9PQYFY

SauravMaheshkar: Ahh yes, makes sense. I found the same in the [codebase](https://github.com/actions/setup-node/blob/5b32c9063c6eb277dacbf3e916a8bc5d2617c5b7/src/cache-utils.ts#L20)

SauravMaheshkar: Fixed in 76d7e92b080c95faeaec29593db88e7769d26843

balazsorban44:
Thanks, but based on #6939 this actually showed a slowdown. 🤔

SauravMaheshkar:
> Thanks, but based on #6939 this actually showed a slowdown. 🤔

Ah sorry missed the PR. Closing this

AlexErrant:
FWIW I left a comment on that PR that described why their analysis was flawed.

balazsorban44:
FYI, I re-added caching as I can see node_modules being installed on every run, and I'm leaning towards that https://github.com/nextauthjs/next-auth/pull/6939#issuecomment-1518709011 is a true statement.

PR #7357, opened 2023-04-24
feat: add  identifier as optional parameters to support advanced verification token generation

### LINKED ISSUE #7356: support passwordless login without a database (same-device only)
### Description 📓

Currently the generateVerificationToken option in the email provider assumes that a _completely **random string**_  should be generated.

This makes sense as the database adapter needs to persist a hash of the verification process `createVerificationToken` and `useVerificationToken` merthods.

This assumption actually prevent advanced and more performant alternatives.

For instance a stateful approach put additional load on the database to persist, check and vacuum previous authentication verification codes.

I have had great success using self-contained, stateless encryption container such as iron-session.

An iron-session seal will contain the identifier of the user in itself as well as TTL. As such it does not need to be stored in database.

Ideally the generate verification method should pass the user identifier, so that stateless tokens can be used as follows:

- createVerificationToken would be a noop
- useVerificationToken would decode and check the stateless token, throw otherwise.

Right now I can't use this method with the existing customization options as no arguments are  passed to generateVerificationToken.

We'd simply need to replace
https://github.com/nextauthjs/next-auth/blob/6132c3fa751b647dd8261ac9b30e24fb140a2f1a/packages/next-auth/src/core/lib/email/signin.ts#L15

```ts
  // Generate token
  const token =
    (await provider.generateVerificationToken?.()) ??
    randomBytes(32).toString("hex")
```

with 

```ts
  const token =
	(await provider.generateVerificationToken?.({
		identifier,
		expires,
		url: _url,
		provider
	})) ?? randomBytes(32).toString('hex');
```

Thoughts?

I'd be happy to help implement this feature.

### How to reproduce ☕️

The email provider is as follow
```
 Providers.Email({
   async generateVerificationToken() {
     return "ABC123"
   }
 })
```

generateVerificationToken expects no argument.

### Contributing 🙌🏽

tclain-pillar:
> _NOTE_:
>

## ☕️ Reasoning

Having access to the expires date and identifier in the generateVerificationToken allows for advanced verification token generation, for instance stateless cryptographically secure payloads.

## 🧢 Checklist

- [X] Add a backward compatible option to generateVerificationToken in the EmailProvider that allow a verification token to be generated using the identifier as source material
- [ ] Tests -> I'd be happy to add tests, where should I put them?
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: #7356 

## 📌 Resources

### DIFF
--- a/docs/docs/guides/providers/email-provider.md
+++ b/docs/docs/guides/providers/email-provider.md
@@ -214,3 +214,18 @@ providers: [
   })
 ],
 ```
+
+If the generated token depends on the identifier, you can access it as an optional first parameter:
+
+```js title="pages/api/auth/[...nextauth].js"
+providers: [
+  EmailProvider({
+    async generateVerificationToken({identifier}) {
+      // example encryption scheme
+      return nacl(identifier)
+    }
+  })
+],
+```
+
+
--- a/packages/core/src/lib/email/signin.ts
+++ b/packages/core/src/lib/email/signin.ts
@@ -11,7 +11,10 @@ export default async function email(
 ): Promise<string> {
   const { url, adapter, provider, callbackUrl, theme } = options
   const token =
-    (await provider.generateVerificationToken?.()) ?? randomString(32)
+    (await provider.generateVerificationToken?.({
+      identifier,
+      expires,
+    })) ?? randomString(32)
 
   const ONE_DAY_IN_SECONDS = 86400
   const expires = new Date(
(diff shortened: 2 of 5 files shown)


balazsorban44 (changes requested):
See https://github.com/nextauthjs/next-auth/issues/7356#issuecomment-1523232982

I think we can make this unnecessary by not requiring an `adapter` when an EmailProvider is added, and saving the hashed token in an encrypted cookie instead.

balazsorban44:
Closing as the PR is outdated. If you could clean this up and re-open, that would be great! :pray:

Issue #6582, opened 2023-02-01
Dynamo adapter can't deploy to edge


arackaf:
### Environment

I'm using SvelteKit so I ran 

```
npx envinfo --system --binaries --browsers --npmPackages "@sveltejs/kit,svelte,@auth/*"
```

```
  System:
    OS: macOS 13.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 35.95 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.11.1 - /usr/local/Cellar/nvm/0.38.0/versions/node/v16.11.1/bin/node
    Yarn: 1.22.17 - /usr/local/Cellar/nvm/0.38.0/versions/node/v16.11.1/bin/yarn
    npm: 9.3.0 - ~/Documents/git/booklist/svelte-kit/node_modules/.bin/npm
  Browsers:
    Chrome: 109.0.5414.119
    Firefox: 109.0.1
```

### Reproduction URL

Not quite a repro, but this shows the error I'm reporting:

https://vercel.com/adam-rackis/booklist-svelte-kit/9KDXU5pMSbXbNsHS3nvk8MKNrqdi

EDIT - repro is here: https://github.com/arackaf/dynamo-adapter-edge-deploy

`npm i` then `npm run build` and there should be an error about `crypto` not existing. See comments below for more info

### Describe the issue

When trying to deploy to edge with Vercel, the dynamo adapter causes things to fail 

>  Error: Build failed with 2 errors:
--
00:37:29.818 | .svelte-kit/output/server/entries/endpoints/api/cover-upload/_server.ts.js:1:17: ERROR: Could not resolve "path"
00:37:29.818 | node_modules/@next-auth/dynamodb-adapter/dist/index.js:4:25: ERROR: Could not resolve "crypto"
00:37:29.818 | at failureErrorWithLog (/vercel/path0/svelte-kit/node_modules/esbuild/lib/main.js:1604:15)

### How to reproduce

Use the dynamo adapter, and try to deploy to the edge (with SvelteKit and the vercel adapter, but presumably with Next, too)

### Expected behavior

Ideally it would be great if this worked. Is it possible for the adapter to use the web crypto api, instead of Node's?

balazsorban44:
The Vercel logs aren't public. Please add a proper reproduction

arackaf:
[...]

PR #7427, opened 2023-05-02
refactor: replace uuid with Node.js crypto.randomUUID

Juneezee:
## ☕️ Reasoning

Starting from Node.js v14.17.0 and v15.6.0, we can use Node.js's `randomUUID` function in the crypto module to generate version 4 UUID. We don't need the `uuid` package anymore.

Reference: https://nodejs.org/docs/latest-v18.x/api/crypto.html#cryptorandomuuidoptions

## 🧢 Checklist

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/adapter-dynamodb/package.json
+++ b/packages/adapter-dynamodb/package.json
@@ -52,11 +52,7 @@
     "@next-auth/adapter-test": "workspace:*",
     "@next-auth/tsconfig": "workspace:*",
     "@shelf/jest-dynamodb": "^2.1.0",
-    "@types/uuid": "^9.0.0",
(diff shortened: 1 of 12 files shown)


balazsorban44:
Thanks, but this is currently blocked by the fact that Node.js 16 only exposes the `crypto` API globally if you enable a flag: https://nodejs.org/docs/latest-v16.x/api/globals.html#crypto_1

Once that is lifted, we can drop the dependency. Otherwise, we would not be compliant with runtimes that do not need to import `crypto`.

See #6582

Issue #7401, opened 2023-04-29
Docs: Request to Add to Sign In Callback Documentation


JSMarsh813:
### What is the improvement or update you wish to see?

Description 🐜

Request to add to the documentation for the Sign In Callback to show how a magic link can be sent to only existing users in a database.

Although this answer helped me figure out how to do this, I would rather add it to the documentation so its clear how to accomplish this for future next-auth users. https://github.com/nextauthjs/next-auth/issues/1229

Why its important: If new users can sign in with the magic link, this can lead to deformed user objects in the database. Instead, this code makes it possible to check your database and force new users to go to the registration page if they do not exist in the database.

Is this a bug in your own project?
No 

How to reproduce ☕️
N/A

Screenshots / Logs 📽
N/A

Link to the original documentation:
https://authjs.dev/guides/basics/callbacks

Link to the pull request:
https://github.com/JSMarsh813/next-auth/pull/1

Visual of what my commit would look like:
![image](https://user-images.githubusercontent.com/101692334/236629369-e9d2e7e3-c3fd-4c0f-b836-541ab3ae562b.png)
Environment 🖥
N/A

Contributing 🙌🏽

Yes, I am willing to help solve this in a PR. Thanks!

### Is there any context that might help us understand?

I want to help make the documentation clearer about how to use signin callback to limit which users will receive a magic login link.

### Does the docs page already exist? Please link to it.

https://authjs.dev/guides/basics/callbacks

balazsorban44:
Closing as fixed in #7663

Issue #1229, opened 2021-02-01
Is it possible to use magic link sign-in only for pre-existing users? 


KerimG:
I would like to migrate our website to Next.js and we already have a database with users. I would like to move to magic links for sign-in but only for those existing users. 

Unfortunately, the docs for the email auth provider don't show how to accomplish that.

iaincollins:
You should be able to use the signIn callback for this:
https://next-auth.js.org/configuration/callbacks#sign-in-callback

You should be able to look up the email address in the database (or just against a hard coded list!), I think it will be in `user.email`. If it doesn't exist in the database you can return `false` or a URL to redirect to (otherwise, returning `true` will allow them to sign in).

This works great for restricting access via email and OAuth providers like Google that have a 'email verified' property. You can even use this with oAuth providers that verify email addresses to provide access control on small sites without needing a database, as you can just have an array of email addresses (or a flat file with a list of them) and only allow listed accounts to create a session.

iaincollins:
NB: This is an area I think we could improve on and reduce the code required, with some sort of "existing accounts only" option that would apply to all providers, but that's going to need a bit of thinking about with regard to should be tied to email address or specific OAuth provider account; email addresses are not trust worthy across all oAuth providers but tying it to an OAuth account would require knowing things like each users Account ID (which they are not going to know).

Ultimately a good way to address this is probably to implement some sort of "invite" option, where people can generate one time links to invite people to activate (much like the email sign in links), though that will also require some thought.

KerimG:
Thanks a lot, @iaincollins! I completely forgot about the sing-in-callback!

PR #7470, opened 2023-05-07
Doc: Sending a Magic Link to Only Existing Users

ghiblimagic:
## ☕️ Reasoning

Request to add to the documentation for the Sign In Callback to show how to send magic links to only existing users in a database.

Although this answer helped me figure out how to do this, I would rather add it to the documentation so its clear how to accomplish this for future next-auth users. https://github.com/nextauthjs/next-auth/issues/1229

Why its important: If new users can sign in with the magic link, this can lead to deformed user objects in the database. Instead, this code makes it possible to check your database and force new users to go to the registration page if they do not exist in the database.

It would edit this doc:
https://github.com/nextauthjs/next-auth/edit/v4/docs/docs/configuration/callbacks.md

image

## 🧢 Checklist

- [X ] Documentation
- [ ] Tests (N/A)
- [X ] Ready to be merged

## 🎫 Affected issues

This was an issue about the same topic: https://github.com/nextauthjs/next-auth/issues/1229 Although it was closed, I'd rather add to the documentation to make it clear to new nextauth users how to accomplish this.

Fixes: It will close this issue I opened: https://github.com/nextauthjs/next-auth/issues/7401

## 📌 Resources

### DIFF
--- a/docs/docs/configuration/callbacks.md
+++ b/docs/docs/configuration/callbacks.md
@@ -62,6 +62,27 @@ callbacks: {
 
   You can check for the `verificationRequest` property to avoid sending emails to addresses or domains on a blocklist (or to only explicitly generate them for email address in an allow list).
 
+  You can ensure that only existing users are sent a magic login link. You will need to grab the email the user entered and check your database to see if the email already exists in the User collection in your database. If it exists, send the user a magic link but otherwise, you can send the user to another page, such as "/register". 
+
+  ```js title="pages/api/auth/[...nextauth].js"
+  import User from "../../../models/User"; 
+  import db from "../../../utils/db"; 
+  ...
+  callbacks: {
+    async signIn({ user, account, email }) { 
+      await db.connect(); 
+      const userExists = await User.findOne({ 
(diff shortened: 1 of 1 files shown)


ndom91 (commented):
First of all - thanks a lot for contributing this back to us. This is a use-case that others have most likely come across as well.

However, I don't think this is the best place for it. It's a relatively niche thing, right, not every `Email` provider using user will want this. 

So I think it would fit better either in the [`Email` provider docs](https://next-auth.js.org/providers/email) themselves. Or if you flesh it out a bit more with more details, then as its own guide/tutorial [here](https://next-auth.js.org/tutorials).

JSMarsh813:
Thanks for the warm welcome and for looking over this pull request ndom91!

After seeing your comment, I had planned on making a guide for this, but I decided that I'd be better to submit it to the email provider docs for now and maybe submit a tutorial one of these days. 

I agree with your assessment that this information would fit better in the email provider docs instead of the callback documentation. I'll close this pull request and submit a new pull request for the email provider docs.

PR #7575, opened 2023-05-17
fix(server actions): origin is under headers object

Kinbaum:
## ☕️ Reasoning

Testing out server actions in Next 13.4 I've noticed that the `experimental` version of next auth complains about `NEXTAUTH_URL` not being set. The result is that upon sign in I'm always redirected to localhost:3000.

This PR pulls the origin from the headers rather than the req object as it is not defined there.

### DIFF
--- a/packages/next-auth/src/core/index.ts
+++ b/packages/next-auth/src/core/index.ts
@@ -140,7 +140,7 @@ export async function AuthHandler<
     authOptions,
     action,
     providerId,
-    origin: req.origin,
+    origin: req.headers.origin,
     callbackUrl: req.body?.callbackUrl ?? req.query?.callbackUrl,
     csrfToken: req.body?.csrfToken,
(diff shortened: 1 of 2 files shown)


ThangHuuVu (commented):
Thanks, we truly appreciate the effort! Could you open a small reproduction so that we can check the validity of this issue first?

lopesleandrorj:
@ThangHuuVu I am also facing the same issue using a custom provider on NextJS 13. When Authenticating from the server side, no matter which URL I am coming from, it will redirect me to NEXTAUTH_URL.

You can also verify it by adding a redirect function to callbacks Object: url will never change.

ndom91:
@Kinbaum are you still experiencing this issue with the latest release of next-auth? Because you're PR is pointed at the `v4` branch which we're not updating much anymore. 

If so, could you please provide a reproduction? Thanks!

PR #7600, opened 2023-05-19
docs(prisma adapter): fix emailVerified field type of all User models in prisma adapter

moinulmoin:
## ☕️ Reasoning

I found out that in user model of all schema and ts files in prisma adapter, we are using `emailVerified DateTime?`, [(screenshot from docs)](https://authjs.dev/reference/adapter/prisma)
image

then replaced them all with `emailVerified Boolean?` and everything else was just the way they were.

## 🧢 Checklist

- [x] Documentation
- [ ] ~~Tests~~
- [x] Ready to be merged

### DIFF
--- a/apps/dev/nextjs-v4/prisma/schema.prisma
+++ b/apps/dev/nextjs-v4/prisma/schema.prisma
@@ -42,7 +42,7 @@ model User {
   id            String    @id @default(cuid())
   name          String?
   email         String?   @unique
-  emailVerified DateTime?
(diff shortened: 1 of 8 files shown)


balazsorban44:
Thanks, but this is correct https://github.com/nextauthjs/next-auth/blob/7711eb07e807bed511aa8e35360c1a057afd1139/packages/next-auth/src/core/lib/callback-handler.ts#L107

PR #7689, opened 2023-05-31
Please change the prisma adapter Account providerType to type and providerId to provider

Kailash8799:
Hello this is a small issue in google login auth in next auth when we are copy code which are given in next-auth docs this will be give you  error please change the below code

model Account {
  id                 String    @id @default(cuid())
  userId             String
  providerType       String
  providerId         String
  providerAccountId  String
  refreshToken       String?
  accessToken        String?
  accessTokenExpires DateTime?
  createdAt          DateTime  @default(now())
  updatedAt          DateTime  @updatedAt
  user               User      @relation(fields: [userId], references: [id])

  @@unique([providerId, providerAccountId])
}

change above code to below code 

model Account{
  id String @id @default(auto()) @map("_id") @db.ObjectId
  userId String @db.ObjectId
  type String
  provider String
  providerAccountId String
  refresh_token String? @db.String
  access_token String? @db.String
  expires_at Int?
  token_type String?
  scope String?
  id_token String? @db.String
  session_state String?

  user User @relation(fields: [userId],references: [id],onDelete:Cascade)
  @@unique([provider,providerAccountId])
}


balazsorban44:
Hi, this PR was likely opened by mistake.

Anyway, the suggested change probably wouldn't make it, since the Account model already has a `provider` field as you can see in https://authjs.dev/reference/adapters#models

And `type` might be a field returned by providers, so it would crash if renamed `providerType`.

Thanks though!

PR #7703, opened 2023-06-01
feat: introduce `@auth/prisma-adapter`

balazsorban44:
Database adapters are not dependent on Next.js features, so it makes sense to republish them under the `@auth/*` scope.

This PR is the first of many, to convert adapters, using `@auth/core` for types.

To upgrade, in your package.json:

```diff
-  "@next-auth/prisma-adapter": "0.0.0",
+  "@auth/prisma-adapter": "0.0.0",  
```

And run `npm install`, `yarn install` or `pnpm install` respectively.

**Note:** This packages is now published ESM-only.

PR #7706, opened 2023-06-01
feat: introduce `@auth/typeorm-adapter`

balazsorban44:
Follow-up of #7703

Database adapters are not dependent on Next.js features, so it makes sense to republish them under the `@auth/*` scope.

This PR is part of a series to convert adapters, using `@auth/core` for types.

BREAKING CHANGE:
If you are coming from the previous adapter, change your `package.json`:

```diff
-  "@next-auth/typeorm-legacy-adapter": "0.0.0",
+  "@auth/typeorm-adapter": "0.0.0",
```

And run `npm install`, `yarn install` or `pnpm install` respectively.

**Note:** This packages is published as ESM-only


packages/adapter-typeorm/package.json
```diff
@@ -38,26 +46,27 @@
...
     "@next-auth/adapter-test": "workspace:*",
     "@next-auth/tsconfig": "workspace:*",
     "jest": "^27.4.3",
     "mssql": "^7.2.1",
     "mysql": "^2.18.1",
-    "next-auth": "workspace:*",
     "pg": "^8.7.3",
     "sqlite3": "^5.0.8",
-    "typeorm": "0.3.7",
+    "typeorm": "0.3.15",
```
balazsorban44, line 60: Locked in on `0.3.15` as something is likely broken in `0.3.16`

I think https://github.com/typeorm/typeorm/pull/10041 would solve the issue.

PR #7856, opened 2023-06-22
fix(adapters): Add .js file extension to relative imports

estevebadia:
## ☕️ Reasoning

Added .js file extension to the relative imports so that node doesn't fail to resolve the internal relative imports in this module (ex: from index.js to ./entities). 

Note that:
1. Node requires file extension to imports in ESM (https://nodejs.org/api/esm.html#mandatory-file-extensions).
2. Typescript does not add/change the file extension of imports when transpiling to js.
3. The other code in this repository is written adding .js extension to relative imports, so this is the convention to use (See for example https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/index.ts).

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources

### DIFF
--- a/packages/adapter-typeorm/src/index.ts
+++ b/packages/adapter-typeorm/src/index.ts
@@ -21,8 +21,8 @@ import type {
   AdapterSession,
 } from "@auth/core/adapters"
 import { DataSourceOptions, DataSource, EntityManager } from "typeorm"
-import * as defaultEntities from "./entities"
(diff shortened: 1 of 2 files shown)


balazsorban44 (approved):
Good catch, thank you!

PR #7900, opened 2023-06-29
Split and build cjs/mjs config for distribution

december1981:
## ☕️ Reasoning

Trying to use mikro-orm cli tool to migrate, and it wouldn't work, as it requires a cjs distribution. I've added support for both mjs/cjs for this adapter package now.

`npx mikro-orm migration:create --initial`

## 🧢 Checklist

- [ ] Documentation (NA)
- [x] Tests
- [x] Ready to be merged


balazsorban44:
Hi, we do not intend to split the distribution, as the ecosystem is already fragmented enough. ESM is the only way we intend to ship future libraries.

balazsorban44:
It's unclear to me though where the issue is coming from, you aren't supposed to use the adapter package for migration. 🤔 Can you open a bug report with an attached reproduction/explanation of the problem first? Thanks!

december1981:
It's a bit of a pain to open up an issue (I need to include a github repo to the code, etc), so I'll give you a heads up here:

It's not uncommon to do this in mikro-orm.config.ts (which is used by mikro-orm and its migration cli ):

`import { defaultEntities } from "@auth/mikro-orm-adapter";`

Here's an example in authjs.dev itself for why this might be done:
https://authjs.dev/reference/adapter/mikro-orm#including-default-entities

For my use case, with a tsconfig like so:

```json
{
  "compilerOptions": {
    "target": "es2020",
    "lib": ["dom", "dom.iterable", "esnext"],
    "allowJs": true,
    "skipLibCheck": true,
    "strict": true,
    "forceConsistentCasingInFileNames": true,
    "noEmit": true,
    "experimentalDecorators": true,
    "emitDecoratorMetadata": true,
    "esModuleInterop": true,
    "module": "esnext",
    "outDir": "out",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "sourceMap": true,
    "jsx": "preserve",
    "incremental": true,
    "plugins": [
      {
        "name": "next"
      }
    ],
    "baseUrl": "./src",
    "paths": {
      "@/*": ["./src/*"]
    }
  },
  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
  "exclude": ["node_modules"]
}
```

And the following in package.json for the mikro-orm cli config

```json
  "mikro-orm": {
    "useTsNode": true,
    "configPaths": [
      "./src/mikro-orm.config.ts",
      "./dist/mikro-orm.config.js"
    ]
  },
```

Now, trying to run `npx mikro-orm migration:create --initial` (or `migration:up`, etc) leads to:

`Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in <snip>/node_modules/@auth/mikro-orm-adapter/package.json`

That happens because the above import fails for the cli tool as it is running with "require", not "import" package exports flag set.  (ie in cjs mode)

I think another solution might be for mikro-orm to support using ts-node-esm? I'm not au fait enough with ts-node to be sure, but if that's viable, it would mean raising an issue their side.

december1981:
After reading up on ts-node, some more, as a work around, I wrote this utility script, which temporarily sets "type": "module", to the my next.js application's package.json, then runs ts-node through mikro-orm's cli with `NODE_OPTIONS="--loader ts-node/esm"`:

```sh
#!/bin/sh

# No set -e, we always want to restore our package.json
# we'll exit with any non zero exit code from mikro-orm manually

tmpjson=$(mktemp /tmp/package.XXXXXX.json)
exec 3>"$tmpjson"
exec 4<"$tmpjson"
exec 5<"$tmpjson"
rm "$tmpjson"

jq '. + {"type": "module"}' < package.json >&3
cat <&4 > package.json

NODE_OPTIONS="--loader ts-node/esm" npx mikro-orm "$@"
RET_CODE=$?

jq 'del(.type)' <&5 > package.json
exit $RET_CODE
```

PR #7914, opened 2023-06-30
chore: remove extra 'if' in comment

fatihsolhan:
Removed the extra 'if' word from the comment of function `withAuth ` to correct a typo and improve the accuracy of the documentation.


ThangHuuVu (commented):
Could you also change in the "core" dir? That's where the code for v5 is 🫶

balazsorban44 (approved):

PR #7936, opened 2023-07-02
Fix: session deletion throws an error, if user has already been deleted

### LINKED ISSUE #4495: Prisma throws on attempted to delete record that does not exist but next-auth core does not expect that

sharknoon:
# Scenario
A user closes his account by deleting himself from the database and all his sessions with it.
Another user in the same browser tries to log in via email.
The client will send the old session token to the server, who is correctly identifying it as the wrong one and wants to delete it:
https://github.com/nextauthjs/next-auth/blob/a79774f6e890b492ae30201f24b3f7024d0d7c9d/packages/core/src/lib/callback-handler.ts#L95-L102
This works well in cases where the first user just logs out, but when the first user deletes his account, the function `deleteSession(...)` will fail on the login attempt of the second user.  

**Thank you so much for creating this awesome library, it is really a joy to work with 😀**
  
## ☕️ Reasoning

My proposed solution is a simple `try catch` block around that statement.

## 🧢 Checklist

- [ ] Documentation (not needed)
- [ ] Tests (not needed)
- [x] Ready to be merged

## 🎫 Affected issues

A similar issue is this one #4495.
Although the issue is closed, the proposed solution (deleting the user via the adapter) does not work in my case.
It still keeps the session token on the client side and throws this error.

Fixes: #4495 

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/callback-handler.ts
+++ b/packages/core/src/lib/callback-handler.ts
@@ -98,7 +98,11 @@ export async function handleLogin(
         // Delete existing session if they are currently signed in as another user.
         // This will switch user accounts for the session in cases where the user was
         // already logged in with a different account.
-        await deleteSession(sessionToken)
+        // Catch any errors in case the other user is already deleted in the database
+        // and therefore the session does not exist anymore.
+        try {
+          await deleteSession(sessionToken)
+        } catch { }
       }
 
       // Update emailVerified property on the user object


Sharknoon:
@ThangHuuVu do you have any questions regarding this PR? Is there something I can help you with?

ThangHuuVu:
Thanks for sending the PR, but I think we don't want to merge it. For this issue, it is not a problem that `next-auth` needs to handle. The core functionalities of next-auth is to handle the interaction between the user and the app. So everything else that happen outside of those flows are not handled intentionally to avoid complexity.

PR #7979, opened 2023-07-07
feat(sveltekit): Add server side signIn and signOut functions

### LINKED ISSUE #7098: Is there a server side version of SvelteKit signIn() / signOut()?
### Question 💬

There are examples of calling `signIn()` or `signOut()` from the browser, but I need to call this from the server in SSR context. Is this possible somehow?

### How to reproduce ☕️

n/a

### Contributing 🙌🏽

ksmoore17:
## ☕️ Reasoning

`@auth/sveltekit` only offers client side `signIn` and `signOut` auth functions. The client side functions cannot be used on the server because they use native `fetch` to make a request (instead of sveltekit's `RequestEvent.fetch`) and `window.location` to redirect. This pr adds `signIn` and `signOut` inside `@auth/sveltekit/server` that internally use the same request to `/auth` endpoints as their client side counterparts (see unexported `@auth/sveltekit/common`). Sign in/out from the server can be useful in some cases:

- Auth.js frontend is not desired
- there is a single provider (or the provider is known for some reason)
- site has custom login form with logic in form actions

It may be possible to enable this for server and client in one function, but you would need to add a fetch function parameter (as I did for server `signIn` and `signOut`) and pass `{ redirect: false }` in `options` to `signIn` (to avoid using `window`) and perform the sveltekit redirect with the url in the returned response value, but I think this pr makes it more clear on the server (and includes examples anyways).

## 🧢 Checklist

- [x] Documentation
- [x] Tests _(in the form of changes to the sveltekit example)_
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: #7098

## 📌 Resources


aakash14goplani:
When are we expecting it to be released?

ThangHuuVu:
hello @ksmoore17 👋 I closed this regarding #9714, where I implemented it differently, making it compatible with [Kit's Form Actions](https://kit.svelte.dev/docs/form-actions). I avoid using `fetch` on the server side, so we don't have to pass the `fetch` object around like in your PR. Thanks for the initiative!

Issue #8188, opened 2023-07-31
Apple Provider not working with provided documentation


ChrGrb:
### Provider type

Apple

### Environment

  System:
    OS: macOS 14.0
    CPU: (8) arm64 Apple M1 Pro
    Memory: 68.98 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.15.0 - ~/.nvm/versions/node/v18.15.0/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 9.5.0 - ~/.nvm/versions/node/v18.15.0/bin/npm
    pnpm: 8.6.10 - ~/.nvm/versions/node/v18.15.0/bin/pnpm
  Browsers:
    Chrome: 110.0.5481.177
    Safari: 17.0

### Reproduction URL

https://authjs.dev/reference/core/providers_apple

### Describe the issue

Using the Apple provider in it's current state gives a userinfo endpoint not configured error, as described in this issue: https://github.com/nextauthjs/next-auth/issues/6788#issue-1595885175

This can be fixed by simply providing a token endpoint (I feel like this should be stated in the error message).

The current Apple provider is not correctly configured. The authorization request is sent using the response mode form_post, which returns the code used for the OAuth flow in the body, instead of the query.  The current approach only sends the query to the handleOAuth function and not the body.

One fix for this is to define `response_mode: "query"`and set `scope: ""`, as the endpoint does not allow the request of additional information in this mode.

This makes the provider work, but fails to get the name or email of the user.

For the case that the body is passed instead of the query the relevant user data is in the authorization response, instead of the token response. This would probably be the behaviour that the library currently expects. This behaviour is also described in this issue: https://github.com/nextauthjs/next-auth/issues/7596#issue-1717389831

There is currently no way to get to this additional data.

Additionally the provider is currently missing the wellKnown and authorization endpoint.

### How to reproduce

* Create a project using the default Apple Provider, as stated in the docs (make sure to use https and not use localhost as the domain): [https://next-auth.js.org/providers/apple](https://next-auth.js.org/providers/apple)
* Enable debug mode: [https://next-auth.js.org/configuration/options#debug](https://next-auth.js.org/configuration/options#debug)
* Try to sign in using your AppleID

### Expected behavior

* The call of handleOAuth(...) inside lib/routes/callback.ts should check if query or body contains the data and pass the correct one
[...]

Issue #6788, opened 2023-02-22
Apple provider broken - blowing up when not finding userinfo endpoint


kyllerss:
### Provider type

Apple

### Environment

  System:
    OS: Linux 5.19 KDE neon 5.27 5.27
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
    Memory: 6.74 GB / 15.34 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 16.18.0 - ~/apps/node-v16.18.0-linux-x64/bin/node
    npm: 9.3.0 - ~/apps/node-v16.18.0-linux-x64/bin/npm
  Browsers:
    Brave Browser: 110.1.48.164
    Chrome: 110.0.5481.100
    Firefox: 110.0

### Reproduction URL

Needed?

### Describe the issue

Seems as if Apple another one of those providers that does not provide user info. 

When the redirect endpoint is invoked I get the following error:

```
[auth][cause]: TypeError: TODO: Authorization server did not provide a userinfo endpoint.
     at handleOAuth (file:///xxx/node_modules/@auth/sveltekit/node_modules/@auth/core/lib/oauth/callback.js:28:19)
     ...
``` 

The line in question is [here](https://github.com/nextauthjs/next-auth/blob/2130765a577570cbb1a85ab3d540bdb40a184b94/packages/core/src/lib/oauth/callback.ts#L52).

I found [a relevant issue](https://github.com/nextauthjs/next-auth/issues/573) that had a specific work-around, but pertained to *Todoist*. 

Doesn't seem to me that requiring a user-info endpoint is part of the OAuth spec, but I could be wrong. I would rather have this restriction be a configuration option or handled explicitly by some callback. I wouldn't be surprised if the push for privacy is going to push additional partners to follow a similar pattern.

### How to reproduce

Try to do a sign-in using Apple.

### Expected behavior

For it to work and have the client specify how to handle the missing information.

jschlesser:
I'm running into this too.  @kyllerss did you find a workaround?

kyllerss:
I've put this task aside for now, so I haven't had a chance to do a workaround.

ChrGrb:
@kyllerss @jschlesser The debug message here is a bit unclear. If you provide a token endpoint, a userinfo endpoint is not necessary. 

[...]

PR #8189, opened 2023-07-31
fix: Add features missing for Apple provider

ChrGrb:
### Changes to routes/callback.ts
- If the response_mode of authorization is set to 'form_post', the information will be returned in the body and not the query. Therefor the callback handler should check for the information there to pass it on to the handleOAuth function

### Changes to providers/oauth.ts
- The Apple provider returns the user information with the authorization response. There should be a way to get to this data in the provider configuration

### Changes to providers/apple.ts
- The apple token endpoint expects authentication data to be passed in the body. Therefor the token_endpoint_auth_method has to be set to 'client_secret_post'
- wellKnown endpoint should be defined in the default provider
- authorization url should be defined in the default provider
- authorization params should be set correctly
- token url should be defined in the default provider
- The profile function should correctly handle the available data
- The documentation should mention the necessary pkceCodeVerifier cookie

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: 
* [https://github.com/nextauthjs/next-auth/issues/8188](https://github.com/nextauthjs/next-auth/issues/8188)
* [https://github.com/nextauthjs/next-auth/issues/6788](https://github.com/nextauthjs/next-auth/issues/6788)

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/oauth/callback.ts
+++ b/packages/core/src/lib/oauth/callback.ts
@@ -25,7 +25,7 @@ import type { Cookie } from "../cookie.js"
 export async function handleOAuth(
   query: RequestInternal["query"],
   cookies: RequestInternal["cookies"],
-  options: InternalOptions<"oauth" | "oidc">,
+  options: InternalOptions<"oauth" | "oidc" | "siwa">,
   randomState?: string
 ) {
   const { logger, provider } = options
@@ -125,7 +125,7 @@ export async function handleOAuth(
   let profile: Profile = {}
   let tokens: TokenSet & Pick<Account, "expires_at">
 
-  if (provider.type === "oidc") {
+  if (provider.type === "oidc" || provider.type === "siwa") {
     const nonce = await checks.nonce.use(cookies, resCookies, options)
     const result = await o.processAuthorizationCodeOpenIDResponse(
       as,
@@ -139,7 +139,12 @@ export async function handleOAuth(
       throw new Error("TODO: Handle OIDC response body error")
     }
 
-    profile = o.getValidatedIdTokenClaims(result)
+    // siwa returns the user with the authorization response
+    if (provider.type === "siwa") {
+      profile = query?.user;
+    } else {
+      profile = o.getValidatedIdTokenClaims(result)
+    }
     tokens = result
   } else {
     tokens = await o.processAuthorizationCodeOAuth2Response(
@@ -177,7 +182,6 @@ export async function handleOAuth(
     provider,
     tokens,
     logger,
-    query?.user
   )
 
   return { ...profileResult, profile, cookies: resCookies }
@@ -189,10 +193,9 @@ async function getUserAndAccount(
   provider: OAuthConfigInternal<any>,
   tokens: TokenSet,
   logger: LoggerInstance,
-  userResponse: string | undefined
 ) {
   try {
-    const user = await provider.profile(OAuthProfile, tokens, userResponse)
+    const user = await provider.profile(OAuthProfile, tokens)
     user.email = user.email?.toLowerCase()
 
     if (!user.id) {
--- a/packages/core/src/lib/oauth/callback.ts
+++ b/packages/core/src/lib/oauth/callback.ts
@@ -122,7 +122,7 @@ export async function handleOAuth(
     throw new Error("TODO: Handle www-authenticate challenges as needed")
   }
 
-  let profile: Profile = {} 
+  let profile: Profile = {}
   let tokens: TokenSet & Pick<Account, "expires_at">
 
   if (provider.type === "oidc") {
@@ -176,7 +176,8 @@ export async function handleOAuth(
     profile,
     provider,
     tokens,
-    logger
+    logger,
+    query?.user
   )
 
   return { ...profileResult, profile, cookies: resCookies }
@@ -187,10 +188,11 @@ async function getUserAndAccount(
   OAuthProfile: Profile,
   provider: OAuthConfigInternal<any>,
   tokens: TokenSet,
-  logger: LoggerInstance
+  logger: LoggerInstance,
+  userResponse: string | undefined
 ) {
   try {
-    const user = await provider.profile(OAuthProfile, tokens)
+    const user = await provider.profile(OAuthProfile, tokens, userResponse)
     user.email = user.email?.toLowerCase()
 
     if (!user.id) {
--- a/packages/core/src/lib/oauth/callback.ts
+++ b/packages/core/src/lib/oauth/callback.ts
@@ -140,6 +140,11 @@ export async function handleOAuth(
     }
 
     profile = o.getValidatedIdTokenClaims(result)
+
+    if (provider.profileConform) {
+      profile = provider.profileConform(profile, query);
+    }
+
     tokens = result
   } else {
     tokens = await o.processAuthorizationCodeOAuth2Response(
@@ -176,8 +181,7 @@ export async function handleOAuth(
     profile,
     provider,
     tokens,
-    logger,
-    query?.user
+    logger
   )
 
   return { ...profileResult, profile, cookies: resCookies }
@@ -189,10 +193,9 @@ async function getUserAndAccount(
   provider: OAuthConfigInternal<any>,
   tokens: TokenSet,
   logger: LoggerInstance,
-  userResponse: string | undefined
 ) {
   try {
-    const user = await provider.profile(OAuthProfile, tokens, userResponse)
+    const user = await provider.profile(OAuthProfile, tokens)
     user.email = user.email?.toLowerCase()
 
     if (!user.id) {
--- a/packages/core/src/lib/oauth/callback.ts
+++ b/packages/core/src/lib/oauth/callback.ts
@@ -25,7 +25,7 @@ import type { Cookie } from "../cookie.js"
 export async function handleOAuth(
   query: RequestInternal["query"],
   cookies: RequestInternal["cookies"],
-  options: InternalOptions<"oauth" | "oidc" | "oidcfp">,
+  options: InternalOptions<"oauth" | "oidc">,
   randomState?: string
 ) {
   const { logger, provider } = options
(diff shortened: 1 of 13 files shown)


packages/core/src/providers/apple.ts
```diff
@@ -158,17 +170,38 @@ export default function Apple<P extends AppleProfile>(
     name: "Apple",
     type: "oidc",
     issuer: "https://appleid.apple.com",
+    wellKnown: "https://appleid.apple.com/.well-known/openid-configuration",
+    checks: ["pkce"],
+    token: {
+      url: `https://appleid.apple.com/auth/token`,
+    },
```
balazsorban44, line 177:
```suggestion
    wellKnown: "https://appleid.apple.com/.well-known/openid-configuration",
    checks: ["pkce"],
    token: {
      url: `https://appleid.apple.com/auth/token`,
    },
```

This is not necessary, as we use the `issuer` to discover the well-known endpoint for this metatdata.

packages/core/src/providers/apple.ts
```diff
@@ -158,17 +170,38 @@ export default function Apple<P extends AppleProfile>(
...
     authorization: {
-      params: { scope: "name email", response_mode: "form_post" },
+      url: 'https://appleid.apple.com/auth/authorize',
+      params: {
+        scope: "name email",
+        response_type: "code",
+        response_mode: "form_post",
+        state: crypto.randomUUID()
+      },
```
balazsorban44, line 185:
```suggestion
      params: {
        scope: "name email",
        response_mode: "form_post",
      },
```

Already added/default.

packages/core/src/providers/apple.ts
```diff
@@ -106,7 +106,8 @@ export interface AppleProfile extends Record<string, any> {
  *
  * #### Configuration
  *
- * Import the provider and configure it in your **Auth.js** initialization file:
+ * Import the provider and configure it in your **Auth.js** initialization file.
+ * Also add the pkceCodeVerifier cookie to the cookies option.
```
balazsorban44, line 110:
```suggestion
 * Import the provider and configure it in your **Auth.js** initialization file:
```

This seems unnecessary, since the option is already added via `checks: ["pkce"]`

packages/core/src/providers/apple.ts
```diff
@@ -115,10 +116,21 @@ export interface AppleProfile extends Record<string, any> {
...
+ *     pkceCodeVerifier: {
+ *       name: "next-auth.pkce.code_verifier",
+ *       options: {
+ *         httpOnly: true,
+ *         sameSite: "none",
+ *         path: "/",
+ *         secure: true,
+ *       },
+ *     },
+ *   },
```
balazsorban44, line 133:
```suggestion
```

See https://github.com/nextauthjs/next-auth/pull/8189/files#r1281198762

packages/core/src/providers/apple.ts
ChrGrb: The reason I added the pkce cookie manually is that I get this error when removing it, even if the pkce check is set:
```
sveltekit-auth-app:dev: [auth][error][InvalidCheck]: Read more at https://errors.authjs.dev#invalidcheck
sveltekit-auth-app:dev: [auth][cause]: InvalidCheck: PKCE code_verifier cookie was missing.
```

Could this be an issue with SvelteKitAuth not correctly calling the init function?

balazsorban44: Sounds like that, yes.

balazsorban44:
I'm hesitant about merging this yet, as we don't want to make exceptions for handling providers specifically in the core library. Apple is making this hard here, not adhering to the OAuth spec... User information should be in `id_token`, and not part of a query param. :grimacing:

I recently figured out how to finally test an Apple provider (had to pay for it :-1:) myself, so I'll think about it a bit more.

_\*Venting\*: Apple really does not want people to use their provider. :sweat_smile:. By far the most convoluted/worst documented provider, and not even free..._

ChrGrb:
I 100% agree with what you say. Figuring out what is going wrong and why is a nightmare with the Apple provider and changing the core did feel off.

I did have 2 possible alternatives to this, that could be considered:

**Make the provider usable with the current core functionality**
By changing the authorization params to
```
scope: "",
response_mode: "query",
``` 
the response is returned in the url, instead of the body.
This would limit the functionality in so far that the call does neither provide an email, nor a name. But the general sign in works and the id_token is still returned.
The email and name could then be requested and stored in the database on first signIn.

**Add the ability to provide a custom `handleOAuth` function in the provider configuration**
This would move the bulk of changes to the Apple provider configuration (except for determining whether the data of the callback is in the body or query). On the other hand I assume that you have a good reason, why this is not yet an implemented feature.


Either way, if you need any assistance or get stuck, just hit me up. I know how hard it is to find documentation or information on this topic.

ChrGrb:
Doing some more digging I found that Apple actually does adhere to the OIDC spec. This is the actual, expected behaviour for an OAuth2.0 form_post request. It is just that apparently no other provider in Auth.js uses this. https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html#FormPostResponseExample 

This does, in my opinion, justify changing the core. It is practically speaking not changing it for a single provider, but instead for all potential future providers using the same form_post spec.

balazsorban44:
My ick is with specifically returning the user data in a `user` search param as s stringified JSON objoect. That is not part of the spec AFAICT.

packages/core/src/lib/oauth/callback.ts
```diff
@@ -139,7 +139,12 @@ export async function handleOAuth(
       throw new Error("TODO: Handle OIDC response body error")
     }
 
-    profile = o.getValidatedIdTokenClaims(result)
+    // oidcfp returns the user with the authorization response
+    if (provider.type === "oidcfp") {
+      profile = query?.user;
```
balazsorban44, line 144: I cannot find this documented anywhere in the spec.

balazsorban44 (changes requested):
Let's revert the last changes, it's not the correct way to handle this IMO.

balazsorban44:
(closed by accident, sorry).

> Add the ability to provide a custom handleOAuth function in the provider configuration

I'm thinking of something similar.

See https://github.com/nextauthjs/next-auth/blob/60c5037ee1dd7b11f180e83de74b66aab88d4c47/packages/core/src/lib/oauth/callback.ts#L111-L115
Here, if a provider is not adhering to the spec, we let them call into this `.conform` method, signaling that this is really on the provider and not on us. Also warning for the users so they can let us know when a provider finally fixed their stuff.

An example is Twitch:

https://github.com/nextauthjs/next-auth/blob/60c5037ee1dd7b11f180e83de74b66aab88d4c47/packages/core/src/providers/twitch.ts#L88-L115

ChrGrb:
I tried my best to implement your idea. I could not figure out a way on how to do it with only the existing conform methods, as the `authorization.conform` is afaict not being used.

What I did instead is create a `provider.profileConform(profile, query)` method that does, what the changes to the `provider.profile(..)` callback have done before. I feel like this is more readable, but in the end it still comes down to basically the same changes to the core.

hesselbom:
May I ask if there's any progress in merging this PR?

I'm currently using the fallback suggested [here](https://github.com/nextauthjs/next-auth/issues/6788#issuecomment-1658765772) but with that the email is reported as null since it can't be part of the scope.

leo3linbeck:
I can successfully sign into Apple, but this issue continues to be a problem for me since the failure to be able to get an email returned from Apple on a Vercel implementation makes the Apple provier unusable for my purposes.

After digging into it, it looks like the fix should be pretty simple: since Apple returns the state in the body of a POST (which one would expect for a POST, and is frankly better security practice) rather than as a GET query parameter (which is what the state check is looking for), the following approach should address the problem:

1. In lib/actions/callback/index.ts:

```
      const { proxyRedirect, randomState } = handleState(
        query,
        body,
        provider,
        options.isOnRedirectProxy
      )

```
2. In the handleState function in lib/actions/callback/oauth/checks.ts, the state can taken from query or body, depending on which is non falsy, and proceed accordingly.

I'd try to fix this myself, except I've never tried to submit a pull request and I'm on a tight time deadline on this project and the other providers (Google, Facebook, Auth0, and Github) all work fine so I can proceed without Apple. But given the ubiquity of iPhones, it would be great if Apple would return an email address in a Vercel implementation.

Love this library, and appreciate all of the hard work that has gone into what is an elegant solution to a tough problem.

Cheers,
L3

tristan-warner-smith:
We've hit this issue ourselves and, short of modifying the core which I don't feel qualified to do, it doesn't seem we have any recourse on this other than to hope this PR exposing the form data vs query approach gets merged.

We'd really appreciate if this could keep moving forward 🙏

JamesWalkerGit:
I've run into the same issue, it seems the Apple Auth simply doesn't work currently. I've tried 5.0.0-beta.20 and 5.0.0-beta.21 following the documentation and the issue is a server error that occurs after successfully logging in and then redirecting to the callback url. 

I've also tried some workarounds that were posted but they did not work either. 

https://github.com/nextauthjs/next-auth/issues/6788#issuecomment-2362096320

blinton-sc:
Is there a reason this wasn't merged? We are blocked due to this and so far none of the workarounds we've tried work.

It would be awesome if we could get an update here @ChrGrb @balazsorban44 @ThangHuuVu

balazsorban44:
> Is there a reason this wasn't merged? We are blocked due to this and so far none of the workarounds we've tried work.

You are not blocked. You can copy-paste the code whenever you want and use something like `pnpm patch`, the source is... open.

I honestly just don't like that we would need a workaround like this, just for the sake of Apple...

My opinion is the same as https://github.com/nextauthjs/next-auth/pull/8189#issuecomment-1664055204, Apple is one of the worst providers to work with, by far...

JamesWalkerGit:
@balazsorban44 

Does this mean that Auth.js will not support Apple login, and that the documentation around Apple OAuth support needs to be removed? 

[Apple Documentation](https://authjs.dev/getting-started/providers/apple)

For me it's a pretty big deal, Apple is one of the largest mobile device markets so to not have their login is a very large negative for me.

balazsorban44:
Never said we don't want to support it. Just not the way it's presented in this PR.

Recently merged https://github.com/nextauthjs/next-auth/pull/11975 which might help resolve this.

JamesWalkerGit:
@balazsorban44 Ok awesome, thanks for the insight, I definitely appreciate your efforts!

PR #7443, opened 2023-05-04
feat(nextjs): introduce `next-auth` v5

balazsorban44:
Next.js 13.4 [is out](https://nextjs.org/blog/next-13-4).

For discussing project-related issues, please use https://github.com/nextauthjs/next-auth/discussions/8487

The new version of NextAuth.js is based on `@auth/core`.

If you want to test it out, you can do so already, installing `next-auth@experimental`:

- **Documentation**: https://auth-docs-git-feat-nextjs-auth-authjs.vercel.app/reference/nextjs
- **Migration guide**: https://auth-docs-git-feat-nextjs-auth-authjs.vercel.app/guides/upgrade-to-v5

**BREAKING CHANGES**

Follow the [migration guide](https://auth-docs-git-feat-nextjs-auth-authjs.vercel.app/guides/upgrade-to-v5)

PR #8204, opened 2023-08-01
feat: `getToken` server component support

### LINKED ISSUE #7913: How do I access my access_token in a server component?
### Question 💬

I'm logging in using NextAuth with Azure AD as the provider, which works fine.

In the NextAuth callbacks, I receive the access_token. In middleware.ts, I can use the getToken(req, secret) method and access the access_token (but what to do with it there?). In my internal API routes, the getToken(req, secret) always returns null.

I read in another question thread that this is the expected behaviour. These values are not passed to the API routes for security reasons. 

NOTE: **The only reasons i created the internal API routes was because there I got access to the request object which is needed to use the getToken() method**

### My question then becomes: 
How do I access my JWT token in my server components? And if they are encrypted by next, how do I decode and use them? 

My NEXTAUTH_SECRET is created with openssl rand -base64 32 (I dont know if you should say that in public  )

### How to reproduce ☕️

auth.ts

```
import { NextAuthOptions } from "next-auth";
import AzureADProvider from "next-auth/providers/azure-ad";
import { CLIENT_ID, CLIENT_SECRET, TENANT_ID } from "./utils/envPublic";

const authOptions: NextAuthOptions = {
    providers: [
        AzureADProvider({
            clientId: CLIENT_ID,
            clientSecret: CLIENT_SECRET,
            tenantId: TENANT_ID,
            authorization: {
                params: {
                    scope: "openid profile email api://myapi/my.scope",
                }
            }
        })
    ],
    secret: process.env.NEXTAUTH_SECRET,
    session: {
        strategy: "jwt"
    },
    jwt: {
        secret: process.env.NEXTAUTH_SECRET
    },
    callbacks: {
        async jwt({ token, account, user }) {
            token.accessToken = account?.access_token;
            return { ...token, ...user, ...account };
        },
        async session({ session, token, user }) {
            session.token = token.accessToken;
            session.user.token = token.accessToken;
            return session;
        },
        async redirect() {
            return "/"
        }
    },

};

export default authOptions;

```

My server component:

```
const Home = async () => {
  // How do I get the access token here?
  let response = await fetch("http://external.api.that.needs.accestoken.to.work/");
  let json = await response.json();

  return (
    <div className="flex flex-col items-center">
      <p className="p-12 hover:bg-red-600 cursor-pointer">Hello there github friends</p>
    </div>
  )
};

export default Home;
```

I've also done some attempts to extend the types/interfaces to store the access_token in the callbacks.
Tbh, im not really sure how this works.
next-auth-d.ts:

```
import NextAuth, { DefaultSession } from "next-auth";
import type { Session, Account } from "next-auth";

declare module "next-auth" {

  interface Session {
    token?: accessToken;
    user: {
      token?: accessToken;
    } & DefaultSession["user"];
  }
  interface Account {
    access_token: accessToken;
    user: {
      token?: accessToken;
    } & DefaultSession["user"];
  }
}
```

### Contributing 🙌🏽

Kevin-McGonigle:
## ☕️ Reasoning

`getToken` currently requires a `req` parameter to access the session cookies and authorization header. This works great for API routes, middleware, `getServerSideProps`, etc. However, this approach means that the method cannot be called from within a server component, as there is no access to a `NextRequest` object.

This PR implements a similar approach to `getServerSession`, using the `cookies()` and `headers()` methods from `next/headers` if the `req` parameter is omitted. 

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: https://github.com/nextauthjs/next-auth/issues/7913

## 📌 Resources

### DIFF
--- a/docs/docs/guides/basics/securing-pages-and-api-routes.md
+++ b/docs/docs/guides/basics/securing-pages-and-api-routes.md
@@ -55,6 +55,8 @@ More details can be found [here](/reference/nextjs/#middleware).
 
 ### Server Side
 
+#### Using Pages Router
+
 You can protect server side rendered pages using the `unstable_getServerSession` method. This is different from the old `getSession()` method, in that it does not do an extra fetch out over the internet to confirm data from itself, increasing performance significantly.
 
 You need to add this to every server rendered page you want to protect. Be aware, `unstable_getServerSession` takes slightly different arguments than the method it is replacing, `getSession`.
@@ -110,6 +112,51 @@ export default function App({
 ```
 
 :::
+
+#### Using App Router
+You can protect client components using  the [`useSession` hook](#client-side).
+
+You can protect server components using the `getServerSession` method.
+
+```js title="app/server-component-get-server-session-example.js"
+import { getServerSession } from "next-auth/next"
+
+export default async function Page() {
+  const session = await getServerSession()
+ 
+  if (session) {
+    return (
+      <>
+        <h1>Protected Page</h1>
+        <p>You can view this page because you are signed in.</p>
+      </>
+    )
+  }
+  
+  return <p>Access Denied</p>
+}
+```
+
+If you are using JSON Web Tokens you can use the `getToken()` helper to protect a server component and access the contents of the JWT without having to handle JWT decryption / verification yourself.
+
+```js title="app/server-component-get-token-example.js"
+import { getToken } from "next-auth/jwt"
+
+export default async function Page() {
+  const token = await getToken()
+ 
+  if (token) {
+    return (
+      <>
+        <h1>Protected Page</h1>
+        <p>You can view this page because you are signed in.</p>
+      </>
+    )
+  }
+  
+  return <p>Access Denied</p>
+}
+```
 
 ## Securing API Routes
 
--- a/packages/next-auth/src/jwt/index.ts
+++ b/packages/next-auth/src/jwt/index.ts
@@ -68,7 +68,7 @@ export interface GetTokenParams<R extends boolean = false> {
  * [Documentation](https://next-auth.js.org/tutorials/securing-pages-and-api-routes#using-gettoken)
  */
 export async function getToken<R extends boolean = false>(
-  params: GetTokenParams<R>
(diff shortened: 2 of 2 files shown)


Kevin-McGonigle:
Neat little security bonus that this should provide - this should reduce the need for forwarding the `access_token` to the session in the `jwt` and [`session` callbacks](https://next-auth.js.org/configuration/callbacks#session-callback) when the access token needs to be exposed (provided it's being exposed in a server component).

balazsorban44:
Thanks, but for the App Router, we only plan to support `getServerSession` and later on a unified `auth()` method. See #7443 for more details. Upcoming docs at https://auth-docs-git-feat-nextjs-auth-authjs.vercel.app/guides/upgrade-to-v5#authenticating-server-side

Using `getServerSession` is the recommended replacement, but if for some reason you want to keep using `getToken`, you can achieve this in user-land as such:

```tsx
import { cookies, headers } from "next/headers"

export default function Page() {
    cookies: cookies(), headers: headers()
  })
}
```

jaredgalanis:
> > https://auth-docs-git-feat-nextjs-auth-authjs.vercel.app/guides/upgrade-to-v5#authenticating-server-side
> 
> @balazsorban44, even on the main branch it looks like passing a request object to `getToken` is not optional. Since RSC's won't have a request object, for those who need to get a token to authenticate to an external API, what is the solution?

You can achieve this by using `getToken` in nextjs middleware and setting that as a header which you can read from an RSC.

vipuljbhikadiya:
> Thanks, but for the App Router, we only plan to support `getServerSession` and later on a unified `auth()` method. See #7443 for more details. Upcoming docs at https://auth-docs-git-feat-nextjs-auth-authjs.vercel.app/guides/upgrade-to-v5#authenticating-server-side
> 
> Using `getServerSession` is the recommended replacement, but if for some reason you want to keep using `getToken`, you can achieve this in user-land as such:
> }
> ```

@balazsorban44 if we are passing token to session side, is that not any security issue ? i mean expose access_token to session callback.

PR #8214, opened 2023-08-02
Update email.md

tmackness:
Better documentation for http email provider

### DIFF
--- a/docs/docs/providers/email.md
+++ b/docs/docs/providers/email.md
@@ -32,6 +32,108 @@ You can override any of the options to suit your own use case.
 
 ## Configuration
 
+Nextauth provides you with the ability to send emails via SMTP or HTTP.
+
+### HTTP
+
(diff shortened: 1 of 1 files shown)


ndom91:
Thanks for the contribution! 

However, I think this would be better suited as a separate "Mailergrid" tutorial as it's very specific to that email provider. 

Something like the Email HTTP page here - https://github.com/nextauthjs/next-auth/blob/main/docs/docs/guides/providers/email-http-api.md 

What we could add directly to this email provider page, is some notes about how easy it is to also use http based cloud email providers and link to the existing one and to your new mailergrid one. 

What do you think?

balazsorban44 (changes requested):
Let's just link to the existing guide, if it's not sufficient, we can edit that instead.

balazsorban44 (approved):

PR #8258, opened 2023-08-08
feat(adapters): Drizzle adapter

balazsorban44:
Finishing up @anthonyshew's PR #7165

Introducing a `drizzle-orm` adapter!

With `drizzle`'s increasing popularity, it's no surprise that [there's a discussion](https://github.com/nextauthjs/next-auth/discussions/7005) for an adapter. This PR adds it.

# Notes
- There is currently an open issue with Drizzle's recent versions that breaks monorepo dependencies. Please use 0.24.x until this issue is resolved: https://github.com/drizzle-team/drizzle-orm/issues/283
- There is currently an open issue with Drizzle truncating millisecond precision. I've added an extra option to the `runBasicTests` utility to be able to provide some custom fixtures to workaround this. https://github.com/drizzle-team/drizzle-orm/pull/668

Closes #7165


ndom91 (approved):
Haven't tested it locally, but in general lgtm. 

Some general feedback though, do we really need separate method implementations for each underlying db type (MySQL, postgres, etc.)? 

Also these inline docs for TS hinting are getting a bit long 😅

stunaz:
@balazsorban44  why all functions from SQLiteDrizzleAdapter are not async/await like the pg/mysql ?
I a ran to a bug without that.

PR #8344, opened 2023-08-17
feat(drizzle-adapter): custom database's tables naming strategy

skyf0l:
> _NOTE_:
>

## ☕️ Reasoning

I'm migrating my database from Prisma to Drizzle, but I got a problem using drizzle adapter cause table names are different (camelCase vs PascalCase). So, I added an option to force the naming strategy :rocket:

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources

### DIFF
--- a/packages/adapter-drizzle/src/index.ts
+++ b/packages/adapter-drizzle/src/index.ts
@@ -16,16 +16,51 @@
  * @module @auth/drizzle-adapter
  */
 
-import { MySqlDatabase, MySqlTableFn } from "drizzle-orm/mysql-core"
-import { PgDatabase, PgTableFn } from "drizzle-orm/pg-core"
-import { BaseSQLiteDatabase, SQLiteTableFn } from "drizzle-orm/sqlite-core"
+import { MySqlDatabase } from "drizzle-orm/mysql-core"
+import { PgDatabase } from "drizzle-orm/pg-core"
+import { BaseSQLiteDatabase } from "drizzle-orm/sqlite-core"
 import { mySqlDrizzleAdapter } from "./lib/mysql.js"
 import { pgDrizzleAdapter } from "./lib/pg.js"
 import { SQLiteDrizzleAdapter } from "./lib/sqlite.js"
-import { SqlFlavorOptions, TableFn } from "./lib/utils.js"
+import type {
+  AnyMySqlDatabase,
+  AnyPgDatabase,
+  AnySQLiteDatabase,
+  SqlFlavorOptions,
+  TableFn,
+} from "./lib/utils.js"
 import { is } from "drizzle-orm"
 
 import type { Adapter } from "@auth/core/adapters"
+
+/** Configure the Drizzle Adapter. */
+export interface DrizzleAdapterConfig<SqlFlavor> {
+  /**
+   * The function used to create database tables.
+   *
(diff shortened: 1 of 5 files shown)


skyf0l:
Up?

GLips:
I'm in a similar boat, migrating from Prisma to Drizzle and about to copy my own DrizzleAdapter out to make these modifications.

nhedger:
Hey @skyf0l, I'm wondering whether giving the option to use either one of `PascalCase`, `camelCase` or `snake_case` would be a better idea. I'd be willing to help implement this. What is your take on this?

skyf0l:
> Hey @skyf0l, I'm wondering whether giving the option to use either one of `PascalCase`, `camelCase` or `snake_case` would be a better idea. I'd be willing to help implement this. What is your take on this?

Hey @nhedger, first it was to keep consistency with the Prima adapter, then to avoid renaming tables in my production database just because of an ORM switch in my main application.

As the Drizzle adapter is new, I think it's a good idea to start keeping consistency between them all.

Finally, even though some database table names are case insensitive, not all are (e.g. PlanetScale).

In any case, the main goal is consistency :)

nhedger:
I understand that; my reasoning was that some users may prefer to use a different case. For example, all my tables use `snake_case`, so for the sake of consistency, I'd like these tables to use `snake_case` too, hence my proposition to make it configurable.

If you personally don't need this, I'd fully understand if you did not want to invest time in it and can submit a PR of my own.

skyf0l:
Oh, I read it sideways, yes indeed your suggestion is definitely the best!
I should have 5 minutes in the week to update this quickly!

nhedger:
Glad to hear; let me know if you need any assistance.

felixnorden:
+1 on what you both have said @nhedger @skyf0l , we have a project where the tables are in `camelCase` and so are the fields in JS-land, but in SQL they are `snake_case`, so starting to work towards allowing the user to configure the naming scheme (as [the docs state](https://authjs.dev/reference/adapters#rdbms-naming-convention) other adapters do) would be grand.
Let me know if you need any help as well!

danclaroni (approved):

ormarek (commented):
Im not in actual power here to accept or reject those changes you've made, but wanted to make a small suggestion

You could create a small naming table like:
```
const names = {
  snake_case: {
    userId: "user_id",
    ...
  },
  camelCase: {
    userId: "userId",
    ...
  },
  pascalCase: {
    userId: "UserId",
    ...
  }
}
```

and then instead of doing tedious trenary everywhere (which is also typo prone) you could use `names[namingStrategy].<name>` e.g. `names[namingStrategy].userId`

ndom91:
In general this looks good to me, tests are still passing too. @balazsorban44 @ThangHuuVu we had something similar with another db adapter in the past, but I can't remember the details. I think technically this is good to go, yall okay with this generally?

ndom91:
Btw see this PR that adds support for custom tables and seems to ahve a lot of overlap: https://github.com/nextauthjs/next-auth/pull/8561

nahtnam:
I have no authority here but just want to chime in that I prefer the other PR over this one. I want to do snake_case but pluralize and I dont think this PR would support that. I assume future developers might want totally different names or a prefix for the table names. It would be better instead to just allow us to pass the schemas as options

PR #8364, opened 2023-08-20
feat(adapter-neo4j): use isolated session per read/write transaction

GregCKrause:
## ☕️ Reasoning

This change utilizes a session-per-transaction approach with the Neo4j adapter, rather than creating and maintaining a single session.

As per [Neo4j docs](https://neo4j.com/docs/api/javascript-driver/current/class/lib6/session.js~Session.html), "multiple sessions are required to execute parallel queries".

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

May affect https://github.com/nextauthjs/next-auth/issues/5849


packages/adapter-neo4j/package.json
```diff
@@ -41,19 +42,19 @@
     "build": "tsc"
   },
   "peerDependencies": {
-    "neo4j-driver": "^4.0.0 || ^5.7.0"
+    "neo4j-driver": "^4.0.0 || ^5.11.0"
```
balazsorban44, line 45: Is this a necessary bump for this PR to be valid?

pnpm-lock.yaml
balazsorban44, line 1: This file seems to include a huge number of changes. Are you sure that you have enabled `corepack` to make sure you use the same pnpm version as per our contributing docs? https://github.com/nextauthjs/.github/blob/main/CONTRIBUTING.md

balazsorban44 (changes requested):
Thanks! I am trying to understand if this PR should be considered breaking. Can you verify?

GregCKrause:
> Thanks! I am trying to understand if this PR should be considered breaking. Can you verify?

Hey, @balazsorban44! You raised some good concerns here.I think the `neo4j-driver` 5.7.0 -> 5.11.0 bump can be deferred

packages/adapter-neo4j/src/index.ts
```diff
@@ -113,20 +111,20 @@ export interface Neo4jOptions extends Session {}
...
- * ON (v:VerificationToken) ASSERT (v.identifier, v.token) IS NODE KEY;
+ * FOR (v:VerificationToken) REQUIRE (v.identifier, v.token) IS NODE KEY;
  *
  * CREATE INDEX verification_token_composite_index IF NOT EXISTS
  * FOR (v:VerificationToken) ON (v.identifier, v.token);
  * ```
  */
-export function Neo4jAdapter(session: Session): Adapter {
-  const { read, write } = client(session)
+export function Neo4jAdapter(driver: Driver): Adapter {
```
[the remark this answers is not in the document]
GregCKrause: Are you referring to v4 of next-auth or the neo4j-driver package?

This would be a breaking change for the Neo4jAdapter, as sessions would no longer be accepted arguments.

pnpm-lock.yaml
GregCKrause: Thank you for catching this. It appears to have been related to me using Node version 20. After switching to Node version 18 (as per contribution docs), `pnpm install` worked smoothly. Newest commit has 0 changes to lock file.

w7tf:
@GregCKrause, thank you very much for this PR. Sad to see that your hard work didn't make it into the package. However I wanted to let you know that I was able to fix the errors mentioned in #5849 on my side with the edits in this PR. Much appreciated!

PR #8379, opened 2023-08-21
fix(adapters): drizzle sqlite `getUserByAccount` - resolve results before calling property

kyujulian:
## ☕️ Reasoning
I was facing the issue that next-auth wasn't able to login again after a signout when using drizzle-adapter connected to a turso.

After investigating a bit, turning on debug mode, these function calls stood out:
```
[next-auth][debug][adapter_getUserByAccount] { args: [ { providerAccountId: '123456789', provider: 'github' } ] }
[next-auth][debug][adapter_getUserByAccount] { args: [ { providerAccountId: '123456789', provider: 'github' } ] }
[next-auth][debug][adapter_getUserByEmail] { args: [ 'example@email.com' ] }
```

So I notice that the adapter called them both, and compared them against each other.
Hm... so one of them must be malfunctioning then, let me pipe those inputs in an isolate call:
```
DrizzleAdapter(db).getUserByAccount ({ providerAccountId: '123456789', provider: 'github' } )
```
In that call, even though the credentials were identical to the database, the function always returned null.
I tried with a couple different users and the same thing. So I dove into the mainstream repo to see the function.

Reading it , the problem was made clear:

```
getUserByAccount(account) {
      const results = client
        .select()
        .from(accounts)
        .leftJoin(users, eq(users.id, accounts.userId))
        .where(
          and(
            eq(accounts.provider, account.provider),
            eq(accounts.providerAccountId, account.providerAccountId)
          )
        )
        .get()

      return results?.user ?? null 
  ```
in the return statement, results is a promise that's still pending, so when trying to access the `.user` property, it would always return a falsy value, causing the function to always return `null`.
so I changed those last couple lines to 
```
      if (!result) {
        return null
      }
      return Promise.resolve(result).then((result) => result.user)
```
And voilá, that solved it !

## 🧢 Checklist

- [x ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Getting  [OAuthAccountNotLinked](https://github.com/nextauthjs/next-auth/issues/8377) 
when trying to login after signout. Even when only one provider is configured.
Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources
- [Drizzle docs](https://orm.drizzle.team)

### CODE REVIEW

File: packages/adapter-drizzle/src/lib/sqlite.ts [hunk 1]
```diff
@@ -159,7 +159,10 @@ export function SQLiteDrizzleAdapter(
         )
         .get()
 
-      return results?.user ?? null
+      if (!results) {
```
pipisso, line 162: This would not work would it?
`results` is a Promise and aren't Promises always truthy?

kyujulian: Oh, I see... In this case, to get the same behavior as the original I should do something like:
```
return Promise.resolve(results).then((results) => results.user ?? null )
```
correct? 

I still feel off about something though...
 is there a reason why the adapter implementations for postgresql and mysql are async, but sqlite isn't?

pipisso: @kyujulian The problem is Promise.resolve is returning a Promise.
So i would just use async await it makes it much simpler.

kyujulian: Hmm got it, that's actually how I've implemented for a custom adapter in another project.

So do you think it might be a useful / valuable PR? to rewrite / refactor the sqlite implementation to be async, like the other ones (mysql and postgres)

FMGordillo:
Bump :cry:

FMGordillo (approved):
Just tested in my end, and it works as expected

image

image

adidoes:
any reason this wasn't merged yet?

joebailey26:
I'm hitting this issue too. using Cloudflare D1 via Drizzle Adapter.

ndom91:
@kyujulian is this ready to merge from your point of view?

socketopp:
> @kyujulian is this ready to merge from your point of view?

I am using drizzle-adapter with D1 on a Sveltekit app and I get `OAuthAccountNotLinked` when I am signing in for the first time with Github. After the error I can see the user being created in the users table. 

Is this connected to this issue?

kyujulian:
@ndom1 I believe so, haven't had any issues after changing that. But bear in mind that I changed it for a simple crud application, that I was building for exploring purposes, so I wouldn't say it's battle tested.

but unless you have concerns or red flags about it, I'd say it is ready :)

ndom91 (approved):

ndom91:
@kyujulian can you please take a look at the merge conflicts? Otherwise this is basically good to go in my view!

kyujulian:
@ndom91 Done  ;)

stunaz:
@kyujulian I had the same issue a Lon time ago... tried to fix it like this - [https://github.com/nextauthjs/next-auth/pull/8636/files](https://github.com/nextauthjs/next-auth/pull/8636/files), so instead of using `async\await`, you use `Promise.resolve`. But you are fixing only the `getUserByAccount` functions, can't we just modify all the others like I tried in my PR, since they are all promises?

Issue #3577, opened 2022-01-06
Session callback params types are misleading


baptisteArno:
### Question 💬

`session` param in session callback is typed as `Session` but in reality, it can also be undefined.

### How to reproduce ☕️

Add a session callback:
```ts
session: async ({ session, token, user }) => {
      token?.user ? (session.user = token.user as User) : (session.user = user)
      if (token.providerType) session.providerType = token.providerType
      return session
    }
```

Error: `Cannot read property 'providerType' of undefined`

### Contributing 🙌🏽

Yes, I am willing to help answer this question in a PR

balazsorban44:
`session` should always be defined.

When using `strategy: "jwt"`, you will have `token`, when using `strategy: "database"`, you will get `user`.

See the docs: https://next-auth.js.org/configuration/callbacks#session-callback

baptisteArno:
Sorry for the misleading issue content, I wanted to talk about `token` and `user` types. They can both be `undefined` but the types don't say they can so it results in potential runtime types errors

balazsorban44:
It is slightly complicated since the existence of those properties is dependent on the `adapter` and `strategy` values in this interface:

https://github.com/nextauthjs/next-auth/blob/49e4af17e2a5436111c322d480de54864b3737fe/src/core/types.ts#L17

I haven't found a nice way of propagating that to the `session` callback, but feel free to have a look at it. I assume it would involve some generics, but they should be totally optional.

PR #8404, opened 2023-08-24
refactor: correct `callbacks` type

yyueniao:
> _NOTE_:
>

## ☕️ Reasoning

Some of the methods in `callbacks` have a wrong params type. There are some of them should be optional.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: https://github.com/nextauthjs/next-auth/discussions/3951 (previous https://github.com/nextauthjs/next-auth/issues/3577)

## 📌 Resources

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -178,6 +178,101 @@ export interface Profile {
   [claim: string]: unknown
 }
 
+export interface SignInCallbackParams<P = Profile, A = Account> {
+  user: User | AdapterUser
+  account: A | null
+  /**
+   * If OAuth provider is used, it contains the full
+   * OAuth profile returned by your provider.
+   */
+  profile?: P
+  /**
+   * If Email provider is used, on the first call, it contains a
+   * `verificationRequest: true` property to indicate it is being triggered in the verification request flow.
+   * When the callback is invoked after a user has clicked on a sign in link,
+   * this property will not be present. You can check for the `verificationRequest` property
+   * to avoid sending emails to addresses or domains on a blocklist or to only explicitly generate them
+   * for email address in an allow list.
+   */
+  email?: {
+    verificationRequest?: boolean
+  }
+  /** If Credentials provider is used, it contains the user credentials */
+  credentials?: Record<string, CredentialInput>
+}
+
+export interface RedirectCallbackParams {
+  /** URL provided as callback URL by the client */
+  url: string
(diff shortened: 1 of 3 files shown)


packages/core/src/types.ts
```diff
@@ -178,6 +178,101 @@ export interface Profile {
...
+   * - [Credentials Provider](https://authjs.dev/reference/core/providers_credentials)
+   * - [User database model](https://authjs.dev/reference/adapters#user)
+   */
+  user?: User | AdapterUser
+  /**
+   * Contains information about the provider that was used to sign in.
+   * Also includes {@link TokenSet}
+   * @note available when `trigger` is `"signIn"` or `"signUp"`
+   */
+  account?: A | null
```
Juneezee, line 250: One drawback of making `user` and `account` optional is that we will have to check them for `undefined` even when `trigger` is `signIn` or `signUp`. This is a breaking change for existing TypeScript users.

Before this PR:

```typescript
callbacks: {
  async jwt({ token, user, account, trigger }) {
    if (trigger === "signIn") {
      console.log(user); // user is not User | AdapterUser
      console.log(account); // account is A | null
    }
  }
}
```

After this PR:

```typescript
callbacks: {
  async jwt({ token, user, account, trigger }) {
    if (trigger === "signIn") {
      console.log(user); // user is User | AdapterUser | undefined
      console.log(account); // account is A | null | undefined
    }
  }
}
```

packages/core/src/types.ts
```diff
@@ -379,7 +385,7 @@ export interface EventCallbacks {
    * - `token`: The JWT for this session.
    * - `session`: The session object from your adapter.
    */
-  session: (message: { session: Session; token: JWT }) => Awaitable<void>
+  session: (message: { session: Session; token?: JWT }) => Awaitable<void>
```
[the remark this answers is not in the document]
yyueniao: When I search through this project, there are only two references in this project, and both of them provide `session`. So at least at this point `session` is not optional.

packages/core/src/types.ts
yyueniao: @Juneezee I use type union to help typescript developers [narrowing down](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions) the types. Unfortunately, this makes jsdocs not DRY anymore. Do you have any advice?

PR #8477, opened 2023-09-02
chore: update NextAuth.js example to use App Router

balazsorban44:
Initially planned to start this in #7443 but since people think v4 does not support App Router (it's been supported since the dawn of time) , I decided to update now to show how it works.

Also adding most of our built-in providers to be able to test them when deployed at https://next-auth-example.vercel.app

> **NOTE:** When merged, these need to be tested as I might have missed adding some environment variables to the deployment.

Ref: https://twitter.com/nextauthjs/status/1689599931810549760


rafalkrupinski:
how to use without `pages/_app`? couldn't find any example

PR #8478, opened 2023-09-03
feat: add `isNewUser` arg on the signIn callback

### LINKED ISSUE #8272: Allow to block authorization depending on whether it is a new user or not
## Problem

Currently, there is no easy and direct way to intervene in the authorization flow based on whether the user is new (signup) or not (login).

A [user has proposed](https://github.com/nextauthjs/next-auth/discussions/1578#discussioncomment-5355744) to verify some property that is not in both situations, but I don't know how reliable it is. Does it work in all circumstances? For all providers? Also, it's not self-evident or something that follows from the documentation. A user a [few comments above](https://github.com/nextauthjs/next-auth/discussions/1578#discussioncomment-530604) in the same thread proposed a similar alternative, which ended up not working. As you can see, it is a workaround that is prone to errors.

## Possible solutions:

1. Bring the [isNewUser property](https://next-auth.js.org/configuration/events#signin) also to the signIn callback ([as requested here](https://github.com/nextauthjs/next-auth/discussions/1578#discussioncomment-5672967)).
2. Allow [events](https://next-auth.js.org/configuration/events) to block the authorization flow in some way (return boolean, throw error, etc.).
3. Extend next-auth to have signup and login as different processes. This would likely involve, but not necessarily be limited to:
   3.1. Add a new Signup callback ([as requested here](https://github.com/nextauthjs/next-auth/discussions/4124)).
   3.2. Add a new Signup function ([as requested here](https://github.com/nextauthjs/next-auth/discussions/945)).

The third point is probably the hardest to implement and the one that would add the most complexity. But I don't know to what extent it would be wrong since users and programmers actually think of creation and login as different concepts.

## Background - Related References

The large number of posts, comments, and upvotes in the login vs signup topic indicate that this is something that users are frequently stumbling upon. Some examples:

- https://github.com/nextauthjs/next-auth/discussions/1578
- https://github.com/nextauthjs/next-auth/issues/736
- https://stackoverflow.com/questions/73369742/how-to-differentiate-nextauth-user-signin-and-signup
- https://github.com/nextauthjs/next-auth/discussions/4124
- https://github.com/nextauthjs/next-auth/discussions/945
- https://github.com/nextauthjs/next-auth/discussions/7164
- https://github.com/nextauthjs/next-auth/issues/1356
- https://stackoverflow.com/questions/70739015/next-auth-check-if-user-signs-in-for-the-first-time
- https://stackoverflow.com/questions/71397331/signup-from-next-auth-react-doesnt-exist

### How to reproduce ☕️

N/A

### Contributing 🙌🏽

kahveciderin:
This PR implements the isNewUser field on the signIn callback, to allow for the rejection of a potential sign up.

## ☕️ Reasoning

Sometimes we might want to reject the signIn event if the user doesn't already have an account.

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes #8272


kahveciderin:
As far as I understand, the "Credentials" provider is intended to support use cases where one has an existing system one needs to authenticate users against, so a user cannot be a new user when using the Credentials provider., thus I've now added `isNewUser: false` to the signIn callback when signing in with the credentials provider.

kahveciderin:
I have exported the "input validation" lines to an external helper function, because the same thing with the `callbackHandler` was being validated in the new `checkIfUserIsNew` function. Also, I simplified the `checkIfUserIsNew` function's logic by first checking the email no matter what provider is used, and it checks if the account exists only after establishing that the email doesn't exist and the user is using oauth.

kahveciderin:
I think I forgot to stage my last changes, will be pushing them now.

kahveciderin:
This PR should now be ready to merge, would be happy to get some reviews.

april83c:
What's going on with this? Will it be merged?

bilalmohib:
Please merge this work

ehenoma:
push

siddharthd0 (approved):

siddharthd0:
merge when

mcnaveen (approved):
LGTM 🫡

garshythoel:
Another bump here! Thanks

PR #8482, opened 2023-09-04
Add more detail to Credentials provider.

anthonyshew:
After speaking with the Auth.js maintainers about a change to the Credentials provider documentation, I'm opening this PR to introduce additional information about:
- *Why* the Credentials provider is discouraged
- How to interact with a database using the Credentials provider

While Auth.js does prefer to encourage use of other providers, the username-password model is (despite its flaws) still widely used today. Over time, I know I've personally seen a handful of users ask about the discouragement of the Credentials provider and express discontent with the way it is documented.

With this PR, I'm hoping to (slightly) close the gap on the community's understanding of the Credentials provider. While we can still (in my opinion, properly) discourage use of the username-password model, we can still bring in the understanding that some folks are going to want to opt to take on those risks, anyway.

## Please note

The psuedo-code that I've added in for interacting with a database is most likely wrong at the time of this PRs authorship. I've never used the Credentials provider myself but am aware that the maintainers have seen successful implementations that communicate with a database. Hoping I can get some maintainer help to document that code correctly!

### DIFF
--- a/docs/docs/guides/providers/credentials-provider.md
+++ b/docs/docs/guides/providers/credentials-provider.md
@@ -7,11 +7,7 @@ The Credentials provider allows you to handle signing in with arbitrary credenti
 
 It is intended to support use cases where you have an existing system you need to authenticate users against.
 
-It comes with the constraint that users authenticated in this manner are not persisted in the database, and consequently that the Credentials provider can only be used if JSON Web Tokens are enabled for sessions.
-
-:::warning
-The functionality provided for credentials-based authentication is intentionally limited to discourage the use of passwords due to the inherent security risks associated with them and the additional complexity associated with supporting usernames and passwords.
-:::
+It comes with the constraint that users authenticated in this manner are not persisted in the database by default, and consequently, that the Credentials provider can only be used if JSON Web Tokens are enabled for sessions.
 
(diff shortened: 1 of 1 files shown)


docs/docs/guides/providers/credentials-provider.md
```diff
@@ -82,15 +92,37 @@ callbacks: {
 }
 ```
 
+### Using a database
+
+You can also use the `jwt()` callback to interact with your database to regain some of the functionality from [more powerful providers](reference/core/providers):
+
+```js
```
anthonyshew, line 99: Would love some help on this pseudo-code!

balazsorban44: All this could live in the `authorize()` callback.

anthonyshew: Ah, right. Switched it up now. Does this make more sense?

balazsorban44 (approved):

docs/docs/guides/providers/credentials-provider.md
arminhupka: We can assume that updating jwt token in callbacks has been fixed?

PR #8581, opened 2023-09-13
Avoid parseDataSourceConfig on each call

gkorland:
Improve performance by avoiding call to parseDataSourceConfig before checking if _dataSource was already initialized, and call for `Object.values()` when not needed.


mayank1513 (approved):
👌

ThangHuuVu (approved):

PR #8599, opened 2023-09-15
fix: Build account object with tokens spread as lowest priority

etrepum:
## ☕️ Reasoning

tokens comes from the provider and could have arbitrary keys, but type, provider, and providerAccountId have reserved meanings to NextAuth that must take priority for correct operation.

I discovered this because I had a tokens endpoint that had a type key in the response, which caused next-auth to fail because it was expecting type to be "oauth". I managed to work around it for now by mutating tokens in the profile callback.

I don't believe that this is a security issue but there's always some chance when untrusted input ends up in the wrong place.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

I didn't find any associated issues and implementing the very small PR was easier than reporting it separately.

### DIFF
--- a/packages/next-auth/src/core/lib/oauth/callback.ts
+++ b/packages/next-auth/src/core/lib/oauth/callback.ts
@@ -160,10 +160,10 @@ async function getProfile({
     return {
       profile,
       account: {
+        ...tokens,
         provider: provider.id,
         type: provider.type,
         providerAccountId: profile.id.toString(),
-        ...tokens,
       },
       OAuthProfile,
     }
--- a/packages/next-auth/src/core/lib/oauth/callback.ts
+++ b/packages/next-auth/src/core/lib/oauth/callback.ts
@@ -160,10 +160,10 @@ async function getProfile({
     return {
       profile,
       account: {
-        ...tokens,
         provider: provider.id,
         type: provider.type,
         providerAccountId: profile.id.toString(),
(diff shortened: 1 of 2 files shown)


mayank1513 (approved):
Good catch!

packages/next-auth/src/core/lib/oauth/callback.ts
ThangHuuVu, line 1: Please revert change in this file, as it is not needed.

ThangHuuVu (changes requested):

packages/next-auth/src/core/lib/oauth/callback.ts
etrepum: It’s the same bug somewhere else, to be discovered later, is there a reason not to?

etrepum: I've reverted the change to that file.

ThangHuuVu: hi @etrepum sorry for not being clear, I meant that the code in `packages/next-auth` on the main branch is not actually published anywhere. we maintain the package in the `v4` branch. For `v5`,  we are working on it at #7443 . Hope it's clear now!

balazsorban44 (approved):

PR #8605, opened 2023-09-15
style: use more idiomatic `@see` annotation

hirehamir:


hamirmahal:
## ☕️ Reasoning


Using
```ts
  /**
   * @see https://next-auth.js.org/configuration/providers/oauth
   */
```

as opposed to

```ts
// https://next-auth.js.org/configuration/providers/oauth
```

gets people hover documentation in VSCode.

image

balazsorban44:
thanks! this was actually unnecessary, since the whole object is typed.

hamirmahal:
You're welcome!

PR #8714, opened 2023-09-25
Update next-auth/react with "use client"

ariannargesi:
Add "use client" directive for better compatibility with next 13


balazsorban44:
Thanks, this would currently break apps using some of the methods in `next-auth/react` server-side, but this is already added in the upcoming v5 version https://github.com/nextauthjs/next-auth/blob/5749b095a284bff912eb3a657e0c6da568c3e9e1/packages/next-auth/src/react.tsx#L8

PR #8767, opened 2023-09-30
fix typo in the v4 docs

AhmedBaset:
[Context](https://next-auth.js.org/getting-started/client#example-2)

The trigger `update` has received `{name: string}`
```tsx
{/* Update the value by sending it to the backend. */}
<button onClick={() => update({ name: "John Doe" })}>
  Edit name
</button>
```

So, the session should contain `name`
```diff
if (trigger === "update" && session?.name) {
   // Note, that `session` can be any arbitrary object, remember to validate it!
-  token.name = session
+  token.name = session.name
}
```


balazsorban44 (approved):
Thanks!

Issue #7638, opened 2023-05-24
Crendtials provider with redirect:false - next-auth/react signIn returns status:200 ok:true with failed authentication.


OdifYltsaeb:
### Provider type

Credentials

### Environment

Using dockerised environment FROM node:18-alpine so
- node 18 latest,
- next 13.4.1
- next-auth 4.22.1
- react 18.2

### Reproduction URL

https://github.com/OdifYltsaeb/next-auth-bug-report

### Describe the issue

I'm following the most basic setup described here https://medium.com/@tom555my/strapi-next-js-email-password-authentication-a8207f72b446 and the next-auth/react signIn (see `console.log('stuff in handlesubmit', stuff);` in the code below) resolves into `{ error: "CredentialsSignin", status: 200, ok: true, url: null }` when authentication fails due to incorrect email/password. The strapi request returns with error 400, and thus the error message in the object that singIn resolves to, but the status is 200, while it should probably be something else.

// app/api/auth/[...nextauth]/route.js

```
import NextAuth from 'next-auth';
import { authOptions } from '../../../../lib/auth';

const handler = NextAuth(authOptions);
export { handler as GET, handler as POST };
```

// lib/auth.js

```
import CredentialsProvider from "next-auth/providers/credentials";
import axios from "../utils/axios";

const signIn = async function({ email, password }) {
    const res = await axios.post(`${process.env.STRAPI_URL}/api/auth/local`, {
        identifier: email,
        password
    });
    return res;
};

const authOptions = {
    session: {
        strategy: "jwt"
    },
    providers: [
        CredentialsProvider({
            name: "Sign in with Email",
            credentials: {
                email: { label: "Email", type: "text" },
                password: { label: "Password", type: "password" }
            },
            async authorize(credentials, req) {
                if (credentials == null) return null;
                try {
                    const { user, jwt } = await signIn({
                        email: credentials.email,
                        password: credentials.password
                    });
                    return { ...user, jwt };
                } catch (error) {
                    console.log('FAIL!', Object.keys(error));
                    // Sign In Fail
                    return null;
                }
            }
        })
    ],
    callbacks: {
        session: async ({ session, token }) => {
            session.id = token.id;
            session.jwt = token.jwt;
            return Promise.resolve(session);
        },
        jwt: async ({ token, user }) => {
            const isSignIn = user ? true : false;
            if (isSignIn) {
                token.id = user.id;
                token.jwt = user.jwt;
            }
            return Promise.resolve(token);
[...]

PR #8775, opened 2023-10-02
fix(next): returns correct status for signing in with `redirect: false` for route handler

### LINKED ISSUE #7725: CredentialsProvider authorize function returns status 200 and response ok when Error is thrown
### Provider type

Credentials

### Environment

System:
    OS: Windows 10 10.0.22621
    CPU: (12) x64 Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
    Memory: 6.52 GB / 15.75 GB
  Binaries:
    Node: 18.16.0 - C:\Program Files\nodejs\node.EXE
    npm: 9.5.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.1702.0), Chromium (113.0.1774.50)
    Internet Explorer: 11.0.22621.1

### Reproduction URL

no-repo-easy-to-reproduce

### Describe the issue

When throwing Error inside authorize method the response is 200 (which presents success).

### How to reproduce

app\api\auth\\[...nextauth]\route.ts file
```
import NextAuth from "next-auth/next";
import CredentialsProvider from "next-auth/providers/credentials";

const providers = [
  CredentialsProvider({
    credentials: {
      email: { label: "Email", type: "text" },
    },
      // Throw error to test the response
      throw new Error("This is error message.");
      
      // Do login stuff and return user object...
      return null;
    },
  }),
];

export { handler as GET, handler as POST };
```

This is reponse:
```
{
    "error": "This is error message.",
    "status": 200,
    "ok": true,
    "url": null
}
```

### Expected behavior

To get response with error status code. Even though it contains error param, status code should be the one which represents an error, instead of success status code.

### LINKED ISSUE #8340: SignIn function always returns "ok", even if the authorization returns null
### Provider type

Credentials

### Environment

  System:
    OS: Linux 6.4 Fedora Linux 38 (Workstation Edition)
    CPU: (8) x64 Intel(R) Core(TM) i5-10300H CPU @ 2.50GHz
    Memory: 9.47 GB / 15.40 GB
    Container: Yes
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.18.1 - /usr/bin/node
    Yarn: 1.22.19 - /usr/bin/yarn
    npm: 8.19.2 - /usr/bin/npm
    pnpm: 8.6.12 - ~/.local/share/pnpm/pnpm

### Reproduction URL

https://github.com/elpablazo/checador-front/

### Describe the issue

If you use @auth/sveltekit and try to use the `signIn() ` function with the credentials provider and set redirect: false so that you can handle the error on the client sade, you always get a response with a status code of 200 and `ok: true`. Doesn't matter if the actual authorization process has a successfull response from the backend or not. This only happens when redirect: false, since when it's set on true it will lead you to the error message. However, the error cannot be customized either, as it states [here](https://authjs.dev/guides/providers/credentials#example---username--password)

### How to reproduce

Example configuration: 

`await signIn('credentials', {
	redirect: false,
	...values
});`

On the authorizize function:

`async authorize({ email, password }) {
    throw new Error("This should be a custom error")
}`

(or even if you just return null)

However, when you try to console.log the signin() response, you get:

`ok: true
​
redirected: false
​
status: 200
​
statusText: "OK"
​
type: "basic"`

My local terminal however, prints this:

`[auth][error][CallbackRouteError]: Read more at https://errors.authjs.dev#callbackrouteerror
[auth][cause]: Error: This should be a custom error.
[auth][details]: {
  "provider": "credentials"
}
`

### Expected behavior

It should return the same error that i threw on Throw new error ("message") or, at least, return an `ok: false`

ThangHuuVu:
> _NOTE_:
>

## ☕️ Reasoning

When signing in with `next-auth/react` `signIn()` function with `redirect: false`, the status is not correctly set in the response. There are two cases that this bug will return 200 OK instead of the expected status:
1. When signing in with the Credential provider and the `authorize` function returns false or throws an error. Expected status: 401
2. When the `signIn` callback returns false. Expected status: 403

Basically, the bug appears every time we return `status` and `redirect` at the same time for one `return` statement in `packages/next-auth/src/core/routes/callback.ts`.

Note that this bug only affects Route Handler. For API Route, we calls `res: NextApiResponse` instead of returning a new `Response` object.

## 🧢 Checklist

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: #7638, fixes #7725, fixes #8340

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/core/lib/cookie.ts
+++ b/packages/next-auth/src/core/lib/cookie.ts
@@ -114,7 +114,7 @@ export function defaultCookies(useSecureCookies: boolean): CookiesOptions {
         path: "/",
         secure: useSecureCookies,
       },
-    }
+    },
   }
 }
 
@@ -161,22 +161,22 @@ export class SessionStore {
     }
   }
 
- /**
+  /**
    * The JWT Session or database Session ID
    * constructed from the cookie chunks.
    */
- get value() {
-  // Sort the chunks by their keys before joining
-  const sortedKeys = Object.keys(this.#chunks).sort((a, b) => {
-    const aSuffix = parseInt(a.split(".").pop() || "0")
-    const bSuffix = parseInt(b.split(".").pop() || "0")
+  get value() {
+    // Sort the chunks by their keys before joining
+    const sortedKeys = Object.keys(this.#chunks).sort((a, b) => {
+      const aSuffix = parseInt(a.split(".").pop() ?? "0")
+      const bSuffix = parseInt(b.split(".").pop() ?? "0")
 
-    return aSuffix - bSuffix
-  });
+      return aSuffix - bSuffix
+    })
 
-  // Use the sorted keys to join the chunks in the correct order
-  return sortedKeys.map(key => this.#chunks[key]).join("")
-}
+    // Use the sorted keys to join the chunks in the correct order
+    return sortedKeys.map((key) => this.#chunks[key]).join("")
+  }
 
   /** Given a cookie, return a list of cookies, chunked to fit the allowed cookie size. */
   #chunk(cookie: Cookie): Cookie[] {
--- a/packages/next-auth/src/next/index.ts
+++ b/packages/next-auth/src/next/index.ts
@@ -101,6 +101,7 @@ async function NextAuthRouteHandler(
     response.headers.delete("Location")
     response.headers.set("Content-Type", "application/json")
     return new Response(JSON.stringify({ url: redirect }), {
+      status: internalResponse.status,
       headers: response.headers,
     })
   }


balazsorban44 (approved):

umair-mirza:
@balazsorban44 Really apologize for my tone in #7725 . I can't really begin to appreciate the work you have done for the community. Hundreds of thousands of developers are powering their apps with this awesome package. I never intended to disregard or disrespect you. Thank you so much for all the hard work you are putting into it.

balazsorban44:
@umair-mirza probably wasn't intentional on your side, no worries. What I tried to clarify in https://github.com/nextauthjs/next-auth/issues/7725#issuecomment-1742903521 is how this is a very widespread issue in OSS, sometimes it's hard to see the people on the other side.

Appreciate your words! 🙏

saraparsa13:
Isn't it fixed for v5?

Issue #8279, opened 2023-08-10
Passkey support (FIDO / WebAuthn)


fabian-michael:
### Description 📓

Hi, I did not find an issue in regards.

Are there any plans on supporting passkey authentication? That would be awesome.

### How to reproduce ☕️

/

### Contributing 🙌🏽

No, I am afraid I cannot help regarding this

balazsorban44:
#6065 started working on it, although it's right now a bit outdated. It is still planned though. Will close this issue for now, you can upvote https://github.com/nextauthjs/next-auth/discussions/3937

PR #6065, opened 2022-12-15
feat: test webauthn platform auth

ndom91:
> _NOTE_:
>
> - It's a good idea to open an issue first to discuss potential changes.
> - Please make sure that you are _NOT_ opening a PR to fix a potential security vulnerability. Instead, please follow the [Security guidelines](../Security.md) to disclose the issue to us confidentially.

## ☕️ Reasoning

Big time WIP

### Note:
- https://medium.com/webauthnworks/introduction-to-webauthn-api-5fd1fb46c285
- https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential#getting_an_existing_instance_of_publickeycredential
- https://simplewebauthn.dev/docs/packages/server

### Screenshot:
image

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Please [scout and link issues](https://github.com/nextauthjs/next-auth/issues) that might be solved by this PR.

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources

- [Security guidelines](../Security.md)
- [Contributing guidelines](../CONTRIBUTING.md)
- [Code of conduct](../CODE_OF_CONDUCT.md)
- [Contributing to Open Source](https://kcd.im/pull-request)

Issue #5663, opened 2022-10-28
FIDO Passkeys (WebAuthn)


DestroyerXyz:
### Description 📓

FIDO recently announced the new [Passkeys](https://fidoalliance.org/passkeys/) authentication standard which would soon be supported cross-platform by Apple, Microsoft, and Google as a more secure replacement for username/password auth.
Dev Docs: https://passkeys.dev

Will there be support for this in NextAuth or is it possible using the Credentials auth option? If the latter, can a guide be added to the docs on how to implement it?

### How to reproduce ☕️

NA

### Contributing 🙌🏽

No, I am afraid I cannot help regarding this

ThangHuuVu:
Indeed, we plan to have Webauthn support in the near future 🙌
cc @ndom91

DestroyerXyz:
Is the work underway or planned?

ndom91:
> Is the work underway or planned?

Just planned as of now 👍

DestroyerXyz:
Gr8, really looking forward to it 😊 👍

KevinPy:
It seems like started #6065

ndom91:
> It seems like started #6065

Yeah I've started experimenting in the above referenced PR. Feel free to comment / chip in there 😊

renchris:
Does the launch of Google Sign-In allowing for passkeys for public users affect the development of this? Using basic services such as YouTube now prompts for passkeys by default but when I checked the Next Auth example with Google Sign-In it only prompts password still.

renchris:
Oh, I tried it again logging out of my Google Account and re-login in and it prompts the passkey

PR #8808, opened 2023-10-06
feat: Passkey / WebAuthn provider (experimental)

Maronato:
### Signing up via built-in signin page
CleanShot 2024-01-23 at 23 29 51

### Signing in via built-in signin page
CleanShot 2024-01-23 at 23 30 19

### Signing in from custom page
CleanShot 2024-01-23 at 23 30 43

### Registering a new Passkey to a logged in user
CleanShot 2024-01-23 at 23 31 02

## ☕️ Reasoning
Passkeys are being adopted very quickly, with all major browsers now supporting it. It makes sense for Auth.js to adopt it as well.

Technically the new provider is called `webauthn`. Passkeys implement a more specialized version of WebAuthn, but the underlying protocol is the same. In this document, I'll use the terms *Passkey* and *WebAuthn* interchangeably.

## Sign in flow
Passkeys require a new login flow and endpoints to work. It requires both server side and client side code. If the user uses the built-in sign in page, the authentication flow follows these steps:
> The flowchart is for the first implementation. It is now a bit outdated, though the important bits remain the same
Untitled-2023-10-06-1506
> Live version of the flowchart [here](https://excalidraw.com/#json=R9qTtgm0jjvdLCzik6bEV,G45gUUk5q-ZZJ7AQ3iOFOw)

### Implementation specs
The entire passkey processing and logic is handled by [SimpleWebAuthn](https://github.com/MasterKale/SimpleWebAuthn). Auth.js' responsibilities here are collecting user data and managing storage of credentials, accounts and sessions.

These are the specs I decided while writing this PR:
- Users can both register and login using a Passkey
- Users can have multiple Passkeys, similar to how they can have multiple accounts
- Users can use their emails to register, but verifying the emails is beyond the scope of this PR
- The built-in sign in page will support Passkeys
- The `signIn` method from `next-auth/webauthn` will support Passkeys

#### Data model
To match the specs, this is the data model I chose:
- A new data structure is needed to hold the user's authenticators: **Authenticator**
- When a user signs up or links a Passkey, a new Passkey account and an Authenticator will be created
- An Authenticator is one-to-one with the a passkey account
- A user can have as many passkey accounts/authenticators as they wish

#### Default sign in page
Currently, the signin page is SSR'd by preact and sent as rendered HTML. Since Passkeys rely on user interaction and JS APIs, I had to find a way to include JS code in the rendered HTML. This is how I did it:
- The code that handles interactivity is inlined into the HTML
- The main library, `@simplewebauthn/browser`, is loaded via a script tag using a customizable version from unpkg.com
- These two scripts are only added to the signin page if the developer enabled a Passkey provider with Conditional UI

#### Custom sign in pages
To allow sign ups and sign ins with Passkeys on custom pages the procedure is identical to other providers. Simply import `signIn` from `next-auth/webauthn` and call it with your provider ID:
```ts
import { signIn } from "next-auth/webauthn"
// ...

// To trigger Conditional UI
signIn("passkey")
// or
signIn("passkey", { action: "authenticate" })

// To handle both existing and new users
signIn("passkey", { email: "myuser@email.com" })

// To only allow users to log in
signIn("passkey", { action: "authenticate" })

// To only allow the user "myuser@email.com" to log in
signIn("passkey", { action: "authenticate", email: "myuser@email.com" })

// To prompt the non existent user "myuser@email.com" to register
signIn("passkey", { action: "register", email: "myuser@email.com" })

// If a user is currently logged in, prompt the user to register a new passkey for their account
signIn("passkey", { action: "register" })
```

#### Email validation
When registering, there is no verification of the user's email. I find this acceptable, since:
- there are many OAuth providers that don't verify emails.
- one of my UX references for Passkey flows was [1Password's Passage](https://console.passage.id), and they don't verify emails.
- developers can extend Passkey option methods such as `getUserInfo` to require existing accounts and verified emails

#### Adapters
Passkeys will require an active database adapter to work, since they need to store the user's credentials.
Since a new data structure had to be created, new Adapter methods were added. That means current adapters are not compatible with Passkeys.

The new methods that must be implemented are:
- `getAccount`: Given an account ID and a provider, return the account object
- `createAuthenticator`: Creates a new authenticator and binds it to an account.
- `getAuthenticator`: Browsers may only provide the authenticator ID during authentication, so this allows us to retrieve it when that happens
- `listAuthenticatorsByUserId`: Allows us to list the authenticators assigned to the user
- `updateAuthenticatorCounter`: Authenticators have a `counter` property that keeps track of the number of logins to flag cloned devices or leaked credentials. This method is called on each login to update the counter

#### Passkey challenge
The Passkey flow requires that the server generate a challenge token that will persist across multiple user requests, similar to how OAuth requires `state` and `pkce` challenges. I decided to follow the current implementation of state and store the challenge in a signed cookie and reading from it on subsequent requests.

#### Passkey options endpoint
The first step of the Passkey flow involves a request from the browser that retrieves the server challenge and information about the server's capabilities. Since passkeys differentiate between registration and authentication, most implementations create two separate endpoints for this task. The normal flow on a new device is then:
1. User supplies their email or username in a form.
2. The browser sends the email or username to the server and asks if the user already exists.
3. If the browser discovers that the user exists, it makes a request to the authentication options endpoint. If the user does not exist, it makes a request to the registration options endpoint.
4. The server sends the required options back and the browser follows the regular Passkey ceremony with the User.

I find that most of these requests can be simplified into one that handles both authentication and registration options generation. Thus, I created a single new API endpoint that takes a few parameters and executes steps 1, 2, and 3 all at once.
Its logic is explained in the flowchart above.

#### Provider config
The Passkey provider takes only required parameter - the Relaying Party identifiers.
In Passkey terminology, the server has the role of Relaying Party, and is identified by the following parameters:
- A name, which is just a regular friendly string
- An ID, which is the server's domain
- It's origin, the website's origin for cross-site validation

There are several optional parameters that you can find documented in the code.

#### Usage
Developers who want to use passkeys need two things:
- Choose a database adapter that implements the new Authenticator methods
- Update their datastore or database to hold the new Authenticator data structure

The provider can be easily setup like any other:
```ts
// [...nextauth].ts
// ...
import Passkey from '@auth/core/providers/passkey'

export const authConfig: AuthConfig = {
  adapter: AdapterWithNewMethods(),
  providers: [
    Passkey({
      relayingParty: {
        name: "My awesome app",
        id: "awesome.example.com",
        origin: "https://awesome.example.com",
      },
    }),
  ],
}
```

#### Security considerations
Passkeys are considered safe without a second factor authentication since they incorporate both expectations of TFA into a single flow. Because of that, they are a good option for users who do not want to keep typing passwords or managing multiple accounts in multiple OAuth providers.

For server administrators, passkeys are also a good alternative since they are useless to hackers. Passkeys are specific to the user and website, so, if compromised, they can't be used to gain access of other services. Having a user's passkey also does not allow a hacker to log in as the user, since it only represents the public part of the asymmetric key.
Finally, since there's no third party server, compromised passkeys do not give attackers access to other services' APIs like authentication tokens would.

As mentioned before, the lack of email validation can be a security caveat, but it doesn't introduce any security flaws that weren't already present, or that are not addressable by developers.

#### RFC
Please comment on the current state of the PR and any concerns you may have. Although this is currently a working implementation of passkeys, I'm open to suggestions and improvements.
After we lock in on the implementation details, I'll update the docs.

Since this is a big PR of a big feature, I decided to avoid touching any existing code to prevent bugs from being introduced. There are, however, some opportunities for reduction of boilerplate and deduplication of code used by multiple providers.
If this PR is approved, I'll submit a new one that implements these optimizations.

#### Acknowledgements
Implementing this would have been impossible without the amazing [SimpleWebAuthn](https://simplewebauthn.dev) library. As great as WebAuthn and passkeys are, their spec is daunting to say the least.
I must also mention the work done in #6065, since it provided me with guidance especially when I first started working on this codebase.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

There doesn't appear to be any open issues related to passkeys or webauthn.

A few closed issues are:
- #5663 
- #8279 
- #7624 

And the discussion #3937 

## 📌 Resources


Maronato:
@linyiru @ThangHuuVu @balazsorban44 
Let me know if you have any concerns/requirements/questions. I'm currently using a version of this in our services, but would love to see it getting adopted

packages/core/src/lib/passkey/options.ts
```diff
@@ -0,0 +1,130 @@
...
+    rpName: provider.relayingParty.name,
+    excludeCredentials: authenticators?.map((a) => ({
+      id: a.credentialID,
+      type: "public-key",
+      transports: a.transports,
+    })),
+    authenticatorSelection: {
+      residentKey: "preferred",
+      userVerification: "preferred",
+      requireResidentKey: true,
```
ThangHuuVu, line 125: In the [spec](https://w3c.github.io/webauthn/#dictionary-authenticatorSelection), it is said that this should only be set to `true` if `residentKey` is `required`
> [Relying Parties](https://w3c.github.io/webauthn/#relying-party) SHOULD set it to true if, and only if, [residentKey](https://w3c.github.io/webauthn/#dom-authenticatorselectioncriteria-residentkey) is set to [required](https://w3c.github.io/webauthn/#dom-residentkeyrequirement-required).

since`residentKey` is being set to `preferred`, I think we can remove this value. 👀

packages/core/src/lib/passkey/verify.ts
```diff
@@ -0,0 +1,249 @@
...
+  if (!verified) {
+    return "Failed to verify response."
+  }
+
+  // Update the authenticator counter
+  try {
+    const newCounter = authenticationInfo.newCounter
+    await adapter.updateAuthenticatorCounter(authenticator, newCounter)
+  } catch (e: any) {
+    logger.error(new AdapterError(e))
```
ThangHuuVu, line 119: Should we tolerate this counter update failure like this, or we abort and return an error?

packages/core/src/lib/routes/passkey-options.ts
```diff
@@ -0,0 +1,125 @@
...
+  switch (selectedAction) {
+    case "authenticate": {
+      if (!loggedIn) {
+        return doAuthenticate(options, queryEmail)
+      }
+
+      break
+    }
+    case "register": {
+      if ((loggedIn !== providedQueryEmail) && !userWithEmailExists) {
```
ThangHuuVu, line 102: Registering could also happens if the same user want to create another authenticator, right? Then I think we don't need to check for `!userWithEmailExist` 🤔

packages/core/src/lib/callback-handler.ts
```diff
@@ -122,116 +124,203 @@ export async function handleLogin(
...
-      const provider = options.provider as OAuthConfig<any>
-      if (provider?.allowDangerousEmailAccountLinking) {
-        // If you trust the oauth provider to correctly verify email addresses, you can opt-in to
-        // account linking even when the user is not signed-in.
-        user = userByEmail
+    // We want to make sure our adapter implements the required methods.
+    // Obs: we use ! here because we know from above that the adapter is not undefined.
+    const adapter: Adapter = options.adapter!
+
+    assertAdapterImplementsMethods(
```
ThangHuuVu, line 244: Same as comment as above, let's throw the error earlier when calling assert config, this would be more useful for the developers IMO

ThangHuuVu (commented):
Done my first round of review. Overall, I agree with most of your decisions, just got some comments/questions for the implementation.
I haven't reviewed the callback route/handle login much, will comeback with more comments there!

packages/core/src/lib/routes/passkey-options.ts
```diff
@@ -0,0 +1,125 @@
...
+    case undefined: {
+      if (!loggedIn) {
+        if (providedQueryEmail === userWithEmailExists) {
+          return doAuthenticate(options, queryEmail)
+        } else {
+          return doRegister(options, queryEmail as string) // TS isn't smart enough to know that this will be defined
+        }
+      }
+
+      break
```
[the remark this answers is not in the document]
Maronato: When a user opens the sign in page and submits their email, the client doesn't know whether or not the user is registered - whether the action should be `authenticate` or `register`.

Alternatively we could have two different buttons in the UI for **sign in** vs **sign up**. However, that adds more complexity and edge cases to the client.
What if the user clicks **sign in**, but their account doesn't exist? The server would return an error and the client would have to deal with it. The same is true for a user that tries to sign up with an existing email.

IMO, by having the client send just the email and no action and letting the server to figure out which action the user meant is the simplest and most intuitive.

I should note that sending the explicit actions is also useful:
- `authenticate` is sent during the sign in flow triggered by the Conditional UI flow - when the client is sure that the user is already registered beforehand.
- `register` is sent when the user wants to add another passkey to their account, or link a passkey account to an existing user.

packages/core/src/lib/cookie.ts
```diff
@@ -242,3 +253,62 @@ export class SessionStore {
...
+  data?: unknown
+): Promise<[Cookie, string]> {
+  const { cookies } = options
+
+  const encoded = encodePayload(value)
+  const payload: SignedCookiePayload = {
+    value: encoded,
+  }
+
+  const maxAge = cookies[type].options.maxAge
```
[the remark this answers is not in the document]
Maronato: Good catch! The spec recommends 5 minutes: https://www.w3.org/TR/webauthn-2/#sctn-createCredential

packages/core/src/lib/routes/passkey-options.ts
```diff
@@ -0,0 +1,125 @@
...
+    providerAccountId: regOptions.user.id,
+    challenge: regOptions.challenge,
+  }
+  const [cookie] = await signCookie("challenge", cookieData, options)
+
+  // Return the options and set the challenge cookie
+  return {
+    status: 200,
+    body: { options: regOptions, action: "register" },
+    cookies: [cookie],
```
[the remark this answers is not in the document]
Maronato: Yes! Fixed in 0860d47e73adb9e4e3639a77fdbda03fc0051e54

packages/core/src/lib/passkey/types.ts
```diff
@@ -0,0 +1,33 @@
...
+  : T extends RegisterOption
+  ? { options: PublicKeyCredentialCreationOptionsJSON; action: RegisterOption }
+  : never
+
+export type PasskeyOptionsQuery = {
+  action: string
+  email?: string
+}
+
+export type Options = InternalOptions<PasskeyProviderType>
```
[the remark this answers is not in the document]
Maronato: Agree. Fixed in f1ae6dd3eb9acc941079af9ce97830bed7e5048e

packages/core/src/lib/passkey/options.ts
```diff
@@ -0,0 +1,130 @@
...
+  const { adapter: _adapter, provider } = options
+  const adapter: Adapter | undefined = _adapter
+
+  // Validate that the adapter is defined and implements the required methods
+  if (!adapter)
+    throw new MissingAdapter(
+      "WebaAuthn getUserAndAuthenticators requires an adapter."
+    )
+  assertAdapterImplementsMethods(
+    "WebaAuthn getUserAndAuthenticators requires an adapter that implements",
```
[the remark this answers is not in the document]
Maronato: You're correct, but my mistake here was not setting `residentKey` to `required`.
I'll remove `requireResidentKey` since it only exists to support the previous spec.

We wan't `required` here because passkeys are always "resident keys" as [Discoverable Credentials](https://passkeys.dev/docs/reference/terms/#discoverable-credential)

Fixed in 2a224e30e67fa041f8093123364d233aaaf62351

packages/core/src/lib/routes/passkey-options.ts
```diff
@@ -0,0 +1,125 @@
...
+  queryEmail?: string
+): Promise<ResponseInternal<PasskeyOptionsReturn> | ResponseInternal<string>> {
+  const { adapter } = options
+
+  let selectedAction: string | undefined = action
+
+  // Get the current session, if it exists
+  // NOTE: this is a bit hacky, but routes.session seems to be
+  // the only place that implements a full session/user check.
+  const { body: currentSession } = await routesSession({
```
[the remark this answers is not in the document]
Maronato: This is one of the things I thought about doing after this PR is merged - extract the session check logic into its own helper function. I haven't done it yet to avoid modifying existing code too much.

I'll leave it up to you to decide if I should extract it now or if this can be done later.

packages/core/src/lib/routes/passkey-options.ts
```diff
@@ -0,0 +1,125 @@
...
+  if (providedQueryEmail) {
+    const existingUser = await adapter?.getUserByEmail(queryEmail)
+    userWithEmailExists = !!existingUser
+  }
+
+  // This logic is explained in the original PR:
+  // https://github.com/nextauthjs/next-auth/pull/8808
+  switch (selectedAction) {
+    case "authenticate": {
+      if (!loggedIn) {
```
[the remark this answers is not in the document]
Maronato: makes sense! Fixed in 5b03ba8093cc35bc4ba51a473a2ec7f93955353b

packages/core/src/lib/callback-handler.ts
Maronato: Fixed in 0635795efddefd044c5d776fb1661127c80215b3

packages/core/src/lib/routes/passkey-options.ts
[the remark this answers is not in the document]
Maronato: Makes sense. Fixed in 0635795efddefd044c5d776fb1661127c80215b3

packages/core/src/lib/passkey/verify.ts
Maronato: We shouldn't. Failure to update the counter will make the authenticator unusable.
Fixed in 4e88ec13e45eeb27d5f21a2e750fc9d02390f2e4

btw, if this does happen (failure to update counter), the dev will probably need to go into the database and manually update using the logged `newCounter` value.

packages/core/src/lib/routes/passkey-options.ts
Maronato: `userWithEmailExists` is true if the value sent via the `email` query param belongs to an existing user.

If a user wants to register a new authenticator, they will make an authenticated request and will omit the `email` param.

We can't accept the `email` query to add an extra authenticator because that would allow anyone to register new authenticators to existing users

packages/core/src/lib/passkey/options.ts
```diff
@@ -0,0 +1,130 @@
...
+): Promise<[Authenticator[] | undefined, AdapterUser | null, AdapterAccount | null]> {
+  const { adapter: _adapter, provider } = options
+  const adapter: Adapter | undefined = _adapter
+
+  // Validate that the adapter is defined and implements the required methods
+  if (!adapter)
+    throw new MissingAdapter(
+      "WebaAuthn getUserAndAuthenticators requires an adapter."
+    )
+  assertAdapterImplementsMethods(
```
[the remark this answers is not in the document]
Maronato: Makes sense. Fixed in 0635795efddefd044c5d776fb1661127c80215b3

packages/core/src/lib/passkey/options.ts
Maronato: You're correct, but my mistake here was not setting `residentKey` to `required`.
I'll remove `requireResidentKey` since it only exists to support the previous spec.

We wan't `required` here because passkeys are always "resident keys" as [Discoverable Credentials](https://passkeys.dev/docs/reference/terms/#discoverable-credential)

Fixed in 2a224e30e67fa041f8093123364d233aaaf62351

packages/core/src/lib/routes/passkey-options.ts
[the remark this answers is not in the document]
ThangHuuVu: agree, let's do it this way. Thanks for the clarification!

packages/core/src/lib/passkey/browser-script.js
```diff
@@ -0,0 +1,225 @@
...
+   * @param {T} method http method to use
+   * @param {PasskeyOptionsAction} action action to submit
+   * @param {unknown | undefined} data optional data to submit
+   * @returns {Promise<T extends "GET" ? Response : void>}
+   */
+  async function submitForm(method, action, data) {
+    const form = getForm()
+    console.log("SUBMITTING FORM", {form, method, action, data})
+    
+    if (method === "GET") {
```
ThangHuuVu, line 74: I don't see the GET method usage here, can we remove it?

packages/core/src/lib/passkey/options.ts
```diff
@@ -0,0 +1,124 @@
...
+  // Generate authentication options
+  const authOptions = await generateAuthenticationOptions({
+    rpID: provider.relayingParty.id,
+    timeout: provider.timeout,
+    allowCredentials: authenticators?.map((a) => ({
+      id: a.credentialID,
+      type: "public-key",
+      transports: a.transports,
+    })),
+    userVerification: "preferred",
```
ThangHuuVu, line 75:
```suggestion
    userVerification: "preferred",
    ...provider.authenticateOptions
```
do you think we should allow the developer to override the authentication/register options?

packages/core/src/lib/routes/passkey-options.ts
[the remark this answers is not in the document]
ThangHuuVu: let's do it in a follow up PR

packages/core/src/lib/routes/passkey-options.ts
ThangHuuVu: got it, thanks for the explanation

packages/core/src/lib/passkey/browser-script.js
```diff
@@ -0,0 +1,225 @@
+//@ts-check
```
ThangHuuVu, line 1: I am still thinking about an easier solution for customized sign in page to have this feature. I think most of our developers are using custom pages. 🤔 maybe we can introduce `@auth/core/browser` and export this script & the SimpleWebauthnBrowser script? cc @balazsorban44

ThangHuuVu (commented):
Hi @Maronato could you update the branch again? 🙏 I dropped some more review comments, lets discuss

ThangHuuVu:
@Maronato FYI, we want to introduce passkey as an "experimental" feature for Auth.js, so we are working on a way to adding such features in the core: https://github.com/nextauthjs/next-auth/pull/9178

Once this is merged, let's pull in the changes and add a new key for this feature!

peterferguson:
We have been using passkeys with a hacked version of next-auth v4 for about a year now & [this restriction](https://github.com/Maronato/next-auth/blob/27b8f00a99a0c0d14ac35cd8abc7d29ceeb9400f/packages/core/src/lib/callback-handler.ts#L258) to only allow a user to have a single passkey account seems strange.

The [webauthn spec](https://www.w3.org/TR/webauthn-3/#user-account) is clear that a user should be able to have multiple accounts. That is a user should be able to have multiple webauthn credentials to a single replying party.

For our use-case the passkey defines an Ethereum account for our users. Each user should be able to define many account & still only be tied to one user.

McPizza0:
@peterferguson After getting this running locally, I can confirm that a single user can have multiple passkeys.
The "Account" in this case is type 'passkey' and the "Authenticator" is for each individual passkey.

With this implementation it's only possible to have 1 single passkey provider per user, with multiple keys for the provider.
This is also why it requires new DB tables and methods

peterferguson:
> @peterferguson After getting this running locally, I can confirm that a single user can have multiple passkeys. The "Account" in this case is type 'passkey' and the "Authenticator" is for each individual passkey.
> 
> With this implementation it's only possible to have 1 single passkey provider per user, with multiple keys for the provider. This is also why it requires new DB tables and methods

I see this makes sense. 

Kind of annoying for us as we used the `providerAccountId` to represent the ethereum address created with each passkey, but in this version the `providerAccountId` would just be directly linked to the original `email` that was used to create the passkey?

Also I see that this implementation is using email as the username for creating the passkey but this isn't a requirement according to the spec, it can be any username. Is the purpose of this just better align with the other auth methods?

McPizza0:
> Kind of annoying for us as we used the `providerAccountId` to represent the ethereum address created with each passkey, but in this version the `providerAccountId` would just be directly linked to the original `email` that was used to create the passkey?

correct

in your current setup, does it all connect back to the same user object?
i.e. if user is logged in, they can access all addresses, or do they need to re-auth for each address?

> Also I see that this implementation is using email as the username for creating the passkey but this isn't a requirement according to the spec, it can be any username. Is the purpose of this just better align with the other auth methods?

I have it working where the passkey name and display name are different than what the server sends back. If you're implementing your own sign in page its trivial to do

peterferguson:
> > Kind of annoying for us as we used the `providerAccountId` to represent the ethereum address created with each passkey, but in this version the `providerAccountId` would just be directly linked to the original `email` that was used to create the passkey?
> 
> correct
> 
> in your current setup, does it all connect back to the same user object? i.e. if user is logged in, they can access all addresses, or do they need to re-auth for each address?
> 
> > Also I see that this implementation is using email as the username for creating the passkey but this isn't a requirement according to the spec, it can be any username. Is the purpose of this just better align with the other auth methods?
> 
> I have it working where the passkey name and display name are different than what the server sends back. If you're implementing your own sign in page its trivial to do

Yea current setup a single user has many accounts each representing an eth address (& passkey). The user only every re-auths to a specific passkey when an action is required on a specific address.

I.e. when I am authed using any of my passkeys, I can see all my accounts linked to the user & then only re-auth to a specific passkey when want to sign a transaction for that public key

> I have it working where the passkey name and display name are different than what the server sends back. If you're implementing your own sign in page its trivial to do

Yea I think this will be the route that most people will go down but for that reason it should probably be the default behaviour?

bestickley:
Excited for this feature! Any updates?

ndom91:
Hey @Maronato if you're still interested in getting this merged, coudl you please take a look at the merge conflicts? :pray: 

Btw the `experimental` options key PR that Thang referred to has been merged now!

Maronato:
Hey everyone!
There were too many conflicts with `main` to reasonably resolve, so I decided to nuke the repo and implement everything from scratch. I think it turned out a bit better, and I took the chance to remove some constraints I created the first time around.

The old repo is still available in my [`passkeys-backup`](https://github.com/Maronato/next-auth/tree/passkeys-backup) branch if we ever need it.

Here are the most important changes between this version and the old one:
- No conflicts with main
- There is a new provider called `webauthn` that implements a WebAuthn-agnostic flow. The Passkey provider inherits from it and adds some default configs that are specific to passkeys
- I removed the "single-passkey-account" constraint. Now, every passkey is a new account.
- I removed the "single-passkey-provider" constraint. Now you can have as many passkey providers as you want, though only one can be used in the default sign in form at a time.
- I added support for passkey providers in the `next-auth/react` `signIn` method. This makes it a lot easier for folks to use passkeys in their custom sign in/register pages. It also supports attaching new passkeys to existing users.
- `@simplewebauthn` was updated to version 9.0.0
- The new `webauthn` and Passkey providers now have a lot of optional configs. Users should now be able to fully customize its behavior.
- The provider is now agnostic of the signup form. Meaning it isn't dependent on the user email, and devs can implement whatever form and validation they wish via the `getUserInfo` config method. Auth.js does require an email, though, so you'll still need to fetch it from some place.
- Passkeys are defined as an *experimental* feature, meaning you need to enable it via `experimental.enableWebAuthn` in your configs.
- This PR adds the required adapter methods to the **Prisma** adapter. 

I think that's pretty much it.
Let me know what you think!
cc: @ThangHuuVu @ndom91 @balazsorban44

ndom91:
@Maronato thanks a lot for continuing to work on this! I tested it locally and it required some slight modifications on our side, so we're working on fixing the ground work, but are hoping to get this in soon.

Maronato:
Hey @ndom91! Just wondering about your setup for local testing. I needed to make some adjustments in `apps/dev/nextjs`, but didn't need to change anything in `core`, `next-auth`, or the Prisma adapter. Did you encounter any issues in those areas?

By the way, I see the irony in my asking, but do you have any estimate on when this PR might be reviewed/merged? Considering it's going to be experimental, I'm guessing there's a bit more leeway?

I'm asking because it's a pretty extensive PR that affects several critical paths, so most of it will likely face conflicts with ongoing changes as time passes. I realize that last time the conflicts were largely due to my own hiatus. This time, however, I'm committed to addressing any feedback quickly to help get this released asap!

packages/core/src/lib/utils/session.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (sessionStrategy === "jwt") {
+    try {
+      const salt = options.cookies.sessionToken.name
+      const payload = await jwt.decode({ ...jwt, token: sessionToken, salt })
+
+      if (!payload || !payload.sub) throw new Error("Invalid JWT")
+
+      return adapter?.getUser(payload.sub) ?? null
+    } catch (e) {
+      logger.error(new JWTSessionError(e as Error))
```
ThangHuuVu, line 33: We need to destroy the broken session cookie here too
https://github.com/nextauthjs/next-auth/blob/9724c208f3eee09bf625e7fd4cd91f3a80fa28b1/packages/core/src/errors.ts#L218-L224

packages/core/src/lib/actions/callback/index.ts
```diff
@@ -363,6 +366,124 @@ export async function callback(
 
       await events.signIn?.({ user, account })
 
+      return { redirect: callbackUrl, cookies }
+    } else if (provider.type === "webauthn" && method === "POST") {
+      // Get callback action from request. It should be either "authenticate" or "register"
+      const action = request.body?.action as unknown
```
ThangHuuVu, line 372:
```suggestion
      const action = request.body?.action
```
we don't need type casting here right?

packages/core/src/lib/utils/webauthn-client.js
```diff
@@ -0,0 +1,224 @@
...
+ * } WebAuthnOptionsReturn
+ */
+
+/**
+ * webauthnScript is the client-side script that handles the webauthn form
+ * 
+ * @param {string} baseURL is the base URL of the auth API
+ * @param {string} providerID is the ID of the webauthn provider
+ */
+export async function webauthnScript(baseURL, providerID) {
```
ThangHuuVu, line 24: Let's name it `authURL` to avoid confusion with `basePath`, an option that we recently added 😁

packages/core/src/lib/utils/webauthn-utils.ts
```diff
@@ -0,0 +1,418 @@
...
+/**
+ * Decides the WebAuthn options based on the provided parameters.
+ * 
+ * @param action - The WebAuthn action to perform (optional).
+ * @param loggedInUser - The logged-in user (optional).
+ * @param userInfoResponse - The response containing user information (optional).
+ * 
+ * @returns The WebAuthn action to perform, or undefined if no decision could be made.
+ */
+export function decideWebAuthnOptions(
```
ThangHuuVu, line 57: nit: I prefer the name 
```suggestion
export function inferWebAuthnOptions(
```

packages/core/src/lib/utils/webauthn-utils.ts
```diff
@@ -0,0 +1,418 @@
...
+type RGetUserInfo = Awaited<ReturnType<GetUserInfo>>
+
+/**
+ * Decides the WebAuthn options based on the provided parameters.
+ * 
+ * @param action - The WebAuthn action to perform (optional).
+ * @param loggedInUser - The logged-in user (optional).
+ * @param userInfoResponse - The response containing user information (optional).
+ * 
+ * @returns The WebAuthn action to perform, or undefined if no decision could be made.
```
ThangHuuVu, line 55: shouldn't it be null
```suggestion
 * @returns The WebAuthn action to perform, or null if no decision could be made.
```

packages/core/src/lib/utils/webauthn-utils.ts
```diff
@@ -0,0 +1,418 @@
...
+ */
+export function decideWebAuthnOptions(
+  action: WebAuthnAction | undefined,
+  loggedIn: boolean,
+  userInfoResponse: RGetUserInfo
+): WebAuthnAction | null {
+  const { user = null, exists = false } = userInfoResponse ?? {}
```
ThangHuuVu, line 62:
```suggestion
  const { user, exists = false } = userInfoResponse ?? {}
```

packages/core/src/providers/passkey.ts
```diff
@@ -0,0 +1,84 @@
...
+ * ### Setup
+ * 
+ * #### Configuration
+ * ```ts
+ * import { Auth } from "@auth/core"
+ * import Passkey from "@auth/core/providers/passkey"
+ *
+ * const request = new Request(origin)
+ * const response = await Auth(request, {
+ *   providers: [Passkey({ relayingParty: { id: "example.com", name: "My Website", origin: "https://example.com" } })],
```
ThangHuuVu, line 27: our current convention is to make this config minimal, I suggest we can infer the relayingParty's fields from env vars prefixed with `AUTH_`, how do you think?

ThangHuuVu (changes requested):
I have some comments for the current state of the PR - very good work so far, thank you @Maronato for the hard work 🫶
I haven't looked at all the files yet, will come back for more later 🙏

packages/core/src/lib/utils/session.ts
```diff
@@ -0,0 +1,50 @@
+import { AdapterUser } from "../../adapters";
+import { JWTSessionError, SessionTokenError } from "../../errors";
+import { InternalOptions } from "../../types";
+import { SessionStore } from "./cookie";
```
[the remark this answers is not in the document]
Maronato: Makes sense, and I changed it to return regular `User` objects instead of `AdapterUser`. Regarding updating the session, I'm not sure if `getLoggedInUser` should do that. Reading from the session shouldn't necessarily update it, wdyt?

packages/core/src/lib/utils/session.ts
```diff
@@ -0,0 +1,50 @@
...
+  const sessionToken = sessionStore.value
+  if (!sessionToken) return null
+
+  // Try to decode JWT
+  if (sessionStrategy === "jwt") {
+    try {
+      const salt = options.cookies.sessionToken.name
+      const payload = await jwt.decode({ ...jwt, token: sessionToken, salt })
+
+      if (!payload || !payload.sub) throw new Error("Invalid JWT")
```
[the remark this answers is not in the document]
Maronato: That shouldn't be necessary here. `in` checks are for narrowing object vs non object types. `jwt.decode` is guaranteed to either return a JWT object, null, or throw an error. Meaning that at this point we can be sure that `payload` is an object, and doing `payload.sub` is valid

packages/core/src/lib/actions/callback/index.ts
Maronato: without it `action` would be `any`. Though, the next line narrows the type enough, so the casting here can be safely removed 👍

ndom91:
I'm getting an error about `NextRequest` isn't a constructor regarding `packages/next-auth/src/lib/env.ts:11`. Changing it to a normal `Request` seems to work :thinking: 

Did next 14 change something abotu NextRequest?

Also I can only get this to work if I set my `AUTH_URL=http://localhost:3000/auth`, i.e. include the `basePath` in the `AUTH_URL`. Is that expected?

ndom91:
Okay so I checked out some other Next.js code, this ^^ _shuold_ definitely be working. Seems to be an issue on my side only. Can anyone confirm that? That they're not seeing this issue about `NextRequest not being a constructor`?

What is still potentially an issue imo is the `AUTH_URL` including the `basePath` thing. @Maronato, is that how you designed it? Expecting the `AUTH_URL` to include the potnetial `basePath`? :thinking:

ndom91:
Update 2: I was able to signin no problem when I **unset the `AUTH_URL` env var**. So when it has to determine the URL itself it seems to work. 

When I set the `AUTH_URL` env var either to `http://localhost:3000` or `http://localhost:3000/auth` (i.e. with and without `basePath`), I start getting errors.

Primarily the `NextRequest` class error I mentioned above. However, when the `AUTH_URL` isn't set, it doesn't go through that code path, so i think its still an issue potentially, it just doesn't get hit when `AUTH_URL` is not set.

I tried for a while to pin down the exact issue with this `AUTH_URL` / `basePath` thing and unfrotunately I wasn't able to find a concrete fix. But I think I narrowed it down to ~`packages/next-auth/src/lib/env.ts` in the `reqWithEnvURL` fn, or in the (new) `setEnvDefaults` function. 

Wuold be awesome if someone could confirm if its only me, or if they also get errors when `AUTH_URL` is set :pray:

ndom91:
Hey @Maronato yeah okay great thanks for confirming you also see that issue. 

It does seem like it's in main, we're taking a look at it as we speak 🙏

smndtrl:
@Maronato, am I correct that this implementation requires the presence of an e-mail address at sign in time and at registration time? I spent way too long trying to build a sign-in experience like on github where one gets prompted automatically to sign-in with a passkey.

Maronato:
Hey @smndtrl ! 

Some kind of user handle is only necessary during sign up.

The experience you refer to is possible through [Conditional UI](https://github.com/w3c/webauthn/wiki/Explainer:-WebAuthn-Conditional-UI). With it, the user will be prompted to authenticate with their passkey as soon as they open the sign in page, without having to type their email address or have any other interaction. Keep in mind that the user will only be prompted to authorize passkeys they have access to on their current. If they don't have access to any passkeys for your website on their device, nothing will happen.

If they have passkeys for your site on other devices, you'll need to get their email address so the server can return a list of the user credentials it knows of. Then the browser will allow you to use a separate device to authorize one of those.

This works by default if you use the default sign in page provided by auth.js, but you can easily implement it on your custom pages.

If you're using react, all you should need to do is call the `signIn` method from `next-auth/react` with the passkey provider id as an argument: `signIn("passkey")`

Do this when the sign in page is ready, or when any other condition is met, and the experience will be the same as on GitHub.
However, you must know that some browsers regulate how often you can use Conditional UIs. Safari, for instance, only allows one Conditional UI prompt per page load. All other Webauthn prompts must be attached to a user interaction, such as clicking a button. Navigation in single apps *doesn't* count as page loads.

packages/core/src/lib/utils/webauthn-utils.ts
```diff
@@ -0,0 +1,440 @@
...
+ * 
+ * @param options - The internal options for WebAuthn.
+ * @param user - The user information.
+ * @returns The registration options.
+ */
+async function getRegistrationOptions(options: InternalOptionsWebAuthn, user: User & { email: string }) {
+  const { provider, adapter } = options
+
+  // Get the user's authenticators.
+  const authenticators = user["id"] ? await adapter.listAuthenticatorsByUserId(user.id) : null
```
ThangHuuVu, line 357: would it make sense to throw if authenticators is an empty array or null? It means that the user attemped to authenticate without prior registering.

packages/core/src/lib/utils/webauthn-utils.ts
```diff
@@ -0,0 +1,440 @@
...
+
+  // Generate a random user ID for the credential.
+  // We can do this because we don't use this user ID to link the
+  // credential to the user. Instead, we store actual userID in the
+  // Authenticator object and fetch it via it's credential ID.
+  const userID = randomString(32)
```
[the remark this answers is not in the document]
Maronato: There's no need for the userId to be the same between client and server, and I think it's better to be consistent in it's definition than to sometimes send the internal userId and sometimes not.

packages/core/src/lib/utils/webauthn-utils.ts
```diff
@@ -0,0 +1,440 @@
...
+  // Verify the response
+  let verification: VerifiedRegistrationResponse
+  try {
+    const relayingParty = getRelayingParty(options)
+    verification = await verifyRegistrationResponse({
+      ...provider.verifyRegistrationOptions,
+      expectedChallenge,
+      response: data as RegistrationResponseJSON,
+      expectedOrigin: relayingParty.origin,
+      expectedRPID: relayingParty.id,
```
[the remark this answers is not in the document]
Maronato: `generateAuthenticationOptions` and `generateRegistrationOptions` only accept a single string, but I agree that there's value in allowing users to have more than one possible RP. I modified the config to accept a custom `getRelayingParty(options, request)` that returns the RP to use for the current request

packages/core/src/lib/utils/webauthn-utils.ts
```diff
@@ -0,0 +1,440 @@
...
+        credentialID,
+        oldCounter: authenticator.counter,
+        newCounter: authenticationInfo.newCounter,
+      })}`, e)
+  }
+
+  // Get the account and user
+  const account = await adapter.getAccount(authenticator.providerAccountId, provider.id)
+  if (!account) {
+    throw new AuthError(`WebAuthn account not found in database: ${JSON.stringify({ credentialID, providerAccountId: authenticator.providerAccountId })}`)
```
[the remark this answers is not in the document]
Maronato: what do you suggest should happen instead? Authentication can't proceed if we don't find the account

packages/core/src/lib/utils/webauthn-utils.ts
Maronato: I don't think so. In this case all that's gonna happen is `allowedCredentials` returns empty and the browser informs the user the server has no authenticators for them. It won't be an invalid state or anything like that, so no need to throw

ndom91:
So I tested this again with the adapter sister [PR]() as well as the next.js dev app changes that were reverted from here (https://github.com/nextauthjs/next-auth/pull/8808/commits/598a0dda661c3c99daadde4438e6b42ce170e93c), and its still looking good!

Something important to note that's not necessarily a part of this PR, but important to know when trying this - Prisma doesn't support the edge runtime and local sqlite DB's.

You can use the Auth.js `middleware` in the Next.js dev app, but then you have to use Prisma's edge specific client (`import { PrismaClient } from "@prisma/client/edge"`) and opt-in to their "Accelerate" platform, afaik. 

So basiclaly long story short, for local dev, you have to delete / rename the `middleware.ts` file.

Anyway, I'm happy moving forward with this under "experimental". I'd also write a little page for the docs for trying this out then :+1:

ndom91:
Regarding the Prisma issue, I did some more research and it looks like they just shipped somethign ~2 days ago that might help.. 

https://github.com/prisma/prisma/issues/22893#issuecomment-1921663841

They apparently updated the prisma client for edge (https://github.com/prisma/prisma/pull/22912).

However, looking into the PR a bit more, it just seems to log the error at query time, instead of at client instantiation :thinking: So maybe I'm missing something.. I tried to intall the aforementioned `5.9.1` version here in the PR and instead just get this error message now:

image


Anyway, lets not let this Prisma edge client compat thing stop us..

packages/core/src/lib/utils/webauthn-utils.ts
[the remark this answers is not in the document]
ThangHuuVu: nevermind, this will be solved once https://github.com/nextauthjs/next-auth/pull/9871/ is merged 😄

packages/core/src/lib/utils/webauthn-utils.ts
ThangHuuVu: perfect!

ThangHuuVu (approved):

ndom91 (approved):

Maronato:
@ndom91  @ThangHuuVu  LGTM 🚀

balazsorban44 (approved):
This is huge! 🎉 Thank you!

packages/next-auth/package.json
```diff
@@ -73,7 +73,8 @@
   ],
   "license": "ISC",
   "dependencies": {
-    "@auth/core": "workspace:*"
+    "@auth/core": "workspace:*",
+    "@simplewebauthn/browser": "v9.0.0"
```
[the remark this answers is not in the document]
ndom91: Updated

kyb3r:
How can you use this in conjunction with an email provider to let only users that have already verified their email to register a passkey. Similar to Github passkeys, which use passkeys as a 2FA method. 

Additionally something Github does to increase security is a sudo mode, which makes you verify your 2fa AGAIN before you perform a sensitive action. I know this can be done by tracking the last time the user has authenticated. Is it possible with this passkeys implementation?

kyb3r:
@Maronato, fantastic work! I've successfully implemented your provider in my project. A few suggestions for future enhancements:

1. **Authenticators Table Expansion**: Consider adding more details to the Authenticators table, such as the passkey type. This would mirror GitHub's approach, providing users visibility into the passkeys linked to their accounts. Is it feasible to identify the device or context for each passkey registration?

   Screenshot of passkey registration example

2. **Utilizing AAGUIDs**: I discovered a [community-sourced list of AAGUIDs](https://github.com/passkeydeveloper/passkey-authenticator-aaguids) for passkey authenticators, which could aid in naming authenticators in user management interfaces. It might be beneficial to store the AAGUID in our Authenticator table.

3. **Registration Origin Tracking**: Recording the origin of passkey registrations (e.g., localhost, preview deployments, production) could streamline management by clarifying where each passkey was registered.

These enhancements could significantly improve the user experience and management of passkeys.

ndom91:
@kyb3r interesting suggestions! 

Could you maybe make these in a new discussion entry instead of at the end of a merged PR? I'd be interested in following some up. 

Do you have further technical suggestions for implementing some of these?

Maronato:
Hi @kyb3r! Glad to know you're enjoying it!

Regarding your suggestions, my opinion is that since they aren't critical to the auth flow and would be used mainly for display purposes, they shouldn't be part of the core implementation. It shouldn't be a problem to implement all 3 of your suggestions in userspace - probably easier even. I think an authentication library benefits from being lean as it reduces the attack surface and cost of maintenance.

One thing that would be helpful and that I agree could be implemented is additional callbacks/hooks during the webauthn lifecycle, like "authenticatorCreated", or "authenticatorUsed". That would allow users to execute or modify stuff mid flow, making the implementation of your suggestions easier.

I'm not the person you must convince, though, so I leave the decision to implement and the actual implementation to authjs' core team.

kyb3r:
@Maronato could you provide guidance on how to obtain the AAGUID in user space? It seems like it’s stuck in the internals of the library.

PR #8833, opened 2023-10-10
Docs: camelCase all property names in Drizzle schemas & make indentation consistent

jasongitmail:
Improvement for Drizzle code samples:

Property names within the schemas are all camel case, except for the accounts table which is different. This PR makes them consistent. I also made indentation consistent.

⚠️ QUESTION: I noticed a potential bug. On Line 91, 154, 211, the `providerAccountId` is the only column that is camel case, the rest are snake case. Should that be made snake case?

E.g.
```js
// current
providerAccountId: text('providerAccountId').notNull(),

// Should it be this, given all other columns use snake_case?
providerAccountId: text('provider_account_id').notNull(),
```
I didn't change it, but can add it to this PR if someone confirms what it should be.

UPDATE: I pushed a commit to change `providerAccountId` to `provider_account_id` because I imagine it's intended to be camel case for consistency.


balazsorban44:
Thanks, but this looks correct to me. NextAuth.js provided properties are camelCase while the ones based on the OAuth standard are following snake_case

jasongitmail:
Surpising, but ok. 

Is it correct for the one database column name to be camelCase too? The rest are snake case.
https://github.com/nextauthjs/next-auth/commit/24093c676121e4d6c1c7a38c1e253d968277624d

Issue #4903, opened 2022-07-11
Include `user.id` as a default value in session


TheoBr:
### Description 📓

I have at least 3 or 4 users a week coming into the T3 Discord to ask specifically about getting a "user's ID from the `session` object"

I've copy-pasted the [snippet detailed in this issue so many times I've lost count](https://github.com/t3-oss/create-t3-app/issues/176). We're thinking of adding this by default to create-t3-app to help smooth out adoption of NextAuth

### How to reproduce ☕️

https://github.com/t3-oss/create-t3-app/issues/176

### Contributing 🙌🏽

Yes, I am willing to help implement this feature in a PR

akinnee:
When I googled how to get the user ID from the next-auth client, I found a bunch of examples of doing exactly this. So it seems like enough people want the user ID stored in the session that it makes sense to do it by default.

balazsorban44:
Thanks! There are a few things here.

First, https://github.com/t3-oss/create-t3-app/issues/176 would only make sense if `NextAuth` was initialized with an adapter (implying `session: { strategy: "database"} `), but that is not our default currently. (Ther is no `user` object in case of a JWT strategy, only `token`, and `token` can already be overwritten by the user, so there is no good default here, without also changing the `jwt` callback's default.)

Our current default is to return a subset of fields that can be used to visually present a logged-in user. The user's id is not part of that list.

https://github.com/nextauthjs/next-auth/blob/5bd00f6ff156efa22f616fd3d047005293a470e1/packages/next-auth/src/core/lib/default-callbacks.ts#L12-L14

https://github.com/nextauthjs/next-auth/blob/5bd00f6ff156efa22f616fd3d047005293a470e1/packages/next-auth/src/core/routes/session.ts#L128-L135

The reason is that every exposed field to the client adds to the privacy/security concerns, and although I cannot dig up the right thread right now, but this was a deliberate decision by Iain, the original author of `next-auth`. The closest answer I could find is this https://github.com/nextauthjs/next-auth/issues/349#issuecomment-650742917.


But if we can make the documentation more clear on this here: https://next-auth.js.org/configuration/callbacks#session-callback I'm all for it!

TheoBr:
@balazsorban44 I'd be willing to put money on the vast majority (as in >70%) of Next-Auth users needing `userId` from session. IMO it would be worth running a survey as the default experience here is rough.

We will be adding this in create-t3-app now as user IDs are _essential_ for the production of an application

[...]

PR #8933, opened 2023-10-24
return entire session from session callback

juliusmarminge:
**proposal only**: 

if you think this is valid i can tidy up the PR and make the tests pass / add more 

---

i've been using this behavior ever since starting to use the v5 alpha way back, as I don't really see the reason for this filtering.

there is no clientside code in @auth/core, so I don't fully understand why it needs to be sanitized? isn't that a job each framework adapter should take.

also, if i'm not mistaken, the default session callback will still only return a subset, so this change should only be apparent to those using a custom session callback and chooses to return more data?

### DIFF
--- a/packages/core/src/lib/routes/session.ts
+++ b/packages/core/src/lib/routes/session.ts
@@ -125,13 +125,7 @@ export async function session(params: {
 
       // Pass Session through to the session callback
       const sessionPayload = await callbacks.session({
-        // By default, only exposes a limited subset of information to the client
-        // as needed for presentation purposes (e.g. "you are logged in as...").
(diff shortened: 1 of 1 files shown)


balazsorban44 (commented):
We should be keeping this. `@auth/core` _can_ be used as-is, without any framework wrapping it, so it's a better default to expose a subset only, since the exposed route at `/api/auth/session` would otherwise leak something unintentionally.

You can do the following if you are fine with exposing everything:

```ts
session: ({ session, user }) => ({ ...session, user })
```

See also: https://github.com/nextauthjs/next-auth/issues/4903

PR #8964, opened 2023-10-26
docs: added callout re getServerSession nextjs.md

Firgrep:
Added an admonition/callout to v4 docs that specifies there is a contrast between the returning object using useSession (client components) and getServerSession (server components). Where the former appears to always return a session object, the latter only returns an object where there a authenticated cookies present, otherwise it returns null. I was expecting getServerSession to behave exactly like useSession, but this appears not to be the case and I would have appreciated if this was highlighted somewhere.

## 🧢 Checklist

- [ x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues / ☕️ Reasoning

Fixes: looking over the issues related to getServerSession, they are usually related to using it in next API calls, and/or they are closed. These discussions did not solve my issue because I did not technically have an error to begin with but the behaviour was just different than expected.

Cheers


ndom91 (approved):
Useful information, thanks

Issue #6226, opened 2022-12-30
releasing packages should respect the dependency graph


xmlking:
### Question 💬

is there any specify reason @auth/sveltekit  depends on older version for ` @auth/core 0.2.2` ?
```shell
pnpm why @auth/core
```
<img width="474" alt="image" src="https://user-images.githubusercontent.com/627951/210030160-40be9531-4a7a-46aa-8d05-0cc9e4ee114a.png">

### How to reproduce ☕️

```shell
pnpm why @auth/core
```

### Contributing 🙌🏽

Yes, I am willing to help answer this question in a PR

balazsorban44:
Looks like #6068 was not fixed after all.

In this: https://github.com/nextauthjs/next-auth/actions/runs/3772067552/jobs/6412870247 SvelteKit Auth got released first.

xmlking:
I will take a look, 
While researching find this options 
https://github.com/changesets/changesets/issues/238#issuecomment-614286787

xmlking:
Temp Workaround: Add to package.json
```json
	"pnpm": {
		"overrides": {
			"@auth/core": "latest"
		}
	},
``` 

https://github.com/xmlking/svelte-starter-kit/blob/main/package.json#L100-L104

jameswoodley:
I've just hit this today, nigh on a year later, anything an outsider can do to help resolve this?

balazsorban44:
I have a PR for this, tested locally: https://github.com/balazsorban44/monorepo-release/pull/9

will do some more testing.

Check out that PR's description to understand how it should work.

PR #9025, opened 2023-11-01
fix: change `@auth/core` dependency to non-experimental

juliusmarminge:
Update @auth/core dependency specifier in package.json and pnpm-lock.yaml

self-explanatory - not sure if there's a rationale why this was set to experimental but it prevents this package from getting the latest updates when core package updates (unless you use a package manager where you can do nested overrides)

### DIFF
--- a/packages/next-auth/package.json
+++ b/packages/next-auth/package.json
@@ -72,7 +72,7 @@
   ],
   "license": "ISC",
   "dependencies": {
-    "@auth/core": "workspace:experimental"
+    "@auth/core": "workspace:*"
   },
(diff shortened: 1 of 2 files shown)


balazsorban44 (approved):
This way we can ask people to re-install `next-auth` only, even if `@auth/core` has changed (which next-auth relies on).

Otherwise during release, the version on `@auth/core` would be locked to a specific version.

This is likely fine now though, since I fixed https://github.com/nextauthjs/next-auth/issues/6226

So `@auth/core` changes will now always trigger a new release for all dependent packages.

PR #9118, opened 2023-11-11
refactor: throw errors when `raw`, reorganize

balazsorban44:
The main change here is the error handling behavior when a framework sets the `raw` config option. We should let the framework handle errors.

For example in NextAuth.js, with Server Actions, throwing an `AuthError` makes it possible to handle these errors better (for example render inline form validation errors, or redirect to a desired page if a specific error occurs).

Other than that, I reorganized the source code, so in the future, we could take advantage of this for various reasons:

- reorganized `routes` (renamed to `actions`): Auth.js will stay focused on being a zero-config solution, but this is some ground work so that these modules could essentially become directly importable submodules in the future, making them better building blocks for more advanced users.
- reorganized the HTML renderer code: This idea has been around for a while. Most of the "GET" actions just use `preact` to server-side render some HTML. By reorganizing this code, we can start thinking about dropping the `preact` dependency by making it optional. This way, frameworks libraries could instead export their own UI primtives instead of hard-to-configure built-in HTML pages, which can also not be optimized by framework primitives (eg.: `next/image`). Think login buttons, profile cards, edit forms, etc. (we will likely have an RFC for the available primitives)


magnusriga:
@balazsorban44 Would it be possible to remove the `isRaw` check in `if (isAuthError && isRaw && !isRedirect) throw error`, to ensure errors are re-thrown also in the second pass through the library in an OAuth sign-in flow, i.e. when `../api/auth/callback/google` is called (using google as an example)?

That will enable us to catch errors in Next.js Route Handlers for OAuth sign-in (since it uses `httpHandler`), just like is already possible when doing a `credentials` sign-in (next-auth sets raw to true).

**From Auth() as it currently stands:**

```ts
  catch (e) {
    const error = e as Error
    logger.error(error)

    const isAuthError = error instanceof AuthError
    if (isAuthError && isRaw && !isRedirect) throw error // It would be helpful is isRaw was not part of the check here.

    // ...
  }
```

If you think it make sense, I would be happy to send a PR.

Thank you!

PR #9131, opened 2023-11-13
docs: added image

nicitaacom:
I added image where user can
Read and write and Direct message


ndom91:
Closing this PR - we're not updating v4 docs much anymore and `main` provider docs can't really support a big screenshot like this, thanks for your interest though!

PR #9271, opened 2023-11-28
feat(providers): support self-hosted github

cometkim:
## ☕️ Reasoning

Add `baseUrl` option to existing github provider, to support self-hosted [GitHub Enterprise Server](https://docs.github.com/en/enterprise-server@3.10/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps)

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

None

## 📌 Resources

### DIFF
--- a/packages/core/src/providers/github.ts
+++ b/packages/core/src/providers/github.ts
@@ -69,9 +69,12 @@ export interface GitHubProfile {
 }
 
 /**
- * You can customize GitHub's baseUrl to your own GitHub Enterprise Server.
+ * Configuration for usage with GitHub Enterprise Server.
  */
-export interface GitHubUserConfig extends OAuthUserConfig<GitHubProfile> {
+export interface GitHubEnterpriseServerConfig {
+  /**
+   * Customize our own GitHub Enterprise Server.
+   */
   baseUrl?: string
 }
 
@@ -127,7 +130,7 @@ export interface GitHubUserConfig extends OAuthUserConfig<GitHubProfile> {
  * :::
  */
 export default function GitHub(
-  config: GitHubUserConfig
+  config: OAuthUserConfig<GitHubProfile> & GitHubEnterpriseServerConfig
 ): OAuthConfig<GitHubProfile> {
   const baseUrl = config.baseUrl || "https://github.com"
   const apiBaseUrl = config.baseUrl
--- a/packages/core/src/providers/github.ts
+++ b/packages/core/src/providers/github.ts
@@ -132,7 +132,7 @@ export interface GitHubEnterpriseServerConfig {
 export default function GitHub(
   config: OAuthUserConfig<GitHubProfile> & GitHubEnterpriseServerConfig
 ): OAuthConfig<GitHubProfile> {
-  const baseUrl = config.baseUrl || "https://github.com"
+  const baseUrl = config.baseUrl ?? "https://github.com"
   const apiBaseUrl = config.baseUrl
     ? `${config.baseUrl}/api/v3`
     : "https://api.github.com"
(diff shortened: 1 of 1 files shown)


packages/core/src/providers/github.ts
```diff
@@ -120,19 +127,24 @@ export interface GitHubProfile {
  * :::
  */
 export default function GitHub(
-  config: OAuthUserConfig<GitHubProfile>
+  config: GitHubUserConfig
 ): OAuthConfig<GitHubProfile> {
+  const baseUrl = config.baseUrl || "https://github.com"
+  const apiBaseUrl = config.baseUrl
+    ? `${config.baseUrl}/api/v3`
```
ndom91, line 134: How often do they update the API version? Do we want to hardcode that here?

cometkim: Never. As it is equivalent to before, the v3 endpoint (`api.github.com`).

ndom91 (approved):

packages/core/src/providers/github.ts
```diff
@@ -68,6 +68,13 @@ export interface GitHubProfile {
   [claim: string]: unknown
 }
 
+/**
+ * You can customize GitHub's baseUrl to your own GitHub Enterprise Server.
+ */
```
ThangHuuVu, line 73: Putting the comment here will make our Typedoc plugin generates all the properties of `OAuthUserConfig` in the Github provider doc page 😅 which is not very nice. To remedy this issue, I suggest we do the following:
```ts

/**
 * Configuration for usage with GitHub Enterprise Server.
 */
export interface GithubEnterpriseConfig { 
	/**
	 * Customize our own GitHub Enterprise Server.
	 */
	baseUrl?: string
 }
```

below:
```ts
  config: OAuthUserConfig<GitHubProfile> & GithubEnterpriseConfig
```

The JSDoc for the interface is necessary for the type to be generated, and the JSDoc for the property is to explain it.

ThangHuuVu (changes requested):
The changes looks good, although I don't have an enterprise server to test it 😁 Could you take a look & fix my comment? 🙏 let me know if you have any questions.
You can start the doc page by command: `pnpm dev:docs`
Then you can check the generated result at: http://localhost:3000/reference/core/providers/github

packages/core/src/providers/github.ts
cometkim: I made the changes as suggested. I understand the intent, but manipulating types for it feels a bit weird.

A more semantically accurate definition would be like
```ts
interface GitHubCloudConfig extends OAuthUserConfig<GitHubProfile> { ... }

interface GitHubEnterpriseServerConfig extends OAuthUserConfig<GitHubProfile> { ... }

// then
GitHubCloudConfig | GitHubEnterpriseServerConfig
```

cometkim:
Maybe I can do the same for GitLab in another PR later

packages/core/src/providers/github.ts
```diff
@@ -120,19 +130,24 @@ export interface GitHubProfile {
  * :::
  */
 export default function GitHub(
-  config: OAuthUserConfig<GitHubProfile>
+  config: OAuthUserConfig<GitHubProfile> & GitHubEnterpriseServerConfig
 ): OAuthConfig<GitHubProfile> {
+  const baseUrl = config.baseUrl || "https://github.com"
```
balazsorban44, line 135:
```suggestion
  const baseUrl = config.baseUrl ?? "https://github.com"
```

packages/core/src/providers/github.ts
balazsorban44: v3 is not v1 though, so they have already changed it twice. Is there a link to documentation about this?

balazsorban44 (approved):
Approved this by moving the config to `enterprise` object, so it is more future-proof, in case other related config options come up. Thanks!

PR #9351, opened 2023-12-09
refactor: improved signin redirection methods

milindgoel15:
## ☕️ Reasoning

So basically, when using the default signin function for CredentialsProvider, we provide a `redirect` boolean which we can use to redirect user to home URL in case of success or error page in case of error response.

But it was missing the flexibility to choose whether to redirect on both success and error response or only either of them. Hence, the PR addresses such changes where the default redirect has been changed to `redirectOnSuccess` and `redirectOnFailure`. 

So 
- in case of `redirectOnSuccess` (defaults to true), the signin method will redirects the user to the callback URL as it did before while also returning a success response. 
- in case of `redirectOnFailure` (defaults to true), the signin method will redirects the user to default API error URL as it did before while returning the error response.

Additionally, they both can be individually changed as per user, so if the user prefers to deal with either of the responses, they can do freely. 

## 🧢 Checklist

Haven't updated the documentation for the new/replaced redirect options as they are taken up from v4 branch. Lemme know if I should do it myself or can be with future change. Will probably require a separate PR ig.

- [ ] Documentation
- [ ] Tests

Although its ready to be merged (according to my POV), but do check if it requires any change.
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources

### DIFF
--- a/docs/docs/concepts/faq.md
+++ b/docs/docs/concepts/faq.md
@@ -118,7 +118,7 @@ Check out the [Session strategies page](/concepts/session-strategies) to learn m
 
 ## Security
 
-Parts of this section have been moved to their [page](/security)](/security).
+Parts of this section have been moved to their [page](/security).
 
 <details>
 <summary>


socketopp:
Would this PR fix what we discuss here? #9322 #9407

milindgoel15:
> Would this PR fix what we discuss here? #9322 #9407

Hi, I tried to reproduce the same signout redirection issue in my project but I was not able to reproduce it. When the redirect value is set to false, the page would not refresh.

However, there is an issue with the useSession hook on my end which does not seem to update the session if the signin/signout is done.

And what my PR does is that it makes the redirect method more flexible when signing in. Previously, if the redirect was set to false, the next auth creates the user session. But it fails to update the session in the useSession hook as said above hence the website requires a page refresh. I couldn't find a way to fix that but I found a more flexible way to fix it, by introducing the redirectOnSuccess and redirectOnFailure which lets users redirect only and only if the session is created.

However, upon checking the 2 mentioned discussions by you, I checked out the code responsible for signout and it seems that despite of whatever value is set to redirect on signout, the page will refresh.

socketopp:
> Hi, I tried to reproduce the same signout redirection issue in my project but I was not able to reproduce it. When the redirect value is set to false, the page would not refresh.

Interesting, was your project also using `@auth/sveltekit` and using a provider such as Github or Gmail? I used Github as mentioned in the aforementioned [Sveltekit guide.](https://authjs.dev/reference/sveltekit)

milindgoel15:
> > Hi, I tried to reproduce the same signout redirection issue in my project but I was not able to reproduce it. When the redirect value is set to false, the page would not refresh.
> 
> Interesting, was your project also using `@auth/sveltekit` and using a provider such as Github or Gmail? I used Github as mentioned in the aforementioned [Sveltekit guide.](https://authjs.dev/reference/sveltekit)

No, I am using nextjs + next-auth with credentials provider. That would explain why you would face the redirection bug issue on sveltekit

milindgoel15:
Closed due to outdated fork

PR #9358, opened 2023-12-10
fix: make sure the user to install the @types/cookie0.6.0

jianqi-jin:
> _NOTE_:
>

## ☕️ Reasoning

This is just a very minor fix. The scenario this PR aims to address occurs when developers want to add options to cookie, but there are no corresponding types/interfaces. The issue arises because the developer may not have installed the same version of @types/cookie as NextAuth.

Therefore, this PR simply moves @types/cookie from devDependencies to dependencies. This adjustment is made because dependencies under devDependencies may not be necessary for developers who wish to use NextAuth. However, in reality, there are common scenarios where users/developers do want to install devDependencies like @types/cookie.

This is the case:

```
// auth.ts

export const {
  handlers: { GET, POST },
  auth
} = NextAuth({
  cookies: {
    sessionToken: {
      name: 'next-auth.session-token',
      options: {
        sameSite: 'none',
        secure: true
      }
    },
    callbackUrl: {
      name: 'next-auth.callback-url',
      options: {
        sameSite: 'none',
        secure: true
      }
    },
    csrfToken: {
      name: 'next-auth.csrf-token',
      options: {
        sameSite: 'none',
        secure: true
      }
    }
  },
...
})
```

image

As you can see, there is no types/interfaces.

When I track into the CookieOptions Types:
image

I can't see the types.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: INSERT_ISSUE_LINK_HERE

## 📌 Resources

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -64,6 +64,7 @@
   "dependencies": {
     "@panva/hkdf": "^1.1.1",
     "cookie": "0.6.0",
+    "@types/cookie": "0.6.0",
     "jose": "^5.1.0",
     "oauth4webapi": "^2.3.0",
     "preact": "10.11.3",
@@ -86,7 +87,6 @@
     "providers": "node scripts/generate-providers"
   },
   "devDependencies": {
-    "@types/cookie": "0.6.0",
     "@types/node": "18.11.10",
     "@types/nodemailer": "6.4.6",
     "@types/react": "18.0.37",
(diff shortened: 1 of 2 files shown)


balazsorban44 (approved):
Thanks! I updated the lockfile too, otherwise looks good to me!

PR #9380, opened 2023-12-13
feat: add default user id generation

balazsorban44:
Ref: https://twitter.com/Jarpiino/status/1734764485829923040

If the `profile()` callback of a provider (that is used to create a new user) does not return an `id` or it's missing before reaching a database Adapter, we should generate one via `crypto.randomUUID()` in the core library. We already did this at the Adapter level, but it's better to do it in the core so it's consistently there.

Do the same for the credentials provider's `authorize()` method too, marking `id` optional when it's not really needed.

This should also fix a rather annoying issue that Adapters need to generate an ID in `createUser`. Instead we will always pass one for certain, and the adapter can choose to either ignore or use it.

~**NOTE:** Using the modified adapters might break next-auth v4, since it would always assume there is an ID coming from `next-auth`. We _can_ make the same change for the v4 package, but it's likely a breaking change for the adapters. 🤔 I'll likely have to move the adapter changes to a different PR, where we can add a *BREAKING CHANGE* note to bump these adapters with a major. Will think about it.~ #9381

### DIFF
--- a/packages/adapter-azure-tables/src/index.ts
+++ b/packages/adapter-azure-tables/src/index.ts
@@ -94,19 +94,24 @@ export function withoutKeys<T>(
 export const TableStorageAdapter = (client: TableClient): Adapter => {
   return {
     async createUser(user) {
+      const id = crypto.randomUUID()
+      const newUser = {
+        ...user,
+        id,
+      }
       await Promise.all([
         client.createEntity({
-          ...user,
+          ...newUser,
           partitionKey: keys.userByEmail,
           rowKey: user.email,
         }),
         client.createEntity({
-          ...user,
+          ...newUser,
           partitionKey: keys.user,
-          rowKey: user.id,
+          rowKey: id,
         }),
       ])
-      return user
+      return newUser
     },
     async getUser(id: string) {
       try {
--- a/packages/adapter-azure-tables/src/index.ts
+++ b/packages/adapter-azure-tables/src/index.ts
@@ -94,24 +94,19 @@ export function withoutKeys<T>(
 export const TableStorageAdapter = (client: TableClient): Adapter => {
   return {
     async createUser(user) {
-      const id = crypto.randomUUID()
-      const newUser = {
-        ...user,
-        id,
-      }
       await Promise.all([
         client.createEntity({
-          ...newUser,
+          ...user,
           partitionKey: keys.userByEmail,
           rowKey: user.email,
         }),
         client.createEntity({
-          ...newUser,
+          ...user,
           partitionKey: keys.user,
-          rowKey: id,
+          rowKey: user.id,
         }),
       ])
-      return newUser
+      return user
     },
     async getUser(id: string) {
       try {
--- a/packages/adapter-d1/src/index.ts
+++ b/packages/adapter-d1/src/index.ts
@@ -248,14 +248,15 @@ export function D1Adapter(db: D1Database): Adapter {
 
   return {
     async createUser(user) {
+      const id: string = crypto.randomUUID()
       const createBindings = [
-        user.id,
+        id,
         user.name,
         user.email,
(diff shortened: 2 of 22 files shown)


pippinmole:
Hi,

Seriously confused why you have removed the id from the user object. How are you meant to uniquely identify users now?

Some of the logic in your initial post content doesn't make any sense:
> If the profile() callback of a provider (that is used to create a new user) does not return an id or it's missing before reaching a database Adapter, we should generate one via crypto.randomUUID() in the core library

This would imply that the user **always** has an id, because the core library generates one if it's missing.

Thanks

balazsorban44:
Yep, that's the case. Users will always have an id, the whole point of this PR. Feel free to open a PR if you are experiencing a hug! 🙏

pippinmole:
Hi,

Would you be able to explain why the user id field is nullable here then? https://github.com/nextauthjs/next-auth/pull/9380/files#diff-4c2458380e50c967c70d7afe819dc7412180f9f817043f27441de3498e4d0479R469

Thanks

balazsorban44:
The User interface is for the developer. Even if they don't set an id, we'll generate one.

The one used in database adapters is `AdapterUser` https://github.com/nextauthjs/next-auth/blob/665f2bda3aa349679de2b4aa8cf273fb6ebfe6cc/packages/core/src/adapters.ts#L179

pippinmole:
By 'user-land' code, do you mean the user object returned from calling `await auth()`:

```
  const session = await auth();
  const user = session?.user; // This
```

Surely if you always generate an ID no matter what, it should never be null on the 'user-land' code.

Screenshot 2024-01-25 at 10 19 19

Further, why is the database user ID never null, but the session user id is nullable?

davidwinter:
Is there a reason why the `user` object isn't available anymore in the `session` callback? 

image

Unless I'm misreading the type information, it _should_ be available? But I may have overlooked something... I'm using the Dynamo database adapter and it had been working up until `5.0.0-beta.5`. It looks like this issue was introduced with this change 🤔

balazsorban44:
`user.id` still can be nullable, since we don't expose it by default:

https://github.com/nextauthjs/next-auth/blob/3f940815d849b3bd53683835f4b9851750a4b1c9/packages/core/src/lib/init.ts#L45-L49

So this is the correct type `Session["user"]` at the moment.

You can follow https://github.com/nextauthjs/next-auth/pull/9702 where we might change this and always return `AdapterUser` (with `id: string`) instead.

Please open issues for bug reports or other concerns, this is a closed PR and not going to be be monitored for further activity.


https://github.com/nextauthjs/next-auth/pull/9380#issuecomment-1910267155 this is alredy fixed just not released yet.

sparanoid:
After upgrading to 1.0.12 this change breaks `PrismaAdapter`:

```
Type 'import("/Users/me/Git/project/node_modules/.pnpm/@auth+core@0.19.0_nodemailer@6.9.8/node_modules/@auth/core/adapters").Adapter' is not assignable to type 'import("/Users/me/Git/project/node_modules/.pnpm/next-auth@4.24.5_next@14.1.0_nodemailer@6.9.8_react-dom@18.2.0_react@18.2.0/node_modules/next-auth/adapters").Adapter'.
  Types of property 'createUser' are incompatible.
    Type '((user: AdapterUser) => Awaitable<AdapterUser>) | undefined' is not assignable to type '((user: Omit<AdapterUser, "id">) => Awaitable<AdapterUser>) | undefined'.
      Type '(user: AdapterUser) => Awaitable<AdapterUser>' is not assignable to type '(user: Omit<AdapterUser, "id">) => Awaitable<AdapterUser>'.
        Types of parameters 'user' and 'user' are incompatible.
          Property 'id' is missing in type 'Omit<AdapterUser, "id">' but required in type 'AdapterUser'.ts(2322)
```

PR #9381, opened 2023-12-13
feat: drop user id generation in `createUser`

balazsorban44:
Since this is a repetitive task that many adapters decided to do themselves, the parent libraries have been updated to not require adapters to generate an `id` themselves.

Before this can be merged:

- [x] Depends on #9380
- [ ] upgrade in v4 (related #9493, #9699)
- [ ] fix https://github.com/balazsorban44/monorepo-release/issues/10

BREAKING CHANGE:

The adapter now requires the parent library (either `@auth/core` or `next-auth` v4) to send the `id` property in the `createUser` method. Make sure to upgrade your parent library to their latest version.

The parent library generates the id via `crypto.randomUUID()`. If your custom adapter needs a different format, you can easily patch the `createUser(method)` to ignore the incoming `id`, or override it with your own type of id.

### DIFF
--- a/apps/dev/nextjs/app/layout.tsx
+++ b/apps/dev/nextjs/app/layout.tsx
@@ -1,4 +1,4 @@
-import { auth, signIn, signOut, update } from "auth"
+import { auth, signIn, signOut, unstable_update as update } from "auth"
 import Footer from "components/footer"
 import { Header } from "components/header"
 import styles from "components/header.module.css"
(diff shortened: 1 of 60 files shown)


ndom91:
Do you still plan on "backporting" this for v4?

I think with that recent "always set UUID to fallback `user.id`" PR in core, this could basically be good to go for v5.

What do you think?

PR #9419, opened 2023-12-19
make session id available for database dapater

### LINKED ISSUE #7150: Add ability to extend database session and use in session callback
### Description 📓

Currently, as you can see [here](https://github.com/nextauthjs/next-auth/blob/36286b1fae4f07fa8c204e88aefb0876479e8b03/packages/core/src/lib/routes/session.ts#L124-L135), only a subset of the session (from the database) is passed to the session callback.

If I want to extend that session to add additional properties and use them within the session callback, I need to go back to the database with the sessionToken (which is not available within the session call so I end up needing to do a [custom initializer](https://authjs.dev/guides/basics/initialization) to pass it in). It's not very efficient to do the extra database query, and the DX is a pain.

I would love to have the ability to extend which session fields we pass (maybe using a whitelist?), or better yet, we could just pass the session to the callback directly how we get it from the DB, and omit any unnecessary fields (like id for example).

### How to reproduce ☕️

N/A

### Contributing 🙌🏽

KATT:
> _NOTE_:
>
> - It's a good idea to open an issue first to discuss potential changes. <-- 👋 I discuss/think clearer with code

## ☕️ Reasoning

> 👋 Not intended to be merged as-is, opening up for discussion - feel free to guide me if I can already make this work or how you'd like it implemented

I'm building a two-factor-authentication thing on top of my app - I want the validity of the 2fa to be stored in the `Session` so they are tied together.

Apparently, I can't access the session id even if it exists within `getSessionAndUser()` and need to use `req.cookies['....']` right now and call `getSessionAndUser` manually, which seems a bit wasteful.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: #7150


packages/core/src/lib/actions/session.ts
```diff
@@ -126,6 +126,7 @@ export async function session(
         // By default, only exposes a limited subset of information to the client
         // as needed for presentation purposes (e.g. "you are logged in as...").
         session: {
+          id: session.id,
```
KATT, line 129: Honestly, I think the whole `adapterSession` should be available in the `session()`-callback

packages/core/src/adapters.ts
```diff
@@ -204,6 +204,10 @@ export interface AdapterAccount extends Account {
  * A session holds information about a user's current signin state.
  */
 export interface AdapterSession {
+  /**
+   * Session id. Only used when using `"database"` `AuthConfig.strategy` option.
+   */
+  id: string;
```
KATT, line 210: If this is 

```suggestion
  id?: string;
```

it should be backwards compatible

packages/core/src/lib/actions/session.ts
balazsorban44: This is not desired. This callback controls what gets exposed to the browser. We don't want to expose anything than a strict subset. session id should not be needed in the client.

KATT: I kinda want to decide what's on my session object myself based on what I have in the db, but I get your point. Is session is alright?

ndom91:
Allowing just the id seems safe enough for me :thinking: What do you think @balazsorban44 ?

ndom91:
@KATT theres still the pnpm.lock file changes in here btw

Issue #9092, opened 2023-11-08
The checks parameter in OIDCConfig causes a type error


nbifrye:
### Environment

```
  System:
    OS: macOS 12.6
    CPU: (8) arm64 Apple M1
    Memory: 103.48 MB / 8.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.asdf/installs/nodejs/20.9.0/bin/node
    npm: 10.1.0 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 119.0.6045.105
    Safari: 15.6.1


  npmPackages:
    next-auth: ^5.0.0-beta.3 => 5.0.0-beta.3
```

### Reproduction URL

https://github.com/nbifrye/76956150-50fe-4232-b78f-dba38e1d6801

### Describe the issue

Specifying the checks option for custom OAuth provider causes a type error.

```
% npx tsc index.ts --noEmit                                           
index.ts:11:12 - error TS2322: Type 'string' is not assignable to type 'never'.

11   checks: ["none"],
```

The type of `checks` is as follows, which means `("pkce" | "state" | "none")[] & "nonce"[]`. The only value that satisfies this is `[]`.

https://github.com/nextauthjs/next-auth/blob/1349ed33da3280162bd8156a80e0a4659cb873dd/packages/core/src/providers/oauth.ts#L242

The expected type of checks looks like `("pkce" | "state" | "none" | nonce")[]`. How about the following code:

```
checks?: Array<(Exclude<OAuth2Config<Profile>["checks"], undefined>)[number] | "nonce">
```

### How to reproduce

```
git clone https://github.com/nbifrye/76956150-50fe-4232-b78f-dba38e1d6801.git
cd 76956150-50fe-4232-b78f-dba38e1d6801
npm install
npx tsc --noEmit
```

### Expected behavior

`checks` such as `["none"]`, `["state", "nonce"]`, etc. do not cause a type error.

lukasz-karolewski:
Unfortunately, this is now causing this

Type 'OAuthConfig' is not assignable to type 'Provider'.
Type 'OIDCConfig' is not assignable to type 'Provider'.
Type 'OIDCConfig' is not assignable to type 'OIDCConfig & InternalProviderOptions'.
[...]

PR #9477, opened 2023-12-27
fix(ts): bad intersection caused type to be 'never'

konstantinblaesi:
There's some magic with unions/intersections resulting in _never_ type with the previous fix.

I tested this change using the nextjs next-auth example project and the google provider

For that project applying the change in these two files made the error go away

node_modules/.pnpm/@auth+core@0.18.4/node_modules/@auth/core/src/providers/oauth.ts
node_modules/.pnpm/@auth+core@0.18.4/node_modules/@auth/core/providers/oauth.d.ts

### DIFF
--- a/packages/core/src/providers/oauth.ts
+++ b/packages/core/src/providers/oauth.ts
@@ -239,9 +239,7 @@ export interface OAuth2Config<Profile>
 export interface OIDCConfig<Profile>
   extends Omit<OAuth2Config<Profile>, "type" | "checks"> {
   type: "oidc"
-  checks?: Array<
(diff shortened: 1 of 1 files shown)


konstantinblaesi:
This change is not needed, but currently people are getting errors due to the previous fix see https://github.com/nextauthjs/next-auth/issues/9092
Why was next-auth not updated to @auth/core@0.19.0 ?

nbifrye:
`OAuth2Config<Profile>["checks"]` is `("pkce" | "state" | "none")[] | undefined`, only undefined should be removed.
Using NonNullable to remove even null seems inappropriate.

konstantinblaesi:
I think it's fine as null is not a valid value anyways

PR #9497, opened 2023-12-29
feat(sveltekit): add `getSessionWithSetCookies` function for carrying over cookies and enabling token refresh in sveltekit

### LINKED ISSUE #8034: Are token cookies updated when getting the session in `@auth/sveltekit`?
### Question 💬

I noticed that when my token expires, every subsequent request refreshes the token, seemingly because that refreshed token doesn't persist. 

I found https://github.com/nextauthjs/next-auth/issues/7025#issuecomment-1479462037 which states that `getServerSession` in the `app/` directory in NextJS only provides read access to cookies. Does the same apply to Sveltekit? Are there any workarounds so that we can support token refresh?

### How to reproduce ☕️

I've setup auth with Cognito with `@auth/sveltekit` per https://authjs.dev/guides/basics/refresh-token-rotation. When the token expires, observe the token be continually refreshed every time it is grabbed via `getSession()` with the expiry time of the original token, not the most recent refreshed token expiry time.

```ts
import type { Provider } from "@auth/core/providers";
import Cognito from "@auth/core/providers/cognito";
import type { TokenSet } from "@auth/core/types";
import { SvelteKitAuth } from "@auth/sveltekit";

import {
  AUTH_SECRET,
  AWS_COGNITO_CLIENT_ID,
  AWS_COGNITO_CLIENT_SECRET,
  AWS_COGNITO_DOMAIN,
  AWS_COGNITO_REGION,
  AWS_COGNITO_USER_POOL_ID,
} from "$env/static/private";

  secret: AUTH_SECRET,
  trustHost: true,
  providers: [
    Cognito({
      clientId: AWS_COGNITO_CLIENT_ID,
      clientSecret: AWS_COGNITO_CLIENT_SECRET,
      issuer: `https://cognito-idp.${AWS_COGNITO_REGION}.amazonaws.com/${AWS_COGNITO_USER_POOL_ID}`,
      authorization: {
        params: {
          scope: "openid profile email phone",
        },
      },
    }),
  ] as Provider[], // See https://github.com/nextauthjs/next-auth/issues/6174.
  callbacks: {
    jwt: async ({ token, user, account, profile, trigger, session }) => {
      if (account) {
        return {
          access_token: account.access_token,
          expires_at: Date.now() + account.expires_in! * 1000,
          refresh_token: account.refresh_token,
          user: {
            id: profile?.sub,
            name: profile?.name,
            email: profile?.email,
            username: profile?.["cognito:username" as keyof typeof profile],
            groups: profile?.["cognito:groups" as keyof typeof profile],
          },
        };
      } else if (Date.now() < (token.expires_at! as number)) {
        // If the access token has not expired yet, return it.
        return token;
      } else {
        // If the access token has expired, try to refresh it.
        console.log("Token expired at", token.expires_at! as number);
        try {
          const response = await fetch(
            `https://${AWS_COGNITO_DOMAIN}.auth.${AWS_COGNITO_REGION}.amazoncognito.com/oauth2/token`,
            {
              headers: { "Content-Type": "application/x-www-form-urlencoded" },
              body: new URLSearchParams({
                client_id: AWS_COGNITO_CLIENT_ID,
                client_secret: AWS_COGNITO_CLIENT_SECRET,
                grant_type: "refresh_token",
                refresh_token: token.refresh_token as string,
              }),
              method: "POST",
            }
          );

          const tokens: TokenSet = await response.json();

          if (!response.ok) {
            throw tokens;
          }

          console.log("Token refreshed: new token expires at", Date.now() + tokens.expires_in! * 1000);

          return {
            ...token,
            access_token: tokens.access_token,
            expires_at: Date.now() + tokens.expires_in! * 1000,
            refresh_token: tokens.refresh_token ?? token.refresh_token,
          };
        } catch (error) {
          console.error("Error refreshing access token", error);
          // From the docs:
          // * Returning `null` will invalidate the JWT session by clearing
          // * the user's cookies. You'll still have to monitor and invalidate
          // * unexpired tokens from future requests yourself to prevent
          // * unauthorized access.
          return null;
          // return { ...token, error: 'RefreshAccessTokenError' as const }
        }
      }
    },
    session: ({ session, token }) => {
      const { access_token, user, expires_at } = token;
      return {
        ...session,
        user: user as any,
        accessToken: access_token as string | undefined,
        tokenExpiresAt: expires_at as number | undefined,
      };
    },
  },
});
```

### Contributing 🙌🏽

ndom91:
## ☕️ Reasoning

PRed from work done by @Benjaminknox (see: https://github.com/nextauthjs/next-auth/issues/8034#issuecomment-1817628064)

When `getSession()` is called it returns only the body from the backend and the header to set the authentication cookie is lost. Ben was able to create a workaround by creating another function (`getSessionWithSetCookies`) to return the `set-cookie` headers and then set them myself in `+layout.server.ts`

Usage in `+layout.server.ts`
```ts
// src/routes/+layout.server.ts
import type { LayoutServerLoad } from "./$types"

  const sessionWithCookies = await event.locals.getSessionWithSetCookies();

  for(const cookieName in sessionWithCookies?.cookies) {
    const { value, options } = sessionWithCookies.cookies[cookieName]
    event.cookies.set(cookieName, value, options)
  }

  return {
    session: sessionWithCookies?.session
  }
}
```

### Todo

- Having issues with the `cookies` types still :thinking: 

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: #8034

## 📌 Resources


packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -226,10 +242,21 @@ export async function getSession(
   const data = await response.json()
 
   if (!data || !Object.keys(data).length) return null
-  if (status === 200) return data
+  if (status === 200) return {
+    session: data,
+    cookies: parseCookies(Array.from(response.headers).filter(([headerName]) => headerName === 'set-cookie'))
```
ThangHuuVu, line 247: can we use the [`getSetCookie`](https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie) here? 😁

packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -302,6 +331,8 @@ declare global {
   namespace App {
     interface Locals {
       getSession(): Promise<Session | null>
+      getSessionWithSetCookies(): Promise<{ session: Session, cookies: Record<string, Record<string, string>> } | null>
```
ThangHuuVu, line 334: I wonder if we should modify the `getSession` function's return instead 🤔 since we're still at 0.x.x versioning.

packages/frameworks-sveltekit/src/lib/index.ts
ndom91: Yeah good call, modified that a bit. Also added `cookie and `@types/cookie` as peer deps to `@auth/svelte` since their dependencies in `@auth/core` already anyway and that'll always be installed together. Does that make sense? Or should I just add them as straight-up dependencies for `@auth/sveltekit` too?

packages/frameworks-sveltekit/src/lib/index.ts
ndom91: Yeah tbh I think thats fine too, will be cleaner in the end, right

ndom91: Removed the new method and made the new behavior behind `getSession` :+1:

ndom91:
Hmm so the `build` step is failing for the `@auth/drizzle-adapter` apparently, but I can't find any similar code issues in my locally checked out copy of htis branch and the build script runs for me locally without issue as well :thinking: 

image

Is it working for yuo locally?

packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -209,6 +209,29 @@ import { env } from "$env/dynamic/private"
...
+    const camelCaseName = camelize(name.trim());
+    if (camelCaseName === 'expires') {
+      const expires = new Date(value);
+      return { ...accumulator2, expires };
+    }
+    return { ...accumulator2, [camelCaseName]: value ?? true };
+  }, { path: '/' });
+
+  return { [name]: { value, options }, ...accumulator }
+}, {} as Record<string, SetCookie>);
```
balazsorban44, line 234: `cookie` has `parse()` already: https://www.npmjs.com/package/cookie#cookieparsestr-options

packages/frameworks-sveltekit/package.json
```diff
@@ -48,6 +48,8 @@
     "@auth/core": "workspace:*"
   },
   "peerDependencies": {
+    "cookie": "0.6.0",
+    "@types/cookie": "0.6.0",
```
balazsorban44, line 52: should not be a peer dependency, we don't want the user to need to install this.

balazsorban44 (changes requested):
I don't like the added work for the user.

A better API might be to wrap `load` with `getSession` instead (maybe rename it to `auth()` to align with `next-auth` too) so it can set the cookies _after_ the user logic has run, merging the results.

This is similar to what we do here: https://github.com/nextauthjs/next-auth/blob/a05872d2245265e7179f5ddc2edfaa9ebcf7bad8/packages/next-auth/src/lib/index.ts#L206-L210


For the user, ideally, the `+layout.server.ts` file could just be `export const load = auth()`. :thinking:

packages/frameworks-sveltekit/package.json
ndom91: Yeah these are deps of `@auth/core` already so i wasn't sure where to put them here exactly. Since folks are almost always installed the two together, know what i mean

ndom91:
Hmm yeah i can test that, i have a sveltekit project with auth.js currently WIP. 

But the wrapping load strategy feels a bit un-sveltekit like, based on the little experience i have at least. I'd be curious what a more seasoned svelte person would say about this 🤔

packages/frameworks-sveltekit/src/lib/index.ts
ndom91: Addressed :+1:

ThangHuuVu (commented):
I agree with @balazsorban44 that a wrapper for `load` could be very nice, it resemble our approach for `next-auth` which is simple usages everywhere. If we go this route, we might want to call it `serverLoad` or something, and reserve the `auth` method just to get session & rotate session expiry, similar to `next-auth` - this helps if the developer want to do "advance initialization" based on the `event` params.

wobsoriano:
Yeah, I think a wrapper for the `load` function is still svelte-like. For reference, Rich suggested using a higher-order function in this [issue](https://github.com/sveltejs/kit/issues/2979#issuecomment-1218522382).

ndom91:
Ah okay good. So how should we proceed with this PR? 

The way i see it, we could: 

1. Merge it more or less as-is, so that the preexisting `getSession` is a bit more useful for the time being and follow up later with more significant api changes 
2. Close this PR and start developing the load wrapper function apis directly

What do y'all think?

aakash14goplani:
> Ah okay good. So how should we proceed with this PR?
> 
> The way i see it, we could:
> 
> 1. Merge it more or less as-is, so that the preexisting `getSession` is a bit more useful for the time being and follow up later with more significant api changes
> 2. Close this PR and start developing the load wrapper function apis directly
> 
> What do y'all think?

Hey there, I have been using SvelteKit with SvelteKitAuth for sometime and I believe option 1 would be good as this bug is not just limited with Refresh Token. If I want session object to be updated from client to server, that is not happening as session-token cookie is not getting updated. Hence this change will resolve 2 issues - token rotation + a  way to update session object from client to server

ndom91:
Yeah I think I lean more towards option 1 as well. We can get this in now and then think further about hwo we want to implement the load fn wrapper and what the API should look like, etc., etc. 

Cool with everyone?

balazsorban44:
Let's merge with a `unstable_` prefix then. :+1: but IMO it needs to go at the end. it should "just work"

ndom91:
> Let's merge with a `unstable_` prefix then. 👍 but IMO it needs to go at the end. it should "just work"
> 
> (build needs to pass first)

Okay so we had amended the pre-existing `getSession` method, you're saying add it as a new method then? So we have `getSession()` and `unstable_getSessionWithCookies()` or something like that?

stalkerg:
Just in case, I don't think "updated from client to server" is a real issue. A decorator can be better. 
btw composition can be more flexible, decorators (wrapper) usually is less flexible. Current getSession should be keep as is I believe.

ndom91:
Feature's been merged via https://github.com/nextauthjs/next-auth/pull/9694 :pray:

aakash14goplani:
@ndom91 - can you please update documentation to specific how to use new method and how it is different then existing get-session method. Thanks!

ndom91:
@aakash14goplani Yeah we're working on a new version of the docs and it'll definitely be in there.

For now, the old method is still available, although marked ~deprecated~. The usage of the `auth()` method is the same as the `getSession` one, the difference is that it sets the cookies from the auth response before returning the session body, keeping expiry updated.

PR #8354, opened 2023-08-18
feat(frameworks): add @auth/express

rexfordessilfie:
## ☕️ Reasoning
The `Auth` handler from `@auth/core` is built around the standard Web API Request and Response types. However the Express framework is built on its own Request-Response types. This PR introduces a translation layer between the two APIs to allow using  Auth.js within Express. It is based on this [demo](https://github.com/rexfordessilfie/authjs-express-demo). It was worked on together with @hillac!

Specifically the changes are as follows:

### Implementation
* Introduce a `toWebRequest` helper that converts an `ExpressRequest` into a web `Request`
* Introduce a `toExpressResponse` helper that converts a web `Request` into an `ExpressResponse`
* Introduce an `ExpressAuth(authConfig)` initializer which returns a callback function (`(req: ExpressRequest, res: ExpressResponse) => Promise<void>`) to fulfill authentication requests. Under the hood, it:
    1. It calls `toWebRequest(req)` to get a web `Request`
    2. Forwards the web request to `Auth` (from `@auth/core`) to get back a web `Response`
    3. Forwards the web response along with Express' `res` to `toWebResponse(response, res)` to respond to fulfill the request

### Tests
* Tests the `toWebRequest` and `toExpressResponse` helpers to ensure that they forward all headers, and body and in the right format (depending on content-type)
* Playwright e2e tests have been [added](https://github.com/rexfordessilfie/next-auth/tree/auth-express-dev-app/apps/dev/express/e2e) on a separate branch [`rexfordessilfie:auth-express-dev-app`](https://github.com/rexfordessilfie/next-auth/tree/auth-express-dev-app). (NB: I placed them in a separate branch to not make this one too big!)

### Documentation
I added documentation to `frameworks-express/src/index.ts`, and then also in `docs/docs/reference/express` with a preview deployment [here](https://auth-docs-n89fjspo3-authjs.vercel.app/reference/express). 

### Demo
* A demo has been worked on on [`rexfordessilfie:auth-express-dev-app`](https://github.com/rexfordessilfie/next-auth/tree/auth-express-dev-app/apps/dev/express), which we hope to add in after this one goes through.
* UPDATE: deployment link: https://authjs-express-dev-app.onrender.com (you can login with any credentials)

## Questions/Notes
* For the request body to be translated correctly, the express router/app must use body-parsers for parsing both JSON and URL-encoded bodies. For example: `app.use(express.json())`, and `app.use(express.urlencoded())`
[...]

PR #9570, opened 2024-01-07
chore(example): add @auth/express example app

rexfordessilfie:
## ☕️ Reasoning

This PR adds an example application showcasing `@auth/express` package, as a follow up from https://github.com/nextauthjs/next-auth/pull/8354.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources

## Notes/Questions
* A similar example has been previously depoyed [here](https://authjs-express-dev-app.onrender.com/) on Render. Any login credentials work.
* The README.md for the example suggests deployment to Render but I am happy to update it to another hosting provider, or none if not appropriate for this
* It also has a section on Node.js compatibility, which I added from my experience working on the package and experimenting with different Node versions.
* The example mirrors closely the implementation of SvelteKit's!
* The Express logo used in the README.md is from https://i.cloudup.com/zfY6lL7eFa-3000x3000.png. I found it had better contrast in dark mode, compared to the Wikepedia commons version as is used in SvelteKit's example README.

## Demo
https://github.com/nextauthjs/next-auth/assets/43508242/07f03555-cec8-4d75-93b0-159933bb6577


apps/examples/express/src/app.ts
```diff
@@ -0,0 +1,72 @@
...
+
+// Parse incoming requests data
+app.use(express.urlencoded({ extended: true }))
+app.use(express.json())
+
+// Set session in res.locals
+app.use(currentSession)
+
+// Set up ExpressAuth to handle authentication
+app.use("/api/auth/*", ExpressAuth(authConfig))
```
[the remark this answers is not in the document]
balazsorban44: Good point by Code scanning. If you host on serverless platforms, this is more or less taken care of. Since people might host this elsewhere, it would be a good idea to suggest a rate limiter/remind them

apps/examples/express/src/config/auth.config.ts
```diff
@@ -0,0 +1,38 @@
...
+  providers: [
+    GitHub({
+      clientId: process.env.AUTH_GITHUB_ID,
+      clientSecret: process.env.AUTH_GITHUB_SECRET,
+    }),
+    Google({
+      clientId: process.env.AUTH_GOOGLE_ID,
+      clientSecret: process.env.AUTH_GOOGLE_SECRET,
+    }),
+    Credentials({
```
balazsorban44, line 15: let's drop the credentials provider from this example

apps/examples/express/src/middleware/auth.middleware.ts
balazsorban44, line 1: I feel like this whole file should come from the `@auth/express` package.

apps/examples/express/src/app.ts
```diff
@@ -0,0 +1,72 @@
...
+
+// Set up views engine and path
+// @ts-ignore (See https://stackoverflow.com/questions/45342307/error-cannot-find-module-pug)
+app.engine("pug", pug.__express)
+app.set("views", path.join(__dirname, "../views"))
+app.set("view engine", "pug")
+
+// Trust Proxy for Proxies (Heroku, Render.com, etc)
+// https://stackoverflow.com/questions/40459511/in-express-js-req-protocol-is-not-picking-up-https-for-my-secure-link-it-alwa
+app.enable("trust proxy")
```
balazsorban44, line 32: Auth.js already has a `trustHost: boolean` option in its config. Isn't that sufficient?

balazsorban44 (dismissed):
Thanks, added a first round of review!

rexfordessilfie (commented):
Thank you @balazsorban44 for the initial review! I have addressed majority of the comments, and I am now attempting to get a working Vercel-deployed version of this! In the meantime I would appreciate any further feedback on the improvements! Thank you. 

Also thanks to @hillac for his comments re: express' trust proxy. His explanation is spot on to the experience I had deploying the example app on render.

apps/examples/express/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "build:watch": "pnpm build:ts -w & pnpm build:css -w",
+    "dev": "concurrently -k -p \"[{name}]\" -n \"Build,Node\" -c \"cyan.bold,green.bold\" \"pnpm run build:watch\" \"pnpm run dev:node\"",
+    "dev:node": "nodemon --experimental-fetch --experimental-global-webcrypto dist/server.js",
+    "lint": "eslint src/*.ts --fix",
+    "prettier": "prettier src/*.ts --write"
+  },
+  "author": "Rexford Essilfie <rexfordessilfie09@gmail.com>",
+  "license": "MIT",
+  "dependencies": {
+    "@auth/core": "workspace:*",
```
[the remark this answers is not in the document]
rexfordessilfie: Update: I was able to access the providers on `@auth/express` 👍

rexfordessilfie:
@balazsorban44, got an exciting update. I have updated the example to be deployable as serverless function on Vercel, just like the other examples! You can find a deployed example here: https://vercel-auth-express-example.vercel.app/ ! 🎉 

I had a couple follow up questions to callout for if you had opinions on!

First is regarding the deployment:
* I noticed that the v20 Node engine/runtime is currently [in Beta mode](https://vercel.com/docs/functions/serverless-functions/runtimes/node-js#default-and-available-versions) on Vercel. Should we still set the node engine to `>=20`? Or might we set to 18 and use the experimental webcrypto flag OR polyfill by default?

Second is regarding the middleware helpers in the example:
* Do you think we can introduce the helpers inside of [`src/middleware/auth.middleware.ts`](https://github.com/nextauthjs/next-auth/pull/9570/files#diff-6c9af75558c5ba2945dcd1344b81ba417b6faba17141b4adee907ee61b0aec7e) as standard helper functions of the library in a follow-up PR? I'd like to give a bit more thought to the API: specifically for `authenticatedUser`. 
  Should we make the response more standard (e.g `{ code: 'NOT_AUTHENTICATED', success: false }`), or even expose a callback for the user to decide what to do next if not authenticated?

apps/examples/express/src/middleware/auth.middleware.ts
rexfordessilfie: @balazsorban44 please do you think we can address this in a follow-up PR? I can open the PR and then we decide which should go in first.

ndom91:
@rexfordessilfie I'm having trouble getting this to work. I added an `AUTH_SECRET` and both Github env vars and ran the `dev` npm script. This is what I get when I select "Github" on the auto-generated auth.s signin screen.

Am I missing something? 

image

rexfordessilfie:
> @rexfordessilfie I'm having trouble getting this to work. I added an `AUTH_SECRET` and both Github env vars and ran the `dev` npm script. This is what I get when I select "Github" on the auto-generated auth.s signin screen. Using `node@20.9.0`
> 
> 
> 
> Am I missing something? 
> 
> 
> 
> image
> 
> 

Hmm it doesn't sound like you're doing anything wrong, @ndom91. Let me try to reproduce the bug and then get back to you with what I find. Will be back in a couple of hours!

rexfordessilfie:
Okay, I have managed to reproduce the issue @ndom91. I reproduced mine when I had an invalid client secret! My hunch is that the OAuth provider provides a response with an empty `access_token`. Could you double-check the client id and secret you used for your environment variables?

Screenshot 2024-01-14 at 18 35 07

ndom91:
@rexfordessilfie I had copied them over from another auth.js app I'm currently working on so I thought they were good, but turns out there was indeed somehow a typo in the last letter of the secret :joy:

Anyway, everythings working now!

Hope you don't mind, I added the sync config to sync this out of the main repo here into `nextuathjs/express-auth-example`, from which we deploy the example :+1:

ndom91 (approved):

rexfordessilfie:
That's great! Glad we could figure it out 😄! I also tend to copy paste the vars across local projects and copied the wrong one once too.

Thanks for adding the sync script. I was wondering how that worked with the deployed examples.

Also in my experimentation I found that using a callback url (http://localhost:3000/auth/callback/github) but without the path worked. And for Express, it worked even with a mismatching port! I'll double check and open an issue and PR to resolve if necessary after more investigation.

Thanks to you and @balazsorban44  for support in getting this in!

rexfordessilfie:
> Also in my experimentation I found that using a callback url (http://localhost:3000/auth/callback/github) but without the path worked. And for Express, it worked even with a mismatching port! I'll double check and open an issue and PR to resolve if necessary after more investigation.

Okay @ndom91, I narrowed this down to being provider-specific behavior. When using the Google provider, the request fails right after clicking the `Sign in with Google` button due to a `redirect_uri_mismatch` error when there is a port mismatch between where the server is hosted and the URL configured in Google's OAuth settings.

However, the GitHub provider actually lets the request through all the way even with a mismatching port. I am hosting the server on `http://localhost:3000` but can set the `Authorization callback url` to any of the following: `http://localhost` or `http://localhost:8080` or `http://localhost:3000/api/auth/` or `http://localhost:/api` even. Kinda weird, but points to not being an implementation error on our end.

PR #9586, opened 2024-01-09
Encode error as URIComponent

warent-bottleneck:
## ☕️ Reasoning

Certain unexpected backend errors that throw may include invalid URI characters. When that happens, the entire error page crashes with an opaque complaint `TypeError [ERR_INVALID_CHAR]: Invalid character in header content ["Location"]`. This ensures that the error is encoded properly so unexpected errors are still rendered.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/pages/index.ts
+++ b/packages/core/src/lib/pages/index.ts
@@ -127,7 +127,7 @@ export default function renderPage(params: RenderPageParams) {
         return {
           redirect: `${pages.error}${
             pages.error.includes("?") ? "&" : "?"
-          }error=${error}`,
+          }error=${encodeURIComponent(error)}`,
           cookies,
         }
       }


balazsorban44 (changes requested):
Thanks, the real solution to this would be to never end up on the error page with unexpected errors though! It sounds like we are missing catching some errors. It likely should end up as `error=Configuration` here, and the actual error should be logged in your terminal. Could you open a bug report with the code that reproduced this error so we can add better/safer error-handling?

warent-bottleneck:
> Thanks, the real solution to this would be to never end up on the error page with unexpected errors though! It sounds like we are missing catching some errors. It likely should end up as `error=Configuration` here, and the actual error should be logged in your terminal. Could you open a bug report with the code that reproduced this error so we can add better/safer error-handling?

I confirmed this only occurs in `^4.22.1`. Cannot reproduce in the new v5 beta

PR #9638, opened 2024-01-14
feat: support advanced initialization in v5

### LINKED ISSUE #9455: Advanced initialization with Auth.Js 5

Missing documentation on advanced initialization for the latest version of Next Auth.

I am trying to specify session max age based on request data. I need to set max to one day session if user is authenticating using only password auth, but if user is having account then longer. Now as per documentation I need to use advanced initialization, but as per latest version of Next Auth, it is not possible to pass req and res to NextAuth.
How to achieve this with latest version?

https://authjs.dev/guides/basics/initialization#advanced-initialization

ThangHuuVu:
## ☕️ Reasoning

Add support for advanced initialization for v5. 

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: #9455 

## 📌 Resources


packages/next-auth/src/index.tsx
```diff
@@ -45,6 +45,26 @@
...
+ *  if (req) {
+ *   console.log(req) // do something with the request
+ *   return { providers: [ GitHub ] }
+ *  }
+ *  return { providers: [ GitHub ] }
+ * })
+ * ```
+ *
+ * :::tip
+ * This is useful if you want to customize the configuration based on the request, for example, to add a different provider based on the request URL's `searchParams`.
```
balazsorban44, line 65:
```suggestion
 * This is useful if you want to customize the configuration based on the request, for example, to add a different provider in staging/dev environments.
```

packages/next-auth/src/index.tsx
```diff
@@ -61,8 +81,12 @@ import type {
   NextApiResponse,
 } from "next"
 import type { AppRouteHandlerFn } from "next/dist/server/future/route-modules/app-route/module.js"
-import type { NextRequest } from "next/server"
-import type { NextAuthConfig, NextAuthRequest } from "./lib/index.js"
+import { NextRequest } from "next/server"
```
balazsorban44, line 84:
```suggestion
import type { NextRequest } from "next/server"
```

balazsorban44 (changes requested):

balazsorban44 (approved):

chrisivo:
Is there any way lazy initialisation can be executed asynchronously? I need to make async calls to obtain the Id's and secrets for the providers. I could do this using v4, but seems like there's no support in v5 beta (so far)?

ndom91:
@chrisivo what do you mean exactly? The first argument to `NextAuth()` can be an async function. So yuo can do this (as is shown in most examples):

```ts
export const { auth, signIn, signOut } = NextAuth({
  providers: [ ... ],
  debug: true
})
```

But yuo can also do this:

```ts
export const { auth, signIn, signOut } = NextAuth(async (req) => {
  return {
    providers: [ ... ],
    debug: true
  }
))
```

chrisivo:
Thanks for the response ndom91. Is there any roadmap at all for introducing asynchronous functions when initialising in `next-auth` ?

maiconcarraro:
Facing same situation from @chrisivo, need to generate secret for apple sign in during initialization :(

nicolaspernoud:
Hello, 
Same here, we need to get provider configuration from prisma (async) ...

ThangHuuVu:
@chrisivo @maiconcarraro @nicolaspernoud I opened a new PR to add support for asynchronous lazy init - thanks for flagging the issue

PR #9646, opened 2024-01-15
fix(types): Session callback params does not include `user` and `token`

### LINKED ISSUE #9633: Type error in session callback for `user` and `token` properties
### Environment

npmPackages:
@auth/core: ^0.* => 0.20.0
@auth/drizzle-adapter: ^0.* => 0.3.14
@auth/sveltekit: ^0.* => 0.5.2
Node: 20.3.1 - /opt/homebrew/bin/node

### Reproduction URL

https://github.com/thomasmol/authjsbugs

### Describe the issue

related to #9437

i get a type error in the session callback for the `user` or `token` property:
```
Property 'user' does not exist on type '({ session: Session; user: AdapterUser; } | { session: Session; token: JWT; }) & { newSession: any; trigger?: "update" | undefined; }'.
```

error claims `user` does not exist, while it does, although it depends on the database strategy used.

error located here on the `user` property in `hooks.server.ts`:
```typescript 
callbacks: {
  session: async ({ session, user }) => {
    if(session.user){
      session.user.id = user.id;
      }
    return session;
  }
},
```
not sure if useful but this is copilot explaining:
"The error message you're seeing is due to TypeScript not being able to determine the exact type of the object that's being passed to the session callback. The type of the object is a union type, which means it could be one of several types. In this case, it could be { session: Session; user: AdapterUser; } or { session: Session; token: JWT; }.

The user property exists on the first type, but not on the second. When you try to access user.id, TypeScript can't guarantee that user will always be present, hence the error."

### How to reproduce

install auth core + auth sveltekit, create hooks.server.ts and add a the session callback. type error occurs on user and token, but not the session property.

### Expected behavior

no type errors

nbifrye:
## ☕️ Reasoning

In session callback, argument type does not contain user and token

```typescript
callbacks: {
  session: async ({ session, user }) => {
    if(session.user){
      session.user.id = user.id;
    }
    return session;
  }
},
```

```
Property 'user' does not exist on type '({ session: Session; user: AdapterUser; } | { session: Session; token: JWT; }) & { newSession: any; trigger?: "update" | undefined; }'.
```

This seems to be related to https://github.com/Microsoft/TypeScript/issues/14094. Look at this [TS Playground example](https://www.typescriptlang.org/play?#code/C4TwDgpgBA8gRgKwIxQLxQN5QIYC4oDOwATgJYB2A5gDRRz5FlVQC+AUKJLIgExqY4GJCjSgBjIU0qs2bMQHtyRKADMAruTH8AFNmKUC+eMigAfbgh4BKNAD5MbKE5z6CAOjj8ARF7bsgA).

```typescript
type Obj1 = { a: string, b: string }
type Obj2 = { a: string, c: string }

const func = (args: Obj1 | Obj2) => {
    args.b = ""
         ^
         Property 'b' does not exist on type 'Obj1 | Obj2'.
           Property 'b' does not exist on type 'Obj2'.
}
```

Using `{ session, user?, token? }` instead of `{ session, user } | { session, token }` seems to fix this.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

 

Fixes: https://github.com/nextauthjs/next-auth/issues/9633

## 📌 Resources

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -248,18 +248,13 @@ export interface CallbacksOptions<P = Profile, A = Account> {
    * @see [`jwt` callback](https://authjs.dev/reference/core/types#jwt)
    */
   session: (
-    params: (
-      | {
-          session: Session
-          /** Available when {@link AuthConfig.session} is set to `strategy: "database"`. */
-          user: AdapterUser
-        }
-      | {
-          session: Session
-          /** Available when {@link AuthConfig.session} is set to `strategy: "jwt"` */
-          token: JWT
-        }
-    ) & {
+    params: {
+      session: Session
+      /** Available when {@link AuthConfig.session} is set to `strategy: "database"`. */
+      user?: AdapterUser
+      /** Available when {@link AuthConfig.session} is set to `strategy: "jwt"` */
+      token?: JWT
+    } & {
       /**
        * Available when using {@link AuthConfig.session} `strategy: "database"` and an update is triggered for the session.
        *
--- a/packages/next-auth/src/lib/index.ts
+++ b/packages/next-auth/src/lib/index.ts
@@ -76,7 +76,6 @@ async function getSession(headers: Headers, config: NextAuthConfig) {
         const session =
           // If the user defined a custom session callback, use that instead
           (await config.callbacks?.session?.(...args)) ?? args[0].session
-        // @ts-expect-error either user or token will be defined
         const user = args[0].user ?? args[0].token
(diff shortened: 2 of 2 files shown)


aakash14goplani:
Same case with **SvelteKitAuth**.
image

cc: @ndom91 and @ThangHuuVu

balazsorban44:
Closed via #9756. Needs more work in the future. See https://github.com/nextauthjs/next-auth/pull/9756#pullrequestreview-1842339467

PR #9659, opened 2024-01-17
Correct Cognito callback url typo

18thletter:
## ☕️ Reasoning

I believe there's a typo in the Cognito providers documentation.

The docs say the callback URL is `https://example.com/api/auth/callback/cognito`, but the actual URL that is being passed as the `redirect_uri` does not have the `api/` portion of that URL.

Here's a screenshot of the debugging in my terminal:
Screenshot 2024-01-16 at 10 51 01 PM

Either the callback is incorrect, or the docs are incorrect.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/src/providers/cognito.ts
+++ b/packages/core/src/providers/cognito.ts
@@ -24,7 +24,7 @@ export interface CognitoProfile extends Record<string, any> {
  *
  * #### Callback URL
  * ```
- * https://example.com/api/auth/callback/cognito
+ * https://example.com/auth/callback/cognito
  * ```
  *
  * #### Configuration


balazsorban44:
This depends on your setup/Auth.js integration of choice. For example if you use NextAuth.js, the `/api` portion is the documented path, but you can configure it in either case.

18thletter:
😂 Total skill issue here then.

However, I still think the docs should reflect this. It should say something like, "Depending on your setup, the callback URL can be different. It can also be configured through [link to configuration parameters]".

PR #9664, opened 2024-01-17
refactor(providers): rebranded Azure AD to Microsoft Entra

alessandrojcm:
## ☕️ Reasoning

Azure Active Directory is now [Microsoft Entra ID](https://learn.microsoft.com/es-es/entra/fundamentals/new-name), this is just a rebrand; no underlying configuration or functionality is affected. This PR simply swaps the names, adds the new logo, and updates the documentation links.

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

 
N/A


balazsorban44 (changes requested):
Thanks! Can we keep the original file that just re-exports everything from the new file and adds the `@deprecated` JSDoc comment?

Good example on how to do it

https://github.com/nextauthjs/next-auth/blob/d85269d1169d068448b831d2ee9517e4673c78b8/packages/core/src/providers/email.ts#L10-L23

panva:
I think there should be two distinct Azure providers

One where you know the tenant you're working with and you're NOT using any of the "common" issuers (the existing one).

And then another one which can be used with one of the "common" issuers

```
  'https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration'
  'https://login.microsoftonline.com/organizations/v2.0/.well-known/openid-configuration'
  'https://login.microsoftonline.com/consumers/v2.0/.well-known/openid-configuration'
```

This one has the workarounds in and requires explicit configuration to

- allow login from all tenants
- allow login from specific tenants (by providing a list of tenant ids)

alessandrojcm:
@balazsorban44 like this?

@panva I think that probably should be a separate PR, happy to do it if you open an issue about it.

panva:
> @panva I think that probably should be a separate PR, happy to do it if you open an issue about it.

Yeah, apologies, I added my comment to the wrong PR :) (it was meant [here](https://github.com/nextauthjs/next-auth/pull/9718#issuecomment-1909893145))

Issue #8491, opened 2023-09-04
`httpOptions` missing on OAuth config in experimental version


patrickpilch:
### Description 📓

We have been using the [`httpOptions`](https://github.com/nextauthjs/next-auth/blob/65043ba4718b293c4a283e3aee7d7e3340e921fa/packages/next-auth/src/providers/oauth.ts#L144) parameter in next-auth's 4.x `OAuthConfig` to set a header we require for our IDP.

I noticed this is missing in @auth/core's [`OAuth2Config`](https://github.com/nextauthjs/next-auth/blob/65043ba4718b293c4a283e3aee7d7e3340e921fa/packages/core/src/providers/oauth.ts#L110), and unfortunately will be a blocker/headache for us to upgrade. 

It seems the 'oauth4webapi' library [supports additional headers](https://github.com/panva/oauth4webapi/blob/main/docs/interfaces/HttpRequestOptions.md#headers).

### How to reproduce ☕️

N/A

### Contributing 🙌🏽

Yes, I am willing to help implement this feature in a PR

balazsorban44:
Can you add a more concrete example? N/A is not a valid reproduction. 🤷‍♂️

patrickpilch:
Usage on 4.x
```typescript
Keycloak({ 
  // ...
  httpOptions: {
    headers: {
      'some-header-name': 'some-header-value'
    }
  }
}),
```

Usage on 5.x
Not possible. The config type is missing the `httpOptions` parameter, or something similar.

As for the reasoning and use-case: there could be all sorts. One concrete example for custom header support would be if the authorization server is behind a security proxy and requires a service token or other authorization header(s). Of course there are other ways of addressing this scenario with varying levels of complexity, but using headers was a very simple & effective way to get going. Let me know if there's anything else I can provide.

PaulWild:
This will be a blocker for us as well unfortunately from upgrading - our use case is similar in that we require additional httpHeaders on the requests made to our OAuth provider

balazsorban44:
`oauth4webapi` recently added https://github.com/panva/oauth4webapi/blob/main/docs/variables/experimental_customFetch.md which might be useful, we will revisit!

bujikun:
I am facing this similar issue. I can not extend the timeout during debugging.
I had this in v4, after upgrading to v5, i can no longer do the same.
`
httpOptions: {
      timeout:15000
},`

Can we please have this back? Or another way of doing this?

leog:
[...]

PR #9675, opened 2024-01-18
fix: should only merge plain objects

g1eny0ung:
## ☕️ Reasoning

Refs: #2509, https://github.com/nextauthjs/next-auth/discussions/3944.

I got the info from https://github.com/nextauthjs/next-auth/discussions/3944#discussioncomment-7042533, which points to the `merge` function as the reason for the loss of some constructed object types, for example, the `Agent` type (`new HttpsProxyAgent()`) will be wrongly overwritten as `object`.

This stack overflow answer doesn't consider more boundary cases, we should let it only merge pure objects and let constructed objects or the native objects overwrite directly, thus preserving its original type.

So in this PR, I use a tool library `is-plain-obj` to replace the existing `isObject` function to resolve this problem. I manually tested with:

```js
GithubProvider({
  clientId: process.env.GITHUB_ID as string,
  clientSecret: process.env.GITHUB_SECRET as string,
  httpOptions: {
    timeout: 10000, // Increase timeout to 10 seconds.
    ...(process.env.http_proxy && {
      // Use a proxy agent if a proxy is defined.
      agent: new HttpsProxyAgent(process.env.http_proxy as string),
    }),
  },
})
```

in my Next.js project, and seems it works.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -63,8 +63,9 @@
   "license": "ISC",
   "dependencies": {
     "@panva/hkdf": "^1.1.1",
-    "cookie": "0.6.0",
     "@types/cookie": "0.6.0",
+    "cookie": "0.6.0",
+    "is-plain-obj": "^4.1.0",
     "jose": "^5.1.3",
     "oauth4webapi": "^2.4.0",
     "preact": "10.11.3",
--- a/packages/core/src/lib/utils/merge.ts
+++ b/packages/core/src/lib/utils/merge.ts
@@ -1,22 +1,20 @@
-// Source: https://stackoverflow.com/a/34749873/5364135
-
-/** Simple object check */
-function isObject(item: any): boolean {
-  return item && typeof item === "object" && !Array.isArray(item)
-}
+import isPlainObject from 'is-plain-obj';
 
 /** Deep merge two objects */
 export function merge(target: any, ...sources: any[]): any {
   if (!sources.length) return target
   const source = sources.shift()
 
-  if (isObject(target) && isObject(source)) {
+  if (isPlainObject(target) && isPlainObject(source)) {
     for (const key in source) {
-      if (isObject(source[key])) {
-        if (!target[key]) Object.assign(target, { [key]: {} })
+      if (isPlainObject(source[key])) {
+        if (!target[key]) {
+          target[key] = {}
+        }
+
         merge(target[key], source[key])
       } else {
-        Object.assign(target, { [key]: source[key] })
+        target[key] = source[key]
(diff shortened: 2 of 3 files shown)


g1eny0ung:
@ThangHuuVu @balazsorban44 Can you guys approve to get workflows running? I've only run some of the tests locally for network reasons, so I'd like to see the results of the CI run.

g1eny0ung:
@balazsorban44 But version `v4` should be considered a current stable release, shouldn't we be working on bug fixes? Looking at the release and documentation updates, I don't see the new core being released anytime soon. The tutorial of Next.js OAuth is still using the `next-auth` package (https://authjs.dev/getting-started/providers/oauth-tutorial). 

And PR #8491 only indicates the absence of `httpOptions` in the new core client implementation, I don't see any correlation with this PR?

g1eny0ung:
Below is the patched `utils/merge.js`, I just copied the main function in https://github.com/sindresorhus/is-plain-obj/blob/main/index.js to prevent install the package:

### Commands

```sh
pnpm patch next-auth
# After the modification
pnpm patch-commit <path>
```

### merge.js

```js
"use strict";

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.merge = merge;

function isPlainObject(value) {
	if (typeof value !== 'object' || value === null) {
		return false;
	}

	const prototype = Object.getPrototypeOf(value);
	return (prototype === null || prototype === Object.prototype || Object.getPrototypeOf(prototype) === null) && !(Symbol.toStringTag in value) && !(Symbol.iterator in value);
}

function merge(target, ...sources) {
  if (!sources.length) return target;
  const source = sources.shift();

  if (isPlainObject(target) && isPlainObject(source)) {
    for (const key in source) {
      if (isPlainObject(source[key])) {
        if (!target[key]) Object.assign(target, {
          [key]: {}
        });
        merge(target[key], source[key]);
      } else {
        target[key] = source[key];
      }
    }
  }

  return merge(target, ...sources);
}
```

flash4174:
Hi All,

I get the error 'this.createSocket is not a function'

Any idea? I've tried both client.js and merge.js edit and also both the same time

g1eny0ung:
> Hi All,
> 
> I get the error 'this.createSocket is not a function'
> 
> Any idea? I've tried both client.js and merge.js edit and also both the same time

🤔 No ideas. I used `"next-auth": "^4.24.5"` and now it works as normal.

PR #9677, opened 2024-01-18
docs: fix the link to Yandex provider options in yandex.md

BaryshevRS:
## ☕️ Reasoning

Fix the link to Yandex provider options.

https://next-auth.js.org/providers/yandex#options

## 🧢 Checklist

+ [x] Documentation

### DIFF
--- a/docs/docs/providers/yandex.md
+++ b/docs/docs/providers/yandex.md
@@ -15,7 +15,7 @@ https://oauth.yandex.com/client/new
 
 The **Yandex Provider** comes with a set of default options:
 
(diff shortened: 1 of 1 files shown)


ndom91:
Hey thanks for the fix, however can you please update it in the `main` branch instead? 

The docs/link should be in the `packages/core/src/providers/yandex.ts` file

BaryshevRS:
> Hey thanks for the fix, however can you please update it in the `main` branch instead?
> 
> The docs/link should be in the `packages/core/src/providers/yandex.ts` file

The main branch in the `packages/core/src/providers/yandex.ts` file is fine. This is only in the current version of the documentation. 

If I understand correctly.

PR #9702, opened 2024-01-21
feat: add experimental `AuthData` type

balazsorban44:
- The `auth()` method will return the `Session` object, as defined via `session` callback, consistent with `useSession()`
- `experimental.structuredAuth` flag
- `unstable_auth()` and `unstable_update()` that always require an `AuthData` type. This so that we can reliably work with these data structures (eg.: update data in DB)

### DIFF
--- a/packages/next-auth/src/index.tsx
+++ b/packages/next-auth/src/index.tsx
@@ -71,6 +71,7 @@ import { Auth } from "@auth/core"
 import { reqWithEnvUrl, setEnvDefaults } from "./lib/env.js"
 import { initAuth } from "./lib/index.js"
 import { signIn, signOut, update } from "./lib/actions.js"
+import { initFutureAuth } from "./lib/future/index.js"
 
 import type { Session } from "@auth/core/types"
 import type { BuiltInProviderType } from "@auth/core/providers"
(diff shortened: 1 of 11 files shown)


aakash14goplani:
@balazsorban44 - will this change be limited to Next.js or could be used with other frameworks as well?

balazsorban44:
first, I'm gonna test it out with Next.js, then when it's ready, we can move this into other framework integrations too!

packages/next-auth/src/index.tsx
```diff
@@ -384,11 +400,15 @@ export default function NextAuth(
...
       },
-      unstable_update: (data) => {
+      update: (data) => {
         const _config = config(undefined)
         setEnvDefaults(_config)
         return update(data, _config)
       },
+      unstable_auth: initFutureAuth(config, setEnvDefaults),
+      unstable_update: () => {
+        throw new TypeError("Not yet implemented.")
```
ndom91, line 410: Quick note, I'd make a mention here of the rename of the previous `unstable_update` to `update`. Just for anyone whos already potentially using it. If they start seeing `"Not yet implemented"` errors, they'll be very confused.

Maybe something like, `The previous implementation has been renamed 'update', this method is now not implemented` :thinking:

PR #9705, opened 2024-01-21
fix(providers): Add email to VK provider

BaryshevRS:
## ☕️ Reasoning

Add email to the VK provider. Now `undefined` isn't saved in the database. 

Implemented similarly to github provider.

### DIFF
--- a/packages/core/src/providers/vk.ts
+++ b/packages/core/src/providers/vk.ts
@@ -10,7 +10,7 @@
  */
 import type { OAuthConfig, OAuthUserConfig } from "./index.js"
 
-// https://dev.vk.com/reference/objects/user
+/** https://dev.vk.com/reference/objects/user */
 export interface VkProfile {
   id: number
   first_name: string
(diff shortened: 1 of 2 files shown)


packages/core/src/providers/vk.ts
```diff
@@ -10,285 +10,285 @@
  */
 import type { OAuthConfig, OAuthUserConfig } from "./index.js"
 
+// https://dev.vk.com/reference/objects/user
```
balazsorban44, line 13:
```suggestion
/** https://dev.vk.com/reference/objects/user */
```

packages/core/src/providers/vk.ts
```diff
@@ -372,13 +372,26 @@ export default function VK<P extends Record<string, any> = VkProfile>(
...
+
+        return profile.response[0];
+      },
+    },
+    profile(profile: P) {
       return {
         id: profile.id,
         name: [profile.first_name, profile.last_name].filter(Boolean).join(" "),
-        email: null,
+        email: profile?.email ? profile?.email:  null,
```
balazsorban44, line 394:
```suggestion
        email: profile.email ?? null,
```

balazsorban44 (changes requested):

PR #9714, opened 2024-01-22
feat: add server `signIn` and `signOut` methods

### LINKED ISSUE #7098: Is there a server side version of SvelteKit signIn() / signOut()?

ThangHuuVu:
Closes #7979, fixes #7098

- Update `svelte-package` to v2. REVIEW: now exporting from `dist`. This makes our setup now simpler and dev watch command is working well, too.
- Add example usages


packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -350,3 +465,26 @@ export function setEnvDefaults(envObject: any, config: SvelteKitAuthConfig) {
...
+    const separator = pathname.endsWith("/") ? "" : "/"
+    return new URL(`${origin}${pathname}${separator}${action}`)
+  }
+  const host = h.get("x-forwarded-host") ?? h.get("host")
+  const protocol = h.get("x-forwarded-proto") === "http" ? "http" : "https"
+  // @ts-expect-error `basePath` value is default'ed to "/api/auth" in `setEnvDefaults`
+  const { origin, pathname } = new URL(basePath, `${protocol}://${host}`)
+  const separator = pathname.endsWith("/") ? "" : "/"
+  return new URL(`${origin}${pathname}${separator}${action}`)
+}
```
balazsorban44, line 490: Could simplify a bit. See: https://github.com/nextauthjs/next-auth/blob/30fb90f9159d8375826695351cf0a4cd490885f4/packages/next-auth/src/lib/actions.ts#L113-L129

`process.env` is not available in SvelteKit, so that needs adjusting.

packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -277,6 +363,9 @@ export function SvelteKitAuth(
 
     event.locals.auth ??= () => auth(event, _config)
     event.locals.getSession ??= event.locals.auth
+    event.locals.signIn ??= (provider, options, authorizationParams) =>
+      signIn(provider, options, authorizationParams, _config, event)
+    event.locals.signOut ??= async (options) => signOut(options, _config, event)
```
balazsorban44, line 368: I feel like this should not be part of the `event`. :thinking: How/where would you call these methods?

A more appropriate solution would be to return more than `Handle` here:

https://github.com/nextauthjs/next-auth/blob/70efa75ec2b1b77fa2920d7f4fe031f108e2b9b3/packages/frameworks-sveltekit/src/lib/index.ts#L357

and introduce the concept of `auth.ts` similar to `next-auth`.

packages/frameworks-sveltekit/src/lib/index.ts
```diff
@@ -215,6 +217,90 @@ export type {
   User,
 } from "@auth/core/types"
 
+type SignInParams = Parameters<App.Locals["signIn"]>
+async function signIn(
```
balazsorban44, line 221: Nit: Should we move `signIn` and `signOut` to a different file (maybe actions.ts)?

balazsorban44 (dismissed):

packages/frameworks-sveltekit/src/lib/index.ts
ndom91: If returning more than `Handle` it would force users to make their `src/hooks.server.ts` more complex though. Also if exporting the `signIn`/`signOut` methods from there folks woudl have to `import { signIn } from './src/hooks.server.ts'` wherever they'd want to use it. Feels a bit weird importing from the hooks file. That'd be like importing from the `middleware.ts` to anywhere in your Next.js 13+ app afaik :thinking:

balazsorban44: We would introduce the concept of `auth.ts`, so `hooks.server.ts` would become `export { handle } from "./auth"` similar to `export { auth as middleware } from "auth" in NextAuth.js

ndom91: Ahh right okay that has much better ergonomics

ThangHuuVu: I think we can use it like this, no? 
```js
import type { LayoutServerLoad } from "./$types"

  await event.locals.signIn("credentials", { password: "pw" }) // should handle redirect
}

```

ndom91:
Excited about this! Will give it a spin tonight

packages/frameworks-sveltekit/src/lib/index.ts
ThangHuuVu: So I added some usage examples in the dev app, let me know what you think. In SvelteKit, we can assign to the `locals`property and just use in many context: server only with `+server.ts`, layout load or page load.

package.json
```diff
@@ -19,7 +19,7 @@
     "dev:docs:adapters": "turbo run dev --filter=docs",
     "email": "cd apps/dev/nextjs && pnpm email",
     "eslint": "eslint --cache .",
-    "lint": "prettier --check .",
+    "lint": "prettier --check ./packages",
```
ThangHuuVu, line 22:
```suggestion
    "lint": "prettier --check .",
```

ndom91:
Hmm so I'm not sure what I'm doing wrong, lets hop on a call together maybe next tmr.

But I'm just getting a `500` response from `POST`-ing to my signin url `/login` when I click on the new `<SignIn>` btn. 

I have some additional logging setup server side and I see the req to the authorization URL and the redirect: 

image

But from my perspective as a user I just lcick on the `<SignIn />` btn and immediately get to redirected to my `+error.svelte` page with a `500` in the network dev tools :thinking: 

---

Anyway, besides that I just had some small nits:

1. We have a bit of a challenge with styling the buttons now because there's 2 elements in the way now, the `form` and the `button`, so like the `form` is where we apply that `className` prop, but the `button type="submit"` is also "in the way" of the `slot` element the user would be passing it. 

More concretely, here's the `<form>` element getting the `w-full` class I'm passing in from the <SignIn className="w-full" .../>`:


image


But the `<button type="submit"..` is still in the way of the form and the children components I'm rendering through the slot:

image


Know what I mean? 

I dont have any good ideas off the top of my head for this unfortunately. We could expose an additoinal class prop that we pass onto that button but that feels awkward :/ Maybe for the time being we just do somethign like add a `w-full m-0` to that `<button type="submit"..`, to make it as unobtrusive as possible, what do you think?

2. We can change that `className` prop to `class` via this little trick:
```ts
<script>
  let className = ""
  export { className as class }
</script>

<button type="submit" class={className}><slot /></button>
```

ndom91:
Okay so I simplified the whole thing a bit to try n match what you had. I had previously had the signin server route be the same path where my custom login page was, but now sigin and signout are just at `/sign-in` and `/sign-out` as expected, without anythign else at those routes other than the form action handlers. Still same behaviour unforutnately

ndom91:
@ThangHuuVu okay hope you don't mind, I pushed some small PRs cleaning up the route `sign-in` -> `signin`, it was a bti awkward and `signout` was also not hyphenized. Also added thtat `width 100%` to the `SignInComponent`'s `<button type="submit"..` as mentioned above ^

ndom91:
Okay so disabling my custom signin page seemed to get it all to work :partying_face: 

Going to try to find the exact issue.

apps/dev/sveltekit/src/auth.ts
```diff
@@ -0,0 +1,160 @@
...
+      clientSecret: TWITCH_SECRET,
+    }),
+    Reddit({
+      clientId: REDDIT_ID,
+      clientSecret: REDDIT_SECRET,
+    }),
+    AzureAD({
+      clientId: AZURE_AD_ID,
+      clientSecret: AZURE_AD_SECRET,
+    }),
```
balazsorban44, line 155: let's use the environment inference version here

package.json
```diff
@@ -15,10 +15,10 @@
     "dev": "turbo run dev --parallel --continue --filter=next-auth-app... --filter=@auth/core --filter=!./packages/adapter-*",
     "dev-v4:db": "turbo run dev --parallel --continue --filter=next-auth-app-v4...",
     "dev-v4": "turbo run dev --parallel --continue --filter=next-auth-app-v4... --filter=!./packages/adapter-*",
-    "dev:kit": "turbo run dev --parallel --continue --filter=sveltekit-auth-app...",
+    "dev:kit": "turbo run dev --parallel --continue --filter=sveltekit-auth-app... --filter=@auth/core",
```
balazsorban44, line 18:
```suggestion
    "dev:kit": "turbo run dev --parallel --continue --filter=sveltekit-auth-app...
```

This should not be necessary, since `sveltekit-auth-app` depends on `@auth/sveltekit` which depends on `@auth/core`, so it should already be building.

ndom91 (approved):

aakash14goplani:
@ThangHuuVu - how do we use these methods?

ThangHuuVu:
@aakash14goplani please refer to this section here: https://authjs.dev/reference/sveltekit#signing-in-and-signing-out
let me know if you have any questions!

aakash14goplani:
@ThangHuuVu - the new version has removed `type SvelteKitAuthConfig`, can we add that back please?
image

Issue #8831, opened 2023-10-10
Auth provider for Linkedin not working


andersengenolsen:
### Environment

System:
    OS: Linux 6.2 Ubuntu 22.04.3 LTS 22.04.3 LTS (Jammy Jellyfish)
    CPU: (12) x64 AMD Ryzen 5 3600X 6-Core Processor
    Memory: 2.63 GB / 15.53 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 20.4.0 - /usr/local/bin/node
    npm: 9.7.2 - /usr/local/bin/npm

### Reproduction URL

https://github.com/nextauthjs/next-auth-example

### Describe the issue

As per the documentation from Linkedin, I've set up a new LinkedIn app, and added "Sign In with LinkedIn using OpenID Connect" as a product.

At first I had some problems when not specifying scope.

``
    providers: [
        LinkedIn({
            clientId: process.env.LINKEDIN_ID,
            clientSecret: process.env.LINKEDIN_SECRET,
        })
    ], pages: {
        signIn: '/register-cv'
    },
``

This returns an 'unauthorized_scope_error' for r_emailaddress. Managed to fix that issue by providing scopes as per the documentation from Microsoft:

> Authenticating Members
New members logging in to your service for the first time will need to follow the Authenticating with OAuth 2.0 Guide. When requesting the authorization code in Step 2 of the OAuth 2.0 Guide, make sure you use the OpenID scope openid to get the ID Token. We are also introducing new scopes profile and email.

> openid	Required to indicate the application wants to use OIDC to authenticate the member.
> profile	Required to retrieve the member's lite profile including their id, name, and profile picture.
> email	Required to retrieve the member's email address.
> After successful authentication, you will receive the member's access token and ID token.

> If your application does not have these permissions provisioned, you can request access through the Developer Portal. Select your app from My Apps, navigate to the Products tab, and request the Sign in with LinkedIn using OpenID Connect product.

Link: https://learn.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/sign-in-with-linkedin-v2

Updated scopes to match the documentation from Microsoft:

`LinkedIn({
            clientId: process.env.LINKEDIN_ID,
            clientSecret: process.env.LINKEDIN_SECRET,
            authorization: { params: { scope: 'openid profile email' } }})`
            
Now getting this error:

[...]

Issue #8374, opened 2023-08-21
AzureADProvider does not work with default (common) endpoint


georgy:
### Provider type

Azure Active Directory

### Environment

"next" : N/A
"react": N/A
"@auth/core": "0.12.0"
"@auth/sveltekit": "0.3.1"

### Reproduction URL

N/A (it is a basic setup)

### Describe the issue

I am using simple setup for AzureADProvider. My goal is to allow login with any valid Microsoft identity (I have app registration setup, etc). I am am not providing `tenantId` will default to `common`

* User is ends up on https://localhost:5173/auth/error?error=SignInError
* There is a error in the server's log:
```
[auth][error][SignInError]: Read more at https://errors.authjs.dev#signinerror
[auth][cause]: OperationProcessingError: "response" body "issuer" does not match "expectedIssuer"
    at Module.processDiscoveryResponse (file:///<...skipped...>/node_modules/oauth4webapi/build/index.js:232:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async getAuthorizationUrl (file:///<...skipped...>/node_modules/@auth/sveltekit/node_modules/@auth/core/lib/oauth/authorization-url.js:19:20)
    at async Module.signin (file:///<...skipped...>/node_modules/@auth/sveltekit/node_modules/@auth/core/lib/routes/signin.js:14:20)
    at async AuthInternal (file:///<...skipped...>/node_modules/@auth/sveltekit/node_modules/@auth/core/lib/index.js:115:36)
    at async Proxy.Auth (file:///<...skipped...>/node_modules/@auth/sveltekit/node_modules/@auth/core/index.js:100:30)
    at async Module.respond (/<...skipped...>/node_modules/@sveltejs/kit/src/runtime/server/respond.js:274:20)
    at async file:///<...skipped...>/node_modules/@sveltejs/kit/src/exports/vite/dev/index.js:505:22
[auth][details]: {
  "provider": "azure-ad"
}
```

The root cause is the way Microsoft implemented oidc.

### How to reproduce

1. Configure AzureADProvider provider following the documentation:
```
export const auth: Handle = SvelteKitAuth({
	trustHost: true,
	secret: AUTH_SECRET,
	providers: [
		AzureADProvider({
			clientId: AZURE_CLIENT_ID,
			clientSecret: AZURE_CLIENT_SECRET,
		}),
	],
});
```

2. Try to login with any Microsoft identity

### Expected behavior

[...]

PR #9718, opened 2024-01-23
fix(providers): Handle Azure AD tenants correctly

### LINKED ISSUE #9635: AzureAD common tenant URLs aren't being checked properly
### Provider type

Azure Active Directory

### Environment

System:
OS: macOS 14.2.1
CPU: (12) arm64 Apple M2 Max
Memory: 65.83 MB / 32.00 GB
Shell: 5.9 - /bin/zsh
Binaries:
bun: 1.0.14 - ~/.bun/bin/bun
Browsers:
Brave Browser: 120.1.61.109
Safari: 17.2.1
npmPackages:
@auth/core: ^0.18.6 => 0.18.6

### Reproduction URL

### Describe the issue

```js
import * as jose from 'jose'

const as = await oauth
  .then((response) =>
      response,
    ),
  )

let tenantId
}

  throw new Error()
}

  {
    ...as,
  },
  client,
  response,
)
```

### How to reproduce

### Expected behavior

JibbityJobbity:
## ☕️ Reasoning

Endpoints returned by Azure AD want us to edit the path so that each request gets routed to their proper tenant IDs.
The old implementation didn't handle this properly when using the "common" tenant ID, which basically lets us use it normally like any other provider. We would receive the template endpoint paths and use those in error checking and stuff rather than the ones we actually use.

I've just thrown something up and I've done some quick testing. It seems to work normally. I'll test custom tenants at some point and I suspect it won't work until I test and fix it.

## 🧢 Checklist

- [X] Documentation
- [X] Tests
- [X] Ready to be merged

## 🎫 Affected issues

Fixes: #9635

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/actions/callback/oauth/callback.ts
+++ b/packages/core/src/lib/actions/callback/oauth/callback.ts
@@ -1,3 +1,4 @@
+import * as jose from "jose"
 import * as checks from "./checks.js"
 import * as o from "oauth4webapi"
 import {
@@ -44,10 +45,15 @@ export async function handleOAuth(
     // We assume that issuer is always defined as this has been asserted earlier
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     const issuer = new URL(provider.issuer!)
-    const discoveryResponse = await o.discoveryRequest(issuer)
+    let discoveryResponse: Response | undefined = await o.discoveryRequest(issuer)
+
+    if (provider.authorization?.conform) {
+      discoveryResponse = await provider.authorization?.conform(discoveryResponse);
+    }
+
     const discoveredAs = await o.processDiscoveryResponse(
       issuer,
-      discoveryResponse
+      discoveryResponse!
     )
 
     if (!discoveredAs.token_endpoint)
@@ -139,6 +145,10 @@ export async function handleOAuth(
 
   let profile: Profile = {}
   let tokens: TokenSet & Pick<Account, "expires_at">
+
+  if (provider.authorization?.serverConform) {
+    as = await provider.authorization?.serverConform(as, codeGrantResponse)
+  }
 
   if (provider.type === "oidc") {
     const nonce = await checks.nonce.use(cookies, resCookies, options)
--- a/packages/core/src/providers/azure-ad.ts
+++ b/packages/core/src/providers/azure-ad.ts
@@ -9,6 +9,7 @@
  * @module providers/azure-ad
  */
 import type { OAuthConfig, OAuthUserConfig } from "./index.js"
+import * as jose from "jose"
 
 export interface AzureADProfile extends Record<string, any> {
   sub: string
@@ -141,6 +142,31 @@ export default function AzureAD<P extends AzureADProfile>(
       params: {
         scope: "openid profile email User.Read",
       },
+      async conform(response) {
+        // Microsoft being special and non-compliant #9635
+        //
+        // MS doesn't follow the spec for some common tenant
+        // ID's ("common", "organizations", "customers"), returning
+        // a different issuer URL than used to make the request.
+        if (response.status !== 200) return response
+
+        let json = await response.json()
+        json.issuer = rest.issuer;
+        return new Response(JSON.stringify(json), response)
+      },
+      async serverConform(authServer, codeGrantResponse) {
+        if (codeGrantResponse.status !== 200) return authServer
+
+        const { id_token } = await codeGrantResponse.clone().json()
+        const jwt: jose.JWTPayload & { tid?: string } = jose.decodeJwt(id_token)
+
+        return {
+          ...authServer,
+          issuer: jwt.tid
+            ? authServer.issuer.replace(tenantId, jwt.tid)
+            : authServer.issuer,
+        }
+      }
     },
     async profile(profile, tokens) {
       // https://docs.microsoft.com/en-us/graph/api/profilephoto-get?view=graph-rest-1.0#examples
--- a/packages/core/src/lib/actions/signin/authorization-url.ts
+++ b/packages/core/src/lib/actions/signin/authorization-url.ts
@@ -24,8 +24,13 @@ export async function getAuthorizationUrl(
     // We check this in assert.ts
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     const issuer = new URL(provider.issuer!)
-    const discoveryResponse = await o.discoveryRequest(issuer)
-    const as = await o.processDiscoveryResponse(issuer, discoveryResponse)
+    let discoveryResponse: Response | undefined = await o.discoveryRequest(issuer)
+
+    if (provider.authorization?.conform) {
+      discoveryResponse = await provider.authorization?.conform(discoveryResponse);
+    }
+
+    const as = await o.processDiscoveryResponse(issuer, discoveryResponse!)
 
     if (!as.authorization_endpoint) {
       throw new TypeError(
--- a/packages/core/src/lib/utils/providers.ts
+++ b/packages/core/src/lib/utils/providers.ts
@@ -151,5 +151,5 @@ function normalizeEndpoint(
       url.searchParams.set(key, String(value))
     }
   }
-  return { url, request: e?.request, conform: e?.conform }
+  return { url, request: e?.request, conform: e?.conform, serverConform: e?.serverConform }
 }
--- a/packages/core/src/providers/microsoft-entra-id.ts
+++ b/packages/core/src/providers/microsoft-entra-id.ts
@@ -9,6 +9,7 @@
  * @module providers/microsoft-entra-id
  */
 import type { OAuthConfig, OAuthUserConfig } from "./index.js"
+import * as jose from "jose"
 
 export interface MicrosoftEntraIDProfile extends Record<string, any> {
   sub: string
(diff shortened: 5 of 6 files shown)


JibbityJobbity:
Testing needed

- Common tenant group
- organizations and consumers tenant groups
- Any other specific tenant ID
- Other providers still work

balazsorban44 (changes requested):
Any non-compliance needs to be handled at the provider level. we don't want branching based on specific providers, it's not maintainable in the long term

This will need a different solution. Something similar to the `conform` method on `token`.


some useful info at the end of the thread here https://github.com/nextauthjs/next-auth/issues/8831

panva:
I think there should be two distinct Azure providers

One where you know the tenant you're working with and you're NOT using any of the "common" issuers (the existing one with the exception that tenant id is required).

And then another one which can be used with one of the "common" issuers

```
  'https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration'
  'https://login.microsoftonline.com/organizations/v2.0/.well-known/openid-configuration'
  'https://login.microsoftonline.com/consumers/v2.0/.well-known/openid-configuration'
```

This one has the workarounds in and requires explicit configuration to

- allow login from all tenants
- allow login from specific tenants (by providing a list of tenant ids), the ID Token's `tid`  would be checked to match the expected tenants.

JibbityJobbity:
While we all know the real solution to this, I think by the time I have a crack at this, we will know what ends up happening. But that's definitely considerable if `conform()` doesn't end up happening.
However in the meantime, I think the `conform()` method is the first way forward with this to try. It provides a way for us to sort out LinkedIn too, but I'd be careful about getting too comfortable with hacking around providers as I can imagine `conform()` would make it easier to do exactly that.

packages/core/src/providers/azure-ad.ts
```diff
@@ -121,20 +122,50 @@ export default function AzureAD<P extends AzureADProfile>(
...
   rest.issuer ??= `https://login.microsoftonline.com/${tenantId}/v2.0`
   return {
     id: "azure-ad",
     name: "Azure Active Directory",
     type: "oidc",
     wellKnown: `${rest.issuer}/.well-known/openid-configuration?appid=${options.clientId}`,
+    token: {
+      async conform(response) {
+        // Microsoft being special and non-compliant #9635
+        response.url.replace("{tenantid}", tenantId);
```
panva, line 144: This doesn't do anything, does it? The Response `url` property is readonly and immutable

packages/core/src/providers/azure-ad.ts
```diff
@@ -121,20 +122,50 @@ export default function AzureAD<P extends AzureADProfile>(
...
+        return response
+      }
+    },
     authorization: {
       params: {
         scope: "openid profile email User.Read",
       },
+      async conform(response) {
+        // Microsoft being special and non-compliant #9635
+        response.url.replace("{tenantid}", tenantId);
```
panva, line 154: same

packages/core/src/providers/azure-ad.ts
JibbityJobbity: You're probably right. I'm not a TS/JS dev but I should've picked something like this up.
I'd hold off on review until I say something.

packages/core/src/lib/actions/callback/oauth/callback.ts
```diff
@@ -129,6 +135,10 @@ export async function handleOAuth(
   let profile: Profile = {}
   let tokens: TokenSet & Pick<Account, "expires_at">
 
+  if (provider.authorization?.serverConform) {
+    as = await provider.authorization?.serverConform(as, codeGrantResponse)
+  }
+
```
JibbityJobbity, line 141: I'm having some trouble here. It might be the way my development environment is set up, but it seems to think `serverConform` isn't a part of the object.

JibbityJobbity:
I've done what I can for now. Completely open to suggestions and feedback.

I know this doesn't quite match the sample, but I wanted the check to work the other way around. Rather than having it try to check against the "fake" URL with `{tenantid}`, I thought it would be cleaner to perform a real check.

packages/core/src/providers/azure-ad.ts
```diff
@@ -135,6 +136,26 @@ export default function AzureAD<P extends AzureADProfile>(
...
+      async serverConform(authServer, codeGrantResponse) {
+        let tenantId: string = "common";
+        if (codeGrantResponse.status === 200) {
+          const { id_token } = await codeGrantResponse.clone().json()
+            ; ({ tid: tenantId } = jose.decodeJwt(id_token))
+        }
+
+        return {
+          ...authServer,
+          issuer: authServer.issuer.replace("common", tenantId),
```
JibbityJobbity, line 156: This isn't right. Need to detect the tenant that's being used. Probably just replace the provider's tenant ID with whatever we expect from the JWT.

panva:
I'm not finding all the individual bits from my [original recommendation](https://github.com/panva/oauth4webapi/issues/100#issuecomment-1890380232) nor the recommended split from https://github.com/nextauthjs/next-auth/pull/9718#issuecomment-1909893145 so I'm assuming you're going with your own workarounds.

JibbityJobbity:
> I'm not finding all the individual bits from my [original recommendation](https://github.com/panva/oauth4webapi/issues/100#issuecomment-1890380232) nor the recommended split from [#9718 (comment)](https://github.com/nextauthjs/next-auth/pull/9718#issuecomment-1909893145) so I'm assuming you're going with your own workarounds.

It's a mix between what you suggested in the other issue and the conform methods. It still achieves the same thing, which is satisfying those underlying checks in your library while bringing the conformity to the provider level, as suggested by @balazsorban44. If this isn't good enough then I'll change the PR to satisfy the solution you gave earlier. The special `serverConform` is a bit icky so I see where you're coming from.

Anyway, I just did some testing with the other grouped tenants. The issuer URL comes back with some weird tenant ID instead of getting the user to replace "{tenantid}" or some other value we can expect. Replacing "{tenantid}" *only* seems to work with "common" and doesn't work with "organizations" or "consumers".  This keeps rearing its ugly head, you almost need something to the effect of this to override the check to bypass it:

```js
    const discoveryResponse: Response | undefined = await o.discoveryRequest(issuer)
    let json = await response.json();
    const as = await o.processDiscoveryResponse(new URL(json.issuer), discoveryResponse!)
```

I'll need another go at this. I'll shoot for tomorrow.

JibbityJobbity:
This seems to work nicely :) I'm largely happy with the current solution now.

JibbityJobbity:
Bump, rebased.

amplicity:
hey! great work here, we could really use this. would love to see this get merged soon.

MrLoh:
It would be great to see this get merged. I'm sure it blocks a lot of people from adopting v5.

jaysin586:
Is there an update on this?

ladparth:
Hello,
Thank you for working on this PR. I am also stuck on this stage of my development. Is there any update on this fix?

JibbityJobbity:
> Hello, Thank you for working on this PR. I am also stuck on this stage of my development. Is there any update on this fix or is there any temporary fix i can do while this PR is in process?

I'll rebase, but I haven't heard anything from any of the maintainers since I got it working.
Do keep in mind they may be busy with their lives, but there are some other hacks to make it work by configuring the provider itself. This PR would just mean that nobody would have to be concerned about this again.

ladparth:
Thank you @JibbityJobbity

murilobd:
Thanks @JibbityJobbity for that PR! Hopefully maintainers of this repo will merge it!

Ghernouz:
Any updates guys?

RonB:
It looks like there is some status quo here. Azure AD also has been rebranded to Microsoft Entra Id and that has been adressed in the code already. I am afraid the train is moving on and a solution is getting out of site. @JibbityJobbity, thank you for your hard work. A request on behalf of all of us if you could adress the remaining open ends here?

MrLoh:
Entra ID has the exact same issue that this PR tries to fix.

JibbityJobbity:
I've rebased and put the fix onto Entra ID. I haven't changed any logic.
Seeing as people have shown interest in getting this fixed, I'd like to kindly ask the community to try testing this out for me. I'm going to have to pull the "life stuff" excuse for not getting this done sooner or testing it.
Hopefully this gets merged once some people say it works.

RonB:
@JibbityJobbity 
Thank you so much for rebasing, I will advocate for this change to be merged but I am having a little trouble getting it to work.
I have a mono-repo with 2 apps and use pnpm. How should i use your repo?
In my package.json something like:

```json
"dependencies":
    "@auth/core": "https://github.com/JibbityJobbity/next-auth.git",
    "@auth/sveltekit": "^1.1.0"
"pnpm": {
    "overrides": {
      "@auth/core": "https://github.com/JibbityJobbity/next-auth.git"
    }
  },
  ```
  
 It errors with: 
 ```
 Error: Failed to resolve entry for package "@auth/core". The package may have incorrect main/module/exports specified in its package.json.
 ```
 Does the package has to be build first and do i need to make a fork of your repo and do that myself or is there a simpler solution use your repo?
 I am not an expert on this depency management stuff so i hope you can point me in the right direction.
 
 Excuse my ignorance :-)
 Thank you for your response

**Comment by michal-boruczkowski** [comment]
@RonB
I was trying to use it like this:
```
dependencies: {
     ...
    "next-auth": "git://github.com/JibbityJobbity/next-auth.git#azure-ad-common-tenant",
}
```

but I'm getting following error when `npm i`:
```
npm error git dep preparation failed
npm error npm warn using --force Recommended protections disabled.
npm error npm error code EUNSUPPORTEDPROTOCOL
npm error npm error Unsupported URL Type "workspace:": workspace:*
```

I'm using `patch` prepared here: https://github.com/nextauthjs/next-auth/issues/8374#issuecomment-1803387771 ATM

Would be great if this fix is merged 🤞

JibbityJobbity:
Yeah, the way I got around it for development was to hack stuff in my `node_modules` folder. Not entirely sure how to deploy it in my own web app actually :/ if I use the sveltekit framework package in my branch then it tries to pull the core package from npm instead of my branch. Everyone I ask about this says "idk publish the package on npm". Sorry I don't have any useful info.

hiromijorge:
> @RonB I was trying to use it like this:
> 
> ```
> dependencies: {
>      ...
>     "next-auth": "git://github.com/JibbityJobbity/next-auth.git#azure-ad-common-tenant",
> }
> ```
> 
> but I'm getting following error when `npm i`:
> 
> ```
> npm error git dep preparation failed
> npm error npm warn using --force Recommended protections disabled.
> npm error npm error code EUNSUPPORTEDPROTOCOL
> npm error npm error Unsupported URL Type "workspace:": workspace:*
> ```
> 
> `pnpm` seems to handle this but actual fix is not applied in runtime
> 
> I'm using `patch` prepared here: [#8374 (comment)](https://github.com/nextauthjs/next-auth/issues/8374#issuecomment-1803387771) ATM
> 
> Would be great if this fix is merged 🤞

Hi Michael,

I've followed the instructions in this [link](https://github.com/nextauthjs/next-auth/issues/8374#issuecomment-1803387771) for applying the patch.

However, I'm still encountering the same errors as before. Could you please explain how you reproduce this issue?

Errors
[auth][error] CallbackRouteError: Read more at https://errors.authjs.dev#callbackrouteerror
[auth][cause]: OperationProcessingError: unexpected JWT "iss" (issuer) claim value

PR #9721, opened 2024-01-23
feat: update account on login

alv92:
### What needed fixing:

Hi!

Currently, when signing in using a provider, we're populating the Account with refresh and access tokens. These are left static for the entire lifetime of the Account, which I think shouldn't be the case as that will turn at some point on revoked access if the user clicks “sign me out of all my devices” on their provider account, their admin triggers a password loss procedure or if we change the scopes we request on our token.

This is why in this PR I aim to update the Account with newly issued refresh and access tokens with each subsequent login from the user.

This is a PoC (tested locally, works) and I'd love to get some insights on how feasible this solution is to be implemented in Next Auth, both on `core` and on `v4` since we need it in several of our projects.

### Changes:

I created a new adapter function called `updateAccount()` (on prisma adapter only as a PoC) which takes an `Account` and a `UserID` as props. This function will search for that account on the database and update it with the updated account information on login. In case that Account does not exist, it will create a new entry, which means this function could potentially be used as well for linking an account as well.

The function is optional to make sure no other adapters are being broken after merging.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


balazsorban44 (changes requested):
This looks interesting, and probably something we would like to consider! However the correct flow is to create a feature request first:

https://github.com/nextauthjs/next-auth/discussions/new?category=ideas

I also recommend splitting this to multiple PRs so the adapter can be addressed separately.

alv92:
@balazsorban44 thanks for the quick reply! I've added the [discussion](https://github.com/nextauthjs/next-auth/discussions/9757) as requested. Should I delete this pull request and create a new one for each change?

balazsorban44:
We can keep the PR for now!

I think if we introduce this behind an `experimental` flag BTW, we could merge this suggestion as-is.

Some code that might be useful. To throw a proper error if the flag is missing, something like:

https://github.com/nextauthjs/next-auth/blob/30fb90f9159d8375826695351cf0a4cd490885f4/packages/next-auth/src/lib/future/index.ts#L60-L65

Then you an extend the `experimental` config here: https://github.com/nextauthjs/next-auth/blob/62f24f54e0f70c36bf3474de168bd0e14812ec57/packages/core/src/types.ts#L589

alv92:
Awesome, sounds great!

Is the [next-auth/packages/next-auth/src/lib/future/index.ts](https://github.com/nextauthjs/next-auth/blob/30fb90f9159d8375826695351cf0a4cd490885f4/packages/next-auth/src/lib/future/index.ts#L60-L65) file available on main? Do I need to fork from your branch instead and point my PR there?

balazsorban44:
It's just to showcase a possible implementation. you can copy paste it and adjust the naming accordingly for this PR

alv92:
@balazsorban44 I added the experimental error handling function for updateAccount 😄 I went ahead and updated the documentation as well

alv92:
Hi @balazsorban44! Is there anything else that should be done before we can proceed with merging these changes? 😄

PR #9746, opened 2024-01-23
fix: use match instead of split when parsing action

### LINKED ISSUE #9743: basePath feature breaks auth0 (and presumably authentik) provider
### Environment

```
  System:
    OS: macOS 14.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 142.61 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.nodenv/versions/20.9.0/bin/node
    Yarn: 3.2.1 - ~/.nodenv/versions/20.9.0/bin/yarn
    npm: 10.1.0 - ~/.nodenv/versions/20.9.0/bin/npm
    pnpm: 8.14.3 - ~/.nodenv/versions/20.9.0/bin/pnpm
    bun: 1.0.25 - /opt/homebrew/bin/bun
    Watchman: 2023.12.04.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 120.0.6099.234
    Safari: 17.1
  npmPackages:
    @auth/sveltekit: ^0.8.0 => 0.8.0 
```

### Reproduction URL

--

### Describe the issue

The recent addition of the [`basePath` feature](https://github.com/nextauthjs/next-auth/pull/9686) has caused any provider that starts with `auth` (e.g. `auth0`, `authentik`) to break when using the default base path (`/auth`).

The problem is here: https://github.com/nextauthjs/next-auth/pull/9686/files#diff-e2783aeceecac8cd1b559372f242d520b894883a0c297fa4829ef4e5a5f35e88R134

`pathname.split(base)` represents `'/auth/signin/auth0'.split('/auth')` when using the Auth0 provider. As such the result of the operation is `['', '/signin', '0']` which causes the function to throw since there's an expected length of 2.

### How to reproduce

--

### Expected behavior

--

chrislambe:
## ☕️ Reasoning

The recent addition of the https://github.com/nextauthjs/next-auth/pull/9686 has caused providers that start with "auth" (e.g. `auth0`, `authentik`) to break when using the default base path (`/auth`).

The problem is here: https://github.com/nextauthjs/next-auth/pull/9686/files#diff-e2783aeceecac8cd1b559372f242d520b894883a0c297fa4829ef4e5a5f35e88R134

`pathname.split(base)` represents `'/auth/signin/auth0'.split('/auth')` when using the Auth0 provider. As such the result of the operation is `['', '/signin', '0']` which causes the function to throw since there's an expected length of 2.

See https://github.com/nextauthjs/next-auth/issues/9743.

**Side note:** auto-closing issues that don't have a "valid" repro URL is frustrating. I think making it an absolute requirement to include a reproduction URL is a mistake.

## 🧢 Checklist

I was unable to get the tests to work. Presumably the addition of all possible provider URLs here would have caught this:
https://github.com/nextauthjs/next-auth/blob/963086b36f4dc9203cbcf6997b5fed148f90edc8/packages/core/test/url-parsing.test.ts

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

 
Fixes: https://github.com/nextauthjs/next-auth/issues/9743

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/utils/web.ts
+++ b/packages/core/src/lib/utils/web.ts
@@ -131,9 +131,9 @@ export function parseActionAndProviderId(
   action: AuthAction
   providerId?: string
 } {
-  const a = pathname.split(base)
+  const a = pathname.match(new RegExp(`^${base}(.+)`))
 
-  if (a.length !== 2 || a[0] !== "")
+  if (a === null)
     throw new UnknownAction(`Cannot parse action at ${pathname}`)
 
   const [_, actionAndProviderId] = a
--- a/packages/core/test/url-parsing.test.ts
+++ b/packages/core/test/url-parsing.test.ts
@@ -46,6 +46,12 @@ describe("parse the action and provider id", () => {
       error: "Cannot parse action at /api/auth/signin/api/auth/signin/github",
       basePath: "/api/auth",
     },
+    {
+      path: "/auth/signin/auth0",
+      action: "signin",
+      providerId: "auth0",
+      basePath: "/auth",
+    },
   ])("$path", ({ path, error, basePath, action, providerId }) => {
     if (action || providerId) {
       const parsed = parseActionAndProviderId(path, basePath)


balazsorban44 (changes requested):
Thanks for the PR! Really appreciated! ~Could you add some tests for it under https://github.com/nextauthjs/next-auth/blob/main/packages/core/test/url-parsing.test.ts :pray:~ Added one so we can merge this ASAP. Very good catch!


Regarding:
> Side note: auto-closing issues that don't have a "valid" repro URL is frustrating. I think making it an absolute requirement to include a reproduction URL is a mistake.

This won't change. In fact, I hope to make it even stricter. See: https://twitter.com/balazsorban44/status/1749360132205465865

chrislambe:
I definitely appreciate the point about wanting to keep your processes tight. I'm just saying that requiring a repro URL might be a naive approach when there are cases like this where someone is able to point to exactly what and where the problem is. Just something to consider! Thanks for merging this so quickly.

balazsorban44:
Your case is the minority. Besides, important enough issues always resurface by someone taking the proper time to create a reproduction.

PR #9756, opened 2024-01-24
fix(ts): show `user` and `token` in `session` callback

### LINKED ISSUE #9633: Type error in session callback for `user` and `token` properties

ajstars1:
## ☕️ Reasoning

In session callback, argument type does not contain user and token
> ### **Before code updation 👇**
Screenshot 2024-01-24 at 8 34 52 PM

> **TypeError:** Property 'token' does not exist on type '({ session: Session; user: AdapterUser; } | { session: Session; token: JWT; }) & { newSession: any; trigger?: "update" | undefined; }'

**Type Intersection:** The original code incorrectly used a union type (`|`), potentially leading to type mismatches. The updated code employs an intersection type (`&`) to guarantee that both `user` and `token` properties are always available within the session callback, regardless of the authentication strategy.

> ### **After code updation 👇**
Screenshot 2024-01-24 at 8 53 38 PM

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues
Fixed: #9633 

## 📌 Resources

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -254,7 +254,7 @@ export interface CallbacksOptions<P = Profile, A = Account> {
           /** Available when {@link AuthConfig.session} is set to `strategy: "database"`. */
           user: AdapterUser
         }
-      | {
+      & {
           session: Session
           /** Available when {@link AuthConfig.session} is set to `strategy: "jwt"` */
           token: JWT


balazsorban44 (approved):
> Type Intersection: The original code incorrectly used a union type (|), potentially leading to type mismatches. The updated code employs an intersection type (&) to guarantee that both user and token properties are always available within the session callback, regardless of the authentication strategy.

This is also incorrect since it usually is either `user` or `token`, based on the session strategy! 

But I think previously we had the same as what this tries to fix, so for backwards compact, we can probably merge this. This needs to be revisited, maybe as part of #9702

Thanks!

PR #9758, opened 2024-01-24
feat: don't override basePath if it's already been set

ndom91:
## ☕️ Reasoning

- While testing https://github.com/nextauthjs/next-auth/pull/8808/ I kept getting errors in `core/lib/utils/web.ts` - `parseActionAndProviderId()` that it couldn't "parse the action at `/session`". The last part of that error msg is the pathname and should be `/auth/session` there :thinking: 
- If yall can confirm this behaviour, we should fix it in the other client libs, I think the `getEnvDefaults` fn is more or less the same there

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/lib/env.ts
+++ b/packages/next-auth/src/lib/env.ts
@@ -6,7 +6,7 @@ export function setEnvDefaults(config: NextAuthConfig) {
   config.secret ??= process.env.AUTH_SECRET ?? process.env.NEXTAUTH_SECRET
   try {
     const url = process.env.AUTH_URL ?? process.env.NEXTAUTH_URL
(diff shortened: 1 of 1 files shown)


balazsorban44:
This is intentional. If you have an environment variable defined, it should be preferred at all times. Otherwise you can just set the `basePath` option

balazsorban44:
The above is incorrect, as @ndom91 pointed out in private messages. We do need to respect whatever `basePath` was set by the user, similar to other config options that we try to infer! We'll work on a new PR that moves `setEnvDefaults` to the core so this is consistent across frameworks!

Issue #8629, opened 2023-09-17
Do we need a unique constraint on User.email?


keitakn:
### Question 💬

I am trying to persist Auth.js data to RDB using the following table structure.

https://authjs.dev/reference/adapters#models

I use PlanetScale.

I am thinking of using `@auth/drizzle-adapter`.

The `MySQL` section of the following document shows no unique constraint on `email`.

https://authjs.dev/reference/adapter/drizzle

But in @auth/prisma-adapter, email is set to unique.

https://authjs.dev/reference/adapter/prisma

My understanding is that Auth.js does not allow sign-in with the same email address.

If it is correct to make the email unique, I would appreciate it if you could correct the following document🙏

https://authjs.dev/reference/adapter/drizzle

### How to reproduce ☕️

```typescript
import {
  int,
  timestamp,
  mysqlTable,
  primaryKey,
  varchar,
} from "drizzle-orm/mysql-core";
import type { AdapterAccount } from "@auth/core/adapters";

export const users = mysqlTable("user", {
  id: varchar("id", { length: 255 }).notNull().primaryKey(),
  name: varchar("name", { length: 255 }),
  // I think this email needs to be UNIQUE
  email: varchar("email", { length: 255 }).notNull(),
  emailVerified: timestamp("emailVerified", {
    mode: "date",
    fsp: 3,
  }).defaultNow(),
  image: varchar("image", { length: 255 }),
});

export const accounts = mysqlTable(
  "account",
  {
    userId: varchar("userId", { length: 255 })
      .notNull()
      .references(() => users.id, { onDelete: "cascade" }),
    type: varchar("type", { length: 255 })
      .$type<AdapterAccount["type"]>()
      .notNull(),
    provider: varchar("provider", { length: 255 }).notNull(),
    providerAccountId: varchar("providerAccountId", { length: 255 }).notNull(),
    refresh_token: varchar("refresh_token", { length: 255 }),
    access_token: varchar("access_token", { length: 255 }),
    expires_at: int("expires_at"),
    token_type: varchar("token_type", { length: 255 }),
    scope: varchar("scope", { length: 255 }),
    id_token: varchar("id_token", { length: 255 }),
    session_state: varchar("session_state", { length: 255 }),
  },
  (account) => ({
    compoundKey: primaryKey(account.provider, account.providerAccountId),
  })
);

export const sessions = mysqlTable("session", {
  sessionToken: varchar("sessionToken", { length: 255 }).notNull().primaryKey(),
[...]

PR #9787, opened 2024-01-25
Remove `notNull()` constraint from `email` in drizzle

nahtnam:
## ☕️ Reasoning

Based on the docs, it is possible for email to be null: https://authjs.dev/reference/core/providers/discord#email

Additionally it is null in other adapters: https://authjs.dev/reference/adapter/prisma

This would not be a breaking change as it's dropping a constraint, not adding one. Although I believe we should also:
1. Make email `unique()`
2. Allow renaming of tables and columns like we do with other adapters
^ But we should do these in another PR

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Related to https://github.com/nextauthjs/next-auth/issues/8629 but doesn't fix the primary concern of that

 

## 📌 Resources

### DIFF
--- a/packages/adapter-drizzle/src/index.ts
+++ b/packages/adapter-drizzle/src/index.ts
@@ -46,7 +46,7 @@ import type { Adapter } from "@auth/core/adapters"
  *   ],
  * })
  * ```
- * 
+ *
  * :::info
  * If you're using multi-project schemas, you can pass your table function as a second argument
  * :::
@@ -75,7 +75,7 @@ import type { Adapter } from "@auth/core/adapters"
  * export const users = pgTable("user", {
  *  id: text("id").notNull().primaryKey(),
  *  name: text("name"),
- *  email: text("email").notNull(),
+ *  email: text("email"),
  *  emailVerified: timestamp("emailVerified", { mode: "date" }),
  *  image: text("image"),
  * })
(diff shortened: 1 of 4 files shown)


nahtnam:
@balazsorban44 The build is failing because:

```
Types of property 'email' are incompatible.
  Type 'string | null' is not assignable to type 'string'.
    Type 'null' is not assignable to type 'string'.ts(2322)
```

on all of the drizzle adapter functions.

I think we have a few options:
- Allow email to be null here: https://github.com/nahtnam/next-auth/blob/main/packages/core/src/adapters.ts#L181 (this requires other changes to the codebase)
- Cast all the of the `return`s to `AdapterUser`, but this would be type unsafe

Looking closer at the Prisma code, it looks like the docs suggest Users to do `email String?` but then in the codebase when generating types it uses `email String`. https://github.com/nahtnam/next-auth/blob/master/packages/adapter-prisma/prisma/schema.prisma#L13

ndom91:
Some providers may potentially not return an Email (i.e. the Discord example you shared), but we should always return at least an empty string. If not, that's a bug in the provider. 

In addition, as you pointed out, some of the adapter schemas seem to be mismatched when it comes to this field. They should also not allow `null` for the `User.email` field. 

For example, the `AdapterUser` (`packages/core/src/adapters.ts:176`) defines the field as `email: string`.

For those reasons, I'm going to close this PR, but thanks for bringing our attention to the issue. There's definitely room for improvement here :blush:

PR #9788, opened 2024-01-25
refactor: update JWT enc

panva:
This PR updates the JWT Content Encryption Algorithm from A256GCM to A256CBC-HS512, the algorithm will be used for all new issued JWTs. During verification both are allowed to allow for a seamless transition (in due time A256GCM may be completely removed).

The reason for this is that given the used secret is static it is safer to use an algorithm with larger initialization vector (GCM uses 96 bits, CBC uses 128) to limit the possibility of iv-reuse.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Auth.js can be used with or without a database.
 - Designed to be secure by default and encourage best practices for safeguarding user data
 - Uses Cross-Site Request Forgery (CSRF) Tokens on POST routes (sign in, sign out)
 - Default cookie policy aims for the most restrictive policy appropriate for each cookie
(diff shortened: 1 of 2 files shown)


balazsorban44 (changes requested):
Just need to update the docs in these two places as well:

https://github.com/nextauthjs/next-auth/blob/dcc55b1bd3cf8efb2ce3b9e66e77dec3c2df47cb/README.md?plain=1#L56

https://github.com/nextauthjs/next-auth/blob/dcc55b1bd3cf8efb2ce3b9e66e77dec3c2df47cb/packages/core/src/jwt.ts#L8

balazsorban44 (approved):
Thanks!

PR #9793, opened 2024-01-25
fix: always use a generated `id` for `User.id`

### LINKED ISSUE #9699: Default ID generation overridden by oauth (google) user profile
### Adapter type

@auth/prisma-adapter

### Environment

```
  System:
    OS: macOS 14.2.1
    CPU: (8) arm64 Apple M2
    Memory: 66.33 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - ~/.nvm/versions/node/v20.11.0/bin/node
    npm: 10.2.4 - ~/.nvm/versions/node/v20.11.0/bin/npm
    bun: 1.0.16 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 120.1.61.120
    Safari: 17.2.1
  npmPackages:
    next: latest => 14.1.0 
    next-auth: beta => 5.0.0-beta.5 
    react: ^18.2.0 => 18.2.0 
```

### Reproduction URL

https://github.com/meruiden/prisma-google-id-issue

### Describe the issue

next-auth doesn't clear the ID provided by the OAuth provider in the handleLoginOrRegister function. Because the _profile variable contains an ID and is not cleared before passed into the database adapter, the default ID that would normally be generated (I had it set to nanoid in Prisma), is overwritten by the ID the OAuth provider passed. I don't want my users to have their ID be the OAuth provider's ID. Not only can this cause collisions (unlikely tho), but it stores the ID 2 times both in the Account.providerAccountId and User.id tables.

Screenshot 2024-01-21 at 11 12 01

### How to reproduce

I have a setup with google as oauth and prisma as database adapter. When you log in, the ID from my google account is stored as ID for my user instead of my nanoid it would generate by default. It's supposed to only be in Account.providerAccountId.

1. ```npm install```
2. ```cp .env.local.example .env.local```
3. Set up google OAuth in .env.local and secret
5. ```npx prisma migrate dev```
6. ```npm run dev```
7. Login with google
8. ```npx prisma studio```
9. And see that the User in the User table used Google's userID instead of the configured nanoid in the prisma schema. 

### Expected behavior

handleLoginOrRegister should clear any id's and allow the user or database adapter to decide how the ID is formed, usually, database (adapters) have a default value for id generation, in my case I used nanoid's, which works for accounts, sessions and verification tokens, but is overwritten for the user table.

balazsorban44:
Fixes #9699

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/actions/callback/oauth/callback.ts
+++ b/packages/core/src/lib/actions/callback/oauth/callback.ts
@@ -197,7 +197,7 @@ async function getUserAndAccount(
     const userFromProfile = await provider.profile(OAuthProfile, tokens)
     const user = {
       ...userFromProfile,
-      id: userFromProfile.id?.toString() ?? crypto.randomUUID(),
+      id: crypto.randomUUID(),
       email: userFromProfile.email?.toLowerCase(),
     } satisfies User
 
--- a/packages/core/src/lib/utils/providers.ts
+++ b/packages/core/src/lib/utils/providers.ts
@@ -1,4 +1,3 @@
-import { OAuthProfileParseError } from "../../errors.js"
 import { merge } from "./merge.js"
 
 import type {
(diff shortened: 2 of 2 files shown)


yordis:
@balazsorban44 do you mind sharing the Reasoning behind overwriting the `id` coming from the `profile` callback?

I wouldn't expect this behavior to happen. I would expect the identity of the user to be managed by the Provider, or I would use the `profile` callback to overwrite it if I really want to change it.

PR #9817, opened 2024-01-26
feat: add `getEnvDefaults` and `createActionURL` to `@auth/core` for use in client libs

ndom91:
## ☕️ Reasoning

- Add `getEnvDefault()` and `createActionURL()` functions to `@auth/core` to be used elsewhere
- Add wrapper to `next-auth` to set its `basePath = '/api/auth'` by default
- Add wrapper to `@auth/sveltekit` to set its `basePath = '${base}/auth'` (sveltekit import)
- Add 2 tests

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


packages/core/src/lib/actions/env.ts
```diff
@@ -0,0 +1,54 @@
+import type { IncomingHttpHeaders } from "http"
+import type { AuthAction, AuthConfig } from "../../types"
+
+/**
+ * Set default env variables on the config object
+ **/
+export function setEnvDefaults(envObject: any, config: AuthConfig) {
+  try {
+    const url = process.env.AUTH_URL
```
balazsorban44, line 9: Cannot leave `process` in the code lib, it needs to be fully runtime agnostic

```suggestion
    const url = envObject.AUTH_URL
```

packages/core/src/lib/actions/env.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (url && !config.basePath) config.basePath = new URL(url).pathname
+  } catch {
+  } finally {
+    config.basePath ??= `/auth`
+  }
+
+  config.redirectProxyUrl ??= process.env.AUTH_REDIRECT_PROXY_URL
+  config.secret ??= process.env.AUTH_SECRET
```
balazsorban44, line 17: Could you incorporate https://github.com/nextauthjs/next-auth/blob/1345a8cfbc484aeda80bbc22fd753003b824f5fc/packages/next-auth/src/lib/env.ts#L14-L22 which was added in #9804?

packages/next-auth/src/lib/env.ts
balazsorban44, line 1: The entire `setEnvDefaults` function can be almost fully replaced by importing the one from core.

https://github.com/nextauthjs/next-auth/blob/3964f0b5c310ab57e38a4c5dd7774e65ea767a2f/packages/next-auth/src/lib/env.ts#L5-L45

The only thing that needs to stay is https://github.com/nextauthjs/next-auth/blob/3964f0b5c310ab57e38a4c5dd7774e65ea767a2f/packages/next-auth/src/lib/env.ts#L11

since `next-auth` will keep using the `/api/auth` `basePath` if none set for historical/backwards compat reason.

https://github.com/nextauthjs/next-auth/blob/cf07bba4f168efa8d3f416c4e0c59e5386213957/packages/frameworks-sveltekit/src/lib/index.ts#L323-L367

and

https://github.com/nextauthjs/next-auth/blob/cf07bba4f168efa8d3f416c4e0c59e5386213957/packages/frameworks-express/src/index.ts#L183-L243

can however be both fully dropped

packages/core/src/lib/actions/env.ts
```diff
@@ -0,0 +1,54 @@
+import type { IncomingHttpHeaders } from "http"
```
balazsorban44, line 1: Cannot use Node.js specific types

```suggestion
```

packages/core/src/lib/actions/env.ts
```diff
@@ -0,0 +1,54 @@
...
+      finalProvider.apiKey ??= envObject[`AUTH_${ID}_KEY`]
+    }
+    return finalProvider
+  })
+}
+
+export function createActionURL(
+  action: AuthAction,
+  protocol: string,
+  headers: IncomingHttpHeaders,
```
balazsorban44, line 43: We need to use runtime-agnostic alternatives

```suggestion
  headers: Headers,
```

packages/core/src/lib/actions/env.ts
```diff
@@ -0,0 +1,54 @@
...
+  basePath?: string
+): URL {
+  let url = process.env.AUTH_URL
+  if (!url) {
+    const host = headers["x-forwarded-host"] ?? headers.host
+    const proto = headers["x-forwarded-proto"] ?? protocol
```
balazsorban44, line 49:
```suggestion
    const host = headers["x-forwarded-host"] ?? headers.host
    const proto = headers["x-forwarded-proto"] ?? protocol
```
```suggestion
    const host = headers.get("x-forwarded-host") ?? headers.get("host")
    const proto = headers.get("x-forwarded-proto") ?? protocol
```

packages/core/src/lib/actions/env.ts
```diff
@@ -0,0 +1,54 @@
...
+  })
+}
+
+export function createActionURL(
+  action: AuthAction,
+  protocol: string,
+  headers: IncomingHttpHeaders,
+  basePath?: string
+): URL {
+  let url = process.env.AUTH_URL
```
balazsorban44, line 46:
```suggestion
  let url = envObject.AUTH_URL
```

Cannot leave `process` in the core lib, it needs to be fully runtime agnostic.

`createActionURL` will need an `envObject` argument similar to `setEnvDefaults`

balazsorban44 (changes requested):
Left some comments!

Can we also add tests for these functions?

I would not put it under the `actions` folder. Maybe https://github.com/nextauthjs/next-auth/tree/main/packages/core/src/lib/utils is a better place for these. I would then just re-export them from `@auth/core`.

ndom91:
@balazsorban44 okay awesome, thanks. I addressed all of the above. I was adding some tests but seemed to have some problems with the `beforeEach` not reseting the `authConfig` between each test :thinking: Will take another look later.

But the rest is pushed if you wanna take antoher look :+1:

packages/core/src/lib/utils/env.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export function createActionURL(
+  action: AuthAction,
+  protocol: string,
+  headers: Headers,
+  envObject: any,
+  basePath?: string
+): URL {
+  let url = envObject.AUTH_URL
```
ThangHuuVu, line 56: Should also honor `NEXTAUTH_URL`
```suggestion
  let url = envObject.AUTH_URL ?? envObject.NEXTAUTH_URL
```

ndom91:
Still have an issue with mismatching headers types with express, since its passing the node `IncomingHTTPHeaders`. I dont think those have the `.get`, etc. methods, let me double chekc..

PR #9829, opened 2024-01-28
fix: allow string return `signIn` callback

### LINKED ISSUE #9827: SignIn Callback doens't accept URL return anymore
### Environment

 System:
    OS: macOS 11.2.3
    CPU: (8) arm64 Apple M1
    Memory: 125.05 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 19.6.1 - ~/.nvm/versions/node/v19.6.1/bin/node
    Yarn: 1.22.10 - /opt/homebrew/bin/yarn
    npm: 9.4.0 - ~/.nvm/versions/node/v19.6.1/bin/npm
    pnpm: 8.10.2 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 121.0.6167.85
    Edge: 121.0.2277.83
    Safari: 14.0.3
  npmPackages:
    @auth/core: ^0.24.0 => 0.24.0 
    @auth/prisma-adapter: ^1.1.0 => 1.1.0 
    next: 14.0.3 => 14.0.3 
    next-auth: ^5.0.0-beta.5 => 5.0.0-beta.5 
    react: ^18 => 18.2.0 

### Reproduction URL

https://github.com/anest-on/aneston

### Describe the issue

I updated my Next-Auth to version 5 (beta), and since then, the SignIn Callback no longer accepts a URL return, it only accepts a boolean value.

Screen Shot 2024-01-27 at 19 15 38
 
In my case, I need that if the user doesn't check the option to grant access to the google calendar, it returns to an error page informing this, with the option for them to click again on the Google account link button and grant the connection. In version 4.24.5, which I was using before, I used to perform this return with a URL, and it worked perfectly.

The documentation about Auth and NextAuth says that I can continue using URL return in SignIn Callback, but it doens't allow it. When I compare the next-auth lib inside node modules I can see that in the older version the SignIn return accepted both String and Boolean, and in version 5 it only accepts Boolean values.

Return in SignIn on version 4.24.5:

Screen Shot 2024-01-27 at 19 23 28

Return in SignIn on version 4.24.5:

Screen Shot 2024-01-27 at 19 22 07

### How to reproduce

Go to http://localhost:3000/register and click on the button "Conectar".
Try to connect your account without giving permission for the application get access to calendar.
Your not going to be redirected to the URL informed.

### Expected behavior

It should be redirect to "http://localhost:3000/register/?error=permissions"

balazsorban44:
## ☕️ Reasoning

This is still a supported feature in `next-auth` v4 which was removed without a good reason and therefore should not break it. Re-adding for backwards compatibility

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes #9827

### DIFF
--- a/packages/core/src/lib/actions/callback/index.ts
+++ b/packages/core/src/lib/actions/callback/index.ts
@@ -390,5 +390,6 @@ async function handleAuthorized(
     throw new AuthorizedCallbackError(e as Error)
   }
   if (!authorized) throw new AuthorizedCallbackError("AccessDenied")
-  if (typeof authorized === "string") return await redirect({ url: authorized, baseUrl: config.url.origin })
+  if (typeof authorized !== "string") return
+  return await redirect({ url: authorized, baseUrl: config.url.origin })
 }
--- a/packages/core/src/lib/actions/callback/index.ts
+++ b/packages/core/src/lib/actions/callback/index.ts
@@ -101,14 +101,15 @@ export async function callback(
         })
       }
 
-      await handleAuthorized(
+      const redirect = await handleAuthorized(
         {
           user: userByAccount ?? userFromProvider,
           account,
           profile: OAuthProfile,
         },
-        options.callbacks.signIn
+        options
       )
+      if (redirect) return { redirect, cookies }
 
       const { user, session, isNewUser } = await handleLoginOrRegister(
         sessionStore.value,
@@ -220,7 +221,8 @@ export async function callback(
         provider: provider.id,
       }
 
-      await handleAuthorized({ user, account }, options.callbacks.signIn)
+      const redirect = await handleAuthorized({ user, account }, options)
+      if (redirect) return { redirect, cookies }
 
       // Sign user in
       const {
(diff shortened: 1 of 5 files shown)


packages/core/src/lib/actions/callback/index.ts
```diff
@@ -376,13 +379,17 @@ export async function callback(
...
-    const authorized = await signIn(params)
-    if (!authorized) throw new AuthorizedCallbackError("AccessDenied")
+    authorized = await signIn(params)
   } catch (e) {
     if (e instanceof AuthError) throw e
     throw new AuthorizedCallbackError(e as Error)
   }
+  if (!authorized) throw new AuthorizedCallbackError("AccessDenied")
+  if (typeof authorized !== "string") return
+  return await redirect({ url: authorized, baseUrl: config.url.origin })
```
ThangHuuVu, line 394: nit: 
```suggestion
  if (typeof authorized === "string") return await redirect({ url: authorized, baseUrl: config.url.origin })
```

Issue #9448, opened 2023-12-22
Next Auth [v5.0.0-beta.4] Middleware and Scope not working 


siinghd:
### Environment

System:
OS: Windows 11 10.0.22621
CPU: (8) x64 Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz
Memory: 2.78 GB / 11.92 GB
Binaries:
Node: 18.18.2 - C:\Program Files\nodejs\node.EXE
Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
npm: 10.2.5 - C:\Program Files\nodejs\npm.CMD
pnpm: 8.10.3 - C:\Program Files\nodejs\pnpm.CMD
Browsers:
Edge: Chromium (120.0.2210.77)
Internet Explorer: 11.0.22621.1
npmPackages:
@auth/core: ^0.19.0 => 0.19.0
@auth/prisma-adapter: ^1.0.12 => 1.0.12
next: 14.0.4 => 14.0.4
next-auth: 5.0.0-beta.4 => 5.0.0-beta.4
react: ^18.2.0 => 18.2.0

### Reproduction URL

https://github.com/siinghd/question-tracker

### Describe the issue

I am using the NextAuth beta to implement a login feature using the Discord provider and Prisma adapter. I am encountering two primary issues:

#### Issue 1: Scope Not Working as Expected
The specified scope in `auth.config.ts` doesn't seem to be applied. Instead, the default scope is used. Here's the relevant code snippet:

```typescript
const scopes = ['identify', 'guilds'];
export default {
  providers: [
    DiscordProvider({
      clientId: process.env.DISCORD_CLIENT_ID || '',
      clientSecret: process.env.DISCORD_CLIENT_SECRET || '',
      authorization: { params: { scope: scopes.join(' ') } },
    }),
  ],
};
```

#### Issue 2: Middleware Not Redirecting Properly
The middleware intended to protect the root path `/` and redirect unauthenticated users to `/login` is not functioning as expected. The `.env` file includes `AUTH_SECRET` and `AUTH_URL`.

### How to reproduce

1. Set up the NextAuth configuration with the Discord provider and Prisma adapter.
2. Define scopes as `['identify', 'guilds']`.
3. Implement middleware to protect the root path and redirect to `/login`.
4. Observe that the specified scopes are not applied and the middleware does not redirect as intended.

### Expected behavior

[...]

PR #9866, opened 2024-01-31
fix: allow Facebook authorization config to be merged

deronek:
## ☕️ Reasoning

Docs suggest that it's possible to easily extend Facebook provider options, due to deep merging of the provided properties with the defaults.
https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/providers/facebook.ts#L63-66
https://github.com/nextauthjs/next-auth/blob/f33a0ace66dab3523d9cd0883d65558b29837b7b/docs/docs/guides/providers/custom-provider.md?plain=1#L41-60

However, `authorization` property cannot be extended this way due to it being a hardcoded string. 
https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/providers/facebook.ts#L87

My PR includes a fix, converting the string into an object (satisfying `AdvancedEndpointHandler` interface), which is mergable in current implementation. Facebook authorization config can then be easily extended, as per the docs example.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues
None

## 📌 Resources

### DIFF
--- a/packages/core/src/providers/facebook.ts
+++ b/packages/core/src/providers/facebook.ts
@@ -84,7 +84,12 @@ export default function Facebook<P extends FacebookProfile>(
     id: "facebook",
     name: "Facebook",
     type: "oauth",
-    authorization: "https://www.facebook.com/v15.0/dialog/oauth?scope=email",
+    authorization: {
+      url: "https://www.facebook.com/v15.0/dialog/oauth",
+      params: {
+        scope: 'email',
+      },
+    },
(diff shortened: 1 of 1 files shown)


balazsorban44 (approved):
This is a temporary fix, but thanks!

The real fix would be to fix the deep merging. See also https://github.com/nextauthjs/next-auth/issues/9448#issuecomment-1909245481

Issue #9099, opened 2023-11-09
allow custom errors to propagate to the client from authorized callback


amit548:
### Environment

 System:
    OS: Windows 11 10.0.22631
    CPU: (12) x64 AMD Ryzen 5 5500U with Radeon Graphics
    Memory: 4.76 GB / 15.38 GB
  Binaries:
    Node: 20.9.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 9.8.1 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.9.2 - ~\AppData\Roaming\npm\pnpm.CMD
  Browsers:
    Edge: Chromium (119.0.2151.44)
    Internet Explorer: 11.0.22621.1

  npmPackages:
    next: ^14.0.1 => 14.0.1 
    next-auth: ^5.0.0-beta.3 => 5.0.0-beta.3 
    react: 18.2.0 => 18.2.0 

 npmPackages:
    @bomdi/codebox: ^2.0.0 => 2.0.0 
    @editorjs/attaches: ^1.3.0 => 1.3.0 
    @editorjs/checklist: ^1.5.0 => 1.5.0 
    @editorjs/code: ^2.8.0 => 2.8.0 
    @editorjs/delimiter: ^1.3.0 => 1.3.0 
    @editorjs/editorjs: ^2.28.2 => 2.28.2 
    @editorjs/embed: ^2.6.0 => 2.6.0 
    @editorjs/header: ^2.7.0 => 2.7.0 
    @editorjs/image: ^2.8.2 => 2.8.2 
    @editorjs/inline-code: ^1.4.0 => 1.4.0 
    @editorjs/link: ^2.5.0 => 2.5.0
    @editorjs/list: ^1.8.0 => 1.8.0
    @editorjs/marker: ^1.3.0 => 1.3.0
    @editorjs/nested-list: ^1.3.0 => 1.3.0
    @editorjs/paragraph: ^2.11.3 => 2.11.3
    @editorjs/quote: ^2.5.0 => 2.5.0
    @editorjs/raw: ^2.4.0 => 2.4.0
    @editorjs/simple-image: ^1.5.1 => 1.5.1
    @editorjs/table: ^2.2.2 => 2.2.2
    @editorjs/underline: ^1.1.0 => 1.1.0
    @editorjs/warning: ^1.3.0 => 1.3.0
[...]

PR #9871, opened 2024-02-01
feat: customizable `authorize()` error

balazsorban44:
This introduces a way to throw custom errors in the `authorize()` callback of the Credentials provider.

Any generic or sensitive server error will continue to return `error=Configuration` (full error is logged on the server), but for custom ones thrown from `authorize()` that extend `CredentialsSignin`, we will let the error propagate. Example:

```ts
import { CredentialsSignin } from "@auth/core/errors" // import is specific to your framework

class CustomError extends CredentialsSignin {
  code = "custom"
}

// ...
authorize() {
  // ...
  throw new CustomError()
}
```

Then, based on your framework, one of the following will happen:

client-side (`fetch` to the sign-in endpoint):

- get redirected to the signin page as such `/signin?error=CredentialsSignin&code=custom`
- if `redirect: false` was set on `signIn`, it returns the error and code properties.

server-side handled form actions:

- `CustomError` is thrown that you would need to handle in a catch block.

Notes:

- The name `CredentialsSignin` was chosen to be backwards compatible with NextAuth.js v4 https://github.com/nextauthjs/next-auth/blob/2072750c017c3d2a7f7374dd683055b79d331319/packages/next-auth/src/core/routes/callback.ts#L336
- I also renamed `AuthorizedCallbackError` to AccessDenied for the same reason
https://github.com/nextauthjs/next-auth/blob/2072750c017c3d2a7f7374dd683055b79d331319/packages/next-auth/src/core/routes/callback.ts#L99
- Support for framework integrations like `next-auth` will be added in a separate PR

Related: #9099 (see https://github.com/nextauthjs/next-auth/issues/9099#issuecomment-1913754338). This PR won't close that issue yet, a `next-auth` PR will be opened after this gets merged


packages/core/src/errors.ts
```diff
@@ -427,3 +441,21 @@ export class Verification extends AuthError {
...
 }
+
+// TODO: Review this list to make sure it's complete and we only send non-hinting errors to the client.
+const clientErrors = new Set<ErrorType>([
+  "CredentialsSignin",
+  "OAuthAccountNotLinked",
+  "OAuthCallbackError",
+  "OAuthSignInError",
+  "EmailSignInError",
+])
```
balazsorban44, line 452: It's important that we go through this list and include all non-sensitive error types here.

packages/core/src/errors.ts
```diff
@@ -427,3 +441,23 @@ export class Verification extends AuthError {
 export class MissingCSRF extends SignInError {
   static type = "MissingCSRF"
 }
+
+// TODO: Review this list to make sure it's complete and we only send non-hinting errors to the client.
+const clientErrors = new Set<ErrorType>([
```
ThangHuuVu, line 446: WDYT if we introduce a type `ClientErrorType` that contains AuthError the list and then above in `ErrorType` we can do:
```ts
type ErrorType = ServerConfigurationErrorType | ClientErrorType | SensitiveErrorType
type ServerConfigurationErrorType = 
  | "MissingAdapter"
  | "MissingAdapterMethods"
  | "MissingAuthorize"
  | "MissingSecret"
  // so on...
```


and then we can go through the smaller lists easier

balazsorban44: I did not want to many types to maintain, but maybe a `ClientErrorType` is a good idea!

ThangHuuVu: Let's just go ahead without it then, we already have this set which should be enough

ThangHuuVu (approved):

A7med3bdulBaset:
Any updates/plans on when this will be merged?

ndom91 (approved):
So in general LGTM. Curious though, why did you add `skip` to a few of the `core/index.test.ts` tests?

ThangHuuVu:
@ndom91 we will enable it again in my pr #100017 as I remember Balazs got some issues with the tests when he added the restart mock function

WINOFFRG:
Hi! Any updates/ETA on this? I am testing our migration to beta v5 and want to display custom error messages on signin flow from backend directly on the page without redirection.

MrOxMasTer:
How do I use it? I do not install the @auth/core package
image
image

MrOxMasTer:
> @MrOxMasTer do you mean how do you use this new feature? Or `next-auth` in general?

How do I use this feature in next.js, which I am currently on pull request. I couldn't find this class anywhere.

ndom91:
@MrOxMasTer you'd be surprised where people ask what :joy:

Anyway, sorry about that! So yeah the `CredentialsSignin` seems to only be exported from `@auth/core/errors` at the moment, which is not ideal as you're not supposed to install that package explicitly. 

I've opened a PR reexporting it in the framework packages here: https://github.com/nextauthjs/next-auth/pull/10200

It also updates the docs example updated in this PR to show throwing the new custom error

MrOxMasTer:
> I've opened a PR reexporting it in the framework packages here: #10200
> 
> It also updates the docs example updated in this PR to show throwing the new custom error

Thank you, I pray for you 🛐🛐🛐

ndom91:
@MrOxMasTer the changes have been merged to `main`. You can now import `CredentialsSignin` from `next-auth`, for example.

ccyen8358:
Hi, I tried the `CustomError` solution suggested by @balazsorban44
```typescript
export class CustomError extends CredentialsSignin {
  code = "custom";
}
```
However, my custom error message will always be appended with an additional string of `" .Read more at https://errors.authjs.dev#credentialssignin"`, which I don't want to be sent to client.

Below is the screenshot of what my `CustomError` instance looks like in debugger:
custom_error

Upon inspecting the source code, it seems the string is appended in the constructor of `AuthError` class, as shown in below link:
https://github.com/nextauthjs/next-auth/blob/239dfcf71fb3267272e2d972f9551f6f8aea0105/packages/core/src/errors.ts#L75-L76

Currently I have to workaround it by adding a constructor to my `CustomError` class to get rid of the unwanted appended string:
```typescript
export class CustomError extends CredentialsSignin {
  code = "custom";
  constructor(message?: any, errorOptions?: any) {
    super(message, errorOptions);
    this.message = message;
  }
}
```
Is there any more elegant way to solve it?
Thanks in advance.

A7med3bdulBaset:
I think you shouldn't depend on `error.code` not `error.message`

sjoukedv:
@balazsorban44 this has broken the redirection for me. Instead of `/<my-signin-page>?error=CredentialsSignin` this now redirects to `/api/auth/<my-signin-page>?error=CredentialsSignin&code=<code>` which returns "Bad Request" 

My config:
```javascript
{
    pages: {
        signIn: '/<my-signin-page>',
        error: '/<my-signin-page>', // Error code passed in query string as ?error=
    },
    ...
    basePath: "/api/auth",
}
```

on build 13 of the beta this works as expected. I don't see any relevant changes in the regexp matching so am curious why this stopped working.

A7med3bdulBaset:
> @balazsorban44 this has broken the redirection for me. Instead of `/<my-signin-page>?error=CredentialsSignin` this now redirects to `/api/auth/<my-signin-page>?error=CredentialsSignin&code=<code>` which returns "Bad Request"
> 
> My config:
> 
> ```js
> {
>     pages: {
>         signIn: '/<my-signin-page>',
>         error: '/<my-signin-page>', // Error code passed in query string as ?error=
>     },
>     ...
>     basePath: "/api/auth",
> }
> ```
> 
> on build 13 of the beta this works as expected. I don't see any relevant changes in the regexp matching so am curious why this stopped working.

Because of `basePath: "/api/auth"`

sjoukedv:
> > @balazsorban44 this has broken the redirection for me. Instead of `/<my-signin-page>?error=CredentialsSignin` this now redirects to `/api/auth/<my-signin-page>?error=CredentialsSignin&code=<code>` which returns "Bad Request"
> > My config:
> > ```js
> > {
> >     pages: {
> >         signIn: '/<my-signin-page>',
> >         error: '/<my-signin-page>', // Error code passed in query string as ?error=
> >     },
> >     ...
> >     basePath: "/api/auth",
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > on build 13 of the beta this works as expected. I don't see any relevant changes in the regexp matching so am curious why this stopped working.
> 
> Because of `basePath: "/api/auth"`

Both with `basePath` not set and  `basePath: "/"` it's giving the same behaviour.

ndom91:
@ccyen8358 hmm interesting, so I can't htink of any other way to solve it currently. Other than adding an additional custom error class designed to be extended from for this use-case :thinking:  I've put up a quick prototype PR of that [here](https://github.com/nextauthjs/next-auth/pull/10231) to hopefully get some discussion going there :pray: 

@sjoukedv hmm interesting, do you have an `AUTH_URL` env var set as well? If so, what to? iirc there shouldn't be any changes related to that that shipped recently, as we're working on soemthing very similar (regarding `basePath`) [here](https://github.com/nextauthjs/next-auth/pull/10094) as we speak.

sjoukedv:
> @sjoukedv hmm interesting, do you have an `AUTH_URL` env var set as well? If so, what to? iirc there shouldn't be any changes related to that that shipped recently, as we're working on soemthing very similar (regarding `basePath`) [here](https://github.com/nextauthjs/next-auth/pull/10094) as we speak.


```javascript
// path next.config.js

/** @type {import("next").NextConfig} */
module.exports = {
    env: {
        AUTH_URL: `${process.env.VERCEL_URL || 'http://localhost:3000'}/api/auth`,
        NEXTAUTH_URL: `${process.env.VERCEL_URL || 'http://localhost:3000'}/api/auth`,
        AUTH_SECRET: `${process.env.AUTH_SECRET || 'somestring'}`,
    },
    // ...
```

```javascript
// path auth.ts

export const config = {
    // ...
    basePath: "/api/auth",
    trustHost: true
} satisfies NextAuthConfig

export const { handlers, auth, signIn, signOut } = NextAuth(config)
```

With e.g. `VERCEL_URL=http://127.0.0.1:3000`

sjoukedv:
> @sjoukedv hmm okay so we auto detect the basePath based on VERCEL_URL (among other things), so if that's available even in dev too, like you've set it yourself there, then you don't need a lot of this actually.
> 
> Can you try removing auth_url and nextauth_url env vars and basePath config?

@ndom91 That redirects me to `api/auth/<custom-page>?error=CredentialsSignin&code=<my-custom-error>`.

Looks like the page becomes relative to `/api/auth` instead of the root

```
    pages: {
        signIn: '/<custom-page>',
        error: '/<custom-page>', // Error code passed in query string as ?error=
    },
 ```

If I remove the leading slash from`pages.error` I get redirected to `api/auth<custom-page>?error=CredentialsSignin&code=<my-custom-error>` no `/` after `auth`. Is the URL properly parsed to be an absolute path instead of relative to `/api/auth`?
 
I have also verified the `authorized` callback is not the issue by always returning true.

**Comment by CyanFlare** [comment]
> > @sjoukedv hmm okay so we auto detect the basePath based on VERCEL_URL (among other things), so if that's available even in dev too, like you've set it yourself there, then you don't need a lot of this actually.
> > Can you try removing auth_url and nextauth_url env vars and basePath config?
> 
> @ndom91 That redirects me to `api/auth/<custom-page>?error=CredentialsSignin&code=<my-custom-error>`.
> 
> Looks like the page becomes relative to `/api/auth` instead of the root
> 
> ```
>     pages: {
>         signIn: '/<custom-page>',
>         error: '/<custom-page>', // Error code passed in query string as ?error=
>     },
> ```
> 
> If I remove the leading slash from`pages.error` I get redirected to `api/auth<custom-page>?error=CredentialsSignin&code=<my-custom-error>` no `/` after `auth`. Is the URL properly parsed to be an absolute path instead of relative to `/api/auth`?
> 
> I have also verified the `authorized` callback is not the issue by always returning true.

I think I'm running into a similar issue here.

This is what I currently have for my pages setting
```typescript
pages: {
  signIn: `/auth`,
  error: `/auth/error`,
},
```

When user gets redirected to error the resulting redirect redirects to ``/api/auth/auth/error`` instead of ``/auth/error``. This started to happened after I upgraded from ``next-auth@5.0.0-beta.4`` to ``next-auth@5.0.0-beta.15``

nikcio:
I've tried adding a custom error but we're using the `signIn` function from `packages/next-auth/src/react.tsx` [Source](https://github.com/nextauthjs/next-auth/blob/5ea8b7b0f4d285e48f141dd91e518c905c9fb34e/packages/next-auth/src/react.tsx#L214) and here the `code` isn't available in the response. 

Is it possible to have it added to the response?

I believe it should just be adding the following:

```diff
  const error = new URL(data.url).searchParams.get("error")
+ const code = new URL(data.url).searchParams.get("code")

  if (res.ok) {
    await __NEXTAUTH._getSession({ event: "storage" })
  }

  return {
    error,
+   code,
    status: res.status,
    ok: res.ok,
    url: error ? null : data.url,
  } as any
```
[Source](https://github.com/nextauthjs/next-auth/blob/5ea8b7b0f4d285e48f141dd91e518c905c9fb34e/packages/next-auth/src/react.tsx#L281-L293)

ndom91:
@CyanFlare @sjoukedv I think both of these are due to an issue we had with `AUTH_URL`/`basePath`. This shuold be fixed in `main` I believe, unfortunately we haven't cut a release yet though. If you're up for testing it in `main` we'd love some more feedback!

MrOxMasTer:
next-auth@beta-16 / next.js 14.1.4. A strange error occurs when imorting a classnext-auth@beta-16 / next.js 14.1.4. A strange error occurs when imorting a class

https://github.com/nextauthjs/next-auth/assets/59291123/41a1d677-9122-49f4-bf10-514ed44666c8

ndom91:
@MrOxMasTer Hmm so it's working "as expected" with `AuthError`, but when yuo extend it from `CredentialsSignin` you get this odd next/fonts error, did I understand that correctly?

MrOxMasTer:
> @MrOxMasTer Hmm so it's working "as expected" with `AuthError`, but when yuo extend it from `CredentialsSignin` you get this odd next/fonts error, did I understand that correctly?

yes

sjoukedv:
With `beta16` this is working for me now (with Next.js 14.2.0-canary.48)

```js
import { CredentialsSignin } from "@auth/core/errors" 

class InvalidLoginError extends CredentialsSignin {
    code = 'Invalid identifier or password'
}



export const config = {
    providers: [
        CredentialsProvider({
            ....
               throw new InvalidLoginError()
            }
        })
    ]
```

MrOxMasTer:
> With `beta16` this is working for me now (with Next.js 14.2.0-canary.48)


Look where I'm importing the class from. It was moved to the main library to avoid downloading "@auth/core", but it doesn't work. But I can say that you have made a good point, that at least some implementation of the class works

ndom91:
@MrOxMasTer Was just trying something with this again in `beta.16` and wanted to report that the above did work for me with `import { CredentialsSignin } from 'next-auth'`, but with `AuthError` it cuased the FE to redirect to a configuration error, while the BE still did log my custom error message / code.

I can't view your video anymore, but it cause anything else weird to happen with `next@14.2.0-canary.58`

LakshanKarunathilake:
> I've tried adding a custom error but we're using the `signIn` function from `packages/next-auth/src/react.tsx` [Source](https://github.com/nextauthjs/next-auth/blob/5ea8b7b0f4d285e48f141dd91e518c905c9fb34e/packages/next-auth/src/react.tsx#L214) and here the `code` isn't available in the response.
> 
> Is it possible to have it added to the response?
> 
> I believe it should just be adding the following:
> 
> ```diff
>   const error = new URL(data.url).searchParams.get("error")
> + const code = new URL(data.url).searchParams.get("code")
> 
>   if (res.ok) {
>     await __NEXTAUTH._getSession({ event: "storage" })
>   }
> 
>   return {
>     error,
> +   code,
>     status: res.status,
>     ok: res.ok,
>     url: error ? null : data.url,
>   } as any
> ```
> 
> [Source](https://github.com/nextauthjs/next-auth/blob/5ea8b7b0f4d285e48f141dd91e518c905c9fb34e/packages/next-auth/src/react.tsx#L281-L293)

We implemented the login page based on the intercepting logic so we can not enable the redirection. As redirection will move out of the intercepted modal. I want to stick in the same modal. Is it possible to attach the code when redirection is disabled

alejandr0pg:
> > I've tried adding a custom error but we're using the `signIn` function from `packages/next-auth/src/react.tsx` [Source](https://github.com/nextauthjs/next-auth/blob/5ea8b7b0f4d285e48f141dd91e518c905c9fb34e/packages/next-auth/src/react.tsx#L214) and here the `code` isn't available in the response.
> > Is it possible to have it added to the response?
> > I believe it should just be adding the following:
> > ```diff
> >   const error = new URL(data.url).searchParams.get("error")
> > + const code = new URL(data.url).searchParams.get("code")
> > 
> >   if (res.ok) {
> >     await __NEXTAUTH._getSession({ event: "storage" })
> >   }
> > 
> >   return {
> >     error,
> > +   code,
> >     status: res.status,
> >     ok: res.ok,
> >     url: error ? null : data.url,
> >   } as any
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > [Source](https://github.com/nextauthjs/next-auth/blob/5ea8b7b0f4d285e48f141dd91e518c905c9fb34e/packages/next-auth/src/react.tsx#L281-L293)
> 
> We implemented the login page based on the intercepting logic so we can not enable the redirection. As redirection will move out of the intercepted modal. I want to stick in the same modal. Is it possible to attach the code when redirection is disabled
> 
> i also need the above change

Es necesario poder obtener el codigo para diferenciar del error.

packages/core/src/index.ts
```diff
@@ -167,23 +176,19 @@ export async function Auth(
...
+    if (error instanceof CredentialsSignin) params.set("code", error.code)
 
-    const url = `${internalRequest.url.origin}${path}?${params}`
+    const pageKind = (isAuthError && error.kind) || "error"
+    const pagePath = config.pages?.[pageKind] ?? `/${pageKind.toLowerCase()}`
+    const url = `${internalRequest.url.origin}${config.basePath}${pagePath}?${params}`
 
     if (isRedirect) return Response.json({ url })
-
     return Response.redirect(url)
```
andresgutgon, line 190: Hi, I'm having trouble managing to get the `res.redirect` URL when using a server action in a magic link

My action looks like this in my custom signin page
``` javascript
import { signIn } from '$/auth'

      <form
        action={async (formData) => {
          'use server'
          await signIn('nodemailer', {
            callbackUrl,
            email: formData.get('email')?.toString()!,
          })
        }}
      >
```

andresgutgon: Debugging I see the issue is on `@auth/core` in this package in combination with `next-auth` package. This `Response.redirect(url)` is received here:
https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/lib/actions.ts#L76

```javascript
  const res = await Auth(req, { ...config, raw, skipCSRFCheck })

  // ... other code

  if (shouldRedirect) return redirect(res.redirect!)
```

This is my case. `res.redirect` is `undefined` 
and I see this when an error occur with Nodemailer (I misconfigure it to see this error)

image

andresgutgon: I added a `logger.debug` and I can see that the right Error URL is there in this file but when doing `Response.redirect(url)` is received a `undefenid` in `next-auth/src/lib/actions.ts`

dwalker93:
[#10943 comment](https://github.com/nextauthjs/next-auth/pull/10943#issuecomment-2143943734)

mohdfaizan5:
> With `beta16` this is working for me now (with Next.js 14.2.0-canary.48)
> 
> ```js
> import { CredentialsSignin } from "@auth/core/errors" 
> 
> class InvalidLoginError extends CredentialsSignin {
>     code = 'Invalid identifier or password'
> }
> 
> 
> 
> export const config = {
>     providers: [
>         CredentialsProvider({
>             ....
>             async authorize(credentials) {
>                throw new InvalidLoginError()
>             }
>         })
>     ]
> ```

but the problem I came across implementing your solution was I'm not able to use the package from "@auth/core/errors"
instead I tried from 
```ts
import { CredentialsSignin } from "next-auth";
```

which is also not giving me expected results, 

thankyou in advance for the answer, I'd appreciate if you could tell me the complete way to do custom error

rmartin93:
> Hi, I tried the `CustomError` solution suggested by @balazsorban44
> 
> ```ts
> export class CustomError extends CredentialsSignin {
>   code = "custom";
> }
> ```
> 
> However, my custom error message will always be appended with an additional string of `" .Read more at https://errors.authjs.dev#credentialssignin"`, which I don't want to be sent to client.
> 
> Below is the screenshot of what my `CustomError` instance looks like in debugger (my custom error message is just "Login failed"): custom_error
> 
> Upon inspecting the source code, it seems the string is appended in the constructor of `AuthError` class, as shown in below link:
> 
> https://github.com/nextauthjs/next-auth/blob/239dfcf71fb3267272e2d972f9551f6f8aea0105/packages/core/src/errors.ts#L75-L76
> 
> Currently I have to workaround it by adding a constructor to my `CustomError` class to overwrite the unwanted appended string:
> 
> ```ts
> export class CustomError extends CredentialsSignin {
>   code = "custom";
>   constructor(message?: any, errorOptions?: any) {
>     super(message, errorOptions);
>     this.message = message;
>   }
> }
> ```
> 
> Is there any elegant way to solve it? Thanks in advance.

This saved me! I was deep in a rabbit hole trying to get a custom error to show to a user if they had logged in with oAuth but then tried logging in with a password and the message had that extra bit on it. Thank you!!!

domkoder:
@rmartin93 please can you shear a code snippet of your implementation of this

Pirulax:
I don't understand how the `redirect: false` option works.
I still only get an `url` in the response, rather than a `code` (and other fields) as the docs and you suggest.
This is what I get with `redirect: false`:
```ts
{
    "url": "http://localhost:3000/login?error=CredentialsSignin&code=InvalidCredentials"
}
```
Is there a way to just get the `code` and other stuff as JSON?

Pirulax:
Hi, it's me again...
In production is `code` supposed to be whatever I provide in my custom error?
Because in dev mode it's all good, but then in prod `code` isn't what's in my custom error, but rather simply `credentials`:
`https://538842fee68c:3000/login?error=CredentialsSignin&code=credentials`
But in dev mode, it's as expected:
`http://localhost:3000/login?error=CredentialsSignin&code=InvalidCredentials`
Though, do note, that the env is not the same, is it possibly because of that? (one's running on the server, the other one (dev) is on my local machine)

hamzaxoxo:
I was wondering if there's a way to add a custom message to the authentication error object without using a switch statement for different error types. Here's how I'm currently handling it. Any suggestions for improvement

```
import { CredentialsSignin } from "next-auth"

class invalidPassword extends CredentialsSignin {
    code = "invalidPassword"
}
class UserNotFound extends CredentialsSignin {
    code = "UserNotFound"
}

export { invalidPassword, UserNotFound }

export function getAuthErrorMessage(errorCode?: string): string {
    switch (errorCode) {
        case "UserNotFound":
            return "User not found. Please check your email and password.";
        case "invalidPassword":
            return "Invalid password. Please try again.";
        default:
            return "Sign in failed. Please try again.";
    }
}
```

`auth.ts`
```
try {
                    if (!email || !password) {
                        throw new Error("Please fill in all fields");
                    }
                    const [rows]: any = await pool.query("SELECT * FROM users WHERE email = ?", [email]);
                    if (!rows.length) {
                        throw new UserNotFound();
                    }
                    const user = rows[0];
                    const isMatched = await bcrypt.compare(password, user.password);
                    if (!isMatched) {
                        throw new invalidPassword();
                    }
                    return user;
                } catch (error) {
                    throw error;
                }
```

```
 const handleSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
        e.preventDefault();
        setIsLoading(true);

            const res = await signIn('credentials', {
                email,
                password,
            });

            if (res?.error) {
                const message = getAuthErrorMessage(res.code);
                toast.error(message);
            } else {
                toast.success("Signed in successfully!");
            }
        }
}
```

PR #9885, opened 2024-02-03
Replace UUID by nanoid in the drizzle adapter.

AMR-21:
Nothing critical, just replacing every use of `crypto.randomUUID` in the drizzle adapter by `nanoid()`.


ndom91:
What's the thinking behind this change? We'd prefer not to add another dependency for something that's covered by the platform

balazsorban44:
Closing as per https://github.com/nextauthjs/next-auth/pull/9885#issuecomment-1925240223. We don't want to introduce unnecessary dependencies.

PR #9896, opened 2024-02-04
refactor(core): export `isAuthAction`

ThangHuuVu:
- isAuthAction is also used in `@auth/sveltekit` so this PR exports it to reuse . see #9714


packages/core/src/index.ts
```diff
@@ -320,7 +321,8 @@ export interface AuthConfig {
    * ```ts
    * // /pages/api/auth/[...nextauth].js
    * import log from "logging-service"
-   * export default NextAuth({
+   * import { isAuthAction } from './lib/utils/actions';
```
ThangHuuVu, line 324: weird auto import 😄

PR #9924, opened 2024-02-06
feat(sveltekit): webauthn provider support

ndom91:
## ☕️ Reasoning

- Copies the changes from the `next-auth` client
- Also made a new `signIn` import which includes the `webauthn` import
- Also bumped to use `@simplewebauth/server@9.0.2` which should fix the `import node:crypto` issue on edge runtimes
- The unstorage "DB" is using the "filesystem" driver, meaning it'll write its "db" to `/apps/dev/sveltekit/tmp-unstorage`. This makes the saving of your webauthn authenticators persist across restarts of your dev server, like HMR, etc. The unstorage tmp dir is a part of the `.gitignore` :+1: 

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


packages/frameworks-sveltekit/package.json
```diff
@@ -47,9 +47,23 @@
...
+  "peerDependenciesMeta": {
+    "nodemailer": {
+      "optional": true
+    },
+    "@simplewebauthn/browser": {
+      "optional": true
+    },
+    "@simplewebauthn/server": {
+      "optional": true
+    }
```
balazsorban44, line 65:
```suggestion
    "@simplewebauthn/browser": {
      "optional": true
    },
    "@simplewebauthn/server": {
      "optional": true
    },
    "nodemailer": {
      "optional": true
    }
```

packages/frameworks-sveltekit/src/lib/client.ts
```diff
@@ -31,6 +32,40 @@ export type SignInAuthorizationParams =
...
+ *
+ * @param providerId provider ID
+ * @param options SignInOptions
+ * @returns WebAuthn response or error
+ */
+async function webAuthnOptions(providerId: string, options?: SignInOptions) {
+  const { startAuthentication, startRegistration } = await import(
+    "@simplewebauthn/browser"
+  )
+  const baseUrl = `${base}/auth/`
```
balazsorban44, line 47: What if I set a different `basePath` in the `SvelteKitAuthConfig`?

packages/frameworks-sveltekit/src/lib/client.ts
```diff
@@ -31,6 +32,40 @@ export type SignInAuthorizationParams =
...
+ * Returns either the completed WebAuthn response or an error request.
+ *
+ * @param providerId provider ID
+ * @param options SignInOptions
+ * @returns WebAuthn response or error
+ */
+async function webAuthnOptions(providerId: string, options?: SignInOptions) {
+  const { startAuthentication, startRegistration } = await import(
+    "@simplewebauthn/browser"
+  )
```
balazsorban44, line 46: I still think we should follow the same pattern as for `next-auth` for consistency. `@auth/sveltekit/webautn` for now. Could you do a bundle analysis and verify that if I don't use WebAuthn, my client-side bundle won't include this import?

balazsorban44 (dismissed):

pnpm-lock.yaml
```diff
@@ -680,6 +680,15 @@ importers:
       '@auth/core':
         specifier: workspace:*
         version: link:../core
+      '@simplewebauthn/browser':
+        specifier: v9.0.0
+        version: 9.0.0
+      '@simplewebauthn/server':
+        specifier: v9.0.0
+        version: 9.0.0
```
balazsorban44, line 688: Should have not been added, `9.0.1` is already in the lockfile/core/next-auth libs. 

Make sure to add `@simplewebauthn/browser` and `@simplewebauthn/server` to dev dependencies with `^9.0.1`

ndom91: Fixed

packages/frameworks-sveltekit/src/lib/client.ts
```diff
@@ -3,6 +3,7 @@ import type {
   RedirectableProviderType,
 } from "@auth/core/providers"
 import { base } from "$app/paths"
+import type { WebAuthnOptionsResponseBody } from "@auth/core/lib/utils/webauthn-utils"
```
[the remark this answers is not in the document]
ndom91: Yeah was having trouble with this, the import worked in `next-auth` but not here for some reason :thinking: Part of the reason this was still draft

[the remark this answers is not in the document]
ndom91: Fixed :+1:

packages/frameworks-sveltekit/src/lib/client.ts
ndom91: Okay, did the same thing as with `next-auth` for now. A new `@auth/sveltekit/webauthn` exported `signIn`. It also still has the logic for the other signin methods so users can only use that 1 and don't need to juggle 2 if they have webauthn + oauth (for example) providers in their app, just like in `next-auth` :+1:

packages/frameworks-sveltekit/src/lib/client.ts
ndom91: Good question.. was giving this some thought and the `SvelteKitAuthConfig` and sveltekit set `base` aren't necessarily in sync.

So I was thinking about implementing something similar to what we have in `next-auth` with the `apiBaseUrl` helper that parses the `__NEXTAUTH` config object throughout. What do you think?

packages/frameworks-sveltekit/src/lib/webauthn.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/**
+ * Client-side method to initiate a webauthn signin flow
+ * or send the user to the signin page listing all possible providers.
+ * Automatically adds the CSRF token to the request.
+ *
+ * [Documentation](https://authjs.dev/reference/sveltekit/client#signin)
+ */
+export async function signIn<
```
ThangHuuVu, line 50: hmm, I can't remember why we didn't edit the `signIn` function but to create a new one when we merged for `next-auth` 🤔 do you think we should update the `signIn` function in `client` instead?

ndom91: Yeah good idea, lets do that in a follow up. We can check out the svelte + next.js use of that additional `signIn` method and potentially iterate on those in a follow-up PR

ThangHuuVu (approved):

packages/frameworks-sveltekit/src/lib/webauthn.ts
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Fetch webauthn options from server and prompt user for authentication or registration.
+ * Returns either the completed WebAuthn response or an error request.
+ *
+ * @param providerId provider ID
+ * @param options SignInOptions
+ * @returns WebAuthn response or error
+ */
+async function webAuthnOptions(providerId: string, options?: SignInOptions) {
+  const baseUrl = `${base}/auth/`
```
sillvva, line 21: There's a mistake here with the ending `/`. Below an additional `/` is added after `baseUrl`.

packages/frameworks-sveltekit/src/lib/webauthn.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  // @ts-expect-error
+  const params = new URLSearchParams(options)
+
+  const optionsResp = await fetch(
+    `${baseUrl}/webauthn-options/${providerId}?${params}`
```
sillvva, line 27: Is there supposed to be a hook here? This URL is returning HTML, not JSON.

CleanShot 2024-05-17 at 09 54 56@2x

sillvva: Looks like there is an error in the console 
> InvalidProvider: Provider must be WebAuthn. Read more at https://errors.authjs.dev#invalidprovider

packages/frameworks-sveltekit/src/lib/webauthn.ts
ndom91: @sillvva thanks for pointing this out, it's been fixed now :+1:

gatleas17 (commented):
Thx

PR #9932, opened 2024-02-06
fix: mismatched `account.providerAccountId`

### LINKED ISSUE #9878: Bug causing OAuthAccountNotLinked (identified in a recent commit)
### Environment

System:
OS: macOS 13.5
CPU: (12) arm64 Apple M2 Pro
Memory: 86.23 MB / 32.00 GB
Shell: 5.9 - /bin/zsh
Binaries:
Node: 20.5.1 - ~/.nvm/versions/node/v20.5.1/bin/node
npm: 9.8.0 - ~/.nvm/versions/node/v20.5.1/bin/npm
pnpm: 8.7.0 - ~/.nvm/versions/node/v20.5.1/bin/pnpm
Browsers:
Chrome: 121.0.6167.139
Safari: 16.6

### Reproduction URL

https://github.com/kevinpiac/next-auth-providerAccountId-issue

### Describe the issue

As a user, when I try to sign in for the second time using an OAuth or OIDC provider such as Google for example, I keep getting the `OAuthAccountNotLinked` error even if I'm using the same Provider and the same account.

It seems this bug was introduced by this recent fix: https://github.com/nextauthjs/next-auth/pull/9793

This fix always overrides the `user.id` to `crypto.randomUUID()`
Then it sets the `providerAccountId` to `user.id` instead of using the id returned by the `profile callback` as specified in the documentation (see below).

This causes the providerAccountId to be randomly generated every time someone tries to login and will always cause OAuthAccountNotLinked error.

The docs say:

```javascript
 /**
   * This value depends on the type of the provider being used to create the account.
   * - oauth/oidc: The OAuth account's id, returned from the `profile()` callback.
   * - email: The user's email address.
   * - credentials: `id` returned from the `authorize()` callback
   */
```   
   
   
______

For more context, see the two incriminated lines below:

```typescript
// core/src/lib/actions/callback/oauth/callback.ts


async function getUserAndAccount(
  OAuthProfile: Profile,
  provider: OAuthConfigInternal<any>,
  tokens: TokenSet,
  logger: LoggerInstance
) {
  try {
    const userFromProfile = await provider.profile(OAuthProfile, tokens)
    const user = {
      ...userFromProfile,
      id: crypto.randomUUID(), << INCRIMINATED LINE
      email: userFromProfile.email?.toLowerCase(),
    } satisfies User

    return {
      user,
      account: {
        ...tokens,
        provider: provider.id,
        type: provider.type,
        providerAccountId: user.id.toString(), << INCRIMINATED LINE
      },
    }
  } catch (e) {
    // If we didn't get a response either there was a problem with the provider
    // response *or* the user cancelled the action with the provider.
    //
    // Unfortunately, we can't tell which - at least not in a way that works for
    // all providers, so we return an empty object; the user should then be
    // redirected back to the sign up page. We log the error to help developers
    // who might be trying to debug this when configuring a new provider.
    logger.debug("getProfile error details", OAuthProfile)
    logger.error(
      new OAuthProfileParseError(e as Error, { provider: provider.id })
    )
  }
}

```

### How to reproduce

In summary :

1. Use a database adapter such as Prisma.
2. Use any OIDC provider such as Google.
3. Login twice with the same account.

More detail :

- On the first sign-in an `Account` is persisted using the adapter with a `providerAccountId` that is not corresponding to the id field returned by the profile callback. It uses a random UUID instead.

- So, the second time the user tries to sign in, it triggers an `OAuthAccountNotLinked` error because the adapter tries to find the matching `Account` using a new random UUID, causing the `OAuthAccountNotLinked` to throw.

### Expected behavior

It should follow the documentation, and the providerAccountId should be equal to the result of the user.id returned by the profile callback.

```javascript
 /**
   * This value depends on the type of the provider being used to create the account.
   * - oauth/oidc: The OAuth account's id, returned from the `profile()` callback.
   * - email: The user's email address.
   * - credentials: `id` returned from the `authorize()` callback
   */
```

ndom91:
## ☕️ Reasoning

- #9793 set the new `user.id: UUID` to `account.providerAccountId` as well. This caused a mismatch for folks who already had an `account` table full of original `providerAccountId`'s.
- Now sets the original provider's accountId to our `account.providerAccountId`
- Also adds a regression test

Fixes #9878

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


packages/core/src/lib/actions/callback/oauth/callback.ts
```diff
@@ -207,7 +207,7 @@ async function getUserAndAccount(
         ...tokens,
         provider: provider.id,
         type: provider.type,
-        providerAccountId: user.id.toString(),
+        providerAccountId: userFromProfile.id?.toString()!,
```
yordis, line 210:
```suggestion
        providerAccountId: userFromProfile.id,
```

The `ProfileCallback` returns a `User` which expects the `id` to be string already.

packages/core/test/authorize.test.ts
```diff
@@ -67,6 +69,30 @@ describe("auth via callbacks.signIn", () => {
     })
   })
 
+  describe("oauth callback", () => {
```
balazsorban44, line 72: Let's replace this with a proper E2E test in a follow-up PR instead.

packages/core/src/lib/actions/callback/oauth/callback.ts
ndom91: Yeah good catch, the `id` was typed `string | undefined` though so we added a fallback to `randomUUID()` :+1:

yordis:
I recommend cutting off a release for this one; I have seen a few discussions related to this issue already.

PR #9935, opened 2024-02-06
chore: improve Zitadel Provider

yordis:
## ☕️ Reasoning

All the other Providers are using `PascalCase` while this one is all uppercase, causing some IDEs to import the default export as `ZITADEL` instead of `Zitadel`

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/packages/core/src/providers/zitadel.ts
+++ b/packages/core/src/providers/zitadel.ts
@@ -111,7 +111,7 @@ export interface ZitadelProfile extends Record<string, any> {
  *
  * :::
  */
-export default function ZITADEL<P extends ZitadelProfile>(
+export default function Zitadel<P extends ZitadelProfile>(
   options: OAuthUserConfig<P>
 ): OIDCConfig<P> {
   return {


packages/core/src/providers/zitadel.ts
```diff
@@ -111,13 +111,21 @@ export interface ZitadelProfile extends Record<string, any> {
...
   return {
     id: "zitadel",
     name: "ZITADEL",
     type: "oidc",
+    profile(profile: P) {
+      return {
+        id: profile.sub,
+        name: profile.name,
+        email: profile.email,
+        image: profile.picture,
```
yordis, line 126: 🙏🏻

yordis: @balazsorban44, out of curiosity, is there any reason not to rename `id` to `sub` and `image` to `picture` for the incoming v5?

Being close to OIDC lingos will be much better in the long term since it is established spec, and this breaking change is the best opportunity for it.

Thoughts?

packages/core/src/providers/zitadel.ts
```diff
@@ -111,13 +111,21 @@ export interface ZitadelProfile extends Record<string, any> {
...
  */
-export default function ZITADEL<P extends ZitadelProfile>(
+export default function Zitadel<P extends ZitadelProfile>(
   options: OAuthUserConfig<P>
 ): OIDCConfig<P> {
   return {
     id: "zitadel",
     name: "ZITADEL",
     type: "oidc",
+    profile(profile: P) {
```
yordis, line 121: Is this require at all? Is by default doing this exact mapping?

ThangHuuVu: It's not required, you can take a look at the other OIDC providers like Auth0 for reference 🙌

packages/core/src/providers/zitadel.ts
ThangHuuVu: IMO, it's an unnecessary breaking change, and not everyone is OIDC compliant.

ThangHuuVu (changes requested):
You can remove the default profile callback, its not needed. Thanks!

yordis:
@ThangHuuVu ready, I removed the extra stuff

ThangHuuVu:
After looking at this PR carefully, I think there's no need for this change - this is a default export, therefore the import destination can easily use `PascalCase` if needed. ZITADEL is branded as all upper case letters, so I think let's keep it, same as the VK provider, for example 😄 nice effort, thanks anyway!

PR #9944, opened 2024-02-07
chore: fix E2E `@playwright/test` tests

ndom91:
## ☕️ Reasoning

- Update `release.yml` GHA workflow 
	- Cache playwright binaries to reduce runtime
	- Correctly upload traces upon failure
	- Run vitest and playwright tests in two separate steps
- Fix `playwright.config.ts` with local webserver running and remove hardcoded `/core` in `testDir`
- Cleaned up the first e2e `basic-auth.spec.ts` test
	- Includes `test.step()` to separate login / logout
	- Added simple test with `Credentials` provider to ensure test was working

### Notes
- Does NOT inclnude anything related to coverage
	- Found some useful blog posts for follow-up PRs though
	- https://dev.to/anishkny/code-coverage-for-a-nextjs-app-using-playwright-tests-18n7
	- https://www.npmjs.com/package/playwright-test-coverage
- The `vitest` tests are still disabled (`if: false`) due to the decorator `tsconfig.json` issue with the mikroorm and typeorm adapters. We'll have to reenable those when we fix that.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


ndom91:
Okay so this is all cleaned up, but its still failing..

I added a super basic credentials provider which should basically always work, which it does in local playwright `test:e2e`. However in GHA when clicking "Sign in with Credentials" on our signin page, it redirects to the error page with `CallbackRouteError` :thinking: 

With the Keycloak test, it never loads the keycloak page when you click on "Sign in with Keycloak". I can't view the env vars, but is the issuer set correctly? Is the keycloak instance up? This one fails at the same spot for me locally too.

ndom91:
@balazsorban44 Okay so this is all cleaned up, but its still failing..

I added a super basic credentials provider which should basically always work, which it does in local playwright `test:e2e`. However in GHA when clicking "Sign in with Credentials" on our signin page, it redirects to the error page with `CallbackRouteError` 🤔

With the Keycloak test, it never loads the keycloak page when you click on "Sign in with Keycloak". I can't view the env vars, but is the issuer set correctly? Is the keycloak instance up? This one fails at the same spot for me locally too.

I got the traces uploading to work. They can be downloaded on the "Summary" page after a failed `release.yml` `Test` run. You can then view them by uploading one of the `trace.zip` files (there are multiple in the zip you download from GH) to https://trace.playwright.dev

PR #9948, opened 2024-02-07
feat(core/providers): add Gitea/Forgejo login provider

rubyowo:
## ☕️ Reasoning

Add Gitea/Forgejo OAuth provider.

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues
--


m42e:
@vercel @ndom91  Could something be done to move this thing forward?

packages/core/src/providers/gitea.ts
```diff
@@ -0,0 +1,108 @@
...
+export default function Gitea<P extends GiteaProfile>(
+  options: OIDCUserConfig<P> & {
+    issuer: string
+  }
+): OIDCConfig<P> {
+  return {
+    id: "gitea",
+    name: "Gitea",
+    type: "oidc",
+    issuer: options.issuer,
```
ndom91, line 104: If yuo're just going to pass options below anyway this line seems unnecessary to me :thinking:

ndom91:
In general this looks pretty good. I had one comment above on the `issuer` key. And in addition it'd be great if you cuold add a docs page under `docs/pages/getting-started/providers/gitea.mdx` for example. Check out any of the other OIDC provider docs in that same directory for an example, probably just have to modify a few small things. 

Oh and you can uncomment the rest of those fields in the `GiteaProfile` :ok_hand:

rubyowo:
why are the docs failing

PR #10011, opened 2024-02-12
chore(sveltekit): use new APIs in `sveltekit-example-app`

ndom91:
## ☕️ Reasoning

- Bring SvelteKit example app up to speed with latest API
	- Using `SignIn` / `SignOut` from `@auth/sveltekit/components`
	- Added `/signin` / `/signout` routes to handle server-side form actions
- Use `vite.config`'s `server: { port: 3000 }` option like sveltekit dev app, to make it easier to re-use oauth configs designed for our next.js apps
- Cleanup the `pnpm-lock` situation
	- Since `apps/examples/*` aren't part of the pnpm workspace, they should have their own `pnpm-lock.yaml` and should be installed with `pnpm install --ignore-workspace` command 

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

## 📌 Resources


packages/frameworks-sveltekit/src/lib/components/SignIn.svelte
```diff
@@ -59,7 +59,9 @@
       />
     </slot>
   {/if}
-  <button style="width: 100%" type="submit"
-    ><slot name="submitButton">Sign in with {provider}</slot></button
-  >
+  <slot name="submitButton">
```
ThangHuuVu, line 62: I'm curious what sort of CSS workaround were required for the button -> slot order? 🤔 can we solve it by putting an `id` on the button instead? 👀

ndom91: Hmm yeah so if I remember correctly, there was a default border (maybe from global css) that I had to get rid of. Also the `width: 100%` was messing up the layout I was trying to achieve. Also it was a bit awkward trying to style the button, not being allowed to put my own button on the page (cause I can't nest buttons and only one is the submit btn anyway). 

I'm just not sure giving folks something that, although mostly unstyled, does come with some styling implications, is a great idea. Folks will be working around it no matter what I think. Imo just the `form` would be sufficient. Folks then will just have to add their own `button` (which is kind of the whole point of the `SignIn` component anyway), ya know.

What do you think?

ThangHuuVu: makes good sense to me 👍

ndom91: Okay nice, what do you think of merging today? I can update all the docs then

ndom91: Or actually.. Lemme play with the api a bit more. 

Just thought of another idea. What if we pass the class name prop onto the button. So that the `SignIn` component import can be styled directly with tailwind classes, inline styles, etc. And that would apply directly to the button? 🤔

ndom91: @ThangHuuVu so I basically just reverted this. No action here beyond the updating of the components that I did in the recent PR you approved (https://github.com/nextauthjs/next-auth/pull/10006)

apps/examples/sveltekit/src/components/header.svelte
```diff
@@ -0,0 +1,114 @@
...
+        src={$page.data?.session?.user?.image ??
+          "https://source.boringavatars.com/marble/120"}
+        class="avatar"
+      />
+      {#if $page.data.session}
+        <span class="signedInText">
+          {$page.data.session.user?.email ?? $page.data.session.user?.name}
+        </span>
+        <SignOut>
+          <div class="buttonPrimary">Sign out</div>
```
ndom91, line 20: The new slot update isn't in `@auth/sveltekit@latest` ofc

Issue #9631, opened 2024-01-13
unable to to use subdomain in AUTH_URL and wrap middleware in auth


trevorpfiz:
### Environment

System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (16) x64 AMD Ryzen 7 3700X 8-Core Processor
    Memory: 4.78 GB / 7.76 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 20.10.0 - ~/.nvm/versions/node/v20.10.0/bin/node
    npm: 10.2.3 - ~/.nvm/versions/node/v20.10.0/bin/npm
    pnpm: 8.14.1 - ~/.local/share/pnpm/pnpm
    bun: 1.0.14 - ~/.local/share/pnpm/bun
    Watchman: 2023.12.04.00 - /home/linuxbrew/.linuxbrew/bin/watchman

### Reproduction URL

https://github.com/vercel/platforms/tree/main

### Describe the issue

```
Error: getaddrinfo ENOTFOUND app.localhost
    at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:118:26)
    at GetAddrInfoReqWrap.callbackTrampoline (node:internal/async_hooks:130:17) {
  errno: -3008,
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'app.localhost'
}
```

### How to reproduce

Use a subdomain for localhost. Wrap middleware in auth. Don't include `app.localhost` in /etc/hosts file.

`AUTH_URL='http://app.localhost:3000'`

### Expected behavior

In v4 I am able to use `NEXTAUTH_URL='http://app.localhost:3000'` without the error in the platforms starter.

trevorpfiz:
I am setting a cookie on localhost and want it to be accessible in the subdomains like app.localhost. It does not seem like that can be done, so I am following this approach: [Session-token cookie not set #4089](https://github.com/nextauthjs/next-auth/discussions/4089#discussioncomment-2290660)

if I set `AUTH_URL='http://local.run:3000'` and go to `localhost:3000`, the `req.url` is `local.run:3000`. I also get an infinite loop when doing 
```
return NextResponse.rewrite(
      new URL(`/home${path === "/" ? "" : path}`, req.url),
    );
```

[...]

PR #10094, opened 2024-02-21
fix(core): cannot parse action at /session

### LINKED ISSUE #10081: Autodetect protocol+host behind proxy without `AUTH_URL` env variable
### Environment

System:
    OS: Windows 11 10.0.22631
    CPU: (12) x64 AMD Ryzen 5 2600 Six-Core Processor
    Memory: 1.05 GB / 15.91 GB
  Binaries:
    Node: 18.17.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 4.1.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 10.4.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.6.1 - ~\AppData\Local\pnpm\pnpm.EXE
  Browsers:
    Edge: Chromium (121.0.2277.128)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @auth/core: latest => 0.27.0
    @auth/prisma-adapter: ^1.4.0 => 1.4.0
    next: latest => 14.1.0
    next-auth: 5.0.0-beta.11 => 5.0.0-beta.11
    react: ^18.2.0 => 18.2.0

### Reproduction URL

https://github.com/robalmeister/next-auth-example/

### Describe the issue

```
[auth][error] UnknownAction: Cannot parse action at /session .Read more at https://errors.authjs.dev#unknownaction
    at parseActionAndProviderId (webpack-internal:///(middleware)/./node_modules/@auth/core/lib/utils/web.js:87:27)
    at toInternalRequest (webpack-internal:///(middleware)/./node_modules/@auth/core/lib/utils/web.js:32:40)
    at Auth (webpack-internal:///(middleware)/./node_modules/@auth/core/index.js:82:103)
    at getSession (webpack-internal:///(middleware)/./node_modules/next-auth/lib/index.js:22:60)
    at handleAuth (webpack-internal:///(middleware)/./node_modules/next-auth/lib/index.js:135:35)
    at Object.eval [as handler] (webpack-internal:///(middleware)/./node_modules/next-auth/lib/index.js:105:20)
    at eval (webpack-internal:///(middleware)/./node_modules/next/dist/esm/server/web/adapter.js:193:27)
    at AsyncLocalStorage.run (node:async_hooks:338:14)
    at Object.wrap (webpack-internal:///(middleware)/./node_modules/next/dist/esm/server/async-storage/request-async-storage-wrapper.js:81:24)
    at eval (webpack-internal:///(middleware)/./node_modules/next/dist/esm/server/web/adapter.js:180:122)
{ x-middleware-next: '1' }
{ x-middleware-next: '1' }
```

### How to reproduce

try to signin

### Expected behavior

signin page

bwallberg:
## ☕️ Reasoning

According to the comment for `createActionURL()` it's supposed to build an URL in the same way for both with or withot ENV variables but currently it doesn't use the basePath provided.

This causes the `createActionURL()` to create an invalid action url for `/session` when one of the ENV variables are set.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes #10081


bestickley:
Thank you for this fix, @bwallberg! I'm using it via PNPM patch for now.

oscarmylla:
Setting AUTH_URL=http://app.localhost:3000 with this change gives me the following error:

Error: getaddrinfo ENOTFOUND app.localhost
    at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:118:26)
    at GetAddrInfoReqWrap.callbackTrampoline (node:internal/async_hooks:130:17) {
  errno: -3008,
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'app.localhost'
}

Any thoughts?

bestickley:
@oscarmylla, looking at the error message, I don't this is PR is related. It works without this PR?

oscarmylla:
@bestickley Without this PR i receive the error message "Cannot parse action at /session"

oscarmylla:
@bestickley Some more context: 
- See this issue https://github.com/nextauthjs/next-auth/issues/9631#issuecomment-1890915340
- Exporting auth from middleware gives me the error, but using auth() inside a regular middleware is no problem

This gives error:
export default auth(async (req: NextRequest) => {}

This is working:
export default auth(async (req: NextRequest) => {
  const session = await auth();
 }

ThangHuuVu:
hmm, I'm thinking about using `createActionURL` from the core and remove the function in `next-auth` instead - as the core function is intended to be used in framework packages. The fix should therefore be in the core's function. Would you be able to update the PR with my suggestion? 🙏

ndom91:
Also @bwallberg regarding your comment [here](https://github.com/nextauthjs/next-auth/issues/10081#issuecomment-1963417957), the goal is not to require the `basePath` on the `AUTH_URL` (like `http://localhost:3000/api/auth`), quite the contrary, we'd like to continue supporting the previous env vars everyone has already set that consist of only the host (i.e. `AUTH_URL=http://localhost:3000`).

That being said, what we all would love is just to expand this fix to the other fraemwork libraries as well (like `@auth/sveltekit`) :blush:

So long story short, that'd look like this imo:
1. There's a "sister" `createActionURL` in `core` at `packages/core/src/lib/utils/env.ts:47` that's meant to replace the framework specific ones, but just hasn't been done yet. Could you apply your change there instead? 
2. Then we can replace the `next-auth` and `@auth/sveltekit`'s own `createActionURL` with this updated `core` one. Basically these two need replaced:
	- SvelteKit version: `packages/frameworks-sveltekit/src/lib/actions.ts:135`)
	- next-auth version: `packages/next-auth/src/lib/actions.ts:118` (this is the one you updated already)

bwallberg:
> Also @bwallberg regarding your comment [here](https://github.com/nextauthjs/next-auth/issues/10081#issuecomment-1963417957), the goal is not to require the `basePath` on the `AUTH_URL` (like `http://localhost:3000/api/auth`), quite the contrary, we'd like to continue supporting the previous env vars everyone has already set that consist of only the host (i.e. `AUTH_URL=http://localhost:3000`).
> 
> That being said, what we're all getting at, I think, is just to expand this fix to the other fraemwork libraries as well (like `@auth/sveltekit`) 😊
> 
> So long story short, that'd look like this imo:
> 
>     1. There's a "sister" `createActionURL` in `core` at `packages/core/src/lib/utils/env.ts:47` that's meant to replace the framework specific ones, but just hasn't been done yet. Could you apply your change there instead?
> 
>     2. Then we can replace the `next-auth` and `@auth/sveltekit`'s own `createActionURL` with this updated `core` one. Basically these two need replaced:
>        
>        * SvelteKit version: `packages/frameworks-sveltekit/src/lib/actions.ts:135`)
>        * next-auth version: `packages/next-auth/src/lib/actions.ts:118` (this is the one you updated already)

Thanks for clarifying. Sounds very reasonable :)

bwallberg:
@ThangHuuVu @ndom91 

I believe I've updated the PR as requested. I've also updated the unit-test to reflect the expectations, which is that the basePath should be set in the config and _not_ in the env variable. I believe setting a custom one only in `AUTH_URL` would break everything regardless.

I've tested the solution in with next.js but not sveltekit, if you approve of the solution I'll setup the local dev app to test it out.

ndom91:
Awesome, thanks a lot! I'll take a look shortly. 

Regarding the basePath, yeah it's a bit tricky with SvelteKit because the svelte.config basePath is designed to be the basePath for your entire application, right. Like if your app was at a sub path of your entire organizations site (like `https://company.com/projectMgmtApp/*`). And the `auth.ts` config `basePath` / `AUTH_URL` url path is for your auth route only 🤔 

So long short short, as a user, if i had set `svelte.config.js` `basePath: "/app"` and the `auth.ts` config `basePath: "/auth"`, I'd expect my whole app to be at `company.com/app/*` and my auth URL to be at `company.com/app/auth`, the auto-generated signin page for example at `compny.com/app/auth/signin`.

bwallberg:
> Awesome, thanks a lot! I'll take a look shortly.
> 
> Regarding the basePath, yeah it's a bit tricky with SvelteKit because the `svelte.config.js` `basePath` is designed to be the `basePath` for your entire application, right. Like if your app was at a sub path of your entire organizations site (like `https://company.com/projectMgmtApp/*`). And the `auth.ts` config `basePath` / `AUTH_URL` url path is for your auth route only 🤔
> 
> So long short short, as a user, if i had set `svelte.config.js` `basePath: "/app"` and the `auth.ts` config `basePath: "/auth"`, I'd expect my whole app to be at `company.com/app/*` and my auth URLs to be at `company.com/app/auth/*`, i.e. the auto-generated signin page at `compny.com/app/auth/signin`

Yeah I see, that makes sense 🤔

bwallberg:
@ndom91 I believe I've solved the use-case you've described as well now.

ThangHuuVu:
@bwallberg could you separate the package changes in different PRs - one for core, one for next-auth, and one for sveltekit? just to keep things tidy and the git history clean. Sorry I should've mentioned this earlier. You can keep this PR for core and create new PRs for framework packages. Many thanks!

ndom91:
@bwallberg thanks, changes in general look good so far! :+1: 

Was just talking with Thang and thinking about how to avoid foot guns of people misconfiguring this since there are quite a few different permutations that are supported now..

Note that `basePath` will default to `/api/auth` in next-auth and `/auth` with all other frameworks if its not set explicitly in the config.

We basically have the following variations we currently support / see used:
1. `AUTH_URL` with and without an actual application base-path (i.e. `/my-app`)
2. `AUTH_URL` with and without the auth.js path attached (i.e. `/api/auth`)
3. Then both of those with `basePath` set to various things in the configuration

Written out, some of those options look like this:
1. `AUTH_URL=http://localhost:3000` & `basePath: '/api/auth'`
2. `AUTH_URL=http://localhost:3000/api/auth` & `basePath: '/api/auth'`
3. `AUTH_URL=http://localhost:3000/my-app/api/auth` & `basePath: '/api/auth'`
4. `AUTH_URL=http://localhost:3000/` & `basePath: '/my-app/api/auth'`

In my opinion, we should encourage `AUTH_URL` to be only the root domain and/or root domain + application base path (`/my-app`) if applicable, and then the configuration `basePath` is the route to your `[...nextauth]/route.ts` file. That's pretty easy to explain and reason about too imo.

I know many people will probably have something else already configured though because we support / supported many things. 

Imo we should pick a "golden path", whatever that looks like, and encourage that everywhere and document it well, and then throw errors / log warnings when folks are setting them to something unsupported. What do you all think?

bwallberg:
> @bwallberg thanks, changes in general look good so far! 👍
> 
> Was just talking with Thang and thinking about how to avoid foot guns of people misconfiguring this since there are quite a few different permutations that are supported now..
> 
> Note that `basePath` will default to `/api/auth` in next-auth and `/auth` with all other frameworks if its not set explicitly in the config.
> 
> We basically have the following variations we currently support / see used:
> 
>     1. `AUTH_URL` with and without an actual application base-path (i.e. `/my-app`)
> 
>     2. `AUTH_URL` with and without the auth.js path attached (i.e. `/api/auth`)
> 
>     3. Then both of those with `basePath` set to various things in the configuration
> 
> 
> Written out, some of those options look like this:
> 
>     1. `AUTH_URL=http://localhost:3000` & `basePath: '/api/auth'`
> 
>     2. `AUTH_URL=http://localhost:3000/api/auth` & `basePath: '/api/auth'`
> 
>     3. `AUTH_URL=http://localhost:3000/my-app/api/auth` & `basePath: '/api/auth'`
> 
>     4. `AUTH_URL=http://localhost:3000/` & `basePath: '/my-app/api/auth'`
> 
> 
> One idea - we should encourage `AUTH_URL` to be only the root domain and/or root domain + application base path (`/my-app`) if applicable, and then the configuration `basePath` is the route to your `[...nextauth]/route.ts` file/path. That's pretty easy to explain and reason about too imo. Also `AUTH_URL` sounds kinda like its the URL to the auth path, but actually its just `URL` since we prefixed everything with `AUTH_`, so if yuo think about it, its just the `URL` env var which should point at the root of your application
> 
> I know many people will probably have something else already configured though because we support / supported many things.
> 
> Imo to avoid more complexity here, we should pick a "golden path", whatever that looks like, and encourage that everywhere and document it well, and then throw errors / log warnings when folks are setting them to something unsupported. What do you all think?

I agree to me that felt intuitive that `AUTH_URL` represented the root of the app itself and `config.basePath` the route of the auth.

I think it's best to avoid multiple sources of truths that needs to align ( eg. `AUTH_URL=http://localhost:3000/api/auth & basePath: '/api/auth'` what if basePath was configured to `/auth/api` ) and keep one config for the root of the app and one for the basePath of the api.

There is a caveat here though that if `AUTH_URL` is not provided the `basePath` must be configured with the application base-path as well. Which of course is how it already works, but good to keep in mind when documenting.

bwallberg:
@ThangHuuVu 

I've rebased this branch to contain the single commit for core, hope that was OK.
Svelte: https://github.com/nextauthjs/next-auth/pull/10219/files
Next-auth: https://github.com/nextauthjs/next-auth/pull/10218/files

DuarteMartinho:
Any ETA on this fix ?

ndom91:
> Any ETA on this fix ? Will this fix proxy issue ?

What proxy issue are you referring to exactly?

ndom91:
@bwallberg awesome, thanks a lot for the quick split up!

I think then we can go ahead with these 3 PRs.. Going off of the state of the tests, that's how I'd expect the env vars / config option to interact with one another too. 

We can follow-up then with some additional helpful logging or error-ing out in a follow up PR imo.

ndom91 (approved):

DuarteMartinho:
> > Any ETA on this fix ? Will this fix proxy issue ?
> 
> 
> 
> What proxy issue are you referring to exactly?

I self host my website behind a reverse proxy 

When I go to mydomain.com/api/auth/providers all callback urls point to localhost:3000

ndom91:
@DuarteMartinho please open an issue with more details, like the version of Next.js and next-auth yuo're using as well as the Auth.js config and any other relevant details :pray:

KaygNas:
> I self host my website behind a reverse proxy
> 
> When I go to mydomain.com/api/auth/providers all callback urls point to localhost:3000

@DuarteMartinho I met the similar situation when using 127.0.0.1:3000 to access my website. In my situation, Next always replace `127.0.0.1` to `localhost`, not sure whether it's caused by the same reason, but may you can checkout this. 

https://github.com/vercel/next.js/blob/5740ef3e1e53c04231f8996ab737738e9b02e4b4/packages/next/src/server/web/next-url.ts#L25

CyanFlare:
Hello,

I've recently upgraded from ``next-auth@5.0.0-beta.4`` to ``next-auth@5.0.0-beta.15`` and have ran into some issues that I think is related to this and I'm hoping someone can assist me with solving.

In the ``next-auth@5.0.0-beta.6`` release notes, there was a line that says ``next-auth basePath should be set to /api/auth by default``.
I believe this may be the cause of the issues I'm currently facing.

The first issue I'm having is with the ``NEXTAUTH_URL`` env variable and the ``basePath`` auth.ts config option.
Setting ``NEXTAUTH_URL`` to ``http://localhost:3000`` and ``basePath`` to ``/api/auth`` would result in a ``Cannot parse action at /session`` error. But setting ``NEXTAUTH_URL`` to ``http://localhost:3000/api/auth`` and leaving ``basePath`` unset would work perfectly fine but I think this is what lead to the second issue below.

The second issue I'm having is with the ``pages`` auth.ts config option.
In ``next-auth@5.0.0-beta.4`` setting the pages error option to ``/auth/error`` would redirect to ``http://localhost:3000/auth/error`` but when I upgraded to ``next-auth@5.0.0-beta.15`` and applied the fix in the first issue it would redirect to ``http://localhost:3000/api/auth/auth/error``.

DuarteMartinho:
@KaygNas @CyanFlare 

Have a look at [this]()

Turns out for me the problem was my next was not updated

CyanFlare:
> @KaygNas @CyanFlare
> 
> Have a look at [this](https://github.com/nextauthjs/next-auth/issues/10081) I made some comments showing my errors and getting it fixed
> 
> Turns out for me the problem was my next was not updated

My next version is 14.1.3. 
I solved the first issue by removing ``NEXTAUTH_URL`` env variable completely as apparently it is no longer needed. But I am still encountering the second issue.

packages/core/src/lib/utils/env.ts
```diff
@@ -51,13 +51,22 @@ export function createActionURL(
...
-  let url = envObject.AUTH_URL ?? envObject.NEXTAUTH_URL
-  if (!url) {
-    const host = headers.get("x-forwarded-host") ?? headers.get("host")
-    if (!host) throw new TypeError("Missing host")
-    const proto = headers.get("x-forwarded-proto") ?? protocol
-    url = `${proto === "http" ? "http" : "https"}://${host}${basePath}`
+  let envUrl = envObject.AUTH_URL ?? envObject.NEXTAUTH_URL
+
+  let detectedHost = headers.get("x-forwarded-host") ?? headers.get("host")
+  let detectedProtocol = headers.get("x-forwarded-proto") ?? protocol ?? "https"
```
ThangHuuVu, line 57: these are only used in the else block - I suggest we can move them there and turn into `const`

packages/core/src/lib/utils/env.ts
```diff
@@ -51,13 +51,22 @@ export function createActionURL(
...
+    url = new URL(`${detectedProtocol}://${detectedHost}`)
   }
 
-  return new URL(`${url.replace(/\/$/, "")}/${action}`)
+  // remove trailing slash
+  const sanitizedUrl = url.toString().replace(/\/$/, "")
+  // remove leading and trailing slash
+  const sanitizedBasePath = basePath?.replace(/(^\/|\/$)/g, "") ?? ""
+
+  return new URL(`${sanitizedUrl}/${sanitizedBasePath}/${action}`)
```
ThangHuuVu, line 71: I think we're missing the case where `url.pathname` equals to `basePath`. If so, we don't need to append the `basePath`. That should also a valid case IMO

packages/core/test/env.test.ts
```diff
@@ -144,14 +144,31 @@ describe("createActionURL", () => {
...
+    {
+      args: {
+        action: "signout",
+        protocol: undefined,
+        headers: new Headers({}),
+        env: { AUTH_URL: "https://sub.domain.env.com/my-app/" },
+        basePath: "/api/auth",
       },
-      expected: "https://env.com/api/auth/signout",
+      expected: "https://sub.domain.env.com/my-app/api/auth/signout",
```
ThangHuuVu, line 171: This test looks weird to me. It should be a configuration error if `envUrl.pathname !== basePath`. Although we are not explicitly checking for it in the config assertion, I think we can pick either one of them and move on. I am thinking that we can prioritize the `basePath` since that's a new option. How do you think?

ThangHuuVu (dismissed):

packages/core/test/env.test.ts
bwallberg: This was discussed earlier with @ndom91 here https://github.com/nextauthjs/next-auth/pull/10094#issuecomment-1977096182

Arguably they env variable defines the app url+path and basepath defines the auth path. Based on earlier discussions anyway

packages/core/src/lib/utils/env.ts
bwallberg: Based on earlier discussions ( https://github.com/nextauthjs/next-auth/pull/10094#issuecomment-1977096182 ) basepath shouldn't be part of the env url.

Seems unnecessary to support that since then you have to keep basepath in sync with envurl

packages/core/src/lib/utils/env.ts
bwallberg: Done

packages/core/src/lib/utils/env.ts
ThangHuuVu: Technically, `basePath` is required in the env URL, but we supported it with or without before. You can see this explanation from @balazsorban44: https://github.com/nextauthjs/next-auth/issues/10081#issuecomment-1960626967

So in my opinion:
- If we detect both `AUTH_URL` and `basePath`, check if `AUTH_URL`'s `pathname` and `basePath` match:
  - If yes, we should not use `url.toString()`, but instead `const sanitizedUrl = url.origin.toString().replace(/\/$/, "")` we should also log a warning - my suggestion: `AUTH_URL and basePath detected. This is likely a configuration mistake - either remove the basePath, or remove the pathname of AUTH_URL.`.
  - Otherwise, construct the URL with `url.origin.toString()`, and log a warning: "AUTH_URL and basePath detected and are mismatching. This is likely a configuration mistake - either remove the basePath, or remove the pathname of AUTH_URL.\n basePath will override AUTH_URL pathname"
Basically, these two options should not be specified together, as they will overlap/ become redundant. 

If only basePath detected, use detectedHost + basePath
If only AUTH_URL detected, use use `url.toString()`

Hope this makes sense, my goal is to keep the existing setup from breaking at least at possible. I also discussed with @ndom91 before commenting 🙏

bwallberg: @ThangHuuVu I believe it now supports the types of configurations expected.

ThangHuuVu (approved):
great work @bwallberg ! I just push a commit to refactor the warning to a code, using the logger instance. I will also follow up with a doc PR to explain these code. LGTM

Issue #9809, opened 2024-01-26
Prerendering causing build error in SvelteKit


WhyAsh5114:
### Environment

```
  System:
    OS: Windows 11 10.0.22621
    CPU: (12) x64 AMD Ryzen 5 3600 6-Core Processor
    Memory: 6.26 GB / 15.91 GB
  Binaries:
    Node: 18.18.2 - C:\Program Files\nodejs\node.EXE
    npm: 10.2.3 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (120.0.2210.144)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @auth/core: ^0.23.0 => 0.23.0
    @auth/sveltekit: ^0.9.0 => 0.9.0
```

### Reproduction URL

https://github.com/WhyAsh5114/authjs-sveltekit-prerender

### Describe the issue

Marking a route as prenderable with `export const prerender = true` creates a build error, usable in dev though. Removing the prerender line fixes it.
```
TypeError: Cannot read properties of undefined (reading 'length')
    at Object.handle (file:///D:/Programming/Web/authjs-sveltekit-prerender/.svelte-kit/output/server/chunks/hooks.server.js:49:24)
    at respond (file:///D:/Programming/Web/authjs-sveltekit-prerender/.svelte-kit/output/server/index.js:2650:43)
    at async visit (file:///D:/Programming/Web/authjs-sveltekit-prerender/node_modules/@sveltejs/kit/src/core/postbuild/prerender.js:202:20)

node:internal/event_target:1033
  process.nextTick(() => { throw err; });
                           ^
Error: 500 /prerender
To suppress or handle this error, implement `handleHttpError` in https://kit.svelte.dev/docs/configuration#prerender
    at file:///D:/Programming/Web/authjs-sveltekit-prerender/node_modules/@sveltejs/kit/src/core/config/options.js:202:13
    at file:///D:/Programming/Web/authjs-sveltekit-prerender/node_modules/@sveltejs/kit/src/core/postbuild/prerender.js:65:25
    at save (file:///D:/Programming/Web/authjs-sveltekit-prerender/node_modules/@sveltejs/kit/src/core/postbuild/prerender.js:403:4)
    at visit (file:///D:/Programming/Web/authjs-sveltekit-prerender/node_modules/@sveltejs/kit/src/core/postbuild/prerender.js:236:3)
Emitted 'error' event on Worker instance at:
[...]

Issue #9436, opened 2023-12-21
SvelteKit 2.0 Build Error - Cannot read values from $env/dynamic/private while prerendering


ubemacapuno:
### Environment

  System:
    OS: macOS 14.2.1
    CPU: (8) arm64 Apple M1
    Memory: 443.92 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.5.0 - /usr/local/bin/node
    npm: 10.2.4 - /usr/local/bin/npm
    pnpm: 8.10.5 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 120.0.6099.109
    Safari: 17.2.1
  npmPackages:
    @auth/sveltekit: ^0.5.0 => 0.5.0 
    @sveltejs/kit: ^2.0.0
    @vite: ^5.0.3
    @typescript: ^5.0.0

### Reproduction URL

https://github.com/ubemacapuno/authjs-sveltekit2-demo.git

### Describe the issue

Hey all, I'm using @auth/sveltekit with the newly-released SvelteKit 2.0. I noticed that my dev environment runs fine, but when I try to build, I run into an error regarding environment variables.

There is a section in the SvelteKit 2.0 migration docs for ["Dynamic environment variables cannot be used during prerendering"](https://kit.svelte.dev/docs/migrating-to-sveltekit-2#dynamic-environment-variables-cannot-be-used-during-prerendering), where it is mentioned, "...dynamic environment variables can no longer be read during prerendering in SvelteKit 2 — you should use the static modules instead." I believe this is why I am running into trouble when running the build script.

Using SvelteKit 2.0 and @auth/sveltekit and building the project locally, I receive this error in my terminal:
```
Error: Cannot read values from $env/dynamic/private while prerendering (attempted to read env.AUTH_SECRET). Use $env/static/private instead
```

Any insight on a fix would be greatly appreciated, because I'm getting stuck just going off of the docs. Thank you!

### How to reproduce

After starting a SvelteKit 2.0 project with @auth/sveltekit and setting up hooks.server.ts and environment variables [per the docs](https://authjs.dev/reference/sveltekit), I run into an error when running `npm run build`:

You can checkout my [demo](https://github.com/ubemacapuno/authjs-sveltekit2-demo.git), set .env for `GITHUB_ID`, `GITHUB_SECRET`, and `AUTH_SECRET` and run `npm run build` to see the error, or recreate yourself:

[...]

PR #10117, opened 2024-02-23
fix(sveltekit): use `env/static/private` for env vars

dievardump:
SvelteKit does not build with the latest version.

Two reasons:

- during `build`, sveltekit can only access "static" private env var, not dynamic
- during build, [`setEnvDefaults`](https://github.com/nextauthjs/next-auth/blob/c0eb55806d28387ef2df6ba7648c229027a1cac3/packages/frameworks-sveltekit/src/lib/env.ts#L6) was not setting `config.basePath` which is read in the `handle` function, throwing an error for reading undefined.


ndom91:
Hmm so my local project using `@auth/sveltekit@0.13.0` does build. Are you saying it builds but then you had errors running the built output?

dievardump:
I am using @sveltejs/kit 2.5.1 and @sveltejs/adapter-node 4.0.1

When building I get the error:

> Error: Cannot read values from $env/dynamic/private while prerendering (attempted to read env.AUTH_URL). Use $env/static/private instead
>     at Object.get (file:///home/dievardump/www/abc/apps/test-app/.svelte-kit/output/server/index.js:3397:11)
>    at createActionURL (file:///home/dievardump/www/abc/apps/test-app/.svelte-kit/output/server/chunks/hooks.server.js:126:25)
>    at auth (file:///home/dievardump/www/abc/apps/test-app/.svelte-kit/output/server/chunks/hooks.server.js:107:22

ndom91:
Yeah so the way that Svelte works, the `AUTH_URL` env var must be set otherwise it'll throw that first error you mentioned, right. 

I'm not super experienced with Svelte though, is using the env vars here out of `static` a better way to go about this?

dievardump:
For me, if you expect the variables to be set at build time, then you go for static. If you expect those variable can change during the life of the app, or they should only be set at runtime, then go for dynamic.

Because of the nature of the current module, accessed mostly as the first think seen in the server through hook, and because you would expect the AUTH vars to be known and set at build time, I would suggest to use static over dynamic

the doc says:

## dynamic 

> This module provides access to runtime environment variables, as defined by the platform you're running on. For example if you're using [adapter-node](https://github.com/sveltejs/kit/tree/main/packages/adapter-node) (or running [vite preview](https://kit.svelte.dev/docs/cli)), this is equivalent to process.env. 
> 
> Dynamic environment variables cannot be used during prerendering.

## static

> Environment variables [loaded by Vite](https://vitejs.dev/guide/env-and-mode.html#env-files) from .env files and process.env. 
> 
> Unlike [$env/dynamic/private](https://kit.svelte.dev/docs/modules#$env-dynamic-private), the values exported from this module are statically injected into your bundle at build time, enabling optimisations like dead code elimination.

ndom91:
@dievardump okay yeah that makes sense. I've updated from `main`, which included a bunch of updates regarding the sveltekit dev app. Do me a favor and take another look, see if it all still works as expected.

If yes, then I think we can go ahead with this!

ThangHuuVu:
@dievardump great attempt, but I don't think this is a good idea. This comment puts it nicely: https://github.com/nextauthjs/next-auth/issues/9436#issuecomment-1869389824

Let us think a better approach to this issue. Is this related? https://github.com/nextauthjs/next-auth/issues/9809

dievardump:
I think #9809 is specifically corrected by my changes to [packages/frameworks-sveltekit/src/lib/env.ts](https://github.com/nextauthjs/next-auth/pull/10117/files#diff-fefe53a2deabc52af0ef0e385e31b95a66eeba9fb7014d3d54202601f3cb9861) 
For some reasons some variables are not set during build time (not sure why). 

About the security of having the vars in the files: I can imagine why a security audit would flag it, but I would consider anyone having access to the "build image" as a trusted party. Because if they have access to the source after build, they have access to the running environment, and therefore to the process.env vars. So using dynamic or static here wouldn't make much difference in term of security

ndom91:
@dievardump after some internal discussion, we've decided to stick with the dynamic env vars for the following reasons:

1. In many projects / companies, the application is often built once and then redeployed across various environments (preview, staging, prod, etc.) which implies requiring dynamic env vars like auth credentials and database urls.
2. In addition, sometimes auth credentials need to be rotated quickly without rebuilding and redeploying an entire application, which also implies requiring dynamic env vars

That being said, I would still like to avoid the issue yuo've raised here about prerendering fialing if one of the env var's we've used in the library aren't defined by the user. As a concrete example, `AUTH_URL is` used in the library, but is an optional env var as we autodetect it in many environments. What do you think?

I'd also like to keep your change moving the `if (building)..` clause a bit lower

dievardump:
The problem is still that it will not be possible to pre-render any route if using SvelteAuth and `adapter-node` if you keep using `$env/dynamic/private`

Would you then consider using `process.env`, which is what you do in the other adapters, which would allow for everything to work smoothly?

WhyAsh5114:
Would really like a solution to this, prerender is really important for quick page loads and even setting one page to ```
export const prerender = true;
```
breaks the build process.

dievardump:
What you can do today to fix that is to only add `SvelteKitAuth.handle` to the `sequence` if `build == false`

But the better solution for the future would be for the plugin to use `process.env` instead of `$env/*/private`

ndom91:
Hey chiming in with an update, so we definitely don't want to go with the `env/static/private` option to bake the variables into the build output for the reasons listed above.

We took another look at this and talked to some other svelte folks and it basically during prerendering you shuoldn'tn rely on any authenticated info to begin with. Therefore, we're tryign to find a way to skip calling the auth routes during build. Like `if (building) skip`.

But the problem is, those routes are all dynamic from the `hooks.server.ts` `handle` fn. So as soonas you have a `<a href="/auth/signin" ..` in one of your pages, it'll try to hit that route during prerender and it'll get a 404 since that route doesn't have a page or page.server file, right.

Anyway, long story short. I don't think we want to merge any of these changes unfortunately. I've opened a new PR attempting to solve the prerender issue and we'd still love yuor help there if yuo're interested :pray: 

https://github.com/nextauthjs/next-auth/pull/10339

Padi2312:
Its still failing :(

PR #10118, opened 2024-02-23
chore: add new example app for nextjs and Docker

ndom91:
## ☕️ Reasoning

- Add a new `examples/nextjs-docker` example app
- Runs via `test:docker` npm script in that example app dir ^
	- Running `pnpm test:docker` locally requires setting `TEST_KEYCLOAK*` and `AUTH_KEYCLOAK_*` env vars to be defined in the `.env`
- Added (but currently disabled) GHA step for running existing playwright tests against the Docker instance as well
	- In CI the GHA secrets / env vars should override it from GHA secrets

## 🧢 Checklist

- [ ] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources


ndom91:
Okay so this is all workign now local, but in GHA apparently port 3000 is already used when we get to the new `Docker E2E Test` step ?! :joy:

image

I'm guessing its from the normal `Test E2E` step right before. Playwright is running the next.js dev app via the `webserver` config options, and then when its done its probably not killing that next.js server process correctly.

PR #10121, opened 2024-02-23
fix(sveltekit): broken redirect to another origin when `callbackUrl` contains a `#`

dievardump:
When a callbackUrl contains a `#`, signIn and signOut are automatically reloading the page, to ensure thre is a reload.

However, if the redirect is supposed to redirect to another origin, `window.location.reload()` should not be performed, else it bypasses the window.location.href

This PR adds a check if the callbackUrl and the current url have the same origin, before allowing or not a `location.reload()`


ndom91:
I'm having trouble reproducing this, can you provide an minimal repro? :pray:

PR #10154, opened 2024-02-28
fix(docs): update example secret generation cmd for more entropy

coreyward:
This is a documentation-only change that adjusts the recommended command for generating a secret to increase entropy and avoid the needless padding at the end of the output.

Due to the way base64 works, it produces 4 output characters for every 3 input bytes, but input lengths that are not evenly divisible by 3 wind up producing an output with padding characters (`=`) at the end of the output string. Asking for 32 bytes from `openssl rand` thus will always produce a base64 encoded value with a single `=` as padding at the end.

Asking `openssl rand` instead for 33 bytes of output increases the entropy and produces a base64 encoded length of 44 characters evenly without padding characters at the end of the string. This is a better recommendation, and it is quite common for this purpose.

For example: 

```
$ openssl rand 32 | btoa
# 6mahCn+mX6SchA3ZGfFCbrwf+BC+HHze0AGF9/5xZqE=
$ openssl rand 33 | btoa
# JFkmrfp8lANp44bbvCIyHPkCfv/V8cfeIBjiyXOjG8tS
```

### DIFF
--- a/docs/docs/configuration/options.md
+++ b/docs/docs/configuration/options.md
@@ -77,7 +77,7 @@ Not providing any `secret` or `NEXTAUTH_SECRET` will throw [an error](/errors#no
 You can quickly create a good value on the command line via this `openssl` command.
 
 ```bash
-$ openssl rand -base64 32
+$ openssl rand -base64 33
 ```
 
 :::tip


ndom91:
@coreyward first of all, thanks for the contribution! Interesting explanation.

I'd like to merge it, however, you've got to make the PR against the `main` branch. I tried updating the target branch here in the PR and there are too many changes. Could yuo reopen it against `main` and I'd be happy to get it in :pray:

coreyward:
@ndom91 Strange — I opened it from the "Edit on Github" link from the docs page—might be useful to update those to point to `main`. It doesn't look like I have sufficient permissions to change the target branch, so I'll close this one and open a separate PR against `main`.

PR #10185, opened 2024-03-02
Fix API Endpoint Path Mismatch in Next.js Example

zaru:
## ☕️ Reasoning

I tried the Next.js example, but it did not work. I believe the cause was a mismatch between the API endpoint's path and the directory structure. Therefore, I have moved the file.

`app/auth/[...nextauth]/route.ts` to `app/api/auth/[...nextauth]/route.ts`

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


ndom91:
@zaru thanks for pointing this out! While your fix wuold theoretically work, I've moved it to `/auth` (instead of `/api/auth`) on purpose in order to be abel to share OAuth configurations across more of our example / dev apps. Since OAuth applications need a registered callback URL, and all the other frameworks (svelte, solid, express, etc.) default to `/auth`. This makes development much simpler for us.

That being said, this example app is still broken and requires the `basePath: '/auth'` to be set in the configuration. I've updated that [here](https://github.com/nextauthjs/next-auth/pull/10199/files#diff-cf1d1e00ee7c00c7c9bc57379a1152e955fa945d0bb0bf31ec3cc6a7a7528c4b).

This should be working shortly with `main` then :+1:

PR #10218, opened 2024-03-05
fix(next-auth): cannot parse action at /session

bwallberg:
## ☕️ Reasoning

All frameworks should use the core util `createActionsURL()`, however since `next-auth` doesn't provide the request itself we have to rely on the `x-forwarded-proto` header that next-auth sets so instead of calling `createActionsURL` directly we wrap it to avoid some unnecessary code duplication.

Depends on https://github.com/nextauthjs/next-auth/pull/10094

## 🧢 Checklist

- [] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes the original issue described in #10081 
Depends on #10094 

## 📌 Resources


ndom91 (approved):

packages/next-auth/src/lib/actions.ts
```diff
@@ -120,16 +125,12 @@ export function createActionURL(
...
-    const separator = pathname.endsWith("/") ? "" : "/"
-    return new URL(`${origin}${pathname}${separator}${action}`)
-  }
-  const host = h.get("x-forwarded-host") ?? h.get("host")
-  const protocol = h.get("x-forwarded-proto") === "http" ? "http" : "https"
-  // @ts-expect-error `basePath` value is default'ed to "/api/auth" in `setEnvDefaults`
-  const { origin, pathname } = new URL(basePath, `${protocol}://${host}`)
-  const separator = pathname.endsWith("/") ? "" : "/"
-  return new URL(`${origin}${pathname}${separator}${action}`)
+  return createActionURLCore(
```
ThangHuuVu, line 128: Let's use the function from coredirectly, and remove this function definition - so all the usages should be `createActionURL` 🙏

ThangHuuVu (changes requested):

packages/next-auth/src/lib/actions.ts
bwallberg: done

ThangHuuVu (approved):

PR #10285, opened 2024-03-11
chore(docs): fix oauth-tutorial link

MehmetBozkir:
The link leads to a not-found site.
## ☕️ Reasoning

- The redirected site does not reach any point.
- I directed you to the site you want to go to.

## 🧢 Checklist

- 
He doesn't think these are necessary since it's not a big problem. I am adding the necessary picture for the result.

not-found

### DIFF
--- a/docs/docs/getting-started/providers/oauth-tutorial.mdx
+++ b/docs/docs/getting-started/providers/oauth-tutorial.mdx
@@ -113,7 +113,7 @@ export default function App({
 Instances of `useSession` (more on it in the next section) will have access to the session data and status. The `<SessionProvider />` also keep the session updated and synced between browser tabs and windows. 💪🏽
 
 :::tip
-Check our [client docs](reference/nextjs/react/) to learn all the available options for handling sessions on the browser.
+Check our [client docs](/) to learn all the available options for handling sessions on the browser.
 :::
 
 ### Consuming the session via hooks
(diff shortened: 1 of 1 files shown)


ndom91 (commented):
Hey thanks for noticing, the page does exist though, it's just missing the initial `/` here. Can you please update that? Thanks 🙏 

See: https://authjs.dev/reference/nextjs/react

ndom91:
@MehmetBozkir thanks for the quick response, however, it shuold just be:

```diff
- Check our [client docs](reference/nextjs/react) to learn all the available options for handling sessions on the browser.
+ Check our [client docs](/reference/nextjs/react) to learn all the available options for handling sessions on the browser.
```

You generally don't want to include the whole domain at the beginning for a URL on your own site, because then it won't work in development, for example, when the host is `localhost:3000`. This way, with a relative link, it'll go to `/reference..` no matter the host.

The issue originally, is that the initial `/` character was missing meaning that instead of going to `[host]/reference/nextjs/react` it would simply append `reference/nextjs/react` onto the end of whatever path the user was currently at, `/getting-started/providers` in the case of your screenshot.

MehmetBozkir:
Thank you again @ndom91  for my inexperience. It is very exciting for me to provide Contributors to such a big project.

I updated the PR and you can see it from this commit no : "e079021954f7191f2457d6cab2a536fb5c30d58c"

ndom91 (approved):
@MehmetBozkir no worries! This looks good to go now :pray:

PR #10287, opened 2024-03-11
Update @auth/solid-start package

birkskyum:
Sync `auth/solid-start` with [solid-mediakit/auth](https://github.com/solidjs-community/mediakit/tree/main/packages/auth) to make it work with latest version of solid-start.

cc @OrJDev


OrJDev:
From what i see here you deleted the docs comments so they need to be added but overall awesome great work.

birkskyum:
I've added back these comments

OrJDev:
@birkskyum updated mediakit auth to use latest version of Auth/core, this requires some major changes.

birkskyum:
@OrJDev okay, nice. I seems like it would be a lot easier if you made a PR here, instead of me trailing behind your changes trying to keep things in up-to-date/ in-sync.

PR #10297, opened 2024-03-12
feat(drizzle): webauthn support

JipSterk:
## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources


ndom91:
@JipSterk thanks for reopening this!

Could you also add some tests for this new model?? See the `adapter-prisma` `Authenticator` tests for an example :pray:

JipSterk:
hi @ndom91, could you run the test & verify pipelines to ensure if the code i've written is any good? i couldn't get the test running locally. 
And would like to close the pr since i would like to integrate WebAuthn in my application using a drizzle adapter

JipSterk:
hi @ThangHuuVu i would like to know whats keeping the team from merging this pr.
Including the previous pr which closed because i accidentally deleted my fork this pr is open for 2 months~ now.
I added the tests but as i stated before i couldn't get the test running locally to verify if the pr is merge ready.
I get that this is an open source repository and everyone involved maintains this project aside from their normal work but i could use a status update on when this feature could get merged.

ndom91:
@JipSterk sorry for the delay, webauthn hasn't been our #1 priority as we'd like to get the core functionality ironed out and v5 + new docs out before we iron out these experimental features. 

That being said, we definitely appreciate this PR and plan to come back to it. The tests do need to be running and passing though before its ready to merge :pray:

JipSterk:
Thanks! I get that you first want to ship v5 and later on focus on experimental features like webauthn. If looked through the repo and saw a few TODO's in the docs subfolder. I will look if i can contribute to them in the weekend.

ndom91:
@JipSterk we successfully shipped the new docs the other day! :partying_face: 

But we also merged the [rewritten Drizzle adapter](https://github.com/nextauthjs/next-auth/pull/10479), so you'll probably have to modify a decent amt of things here, sorry about that :sweat_smile:

JipSterk:
Hi @ndom91 i've updated my pr to make use of the updated drizzle adapter. Still couldn't get the tests to run locally so im eager to see if i've implemented them correctly.

ndom91:
@JipSterk awesome thanks for the quick turnaround!

So I ran the tests locally as well and ran into the following issues:

1. In both the `pg` and `mysql` suites the `getAUthenticator` test fails
2. In the `sqlite` suite the `testWebAuthnMethods: true` is missing. If I added this, they all passed except for that same `getAUthenticator` test. 

So clearly something up there.. The error message is: 

image

ndom91:
Hope you don't mind, I pushed a little PR updating the `getAuthenticator` method in `pg`,`mysql` and `sqlite` files.

```diff
    async getAuthenticator(credentialID) {
      const authenticator = await client
        .select()
        .from(authenticatorsTable)
        .where(eq(authenticatorsTable.credentialID, credentialID))
-        .then((res) => fromDBAuthenticator(res[0]) ?? null))
+        .then((res) => (res.length ? fromDBAuthenticator(res[0]) : null))
      return authenticator ? authenticator : null
    },

```

JipSterk:
Thanks for updating the pr!
I just tested again and both `pg` as `sqlite` run their tests & succeed. mysql however does not.
instead it throws the following error:
Screenshot 2024-04-17 at 20 03 49

ndom91:
> Thanks for updating the pr!
> I just tested again and both `pg` as `sqlite` run their tests & succeed. mysql however does not.
> instead it throws the following error:
> Screenshot 2024-04-17 at 20 03 49
> 

Hmm well that seems like either the migration didn't run, or didn't complete by the time the tests started (maybe wait a bit longer in test.sh) or maybe table names are wrong (were they renamed to `auth.*`?)

JipSterk:
adjusting the time did allow the mysql test to succeed. however after re-running the command again it would error like the folllowing: 
Screenshot 2024-04-17 at 21 47 16 it seems the test are flakey at least for my machine (i7 macbook pro 2019 16 inch)
 [here](https://gist.github.com/JipSterk/66a024b47540dee78a241098c5e1b785) you can see both postgres and mysql sql generated tables. i can't find any issues with either of them but might be overlooking something.

ndom91:
Yeah I saw this failure once too actually, but if you look a bit closer, isn't this the correct result? Just the array is in the wrong order, it's expecting `[A, B]` and it got `[B, A]`, right? :thinking:

JipSterk:
hi @ndom91 i updated the `packages/utils/adapter.ts` test case to make use of the `expect.arrayContaining` this will check if the members of the array are the same instead of checking the array as a whole including their order that their in. i assume this is also what we would like to test. i have tried you suggestion on switching the authenticators around but they would return in a different order every few test runs. i don't know what the cause is for this behaviour.

ndom91:
> hi @ndom91 i updated the `packages/utils/adapter.ts` test case to make use of the `expect.arrayContaining` this will check if the members of the array are the same instead of checking the array as a whole including their order that their in. i assume this is also what we would like to test. i have tried you suggestion on switching the authenticators around but they would return in a different order every few test runs. i don't know what the cause is for this behaviour.

Yeah no I think yuor `arrayContaining` is perfect for this :ok_hand:, good stuff!

Unfortunately there's been some minor bugs found in the new Drizzle adapter rewrite, so folks are working on updating it as we speak. If you're curious, there's some discussion in the discord:

- Invite: [discord.authjs.dev](https://discord.authjs.dev/)
- Drizzle Adapter Thread: [discord.com/channels/1200116961590399008/1231534836859146241](https://discord.com/channels/1200116961590399008/1231534836859146241)

But basically, its just going to be updating a few things that were pointed out towards the bottom of the original ["rewrite Drizzle adapter PR"](https://github.com/nextauthjs/next-auth/pull/10479).

We should wait with this Drizzle support until the adapter itself is back to a stable state. Sorry its taking so long :sweat_smile:. I'll write back here once that's all taken care of

packages/adapter-drizzle/src/lib/mysql.ts
```diff
@@ -439,9 +537,72 @@ export type DefaultMySqlVerificationTokenTable = MySqlTableWithColumns<{
...
+  dialect: "mysql"
+  schema: string | undefined
+}>
+
 export type DefaultMySqlSchema = {
   usersTable: DefaultMySqlUsersTable
   accountsTable: DefaultMySqlAccountsTable
   sessionsTable: DefaultMySqlSessionsTable
   verificationTokensTable: DefaultMySqlVerificationTokenTable
+  authenticatorsTable: DefaultMySqlAuthenticatorTable
```
juliusmarminge, line 607: at least the bottom 3 here should be optional imo. they are not required for every setup

packages/adapter-drizzle/src/lib/mysql.ts
```diff
@@ -252,21 +280,91 @@ export function MySqlDrizzleAdapter(
...
+        .select()
+        .from(authenticatorsTable)
+        .where(eq(authenticatorsTable.credentialID, credentialID))
+        .then((res) => fromDBAuthenticator(res[0]) ?? null)
+    },
+  }
+}
+
+type DrizzleAuthenticator = Required<
+  DefaultMySqlSchema["authenticatorsTable"]["$inferInsert"]
```
juliusmarminge, line 337: use official `InferInsertModel` helper instead of accessing internal types directly?

packages/adapter-drizzle/test/sqlite/schema.ts
```diff
@@ -5,6 +5,7 @@ export {
   sqliteAccountsTable,
   sqliteSessionsTable,
   sqliteVerificationTokensTable,
+  authenticatorsTable
```
ndom91, line 8:
```suggestion
  sqliteAuthenticatorsTable
```

packages/adapter-drizzle/test/sqlite/index.test.ts
```diff
@@ -6,18 +6,21 @@ import {
   sqliteSessionsTable as sessions,
   sqliteUsersTable as users,
   sqliteVerificationTokensTable as verificationTokens,
+  authenticatorsTable as authenticators,
```
ndom91, line 9:
```suggestion
  sqliteAuthenticatorsTable as authenticators,
```

ndom91:
So the `mysql` and `postgres` tests both passed for me. The `sqlite` schema had a small typo

packages/adapter-drizzle/src/lib/mysql.ts
```diff
@@ -76,17 +80,41 @@ export const mysqlVerificationTokensTable = mysqlTable(
...
 ): Adapter {
-  const { usersTable, accountsTable, sessionsTable, verificationTokensTable } =
-    schema
+  const {
+    usersTable,
+    accountsTable,
+    sessionsTable,
+    verificationTokensTable,
+    authenticatorsTable,
+  } = schema
```
juliusmarminge, line 117: Would it simplify stuff it you dont default assign the entire object in the param but rather each table on its own here? Then you shouldnt need all the `...sessionTable && { ... }` spaghetti

JipSterk: so i tried the following:
```tsx
const {
    usersTable = mysqlUsersTable,
    accountsTable = mysqlAccountsTable,
    sessionsTable = mysqlSessionsTable,
    verificationTokensTable = mysqlVerificationTokensTable,
    authenticatorsTable = mysqlAuthenticatorsTable,
  } = schema
```

but when running the test i would get the following error: `TypeError: Cannot read properties of undefined (reading 'usersTable')` so i don't think this is possible

ndom91:
We also shipped the new docs during the lifetime of this PR (:see_no_evil:), so could you include the `authenticator` schema updates in `/docs/pages/getting-started/adapters/drizzle.mdx`? :pray:

ndom91:
Can you update from `main`? 

I recently removed the JSDoc docs from the `drizzle/index.ts` file in favor of the `/docs/getting-started/adapters/drizzle.mdx` file, so we don't maintain two copies of the same docs. So could you add the "add 'authenticator' schema.." stuff to that mdx file isntead? Thanks!

packages/adapter-drizzle/src/lib/mysql.ts
```diff
@@ -270,21 +302,92 @@ export function MySqlDrizzleAdapter(
...
+  >
+
+function fromDBAuthenticator(
+  authenticator: DrizzleAuthenticator
+): AdapterAuthenticator {
+  const { transports, id, ...other } = authenticator
+
+  return {
+    ...other,
+    transports: transports || undefined,
```
juliusmarminge, line 370: Is there a reason not to allow this to be null? Why do we need to turn null -> undefined in a bunch of places?

Imo it would be better to change this in core lib to handle both null and undefined rather than having this `fromDBAuthenticator` helper

JipSterk: the `@auth/prisma-adapter` had a similar helper function to transform the object upon retrieval, which i based this drizzle pr on.
i tried putting in a helper function in `utils.ts` but don't know from where i would import `defineTables` from to infer the shape of the drizzle table.

juliusmarminge: my point is that this could be removed altogether if the core just accepted `string | undefined | null`, instead of just `string | undefined`. We could just return the response from the db call directly. Same is true for the Prisma adapter

ndom91: Yeah its really not doing anything else, right. 

How about this? It's just loosening up the type so it shouldn't be breaking anything 

https://github.com/nextauthjs/next-auth/pull/10861

ndom91: Okay so I also removed that fn from the prisma adapter in that PR ^^. Let me know what yall think

ndom91: @jipsterk We also dropped the `aurhwnticator.id` as the primary key and made the PK out of `credentialID` and `userId` (see PR above)

packages/adapter-drizzle/src/lib/mysql.ts
```diff
@@ -270,21 +302,64 @@ export function MySqlDrizzleAdapter(
...
+        .where(eq(authenticatorsTable.credentialID, data.credentialID))
+        .then((res) => res[0] ?? null)
+    },
+    async getAuthenticator(credentialID: string) {
+      const authenticator = await client
+        .select()
+        .from(authenticatorsTable)
+        .where(eq(authenticatorsTable.credentialID, credentialID))
+        .then((res) => (res.length ? res[0] : null))
+      return authenticator ? authenticator : null
```
juliusmarminge, line 322: authenticator??null

packages/adapter-drizzle/src/lib/pg.ts
```diff
@@ -240,21 +264,59 @@ export function PostgresDrizzleAdapter(
...
+
+      return user
+    },
+    async getAuthenticator(credentialID: string) {
+      const authenticator = await client
+        .select()
+        .from(authenticatorsTable)
+        .where(eq(authenticatorsTable.credentialID, credentialID))
+        .then((res) => (res.length ? res[0] : null))
+      return authenticator ? authenticator : null
```
juliusmarminge, line 282: Same as above

ndom91:
@JipSterk it looks like the `getAccount` method was missing from the 3 DB types and the `updateAuthenticatorCounter` method is supposed to throw, which it wasn't (even with `return await ..` syntax)

So I've updated those two issues in all 3 files (`sqlite.ts`, `pg.ts`, `mysql.ts`). I also cleaned up a few of the `return await ..` statements.

Hope thats okay!

JipSterk:
@ndom91 Thanks for your review/update any help it has to ship this feature out te happier i would be :) i thought of adding an entry to https://authjs.dev/getting-started/authentication/webauthn for drizzle, perhaps after that as well as all the tests clearing do you feel like we could ship it out? I’m eager to use it in my projects!

ndom91 (approved):

ndom91:
Hmm something seems to be breaking the E2E playwright tests here ([GHA](https://github.com/nextauthjs/next-auth/actions/runs/9127158836/job/25096872274?pr=10297)) :thinking: 

It's using the `/apps/dev/nextjs` app and playwright traces can be downloaded [here](https://github.com/nextauthjs/next-auth/actions/runs/9127158836/artifacts/1512922386).

ndom91:
Decided to take a closer look and it definitely passes for me locally, so I think this is good to go!

image

ndom91 (approved):

JipSterk:
Thanks @ndom91 for resolving & merging this!

PR #10305, opened 2024-03-13
fix: throw `MissingSecret` when secret missing

ndom91:
## ☕️ Reasoning

- We had thrown `MissingSecret` if `config.secret` wasn't set, but not early enough in the flow, so a missing `AUTH_SECRET` got through in early signin and threw a mysterious error when passed to hkdf
- Added another throw of this error if `AUTH_SECRET` isn't set in our core `setEnvDefaults` fn, since that's the only core place where one of `config.secret` or `AUTH_SECRET` should definitely have been set already. If you know a better place for it I'm all ears though 👍 

Previous error msg on missing `AUTH_SECRET`:

image

New error msg:

image

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/utils/env.ts
+++ b/packages/core/src/lib/utils/env.ts
@@ -1,4 +1,5 @@
 import type { AuthAction, AuthConfig } from "../../types.js"
+import { MissingSecret } from "../../errors.js"
 
 /** Set default env variables on the config object */
 export function setEnvDefaults(envObject: any, config: AuthConfig) {
@@ -18,6 +19,12 @@ export function setEnvDefaults(envObject: any, config: AuthConfig) {
       const secret = envObject[`AUTH_SECRET_${i}`]
       if (secret) config.secret.unshift(secret)
     }
+  }
+
+  if (!config.secret?.length) {
+    throw new MissingSecret(
+      "Missing secret, please set AUTH_SECRET or config.secret"
+    )
   }
 
   config.redirectProxyUrl ??= envObject.AUTH_REDIRECT_PROXY_URL
--- a/packages/core/test/env.test.ts
+++ b/packages/core/test/env.test.ts
@@ -18,6 +18,7 @@ beforeEach(() => {
 describe("config is inferred from environment variables", () => {
   it("providers (client id, client secret, issuer, api key)", () => {
     const env = {
+      AUTH_SECRET: "asdf",
       AUTH_AUTH0_ID: "asdf",
       AUTH_AUTH0_SECRET: "fdsa",
       AUTH_AUTH0_ISSUER: "https://example.com",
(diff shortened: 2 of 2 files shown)


ThangHuuVu (approved):

DonikaV:
I have AUTH_SECRET in my .yaml file and it works in beta.5 but in beta.16 I am getting 
**Missing secret, please set AUTH_SECRET or config.secret.**

ndom91:
Hey folks, yeah so setting the `secret` option on your main config or the `AUTH_SECRET` env var should be enough to not trigger this. Looking into it..

DuarteMartinho:
Checked my production container for environment variable and AUTH_SECRET is set 

image

@ndom91 Let me know what to do as this is kind of urgent, thanks!

DuarteMartinho:
@DonikaV Found a fix for now 

Use beta15 instead of beta16

LoisDuplain:
Please revert this merge.

ndom91:
I can't reproduce this in dev and prod builds of SvelteKit apps with Auth.js nor in dev / prod Next.js apps :thinking:

Can you all provide some more details about your setups?

bpo-impact:
Having same issue as @adamspotlite  we keep our secrets in aws secret manager and they are not available at build step only at runtime. We need to be able to build the code and then at runtime pass the secret.

jdmnk:
Facing the same issue where secret is not known at build time and this change breaks the build if the secret is only known at runtime.

laeo:
Can't preview or deploy to Cloudflare Pages, when trying preview it reports same ERROR message.

matthawk60:
The only solution for me other than hard coding the key into the auth.ts file was to switch to NEXT_PUBLIC_AUTH_SECRET as the environment variable.  This environment variable is allowed in a client context on next, no other variable without NEXT_PUBLIC will work.  This leads me to believe NEXT is preventing it from being included when it thinks key will be shown on a client.

jacogasp:
Hello everyone,

> ump! I m also getting "Missing secret, please set AUTH_SECRET or config.secret. Read more at

Same behaviour here. I'm receiving the error message even if I set up `AUTH_SECRET` in `.env` file.
I tried the following cases:

1. Simply set `AUTH_SECRET` as env variable: not working
2. Set `AUTH_SECRET` env variable and in auth config I set `secret: process.env.AUTH_SECRET`: not working
3. Hardcoding the secret `secret: "foo"` into the auth config: it works.

What is really wired to me is that if I do a `console.log(process.env.AUTH_SECRET)` it correctly prints the secret:

```ts
// auth.config.ts

console.log("secret:", process.env.AUTH_SECRET);

export const authConfig: NextAuthConfig = {
  providers: [...],
  secret: process.env.AUTH_SECRET,
  ...
}
```
and the output is

```
> next dev -p 8080

  ▲ Next.js 14.2.1
  - Local:        http://localhost:8080
  - Environments: .env

 ✓ Starting...
 ✓ Ready in 5.6s
 ○ Compiling /src/middleware ...
 ✓ Compiled /src/middleware in 928ms (220 modules)
secret: foo
 ○ Compiling / ...
 ✓ Compiled / in 3s (1894 modules)
secret: foo
secret: foo
 GET / 200 in 3956ms
 GET / 200 in 3673ms
```
proving that the secret is valid each time the config is invoked. But still I receive the missing AUTH_SECRET error.
```
Unhandled Runtime Error

MissingSecret: Missing secret, please set AUTH_SECRET or config.secret. Read more at https://errors.authjs.dev#missingsecret


src/auth.ts (4:61) @ authConfig

  2 | import { authConfig } from "./auth.config";
  3 |
> 4 | export const { auth, handlers, signIn, signOut } = NextAuth(authConfig);
    |                                                             ^
  ```

Tried also with `NEXTAUTH_SECRET` but same result.

Apparently hardcoding the secret in the auth config is the only working method.
I'm currently testing v5 beta 16

PR #10339, opened 2024-03-16
fix(sveltekit): skip auth routes on prerender

ndom91:
## ☕️ Reasoning

- Make SvelteKit Auth work during prerender
- Skip calling the actions during prerender (`if (building === true)`)
- Currently not working because the dynamic routes throw a `404` during prerender still. If theres a `<a href="/auth/signin" ..` anywhere on your page, during prerender sveltekit will try to render that route and it doesn't exist, since we handle it dynamically via the `handle` we export and use in the `hooks.server.ts`.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


dievardump:
SvelteKit has a configuration setting that allows to *force a prefix to private variables*. https://kit.svelte.dev/docs/configuration#env

Plugins can not safely rely on importing variables from `$env/dynamic/private`, because if this is set for the project, those variables will NOT appear in `$env/dynamic/private`

You should use `process.env`, else SvelteKitAuth is sure to have other bug reports in the future

ThangHuuVu:
@dievardump With the env object from `$env/dynamic/private` we are guaranteed to have support for different adapters - that's the main reason we want to use it in the sveltekit auth package 👀
I understand your concern with the prefix, and lets fix it in a separate issue/PR 🤔

ndom91:
Yeah sorry, didn't mention that in my previous post. 

If we use process.env we lose support for other certain environments like cloudflare. Which the env/dynamic/private would handle with the cf-adapter

WhyAsh5114:
Could I please request an update on this? I would really appreciate it.

ndom91:
@WhyAsh5114 yeah so unfortunately theres still this bug:

> Currently not working because the dynamic routes throw a 404 during prerender still. If theres a <a href="/auth/signin" .. anywhere on your page, during prerender sveltekit will try to render that route and it doesn't exist, since we handle it dynamically via the handle we export and use in the `hooks.server.ts`.

WhyAsh5114:
But why would someone do that, don't we use the `signIn()` function in the client side API?

ndom91:
@WhyAsh5114 good question, I ran into the issue a few times initially when working on this PR, but I can't get it to happen again..

Both the package `@auth/sveltekit` and the dev app (`apps/dev/sveltekit`) seem to build fine in this PR :thinking: 

@ThangHuuVu do you remember exactly what the issue was?

basaran:
@ndom91, it seems sveltekit dev example is not using prerender in the layout.ts. That might be the reason why it's building for you. On the latest authjs and sveltekit, I'm still getting the compile errors:

1. If I don't have `basePath: "/api/auth",` in the handler, I get the:

```
TypeError: Cannot read properties of undefined (reading 'length')
// .svelte-kit/output/server/chunks/auth.js:6447:26)
```

2. If I add the basePath, then I get the env error.

```
Error: Cannot read values from $env/dynamic/private while prerendering (attempted to read env.AUTH_URL). Use $env/static/private instead
```

dievardump:
For me to fix all this, I simply remove SvelteKitAuth from the sequence() in hook.server when in build mode.

> `Error: Cannot read values from $env/dynamic/private while prerendering (attempted to read env.AUTH_URL). Use $env/static/private instead`

This is another error. Typically `$env/*` should not be accessed directly by tools around svelte-kit because the prefixes are configurable and those tools can not know that. Any expected configuration should be passed as configuration. It's an error from sveltekitauth to access that directly. 

I opened an issue in SvelteKit to add that to the doc. It should be mentioned there that it's not safe to rely on `$env` anywhere outside the project codebase.

basaran:
Could you paste the logic you have in the server hooks please?

basaran:
For `testing` purposes, I was able to build when prerender is enabled.

```
// hooks.servers.ts
export const handle: Handle = building
    ? ({ event, resolve }) => resolve(event)
    : authenticationHandle;
```

```
// layout.server.ts
    if (building) {
        console.log("building");
        return {};
    }

    const session = await event.locals.auth();
    return {
        session,
    };
};
```

While doing this, I didn't have to provide the basePath in auth.ts.

```
mport { SvelteKitAuth } from '@auth/sveltekit';
import GitHub from '@auth/sveltekit/providers/github';
 
export const { handle, signIn, signOut } = SvelteKitAuth({
  providers: [GitHub]
});
```
I hope this would be helpful to you to fix this properly.

ndom91:
@dievardump @basaran okay so this PR would get you building then though? WIthout having to manually skip SvelteKitAuth handle during build? 

Also thanks for the `$env/..` pointer. We'll take a closer look at that..

ndom91:
> @ndom91, I couldn't succeed unfortunately. This is what I did for building:
> 
> 1. Cloned the branch you provided and built.
> 2. Pulled the test repo from https://github.com/basaran/authjs-debug
> 3. This is the repo in the app folder, but with prerender enabled.
> 4. pnpm linked to the built.
> 
> During the sveltekit build, I received: 
> Error: 404 /auth/signin (linked from /protected)
> 
> I also received a debug error from authjs during dev (while using the linked package), and when clicked on signup button. 
> 
> ```
> Redirect {
>   status: 302,
>   location: 'http://localhost:3000/auth/signin?callbackUrl=http%3A%2F%2Flocalhost%3A3000%2F'
> }
> ```
> 
> Perhaps my package built was wrong?

Hmm yeah so that package / example app actually does have an anchor tag linking to `/auth/signin`, triggering the bug I mentioned earlier. 

But as someone else in this thread mentioned, that's going to be extremely unlikely. @basaran what if you remove that anchor tag on the `/protected` route? Does it build then?

WhyAsh5114:
Instead of completely getting rid of the route, I'd also like to know what happens if we instead use the `signIn()` function as suggested by the docs.

ndom91:
> Instead of completely getting rid of the route, I'd also like to know what happens if we instead use the `signIn()` function as suggested by the docs.

That's used throughout the example app, seems to work, right? I'd love some addition confirmation too though 🙏

ndom91:
So I'm back at the computer, and I tried this again. This time without the anchor tag in the `../routes/protected/+page.svelte` route, and i still got the error(s) you mentioned regarding the basePath (`length`) and then the same `env.AUTH_URL`.. :thinking:

basaran:
I am also unable to use server-side login with the built. I get a 500 and the following in the terminal console.

```
[auth][warn][debug-enabled] Read more: https://warnings.authjs.dev#debug-enabled
[auth][debug]: CREATE_PKCECODEVERIFIER {
  "value": "...removed...",
  "maxAge": 900
}
[auth][debug]: authorization url is ready {
  "url": "https://accounts.google.com/o/oauth2/v2/auth?response_type=code&client_id=...removed...&redirect_uri=http%3A%2F%2Flocalhost%3A5173%2Fauth%2Fcallback%2Fgoogle&code_challenge=...removed...&code_challenge_method=S256&scope=openid+profile+email",
  "cookies": [
    {
      "name": "authjs.pkce.code_verifier",
      "value": "....removed...",
      "options": {
        "httpOnly": true,
        "sameSite": "lax",
        "path": "/",
        "secure": false,
        "maxAge": 900,
        "expires": "2024-04-17T22:10:36.590Z"
      }
    }
  ],
  "provider": {
    "id": "google",
    "name": "Google",
    "type": "oidc",
    "issuer": "https://accounts.google.com",
    "style": {
      "brandColor": "#1a73e8"
    },
    "clientId": "...removed....",
    "clientSecret": "...removed....",
    "signinUrl": "http://localhost:5173/auth/signin/google",
    "callbackUrl": "http://localhost:5173/auth/callback/google",
    "wellKnown": "https://accounts.google.com/.well-known/openid-configuration",
    "checks": [
      "pkce"
    ]
  }
}
Redirect {
  status: 302,
  location: 'https://accounts.google.com/o/oauth2/v2/auth?response_type=code&client_id=...removed...&redirect_uri=http%3A%2F%2Flocalhost%3A5173%2Fauth%2Fcallback%2Fgoogle&code_challenge=...removed...&code_challenge_method=S256&scope=openid+profile+email'
}
````

I confirmed it works fine in dev mode with the npm package.

basaran:
It's possible I am building this the wrong way, are there any instructions I can follow to build the sveltekit package and link it to my local repository?

ndom91:
> It's possible I am building this the wrong way, are there any instructions I can follow to build the sveltekit package and link it to my local repository?

For server-side signin with SvelteKit, it requires those additional `+page.server.ts` actions, you have those? (Check out the "server side" svelte section here: https://authjs.dev/getting-started/session-management/login?framework=sveltekit)

basaran:
yep, I already have those. My test repo works fine if I switch over to the npm installed package. But same thing happens, if I build the master branch and link it. I will attempt again this evening.

ThangHuuVu (approved):
Tested this PR, it works (build successful) for our example project with a prerender page. For the issue with `<a>` tag to the default page like signin, let's put a line in the doc to explain that it is expected. 

@dievardump feel free to open a new PR to fix the envPrefix thing!

ThangHuuVu (approved):

basaran:
@ThangHuuVu, could you share how you built and installed the local package?

basaran:
I had to pack and then install @auth/core and @auth/sveltekit. Thank you.

PR #10360, opened 2024-03-19
fix: api routes not receiving real host, leading to localhost redirect after login

MichaelErmer:
This fix resolves next auth not using the correct hostname for requests using the API methods, which ultimately led to users being redirected to localhost:3000 after login if next run behind a reverse proxy.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources


MichaelErmer:
Anything I can do to speed this up?

ndom91:
Did you enable `AUTH_TRUST_HOST` (see [docs](https://authjs.dev/getting-started/deployment#auth_trust_host)) or try the `AUTH_REDIRECT_PROXY_URL` ([docs](https://authjs.dev/getting-started/deployment#auth_redirect_proxy_url))?

This should be covered by those two options. Let me know if those still don't work for you :pray:

MichaelErmer:
Yes, set both, but still the redirect after idp led to localhost.

ndom91:
Are you sure you updated the `callbackUrl` in your OAuth provider? 

We've got an example for Docker up at https://nextjs-docker-example.authjs.dev which does work behind a reverse proxy with all providers :thinking:

MichaelErmer:
> Are you sure you updated the `callbackUrl` in your OAuth providers dashboard?
> 
> We've got an example for Docker up at https://nextjs-docker-example.authjs.dev which does work behind a reverse proxy with all providers 🤔

Do you suggest using an absolute callbackUrl in the signIn call? 
We are using multiple domains and I think the trust-proxy-option should enable the auth to use the domain from x-forwarded-for 🤔

ndom91:
@MichaelErmer ah okay I think I see what you're saying.

No so you don't have to pass an absolute URL to the `signIn()` call, but you do have to modify your `callbackUrl` to be the correct (prod?) URL instead of local dev. I'd recommend two apps in your IdP.

We ran into this a few times, I forget if Cognito was one but I assume from what you're saying it is - Twitch, for example, allows adding multiple callbackUrls, but it'll always redirect to the first one. So if you add a local dev callbackUrl (`http://localhost:3000/api/auth/twitch`) and a prod one (`https://app.company.com/api/auth/callback/twitch`), it'll always only set the first (local) one as callbackUrl. 

If you create two separate OAuth applications in your IdP, you'll just have to use two separate `clientId`/`clientSecret` env vars, 1 for dev 1 for prod, just like with Github for example.

MichaelErmer:
I understand what you are saying, we wouldn't want our customers to have to setup multiple Apps on their sso for different systems of us, they whitelist only our tld or a list of domains.

This change fixes the issue, in a generic way, by initiating the signIn at the IDP using the (from users perspective) correct return url, whilst ensuring and respecting trust host is enabled etc.

MichaelErmer:
@ndom91 tried with the latest version, both set `AUTH_TRUST_HOST` and `AUTH_REDIRECT_PROXY_URL`, the "login" itself works, BUT after the "login" result is handled at `/api/auth/callback/cognito` the user is then redirected to `https://0.0.0.0:3000/dashboard`

dprothero:
I can confirm this issue, as I'm running into it too. To be consistent, however, I believe this PR should check the `x-forwarded-host` like [here](https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/lib/utils/env.ts#L88)

@MichaelErmer are you still looking to get this merged? Do you want to update your PR to use that header?

dprothero:
Without some kind of fix like this, running your Next.js site with NextAuth behind a proxy that doesn't pass the `Host` header isn't possible that I've been able to determine. Even with `AUTH_TRUST_HOST`, this part of the code isn't even bothering to check the `X-Forwarded-Host` header.

dprothero:
UPDATE: I just discovered if I set the `AUTH_URL`, then I can get it to work.

This PR would still be nice to make things work automatically based on headers, however.

PR #10376, opened 2024-03-21
feat(email) allow for custom mail implementations

JipSterk:
## ☕️ Reasoning

currently solutions using a custom configuration of `sendVerificationRequest` for the `nodemailer` provider still require a `server` config object passed even though it's type is not mandatory which is not need for implementations that for example use a sdk like `aws-ses` or `react-email`

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/src/providers/nodemailer.ts
+++ b/packages/core/src/providers/nodemailer.ts
@@ -1,13 +1,13 @@
 import { createTransport } from "nodemailer"
-import { html, text } from "../lib/utils/email.js"
 import { AuthError } from "../errors.js"
+import { html, text } from "../lib/utils/email.js"
 
 import type { Transport, TransportOptions } from "nodemailer"
 import * as JSONTransport from "nodemailer/lib/json-transport/index.js"
 import * as SendmailTransport from "nodemailer/lib/sendmail-transport/index.js"
 import * as SESTransport from "nodemailer/lib/ses-transport/index.js"
-import * as SMTPTransport from "nodemailer/lib/smtp-transport/index.js"
 import * as SMTPPool from "nodemailer/lib/smtp-pool/index.js"
+import * as SMTPTransport from "nodemailer/lib/smtp-transport/index.js"
 import * as StreamTransport from "nodemailer/lib/stream-transport/index.js"
 import type { Awaitable, Theme } from "../types.js"
 import type { EmailConfig } from "./email.js"
@@ -40,7 +40,7 @@ export interface NodemailerConfig extends EmailConfig {
     theme: Theme
     request: Request
   }) => Awaitable<void>
-  options: NodemailerUserConfig
+  options?: NodemailerUserConfig
 }
 
 export type NodemailerUserConfig = Omit<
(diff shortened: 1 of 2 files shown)


packages/core/src/providers/nodemailer.ts
```diff
@@ -51,32 +51,35 @@ export type NodemailerUserConfig = Omit<
...
+  if (!config.sendVerificationRequest && !config.server)
+    throw new AuthError(
+      "Nodemailer requires a `server` configuration when sendVerificationRequest isn't overridden"
+    )
 
   return {
     id: "nodemailer",
     type: "email",
     name: "Nodemailer",
-    server: { host: "localhost", port: 25, auth: { user: "", pass: "" } },
```
JipSterk, line 61: i couldn't figure out why this server object was needed outside of development purposes

ndom91:
I think this PR is no longer necessary. We shipped multiple http request Email providers (which can be extended). Can you check out these links and let me know if that covers it for yuo?

- Custom Email Provider Guide: https://authjs.dev/guides/configuring-http-email
- Example new HTTP Email Provider (Resend) guide + customization examples: https://authjs.dev/getting-started/providers/resend

JipSterk:
hi @ndmom thanks for the suggestion, i haven’t stumbled upon this page in the docs and it seems like this is exactly what i want! however when i define an object with the same properties the typescript compiler will complain that the following properties are missing: `name`, `from`, `maxAge` and `options` which is of the `EmailUserConfig` type with basically the same options. i assume these aren't required? if so i can update the pr which makes these properties optional to reflect the example on the docs. just to make sure but these are the versions i use:
```
"next-auth": "5.0.0-beta.18",
"@auth/core": "0.31.0",
```

ndom91:
@JipSterk which object are you referring to? Could you provide some code example or point to the relevant piece in the example pages I linked to above? 

Also what are you using `@auth/core` for exactly? That shouldn't be necessary to install, only `next-auth`, for example 🤞

JipSterk:
that would be [EmailConfig](https://github.com/nextauthjs/next-auth/blob/a616aad328054a3fa65541e6baa4d8cfbc7171e0/packages/core/src/providers/email.ts#L29) it requires the properties mentioned above which are: `name`, `from`, `maxAge` and `options`. in the example shown on [the page](https://authjs.dev/guides/configuring-http-email) you mentioned these properties are not defined so is the example wrong or could these types become nullable?

ndom91:
@JipSterk thanks for pointing this out, it looks like both could use a bit of updating.

1. The example should include most of those fields. 
	- `maxAge` is important for defining how long the verificationTokens should be valid. Although we could safely default to something like 24hrs in the `@auth/core` code..
	- `name` is just the name of the provider that by default gets put into the btn on the login page, like "Login with X"
	- `options` we don't really need in this situation actually and can be made optional
	- `from` is also potentially optional depending on how you implement the `sendVerificationRequest` function yourself. If you hardcode a `from` address, like in the example in the guide link from above, then you don't need to pass it in at the provider level here.
	
2. So regarding the types..
	- `from` can be made optional
	- `optoins` can be made optional


I've updated the guide to show the use of `maxAge` and `name`. Could you cover the type updates? 

I dont thinkt he changes to the `nodmeailer` provider in this PR are necessary anymore, right? What do you think?

PR #10391, opened 2024-03-24
fix(docs): modify import path

hjick:
I fixed the App Router example documentation.
The import path was for the Pages Router.


balazsorban44:
Thanks, but this should be correct. In fact, from an App Router `route.ts` file, you cannot export anything other than the named HTTP handler functions (GET, POST, etc), and some Next.js config like `runtime` etc.

PR #10411, opened 2024-03-26
chore(docs): update credentials-provider tutorial svelte example

### LINKED ISSUE #10406: Sveltekit documentation using depreceted method

route/+layout.server.ts
```typescript
import type { LayoutServerLoad } from './$types';

  return {
    session: await event.locals.getSession()
  };
};
```

**getSession() is depreceted**

no

https://authjs.dev/getting-started/providers/credentials-tutorial

AlbertMarashi:
Documentation fix using new svelte recommended syntax

Fixes #10406

### DIFF
--- a/docs/docs/getting-started/providers/credentials-tutorial.mdx
+++ b/docs/docs/getting-started/providers/credentials-tutorial.mdx
@@ -124,7 +124,7 @@ export default NextAuth({
   ```ts title="route/+layout.server.ts"
   export async function load({ locals }) {
       return  { 
-         session = await locals.auth();
+         session: await locals.auth();
       }
   }
   ```
--- a/docs/docs/getting-started/providers/credentials-tutorial.mdx
+++ b/docs/docs/getting-started/providers/credentials-tutorial.mdx
@@ -124,7 +124,7 @@ export default NextAuth({
   ```ts title="route/+layout.server.ts"
   export async function load({ locals }) {
       return  { 
-         session = await locals.getSession();    
+         session = await locals.auth();
       }
   }
   ```
--- a/docs/docs/getting-started/providers/credentials-tutorial.mdx
+++ b/docs/docs/getting-started/providers/credentials-tutorial.mdx
@@ -122,13 +122,11 @@ export default NextAuth({
   Auth.js provides us a getSession, function to access the session data and status, to call from the `event.locals` variable. We can now just call it and add it to our `$page` store.
 
   ```ts title="route/+layout.server.ts"
-  import type { LayoutServerLoad } from './$types';
-  
-  export const load: LayoutServerLoad = async (event) => {
-    return {
-      session: await event.locals.getSession()
-    };
-  };
+  export async function load({ locals }) {
+      return  { 
+         session = await locals.getSession();    
+      }
+  }
   ```
 
   ### Consuming the session via page store


docs/docs/getting-started/providers/credentials-tutorial.mdx
```diff
@@ -122,13 +122,11 @@ export default NextAuth({
...
-  import type { LayoutServerLoad } from './$types';
-  
-  export const load: LayoutServerLoad = async (event) => {
-    return {
-      session: await event.locals.getSession()
-    };
-  };
+  export async function load({ locals }) {
+      return  { 
+         session = await locals.getSession();
```
ndom91, line 127:
```suggestion
         session = await locals.auth();
```

AlbertMarashi:
@ndom91  I just ran through this tutorial and the `session` is null which is bizzare.. Any idea why this might be happening?

ndom91:
> @ndom91  I just ran through this tutorial and the `session` is null which is bizzare.. Any idea why this might be happening?

No, could you share your config?

AlbertMarashi:
@ndom91 does credentials work when using a database adapter? When I disable the DB adapter, the `.auth()` returns the session, but not when the adapter is enabled

ndom91:
> @ndom91 does credentials work when using a database adapter? When I disable the DB adapter, the `.auth()` returns the session, but not when the adapter is enabled
> 
> If so, that should probably be noted in the docs.
> 
> Happy to note that if you can confirm

Yeah it should. When you add a DB adapter though, you automatically use the DB sessions strategy. You can force it to use jwts again (while the DB adapter is enabled) with something like this: 

``` 
session: {
strategy: "jwt"
}
```

AlbertMarashi:
Hmm thats weird because it wasn't working for me

ndom91:
Does it work if you set the session back to `jwt` strategy? (See: https://authjs.dev/reference/core#session)

AlbertMarashi:
> Does it work if you set the session back to `jwt` strategy? (See: https://authjs.dev/reference/core#session)

Yes

docs/docs/getting-started/providers/credentials-tutorial.mdx
```diff
@@ -122,13 +122,11 @@ export default NextAuth({
...
 
   ```ts title="route/+layout.server.ts"
-  import type { LayoutServerLoad } from './$types';
-  
-  export const load: LayoutServerLoad = async (event) => {
-    return {
-      session: await event.locals.getSession()
-    };
-  };
+  export async function load({ locals }) {
```
ndom91, line 125: How would you type this then?

themixednuts: Sorry if I'm misunderstanding but SvelteKit generates the types automatically through 

```
+page.ts -> PageLoad
+page.server.ts -> PageServerLoad
+layout.ts -> LayoutLoad
+layout.server.ts -> LayoutServerLoad
```

AlbertMarashi: That's correct

ndom91: Yeah, I tried that, it didn't pick up the `.svelte-kit/*` autogenerated types, but maybe it's just an issue with my IDE then

AlbertMarashi: You need to have Svelte extension installed - in case it wasn't. Sometimes reloading the window can also fix those issues.

ndom91:
Yeah I use neovim and normally don't have any other problems with sveltekit projects, but okay if it worked for you all :thumbsup:

ndom91:
We're about to ship a new set of docs, so I'm going to close this PR, thanks for your interest though :pray: 

Preview deploy here: https://authjs-nextra-docs.vercel.app

PR #10421, opened 2024-03-27
fix: `createActionURL` set detectedProtocol correctly

Carlos-err406:
## ☕️ Reasoning
the newest implementation of `createActionURL` occasionally throws ERR_INVALID_URL because of double colons in protocol

this fix just checks if the detectedProtocol ends with a colon, and if not it adds it before creating the url with the detected host

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues
Fixes: #10408

## 📌 Resources

- [Security guidelines](https://github.com/nextauthjs/.github/blob/main/SECURITY.md)
- [Contributing guidelines](https://github.com/nextauthjs/.github/blob/main/CONTRIBUTING.md)
- [Code of conduct](https://github.com/nextauthjs/.github/blob/main/CODE_OF_CONDUCT.md)
- [Contributing to Open Source](https://kcd.im/pull-request)

PR #10439, opened 2024-03-29
URL: protocol always ends with colon

adainrivers:
## ☕️ Reasoning

URL.protocol always ends with colon but the implementation adds another colon symbol, causing the URL constructor to throw invalid URL

See: https://developer.mozilla.org/en-US/docs/Web/API/URL/protocol


adainrivers:
Duplicate for https://github.com/nextauthjs/next-auth/pull/10421

PR #10542, opened 2024-04-11
fix(pages): cleanup oauth provider branded buttons

ndom91:
## ☕️ Reasoning

- Cleanup default `/signin` page provider btns, using dynamic colors to make themes work in more cases

### Before: 

<table>
	<tr>
		<td>
			
		</td>
		<td>
			
		</td>
	</tr>
</table>

Notice that the bottom few providers (`WorkOS`, `Twitter`, `Twitch`) all do have logos, just they're invisible because they're the same color as the brand color, which is being used as the button background color atm.

### After: 

<table>
	<tr>
		<td>
			
		</td>
		<td>
			
		</td>
	</tr>
</table>

So this cleans up a good bit:

- First of all, removes unnecessary `logoDark`, `textDark` variables (were only being used by 1 provider each). Also not a user-facing API, so shouldn't be "breaking", right?
- Background colors are just white / default dark colors in light/dark mode. And then we can use the `bg`/`brandColor` provider `style` option (which most providers have set, I've filled in a few more) and it automatically sets the hover color to that a low opacity version of the brand color with [color-mix](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix). `color-mix` is in the mdn "baseline 2023" feature set, meaning all browsers do [support it](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix#browser_compatibility) now :pray: 

If yall are cool with this, I can clean up the rest of the providers `style` objects here before merging this :+1: 

### After with Hover Brand Colors

<table>
	<tr>
		<td>
			
		</td>
		<td>
			
		</td>
	</tr>
	<tr>
		<td>
			
		</td>
		<td>
			
		</td>
	</tr>
</table>

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


0ubbe (commented):
Looking good 🙌🏽 

Two small suggestions:

Screenshot 2024-04-14 at 11 08 08

We can make the white box a bit wider to take advantage of the space on the page, and like that the Azure button won't have 2 lines of text (_which looks slightly odd_) ?

Screenshot 2024-04-14 at 11 10 10

For providers where we don't display an icon, we could display a default icon? 🤔 , maybe [the login one from phospor](https://phosphoricons.com/?q=%22login%22)?

ndom91:
@ubbe-xyz that's the "before", the "after" is below - the all white / dark btns with the brand color only appearing on hover. 

Also a part of the problem with the existing one, was your last comment - those providers actually do have icons, its just their icons are just their brand color so they're totally invisible against a background of the same color :joy:

0ubbe (approved):
Looking good! We can merge and iterate later 🙌🏽

Issue #10009, opened 2024-02-12
Sign In page Buttons don't respect basePath when clicked


jenewland1999:
### Environment

System:
    OS: macOS 14.3.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 1.34 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - ~/Library/Caches/fnm_multishells/47131_1707656901311/bin/node
    Yarn: 1.22.21 - ~/Library/Caches/fnm_multishells/47131_1707656901311/bin/yarn
    npm: 10.2.4 - ~/Library/Caches/fnm_multishells/47131_1707656901311/bin/npm
    pnpm: 8.15.1 - ~/Library/pnpm/pnpm
    bun: 1.0.26 - ~/.bun/bin/bun
  Browsers:
    Chrome: 121.0.6167.160
    Safari: 17.3.1
  npmPackages:
    next: ^14.1.0 => 14.1.0 
    next-auth: 5.0.0-beta.9 => 5.0.0-beta.9 
    react: ^18.2.0 => 18.2.0

### Reproduction URL

https://github.com/jenewland1999/next-auth-bug-repro

### Describe the issue

Using the `basePath` option in Next.js does not currently work with the latest beta of NextAuth.js.

This is because when you're taken to the sign in or sign out page, clicking the button will send you to a URL that does not include the `basePath`. Setting the `basePath` option in NextAuth.js does not affect this.

Furthermore, the use of the `AUTH_URL` environment variable causes all the endpoints NextAuth.js creates to return the following error message:

> "Error: This action with HTTP GET is not supported."

### How to reproduce

1. Clone the example
2. Create a .env file with the AUTH_SECRET and your Provider secrets of choice (in my case that's Azure AD)
3. If you've chosen a different provider, configure accordingly in the auth.ts file
4. Start the dev server and navigate to `http://localhost:3000/foo`
5. Click the login button

You will now be on a 404 page.

### Expected behavior

The `basePath` is taken into account when clicking the sign in or sign out button on the inbuilt pages. And specifying AUTH_URL also doesn't return the above error message.

sx-bpreck:
I am facing the same issue on version `4.24.7`

smultar:
I have a similar issue, that leads me to be constantly redirected back to my sign-in page or my local host.

My production environment is a docker swarm, with caddy infront of it. It works locally, but not in my prod.

[...]

Issue #9984, opened 2024-02-10
Adding base path to nextjs breaks oauth/oidc signin


amsal:
### Environment

  System:
    OS: macOS 14.1
    CPU: (11) arm64 Apple M3 Pro
    Memory: 1.84 GB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - ~/.nvm/versions/node/v20.11.0/bin/node
    npm: 10.2.4 - ~/.nvm/versions/node/v20.11.0/bin/npm
  Browsers:
    Chrome: 121.0.6167.160
    Safari: 17.1
  npmPackages:
    next: 14.1.0 => 14.1.0 
    next-auth: ^5.0.0-beta.9 => 5.0.0-beta.9 
    react: ^18 => 18.2.0 

### Reproduction URL

https://github.com/amsal/next-auth-basepath

### Describe the issue

Setting basePath for next app in next config does not set the base path for next-auth sign in.

### How to reproduce

Run the project (Next basePath is set to `myapp`). Go to sign in page `http://localhost:3000/myapp/api/auth/signin` which outputs 
```
"Error: This action with HTTP GET is not supported."
```

Changing the basePath from `'/myapp/api/auth'` to ``/api/auth'` in auth.ts shows the sign in page again but the urls are not prefixed with base path. Form action url is `http://localhost:3000/api/auth/signin/github` and callback url is `http://localhost:3000`

Also setting the environment variable `AUTH_URL` did not solve the issue

### Expected behavior

Should display the login page including urls with correct base path set

dichioniccolo:
The problem seems to be present from beta.9, having the same issue with Magic Links after upgrading

stunaz:
exactly... there is a nasty bug on beta.9

jenewland1999:
> Funnily enough this happens regardless of if you set the `basePath` option in either Next.js or Auth.js. If I had to guess, there's either a bug or they've purposely removed the ability to hit that route and have a page returned and instead it's just a POST route now.
> 
> I've tried this with beta.9 both with the latest next release and the latest next canary.

Turns out if I removed the AUTH_URL (or NEXTAUTH_URL) and restarted the dev server things worked again. Not sure if this is the case with basePath too, but worth a try.

jenewland1999:
Seems if you set AUTH_URL everything breaks.
[...]

Issue #8841, opened 2023-10-11
Custom basePath not working in Vercel


kldavis4:
### Environment

It is possible to use a custom base path for the auth handler endpoint in a local environment by setting the NEXTAUTH_URL to the full path (https://next-auth.js.org/configuration/options#nextauth_url), but when deployed to Vercel, there is no way to configure the custom base path and NEXTAUTH_URL is ignored.

### Reproduction URL

https://github.com/kldavis4/nextauth-custom-path-issue

### Describe the issue

In the reproduction project, I have moved the auth endpoint from `/api/auth` to `/api/foobar/auth`. I've setup a discord auth provider and specified the `NEXTAUTH_URL=http://localhost:3000/api/foobar/auth`. With this setup, I am able to successfully login using discoard oauth. If I deploy the project to Vercel, the login form action has the default base path, and not the one specified in the NEXTAUTH_URL.

It seems like the culprit is [detectOrigin](https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/utils/detect-origin.ts#L2) which will return the full NEXTAUTH_URL when not running in Vercel, but only return the protocol+host in Vercel. There should be a way to configure the custom base path or detectOrigin should respect the path specified in NEXTAUTH_URL regardless of whether it is in Vercel.

### How to reproduce

- Clone the reproduction repository. Install deps (`pnpm i`).
- Create .env.local and populate it with env vars for an oauth provider and update the auth.ts to specify the auth provider being used. 
- Set `NEXTAUTH_URL=http://localhost:3000/api/foobar/auth` in .env.local.
- Run the project locally `pnpm run dev` and verify it is possible to login
- Add `VERCEL=1` to .env.local and run the project again and verify that it is no longer possible to login

### Expected behavior

Using a custom base path should work whether running in Vercel or not.

NikitaKlimuk:
Hello, were you able to solve this problem? 
I'm having the same problem right now with a redirect that only works on localhost

kldavis4:
> Hello, were you able to solve this problem? I'm having the same problem right now with a redirect that only works on localhost

I have not - let me know if you find a way to work around it

NikitaKlimuk:
Hi, I was able to change the default base path api/auth to next/api/auth (the first option was occupied by my backend server). In my case I had to move nextauth config to app routing (route.ts).

NEXTAUTH_URL=http://localhost:3000/next/api/auth
[...]

PR #10571, opened 2024-04-13
fix(core): use NEXTAUTH_URL when getting the url during initialization

ammmze:
## ☕️ Reasoning

When using an oauth based authentication process and the app is configured with a `basePath`, we need to set the `NEXTAUTH_URL` environment variable to tell next-auth / auth.js the base authentication path. However, it does not seem to respect it in all scenarios. For example, I am running next-auth v4.24.7 with a basePath configured as `/sso` and a `NEXTAUTH_URL=http://localhost:3000/sso/api/auth`. When I perform a `signIn('keycloak')`, it does send me to my auth provider (keycloak) BUT the `redirect_uri` gets set to `http://localhost:3000/api/auth/callback` (note the missing `/sso` base path).

## 🧢 Checklist

- [ ] ~Documentation~
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

I _think_ this fixes the following issues (at least for users using v4...sounds like the v5 beta stuff has this issue too): 
* #10009
* #9984
* #8841

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/core/init.ts
+++ b/packages/next-auth/src/core/init.ts
@@ -42,7 +42,7 @@ export async function init({
   options: InternalOptions
   cookies: cookie.Cookie[]
 }> {
-  const url = parseUrl(origin)
+  const url = parseUrl(process.env.NEXTAUTH_URL ?? origin)
 
   const secret = createSecret({ authOptions, url })
 


ndom91:
Hey thanks for your interest! However, we'd like to avoid making changes like this to v4 atm. 

Does setting `NEXTAUTH_URL` to include only the basePath (i.e. `http://localhost:3000/sso`) work for yuo?

ammmze:
> Does setting `NEXTAUTH_URL` to include only the basePath (i.e. `http://localhost:3000/sso`) work for yuo?

No...that just makes things worse. Doing that makes it so that all the other calls that it makes (i.e. `/session`, `/providers`, etc) no longer have the `/api/auth` base path. For example, it tries to load `/sso/session`, but should be `/sso/api/auth/session`.

It is also explained the [documentation](https://next-auth.js.org/getting-started/client#custom-base-path) that if you are using a custom base path that you set this to the full path to the next auth endpoints.

ndom91:
@ammmze okay gotcha, yeah we changed the default behaviour of that in v5 but I couldn't remember how v4 behaved in detail. Okay let me have a think abuot this..

ammmze:
@ndom91 I can make the change a little more involved if we want by changing the behavior of [parse-url](https://github.com/nextauthjs/next-auth/blob/a83573ed2f696108ad2294a1892aa204cfaf7c71/packages/next-auth/src/utils/parse-url.ts#L15) so that it still uses the protocol and hostname of the given url (the origin) but then also parses the `NEXTAUTH_URL`. If the given url has a pathname **other** than `/`, then we use that as the default path instead of `/api/auth`. If the pathname of the given url has a pathname of `/`, then it would preserve the existing behavior.

The current proposed fix would always result in the `NEXTAUTH_URL` being used as the protocol and hostname. But the change I described above would preserve the origin information, while only taking the path from the `NEXTAUTH_URL` if it is something with a base path.

ammmze:
For clarity in the proposed alternative solution, i've implemented it on a separate branch found [here](https://github.com/nextauthjs/next-auth/commit/ebfd42d5cac179d9895ab9b017ea052138d66998). Effectively it just makes it so that the `parseUrl` function uses the `NEXTAUTH_URL` to get the default path instead of hard coding it to `/api/auth`

ammmze:
Oye...as I continue to dig at this I have made another discovery...

Previously I had assumed the request "origin" would never have a path*. However, apparently we [override the `origin` of the request when making it an "InternalRequest"](https://github.com/nextauthjs/next-auth/blob/bd1d91383ae96e1710cd38bdb55881b43e6ea3d3/packages/next-auth/src/core/index.ts#L83). And then while digging into that [detectOrigin](https://github.com/nextauthjs/next-auth/blob/bd1d91383ae96e1710cd38bdb55881b43e6ea3d3/packages/next-auth/src/utils/detect-origin.ts#L2-L9) method I discovered that having ANY value set for `AUTH_TRUST_HOST` will actually enable `AUTH_TRUST_HOST`. So setting it to `'false'` (string because its an env var) will enable that. And so then it falls into that condition which just uses the forwarded host from the request instead of the `NEXTAUTH_URL`.

So I see 2 issues here:

1. AUTH_TRUST_HOST doesn't allow you to set it to typical falsy value to disable it.
2. `detectOrigin` effectively does not respect the base path configured via setting `NEXTAUTH_URL` (and neither does `parseUrl`, which may be fixed with something like [this](https://github.com/nextauthjs/next-auth/commit/ebfd42d5cac179d9895ab9b017ea052138d66998).

But I guess the good news is I think I am now un-blocked from this issue by removing the `AUTH_TRUST_HOST` environment variable.

> [!NOTE]
> *`origin` typically just contains protocol, host, and sometimes the port. Example:
> ```
> > new URL('https://example.com/foo').origin
> 'https://example.com'
> ```

smultar:
I have a similar issue, that leads me to be constantly redirected back to my sign-in page or my local host.

My production environment is a docker swarm, with caddy infront of it. It works locally, but not in my prod.

TomWonder:
I have the very same issue. Cannot use either signIn() or signOut(). I need signOut() to implement the logout

PR #10582, opened 2024-04-14
fix(providers): update providers config from verification walk-throughs

ndom91:
## ☕️ Reasoning

- Update 3 providers based on findings from our verification of the example app providers

1. **Netlify** - Simply append `?scope` to the `authorizationUrl`
2. **Slack** - Set `checks: "nonce"`
3. **WorkOS** - Add `connection` argument / env var which gets appended to `authorziationUrl` as a search param

@balazsorban44 two question:
1. I saw in the types that `checks: ("none" | "state" | "pkce")`, are the types wrong or is `nonce` really an invalid value?
2. Regarding WorkOS, I wasn't sure how best to pass this unique value, `AUTH_WORKOS_CONNECTION`. Using `process.env` directly in the provider seems wrong since that won't work in all environments.. Do you have a better idea for this one? :thinking: 

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


balazsorban44 (approved):

docs/pages/getting-started/providers/workos.mdx
```diff
@@ -35,7 +35,7 @@ https://example.com/auth/callback/workos
 ```
 AUTH_WORKOS_ID
 AUTH_WORKOS_SECRET
-AUTH_WORKOS_ISSUER
+AUTH_WORKOS_CONNECTION
```
balazsorban44, line 38: Let's just ask the user to pass `connection` as a property.

packages/core/src/providers/workos.ts
```diff
@@ -144,15 +144,20 @@ export interface WorkOSProfile extends Record<string, any> {
  * :::
  */
 export default function WorkOS<P extends WorkOSProfile>(
-  options: OAuthUserConfig<P>
+  options: OAuthUserConfig<P> & { connection?: string }
 ): OAuthConfig<P> {
-  const { issuer = "https://api.workos.com/" } = options
+  const {
+    issuer = "https://api.workos.com/",
+    connection = process.env.AUTH_WORKOS_CONNECTION ?? "",
```
balazsorban44, line 151: we cannot use `process.env` in core as it's non-standard. let's just ask the user to pass `connection` for now.

balazsorban44 (changes requested):

balazsorban44:
Can you also modify the [`providers` list](https://github.com/nextauthjs/next-auth/blob/599bdf2e2dde1cee9b4ae53ca1d3a7996bb9cb09/apps/examples/nextjs/auth.ts#L34) accordingly?

packages/core/src/providers/workos.ts
ndom91: Yeah thats what i wrote in the description as well. Sounds good, i'll just change it to take a connection option then :+1:

ndom91: Also the other question I had in the description - for `checks` we had an enum including the option `none`, but not `nonce`. Is `nonce` correct? Should I update the `checks` enum?

balazsorban44 (approved):

PR #10616, opened 2024-04-17
chore(docs): update WorkOS docs 

MoritzKn:
## ☕️ Reasoning

The `domain` used in the authorization param used in the example is deprecated and no longer supported. If it is still used, WorkOS returns the error code `domain_connection_selector_not_allowed` with the description:

> 'The `domain` connection selector is deprecated and may no longer be used. Use the `connection` or `organization` connection selectors instead.'

See reference: https://workos.com/docs/reference/sso/get-authorization-url/error-codes#:~:text=domain_connection_selector_not_allowed

## 🧢 Checklist

- [x] Documentation
- [x] Tests (does not apply)
- [x] Ready to be merged

## 🎫 Affected issues

Couldn't find any open WorkOS related issues.

### DIFF
--- a/docs/docs/providers/workos.md
+++ b/docs/docs/providers/workos.md
@@ -35,69 +35,17 @@ providers: [
 
 WorkOS is not an identity provider itself, but, rather, a bridge to multiple single sign-on (SSO) providers. As a result, we need to make some additional changes to authenticate users using WorkOS.
 
-In order to sign a user in using WorkOS, we need to specify which WorkOS Connection to use. A common way to do this is to collect the user's email address and extract the domain.
+In order to sign a user in using WorkOS, we need to specify which WorkOS Connection to use. You should use the `organization` or `connection` `authorizationParams` to specifiy which connection to use:
 
-This can be done using a custom login page.
+```js
+import { signIn } from "next-auth/react"
 
-To add a custom login page, you can use the `pages` option:
+// [...]
 
-```javascript title="pages/api/auth/[...nextauth].js"
-...
-  pages: {
-    signIn: "/auth/signin",
-  }
+const organization = 'ORGANIZATION_ID';
+signIn(provider.id, undefined, {
+  organization,
+});
 ```
 
-We can then add a custom login page that displays an input where the user can enter their email address. We then extract the domain from the user's email address and pass it to the `authorizationParams` parameter on the `signIn` function:
-
-```jsx title="pages/auth/signin.js"
-import { useState } from "react"
-import { getProviders, signIn } from "next-auth/react"
-
-export default function SignIn({ providers }) {
-  const [email, setEmail] = useState("")
-
-  return (
-    <>
-      {Object.values(providers).map((provider) => {
-        if (provider.id === "workos") {
-          return (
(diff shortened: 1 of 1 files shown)


MoritzKn:
The docs on the main branch ([docs/pages/getting-started/providers/workos.mdx](https://github.com/nextauthjs/next-auth/blob/main/docs/pages/getting-started/providers/workos.mdx)) probably need to be updated too but I do not know what your process is for this:

ndom91:
@MoritzKn the `main` / v5 docs have been updated with WorkOS to point out that one needs to pass a `connectionId` (check out: https://authjs.dev/getting-started/providers/workos#environment-variables). 

Is that correct in yuor view?

ndom91:
@MoritzKn also why did you remove the rest of the example code?

MoritzKn:
@ndom91 Regarding your second comment:

The whole point of the example code is to create a login page that splits the domain name from the email address, and then use that "domain" so that WorkOS can tell which connection you want to use. If you simply specify the connection using the `organization` or` connection` id, this logic is not needed. 

Further, while working, the example was a little bit contrived. It's a bit strange to prompt your users to enter an email address and then only use the domain part of that email. It was simply necessary to work around the shortcoming in the WorkOS API. Just specifying the organization seems like the easiest to me.
Note: In most cases, organization and connection have a 1 to 1 relation in WorkOS. It seems that in some rare cases, an organization can have multiple connections.

Regarding your first comment:
In this version, the `signIn` function still uses the deprecated `domain` param. The necessary `connection` param gets mentioned but (to my knowledge) used incorrectly in the provider options. At last according to TypeScript you can not use `connection` as provider option like it is used here. But maybe this is because I'm on v4 and this is an v5 API..?

In my opinion, the entire "notes" section in the v5 docs is confusing and should be omitted. But maybe this is because in my use case I'm using WorkOS with just one backend providers, simply as a SAMl <> OAuth bridge.
To my understanding, for this section to make sense with the current API for multi provider user, it would need to include some logic to map from the email address to the `connection` ID (instead of simply using the domain).
Another option would be to just show multiple login buttons that all call the `signIn("workos", undefined, { connection: "con_123" })` function but with a different `connection` ID.
However, I think it's fine to leave it to the developers to figure out what solution makes sense to them.

docs/docs/providers/workos.md
```diff
@@ -35,69 +35,17 @@ providers: [
...
 
-In order to sign a user in using WorkOS, we need to specify which WorkOS Connection to use. A common way to do this is to collect the user's email address and extract the domain.
+In order to sign a user in using WorkOS, we need to specify which WorkOS Connection to use. You should use the `organization` or `connection` `authorizationParams` to specifiy which connection to use:
 
-This can be done using a custom login page.
+```js
+import { signIn } from "next-auth/react"
 
-To add a custom login page, you can use the `pages` option:
+// [...]
```
ndom91, line 43:
```suggestion
```

ndom91 (approved):
Okay yeah that makes a lot of sense. I've removed that part from the current (v5) docs as well since we support the `connection` param directly when using the provider there.

PR #10624, opened 2024-04-17
fix(next-auth): improve typing for using Dynamic Route Segments with auth()

### LINKED ISSUE #9344: How to implement Dynamic Route Segments with Auth Beta

I can't find how to implement Dynamic routes with the latest beta of authjs

with auth handlers it only accept reqeust as params

```
export const GET = auth(async (req) => {
  const session = req.auth;
  if (!session) {
    return Response.json({ error: 'Unauthorized' });
  }
});
```

With next handlers

`app/api/[slug]/route.ts`

```
export async function GET(
  request: Request,
  { params }: { params: { slug: string } }
) {
  const slug = params.slug // 'a', 'b', or 'c'
}
```

_No response_

ethernet8023:
## ☕️ Reasoning
all the plumbing to pass context is already in place,
but the typescript definition doesn't allow `context` to be passed.
this allows dynamic route segments to be used in auth-wrapped api routes without hacky typescript `as any` overrides.
    

this simply adds the typescript typings that reflect what's already being done under the hood.
## 🧢 Checklist

- [x] Documentation (behavior hasn't changed, this is already expected behavior, as seen by confusion in the above issue)
- [x] Tests (behavior hasn't changed)
- [x] Ready to be merged (in my opinion, yes. i'm already using this patch locally and it works great.)

## 🎫 Affected issues
Fixes: https://github.com/nextauthjs/next-auth/issues/9344

### DIFF
--- a/packages/next-auth/src/index.tsx
+++ b/packages/next-auth/src/index.tsx
@@ -79,7 +79,7 @@ import type {
   NextApiRequest,
   NextApiResponse,
 } from "next"
-import type { AppRouteHandlerFn } from "./lib/types.js"
+import type { AppRouteHandlerFn, AppRouteHandlerFnContext } from "./lib/types.js"
 import type { NextRequest } from "next/server"
 import type { NextAuthConfig, NextAuthRequest } from "./lib/index.js"
 export { AuthError, CredentialsSignin } from "@auth/core/errors"
@@ -233,7 +233,7 @@ export interface NextAuthResult {
     ((...args: []) => Promise<Session | null>) &
     ((...args: [GetServerSidePropsContext]) => Promise<Session | null>) &
     ((
-      ...args: [(req: NextAuthRequest) => ReturnType<AppRouteHandlerFn>]
+      ...args: [(req: NextAuthRequest, ctx: AppRouteHandlerFnContext) => ReturnType<AppRouteHandlerFn>]
     ) => AppRouteHandlerFn)
   /**
    * Sign in with a provider. If no provider is specified, the user will be redirected to the sign in page.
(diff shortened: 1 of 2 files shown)


packages/next-auth/src/index.tsx
```diff
@@ -233,7 +233,7 @@ export interface NextAuthResult {
     ((...args: []) => Promise<Session | null>) &
     ((...args: [GetServerSidePropsContext]) => Promise<Session | null>) &
     ((
-      ...args: [(req: NextAuthRequest) => ReturnType<AppRouteHandlerFn>]
+      ...args: [(req: NextAuthRequest, ctx: AppRouteHandlerFnContext) => ReturnType<AppRouteHandlerFn>]
```
ThangHuuVu, line 236:
```suggestion
      ...args: [(..._args: Parameters<AppRouteHandlerFn>) => ReturnType<AppRouteHandlerFn>]
```
Then no need to export the `AppRouteHandlerFnContext` type

ThangHuuVu (changes requested):
Thanks! I left a suggestion

ThangHuuVu:
Ref: Next.js doc link for [Dynamic Route Segments](https://nextjs.org/docs/app/building-your-application/routing/route-handlers#dynamic-route-segments)

packages/next-auth/src/index.tsx
arilotter: That won't work, because the AppRouteHandlerFn takes a `NextRequest`, not a `NextAuthRequest`, so the `.auth` param won't be available:
https://github.com/nextauthjs/next-auth/blob/5c3cc2e9facd0ace1da4aac64cacd8c726a837ee/packages/next-auth/src/lib/types.ts#L14

ThangHuuVu: I see, then this is fine, resolving.

ThangHuuVu (approved):

Issue #10276, opened 2024-03-11
NextAuth.js middleware does not behave as documented


goleary:
### Environment

  System:
    OS: macOS 14.2.1
    CPU: (10) arm64 Apple M2 Pro
    Memory: 139.33 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.17.1 - ~/.nvm/versions/node/v18.17.1/bin/node
    Yarn: 1.22.21 - ~/.nvm/versions/node/v18.17.1/bin/yarn
    npm: 9.6.7 - ~/.nvm/versions/node/v18.17.1/bin/npm
  Browsers:
    Chrome: 122.0.6261.112
    Edge: 122.0.2365.80
    Safari: 17.2.1
    Safari Technology Preview: 17.4
  npmPackages:
    @auth/prisma-adapter: ^1.4.0 => 1.5.0 
    next: ^14.1.0 => 14.1.3 
    next-auth: ^4.24.6 => 4.24.7 
    react: 18.2.0 => 18.2.0 

### Reproduction URL

https://github.com/goleary/next-auth-repro

### Describe the issue

Next.js middleware used [as documented](https://next-auth.js.org/configuration/nextjs#middleware) causes an infinite loop, redirecting to the login route even if the user has been logged in.

### How to reproduce

Pull down repro, provide required secrets. Trying to access any page causes a redirect to `http://localhost:3000/api/auth/signin` even after the user has successfully logged in.

### Expected behavior

The user should only be redirect if they are not logged in.

I obtained a solution from stack overflow that actually behaves the way I expect the default middleware should behave: https://stackoverflow.com/a/76749457/5186877

JasperAlexander:
Currently, the middleware only supports jwt sessions. When using an adapter, the default session strategy is database, so you need to set the session strategy to jwt in order to get the middleware working.

ndom91:
The stackoverflow answer you linked just checks if session token cookie is there or not, which might work in certain situations, but that session might be invalid, expired, etc. I wouldn't recommend relying on that. 

Also just in case anyone else comes across this, middleware does support database sessions as well as long as your database adapter can run in an edge-runtime (like the latest Prisma releases can, etc.) For more details see:

- [authjs.dev/getting-started/migrating-to-v5#edge-compatibility](https://authjs.dev/getting-started/migrating-to-v5#edge-compatibility)
[...]

Issue #8161, opened 2023-07-28
Redirects to signin page using Next.js v13.4.12, adapter and session strategy: database


std4453:
### Environment

  System:
    OS: Linux 6.4 Arch Linux
    CPU: (16) x64 12th Gen Intel(R) Core(TM) i5-12600KF
    Memory: 6.50 GB / 62.66 GB
    Container: Yes
    Shell: 3.6.1 - /usr/bin/fish
  Binaries:
    Node: 16.17.1 - ~/.local/share/nvm/v16.17.1/bin/node
    Yarn: 1.22.19 - ~/.local/share/nvm/v16.17.1/bin/yarn
    npm: 8.15.0 - ~/.local/share/nvm/v16.17.1/bin/npm
    pnpm: 6.7.1 - ~/.local/share/nvm/v16.17.1/bin/pnpm

### Reproduction URL

https://github.com/std4453/next-auth-bug/

### Describe the issue

`next-auth` does not work when:

- `next.js` v13.4.12 (other versions not tested)
- Using app router
- Using middlewares
- Using adapters
- Using `database` as session strategy

When entering `/`, it jumps to `/api/auth/signin`, you login, it jumps back to `/api/auth/callback/keycloak`, then to `/`, then to `/api/auth/signin` again.

As a result, you never get to use the app.

It works when using `jwt` as session strategy.

**I'm not sure whether this is a duplicate of #5008 , that one claims to be fixed in Next.js `v12.2.5-canary.2` or newer, yet there has since been more feedback about it not working, and it might be worth to take another look.**

### How to reproduce

See [`README` ](https://github.com/std4453/next-auth-bug/).

### Expected behavior

It should redirect to the OAuth provider, then back to the app, and you should be able to use it.

tmackness:
I am having the same issue as you. I did notice in your code you didn't have SessionProvider context though. Not sure if you missed that in the example only. 

Otherwise have you found a solution.

tmackness:
I too am having this issue. In addition I'm getting two callbackUrl query's attached.

tmackness:
> Same issue here. Have you found any workaroud?

I'm still trying to figure out the issue with limited time. Yet to try previous next Auth version. 

Does appear this has been an issue in the past too. 

To clarify I am using App directory and middleware. I think it works without the middleware from memory.

std4453:
[...]

PR #10667, opened 2024-04-21
fix(docs): add middleware prerequisite

JasperAlexander:
## ☕️ Reasoning

Currently, the Next.js middleware only supports the jwt session strategy. When using an adapter, the default session strategy is database. It is not well documented that you need to set the session strategy to jwt when using an adapter in order to use the middleware. I added this to the docs.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Fixes: https://github.com/nextauthjs/next-auth/issues/10276

## 📌 Resources

### DIFF
--- a/docs/docs/configuration/nextjs.md
+++ b/docs/docs/configuration/nextjs.md
@@ -131,6 +131,8 @@ You must set the same secret in the middleware that you use in NextAuth. The eas
 
 Alternatively, you can provide the secret using the [`secret`](#secret) option in the middleware config.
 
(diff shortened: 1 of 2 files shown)


ndom91:
@JasperAlexander unfortunately this is not the case, you can use the database session with middleware, as long as you use a database who is edge runtime compatible. 

Check out my reply to the other thread you posted in: https://github.com/nextauthjs/next-auth/issues/8161#issuecomment-2068112321

PR #10479, opened 2024-04-04
fix(drizzle): reimplement Drizzle adapter

realmikesolo:
##  Reasoning

The previous Drizzle adapter for NextAuth was implemented incorrectly. I have rewritten it from the ground up to fix the issues and align it with Drizzle's best practices. The adapter passed all tests. Documentation was updated according to new functionality.

## 🧢 Checklist

- [X] Documentation
- [X] Tests
- [ ] Ready to be merged

PR #10682, opened 2024-04-22
fix(drizzle): revert breaking changes and cleanup

realmikesolo:
## ☕️ Reasoning

Fix breaking changes, remove redundant functionality and update types for tables.

## 🧢 Checklist

- [X] Documentation
- [X] Tests
- [ ] Ready to be merged

This PR fixes issues raised by the [previous](https://github.com/nextauthjs/next-auth/pull/10479) one.


packages/adapter-drizzle/src/lib/mysql.ts
```diff
@@ -1,29 +1,29 @@
...
 import type {
   Adapter,
-  AdapterUser,
   AdapterAccount,
   AdapterSession,
+  AdapterUser,
   VerificationToken,
 } from "@auth/core/adapters"
 
 import { randomUUID } from "crypto"
```
juliusmarminge, line 24: i think you wanna use the global crypto module instead of importing to support as many runtimes as possible:

```ts
// instead of this
import { randomUUID } from "crypto";
const id = randomUUID()

// do this
const id = crypto.randomUUID()
```

packages/adapter-drizzle/src/lib/mysql.ts
```diff
@@ -130,14 +118,12 @@ export function MySqlDrizzleAdapter(
       userId: string
       expires: Date
     }) {
-      const id = randomUUID()
-
-      await client.insert(sessionsTable).values({ ...data, id })
```
juliusmarminge, line 135: this is still a breaking change (if we compare the last 0.x release) that now requires the user to have the `.$defaultFn()` or `.default()`. Is there some way to check if the table has a default? There should be some props on `sessionsTable._` I think

juliusmarminge: (commented on the rong table, meant to comment on the `createUser` call)

packages/adapter-drizzle/src/lib/pg.ts
```diff
@@ -239,18 +227,295 @@ export function PostgresDrizzleAdapter(
...
-export type PostgresTableFn<T extends TableConfig> = PgTableWithColumns<{
-  name: T["name"]
-  columns: T["columns"]
-  dialect: T["dialect"]
+export type DefaultPostgresUsersTable = PgTableWithColumns<{
+  name: string
+  columns: {
+    id: PgColumn<{
+      name: string
+      columnType: "PgVarchar" | "PgText"
```
lucasfeliciano, line 235: If I understood correctly, this is restricting the column type to these two types.

What if the user wants to use an `autoincrement` id for example?

packages/adapter-drizzle/src/lib/mysql.ts
realmikesolo: Now it is not obligatory to have `.$defaultFn()` or `.default()` if we speak about types. We can generate `id` in the code and pass it to insert values or just use `.$defaultFn()` or `.default()` regarding to `id` has default or not.

 `.$defaultFn(() => randomUUID())` is the same as

```ts
const id = randomUUID()

await client.insert(sessionsTable).values({ ...data, id })
```

juliusmarminge: it's not quite, consider someone doesn't have a default set on their column definition, then inserts would now start failing cause the lib isn't generating one like it used to do.

but if the lib is generating and sending to the insert call, then it doesn't matter if i have or haven't got a default on the table definition since it'll never be used since the param is always passed to the insert values.

so you want to generate an id in the lib code __if and only if__ the column definition doesn't have a `.default()` or `.$defaultFn()` set

juliusmarminge: something like

```ts
if (usersTable._.columns.id.hasDefault) {
  await db.insert(usersTable).values(data)
} else {
  const id = crypto.randomUUID();
  await users.insert({ ...data, id });
}
```

packages/adapter-drizzle/src/lib/pg.ts
realmikesolo: Honestly, I am not sure if it is should be allowed to specify `non-string` type.


@ndom91  Can you please specify what type the `id` might be in the `users` table? Unfortunately, the information about the models at this [link](https://authjs.dev/getting-started/adapters#models) is currently unavailable, but as far as I remember, the `id` should be a `string`. In the TypeScript interface `AdapterUser`, the `id` is also a `string`.

juliusmarminge: probably safest to only allow a single typescript datatype. could be a potential maintenance hell if the core lib should start allowing different data types if they reference user.id and can't expect it to always be of a certain datatype.

on the typesafety side, if you'd accept number here, you'd also want to restrict that `user.id` has the same type as `account.userId` etc etc and that can quickly escalate in complexity

ndom91: There's a more general page regarding DB adapters now available here: https://authjs.dev/concepts/database-models

I agree that `user.id: String` makes most sense :+1:

ndom91: Also more useful info here as well: https://authjs.dev/guides/creating-a-database-adapter

lucasfeliciano: You are right, this change was introduced on next-auth v4.
Before that we could use number there.

Sorry for the confusion.

packages/adapter-drizzle/src/lib/mysql.ts
rodrigonzalz: Currently can't use Turbopack because of this.

packages/adapter-drizzle/src/lib/mysql.ts
ndom91: So this is still an active breaking change, if I understand correctly, right?

realmikesolo: I have implemented this workaround:

```ts
      const hasDefaultId = getTableColumns(usersTable)["id"]["hasDefault"]
      
      await client
        .insert(usersTable)
        .values(hasDefaultId ? data : { ...data, id: crypto.randomUUID() })
```

packages/adapter-drizzle/src/lib/mysql.ts
ndom91: This is fixed in this PR

docs/pages/getting-started/adapters/drizzle.mdx
```diff
@@ -196,7 +198,9 @@ import type { AdapterAccount } from "next-auth/adapters"
 import { randomUUID } from "crypto"
 
 export const users = mysqlTable("user", {
-  id: varchar("id", { length: 255 }).primaryKey().$defaultFn(() => randomUUID()),
+  id: varchar("id", { length: 255 })
+    .primaryKey()
+    .$defaultFn(() => randomUUID()),
```
juliusmarminge, line 203: docs should show usage using global crypto too to prevent confusion regarding runtime compatibility

docs/pages/getting-started/adapters/drizzle.mdx
```diff
@@ -233,16 +237,22 @@ export const accounts = mysqlTable(
...
-  expires: timestamp("expires", { mode: "date" }).notNull(),
-}, (s) => ({
-  userIdIdx: index("Session_userId_index").on(s.userId),
-}))
+export const sessions = mysqlTable(
+  "session",
+  {
+    id: varchar("id", { length: 255 })
+      .primaryKey()
+      .$defaultFn(() => randomUUID()),
```
juliusmarminge, line 245: remove

docs/pages/getting-started/adapters/drizzle.mdx
```diff
@@ -233,16 +237,22 @@ export const accounts = mysqlTable(
...
+    sessionToken: varchar("sessionToken", { length: 255 }).notNull().unique(),
```
juliusmarminge, line 246: primary key

docs/pages/getting-started/adapters/drizzle.mdx
```diff
@@ -270,13 +280,16 @@ export {
...
 ```ts filename="schema.ts"
 import { integer, sqliteTable, text, primaryKey } from "drizzle-orm/sqlite-core"
 import type { AdapterAccount } from "next-auth/adapters"
 import { randomUUID } from "crypto"
 
 export const users = sqliteTable("user", {
-  id: text("id").primaryKey().$defaultFn(() => randomUUID()),
+  id: text("id")
+    .primaryKey()
+    .$defaultFn(() => randomUUID()),
```
juliusmarminge, line 292: same as above

docs/pages/getting-started/adapters/drizzle.mdx
```diff
@@ -308,16 +321,22 @@ export const accounts = sqliteTable(
...
-}, (s) => ({
-  userIdIdx: index("Session_userId_index").on(s.userId),
-}))
+export const sessions = sqliteTable(
+  "session",
+  {
+    id: text("id")
+      .primaryKey()
+      .$defaultFn(() => randomUUID()),
+    sessionToken: text("sessionToken").notNull().unique(),
```
juliusmarminge, line 330: same as above

packages/adapter-drizzle/src/index.ts
```diff
@@ -28,23 +28,23 @@ import { DefaultSchema, SqlFlavorOptions } from "./lib/utils.js"
 import type { Adapter } from "@auth/core/adapters"
 
 export {
-  postgresUsersTable,
+  mysqlAccountsTable,
+  mysqlSessionsTable,
+  mysqlUsersTable,
+  mysqlVerificationTokensTable,
```
juliusmarminge, line 34: why are these exported 🤔

packages/adapter-drizzle/src/lib/mysql.ts
```diff
@@ -100,15 +88,17 @@ export function MySqlDrizzleAdapter(
     schema
 
   return {
-    async createUser(data: Omit<AdapterUser, "id">) {
-      const id = randomUUID()
+    async createUser(data: AdapterUser) {
+      const hasDefaultId = getTableColumns(usersTable)["id"]["hasDefault"]
```
juliusmarminge, line 92: is the `getTableColumns` needed? Do Drizzle recommend not using the `_` attribute on the table directly?

packages/adapter-drizzle/src/lib/mysql.ts
```diff
@@ -100,15 +88,17 @@ export function MySqlDrizzleAdapter(
...
-      await client.insert(usersTable).values({ ...data, id })
+      await client
+        .insert(usersTable)
+        .values(hasDefaultId ? data : { ...data, id: crypto.randomUUID() })
 
       return client
         .select()
         .from(usersTable)
-        .where(eq(usersTable.id, id))
+        .where(eq(usersTable.email, data.email))
```
juliusmarminge, line 101: there are no docs recommending putting an index on this column, so this lookup might be expensive? sucks that we don't have `insert().returning()`. can you run the defaultFn manually?

```ts
const id = usersTable._.$defaultFn() ?? crypto.randomUUID();
await client
  .insert(usersTable)
  .values({ ...data, id: crypto.randomUUID() })
```

wouldn't work with default's set on the db level though :/

packages/adapter-drizzle/src/lib/mysql.ts
```diff
@@ -266,18 +254,193 @@ export function MySqlDrizzleAdapter(
...
+
+export type DefaultMySqlUsersTable = MySqlTableWithColumns<{
+  name: string
+  columns: {
+    id: DefaultMyqlColumn<{
+      data: string
+      dataType: "string"
+      notNull: true
+      columnType: "MySqlVarChar" | "MySqlText"
+    }>
```
juliusmarminge, line 284: nice type helper, easier to read than the original 👌

packages/adapter-drizzle/src/lib/mysql.ts
```diff
@@ -266,18 +254,193 @@ export function MySqlDrizzleAdapter(
...
+}>
+
+export type DefaultMySqlUsersTable = MySqlTableWithColumns<{
+  name: string
+  columns: {
+    id: DefaultMyqlColumn<{
+      data: string
+      dataType: "string"
+      notNull: true
+      columnType: "MySqlVarChar" | "MySqlText"
```
juliusmarminge, line 283: is it required specifying the exact database type here? this is fine for mysql but for example postgres got the UUID datatype that would also work. isn't it enough to specify the `dataType: "string"`?

packages/adapter-drizzle/src/lib/mysql.ts
```diff
@@ -266,18 +254,193 @@ export function MySqlDrizzleAdapter(
...
-  usersTable: MySqlTableFn<(typeof mysqlUsersTable)["_"]["config"]>
-  accountsTable: MySqlTableFn<(typeof mysqlAccountsTable)["_"]["config"]>
-  sessionsTable: MySqlTableFn<(typeof mysqlSessionsTable)["_"]["config"]>
-  verificationTokensTable: MySqlTableFn<
-    (typeof mysqlVerificationTokensTable)["_"]["config"]
-  >
+  usersTable: DefaultMySqlUsersTable
+  accountsTable: DefaultMySqlAccountsTable
+  sessionsTable: DefaultMySqlSessionsTable
+  verificationTokensTable: DefaultMySqlVerificationTokenTable
```
juliusmarminge, line 445: nit - but should the default columns have some guard that they adhere to these types to avoid drifting away from public interface?

```ts
export const defaultUsersTable = mysqlTable(...) satisfies DefaultMySqlUsersTable;
```

packages/adapter-drizzle/src/index.ts
realmikesolo: If users want to use default tables they can just import them instead of copying

docs/pages/getting-started/adapters/drizzle.mdx
realmikesolo: thanks, will update docs

packages/adapter-drizzle/src/index.ts
balazsorban44: let's not, for now. it wasn't there before. let's minimize the public api surface

packages/adapter-drizzle/src/lib/mysql.ts
realmikesolo: I have tested now and it works with `$defaultFn()` and `default()` (db level)

```ts
const defaultId = getTableColumns(usersTable)["id"]['default'] ?? getTableColumns(usersTable)["id"]['defaultFn']!()
```
but if user pass some function for generating id (e.g. `UUID()`)  then it won't work cause we can't get the result of this function before using in insert and in select statements. I also thought about selecting db default (e.g. `select (UUID())`) and assign it to the variable and then pass to all statements.

As for now, might be the best way to add information in docs about index on `email` field. What do you think about this?

juliusmarminge: Sounds good to me

packages/adapter-drizzle/src/lib/mysql.ts
realmikesolo: If we use `_` then we get `undefined`. That's why I chose to use `getTableColumns`

packages/adapter-drizzle/src/lib/mysql.ts
realmikesolo: I added uuid type for `columnType`. Unfortunately, `dataType` won't be enough because some `non-string` types with `mode: 'string'` can have the same `dataType` as `string` types.

packages/adapter-drizzle/src/lib/mysql.ts
realmikesolo: Added it  🚀

packages/adapter-drizzle/src/lib/pg.ts
```diff
@@ -239,18 +227,192 @@ export function PostgresDrizzleAdapter(
...
+  }
+  dialect: "pg"
+  schema: string | undefined
+}>
+
+export type DefaultPostgresAccountsTable = PgTableWithColumns<{
+  name: string
+  columns: {
+    userId: DefaultPostgresColumn<{
+      columnType: "PgVarchar" | "PgText"
```
juliusmarminge, line 291: Should allow UUID

packages/adapter-drizzle/src/lib/pg.ts
```diff
@@ -239,18 +227,192 @@ export function PostgresDrizzleAdapter(
...
+  name: string
+  columns: {
+    sessionToken: DefaultPostgresColumn<{
+      columnType: "PgVarchar" | "PgText"
+      data: string
+      notNull: true
+      dataType: "string"
+    }>
+    userId: DefaultPostgresColumn<{
+      columnType: "PgVarchar" | "PgText"
```
juliusmarminge, line 371: Should accept UUID

packages/adapter-drizzle/src/lib/pg.ts
realmikesolo: my bad, will fix it now. Thank you!

ndom91:
Checking it out now :eyes:

ndom91 (approved):
So theres some small docs related stuff I'd like to clean up, but generally, it looks good to me!

- All tests pass
- I setup a test app with `next-auth@5.0.0-beta.16` and installed the previous verison of the adapter (`0.9.0`), got a DB all setup with a user, and then upgraded with this PR, and there were no migrations required by drizzle and everything in the app still worked :crossed_fingers: 

I'd like to get this out in a `1.0.1` so folks can continue working with the drizzle adapter and then any enhancements / docs cleanup we can do afterward.

ndom91:
The package has been published under `@auth/drizzle-adapter@1.0.1` :pray:

PR #10719, opened 2024-04-24
fix(dgraph): id generation overlap

korzhanov:
## ☕️ Reasoning

Dgraph will generate the ID by itself. In Dgraph, every node has a unique 64-bit identifier that you can expose in GraphQL using the ID type. An ID is auto-generated, immutable and never reused. Each type can have at most one ID field.

The ID type works great when you need to use an identifier on nodes and don’t need to set that identifier externally 
[DGraph type docs ](https://dgraph.io/docs/graphql/schema/types/)

Since next-auth.js or auth.js creates a new user id, this causes an error and does not save the entry to dgraph

## 🧢 Checklist

- [X] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources


ndom91:
Just curious why the current tests pass and why this isn't necessary for `createSession` or `createVerificationToken`, for example, then? :thinking:

korzhanov:
@ndom91 Sorry, I didn't see your comment in this thread without mention.
> Just curious why the current tests pass and why this isn't necessary for `createSession` or `createVerificationToken`, for example, then? 🤔

The test uses a predefined id value ```id: () => "0x0a0a00a00"``` that falls under the DGraph requirements.
https://github.com/nextauthjs/next-auth/blob/770112c7a12c30f9a294bd8e0bbaef5274c4ad4f/packages/adapter-dgraph/test/index.test.ts#L22-L43

And in reality, the random UUID is invalid, it doesn't match the DGraph UID conditions.

Situation similar to the FaunaDB adapter
https://github.com/nextauthjs/next-auth/blob/770112c7a12c30f9a294bd8e0bbaef5274c4ad4f/packages/adapter-fauna/test/index.test.ts#L17-L26

And I think maybe `createSession` and `createVerificationToken` don't use external generated ids, only internal ones

adriangalilea:
I spent several hours trying to get the dgraph adapter to work on next-auth only to realize that: `DgraphClientError: unknown field` really meant it's the overlapping of ID's, and finally arrived here...

[I tried discussing on the forum but seems dead.](https://discuss.dgraph.io/t/next-auth-dgraph-dql-graphql-database-adapter/14760)

What's the status of this PR?

I've noticed Dgraph adapter and Dgraph adapter docs are outdated in several ways:

Dgraph now allows HS512 so [this is wrong now](https://authjs.dev/getting-started/adapters/dgraph#jwt-session-and-auth-directive) on the dgraph adapter docs.

Also I faced some edge runtime issues when trying to call `auth()` from the middleware.

Can we make this PR cover all graph adapter fixes so it's easier to track? @korzhanov

korzhanov:
@adriangalilea same story bro. 
I spent a very long time trying to figure out what the problem was. I tried the `@auth/dgraph-adapter`, but it didn't work at all. I had to install `@next-auth/dgraph-adapter`
It worked, but with errors.
I ended up spending a few hours to understand how the adapter works, fixed one line as described in this PR. And instead of a dependency in package.json, I copied the code separately and plugged it in as a component until the official package update comes out. Some time :)

May be @ndom91 help us

adriangalilea:
@korzhanov I created a issue with a repro repo now so it stays open and let's us track it, I also reached out to the Dgraph team both on the forum(which seems dead) and their support, who replied previously so hopefully they can chime in as I pointed out both the the Github issue and this very PR.

adriangalilea:
Hey @korzhanov I fixed:
id issues, in this PR you are only fixing user creation, but you may have issues with sessions and linking accounts. I linked the repo on the main issue. (code is just proof of concept)

edge runtime issues, which seems to be not related to dgraph.

The JWT algo encoding is outdated and should be ideally updated so that it uses the same that next-auth uses by default, so overall would be simpler, and docs can be trimmed down too, [this part should be removed](https://authjs.dev/getting-started/adapters/dgraph#jwt-session-and-auth-directive).

Check it out if you wish, I think we can fix it fully, I may be able to also sort the documentation too, would love someone from the Dgraph  team to chime in as I mentioned.

@korzhanov Do you want me to create a new PR so I cover all the mentioned issues, or do you prefer to update this one together?

korzhanov:
@adriangalilea You've done a great job, that's very cool! Your approach is very thorough, thank you! 
I will also try your solution to test on some project of mine in the coming weeks

PR #10733, opened 2024-04-26
feat(adapter-pg): define foreign keys, use single queries

kevinji:
## ☕️ Reasoning

- Foreign keys are added linking `userId` to the `id` in `users` with cascading delete. The delete user query can now be a single one due to cascading deletes.
- The `id` in `sessions` is deleted and the primary key replaced with `sessionToken`, which needs to be unique anyways. Turning it into the primary key should also speed up `WHERE sessionToken` queries.
- Queries that were previously a `SELECT` then `UPDATE` are now a single `UPDATE` to ensure atomicity. This required creating a rather ugly `createParameterizedUpdate` function to only join together `key = $2` if it exists in the passed-in object. I was unable to figure out the right incantation of TypeScript types to make `createParameterizedUpdate` type-safe but happy to accept any suggestions to fix this.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/adapter-pg/schema.sql
+++ b/packages/adapter-pg/schema.sql
@@ -1,4 +1,15 @@
-\set ON_ERROR_STOP true
+BEGIN TRANSACTION;
+
+CREATE TABLE users
+(
+  id SERIAL,
+  name VARCHAR(255),
+  email VARCHAR(255),
+  "emailVerified" TIMESTAMPTZ,
+  image TEXT,
+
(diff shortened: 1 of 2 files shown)


kevinji:
Closing due to lack of attention from reviewers and also that I've switched to using https://github.com/lucia-auth/lucia instead.

PR #10750, opened 2024-04-28
fix(drizzle): update schema type & fix issue with default id

realmikesolo:
Made tables for adapter optional cause not all of them are used for authentication strategies. Also I exlcuded `id` from data coming from Authjs in `createUser` function to enable setting default id as defined in Drizzle table.

## ☕️ Reasoning

1. Default id was overwritten by one coming from Authjs;
2. Not all tables are used for `next-auth`.

## 🧢 Checklist

- [X] Documentation
- [X] Tests
- [ ] Ready to be merged


srigi:
+1 on this, as I'm also not using INT or UUID for PKs in my tables and this would unblock the whole project moving forward.

packages/adapter-drizzle/src/lib/pg.ts
```diff
@@ -19,83 +19,100 @@ import type {
...
   return {
     async createUser(data: AdapterUser) {
+      const { id, ...insertData } = data
       const hasDefaultId = getTableColumns(usersTable)["id"]["hasDefault"]
 
       return client
         .insert(usersTable)
-        .values(hasDefaultId ? data : { ...data, id: crypto.randomUUID() })
+        .values(
+          hasDefaultId ? insertData : { ...insertData, id: crypto.randomUUID() }
```
juliusmarminge, line 114: could save some cycles and re-use the id coming from auth.js as the fallback?
```suggestion
          hasDefaultId ? insertData : { ...insertData, id }
```

ndom91: True, the `id` coming from auth.js will already be the result of a `crypto.randomUUID()` call. Can you do this update in all db types (postgres, mysql, sqlite)? :pray:

realmikesolo: yep, will update now

ndom91:
Okay nice, looks better to me. Tests all pass as well. @juliusmarminge you want to take one last look over it before I merge it? :pray:

packages/adapter-drizzle/src/index.ts
```diff
@@ -100,7 +99,7 @@ import type { Adapter } from "@auth/core/adapters"
  *   userId: text("userId")
  *     .notNull()
  *     .references(() => users.id, { onDelete: "cascade" }),
- *   type: text("type").notNull(),
+ *   type: text("type").$type<AdapterAccount["type"]>().notNull(),
```
juliusmarminge, line 102: can use the new `AdapterAccountType` introduced in https://github.com/nextauthjs/next-auth/pull/10832 to avoid circular refernce issues

juliusmarminge (commented):
Rest looks good to me :)

ndom91 (commented):
One last thing, lets also update the example(s) in the actual doc page at `/docs/pages/getting-started/adapters/drizzle.mdx`.

Ideally we wouldn't have two copies of the example setup code. One in the JSDoc in the `index.ts` **and** this docs page, but we just migrated to these new docs recently. We'll probably cleanup the schema stuff from the jsdoc shortly.

JipSterk:
Do we still want to use `AdapterAccount['type']` in the tests?

ndom91 (approved):

PR #10782, opened 2024-05-01
chore: fix casings

balazsorban44:
## ☕️ Reasoning

Continues on #10781

Using [`npx case-police --fix`](https://github.com/antfu/case-police), I fixed the rest of the casing issues


ndom91 (approved):
Nice!

PR #10797, opened 2024-05-02
fix(next-auth): Add support for Next.js basePath configuration

k3k8:
## ☕️ Reasoning

When I add basePath configuration to Next.JS config, the API routes are broken.
If both AUTH_URL and NextAuth basePath configurations are empty,  NextAuth will look like it works correctly. But, OAuth redirects callback URI is omitted basePath, so will fail.

In case of settings with AUTH_URL with full path: http://localhost:3000/basepath/api/auth, or adding NextAuth basePath configuration like /basepath/api/auth, NextAuth will raise Unknown Action Error.

I’ve investigated this error and figured out that it is caused by the Next.JS specification or bug. NextAuth parses the NextRequest URL and creates an action with the request URL. But, Next.JS with basePath configuration reports that the URL does not contain basePath. The simple Next.JS project here reproduces this issue. 
 https://github.com/k3k8/nextjs-base-path-issue

This issue has been reported in the Next.JS repository, but has not yet received any responses from the members. (https://github.com/vercel/next.js/issues/60956)
I suspect that the Next.JS app router should rewrite from an inbound URL to an internal one. Thus this is likely to be a specification, in my opinion.

Consequently, I’ve decided to fix this problem with next-auth components. The problem is caused by the Next.JS specification; my commits are only on the next-auth package. No commits are on core libraries.

I’ve created reqWithBasePathURL and estimateBasePath on packages/next-auth/src/lib/env.ts and modified reqWithEnvURL call point.

Configuration sample:
 Next.Js baesPaht: “/base_path”, NextAuth config.basePath: “/base_path/api/auth” 
The reqWithEnvURL called POST or GET request, and req: NextRequest contains like this:
http://localhost:3000/api/auth/session  The estimateBasePath function estimates from the request URL and config.BasePath.

If you set a Next.JS `basePath` such as `/base_path`, the NextAuth config	should be contain full path to auth dir. AUTH_URL: http://localhost:3000/base_path/api/auth, or basePath: /base_path/api/auth.
Configuring the full path to AUTH_URL is a v4 specification, and this correct configuration also works. AUTH_URL: http://locahost:3000/, basePath: /base_path/api/auth

These configuration rules are obeyed on document one (https://authjs.dev/reference/warnings#env-url-basepath-redundant).

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

#9274 
#9984 
#10009 

## 📌 Resources

### DIFF
--- a/docs/pages/getting-started/deployment.mdx
+++ b/docs/pages/getting-started/deployment.mdx
@@ -47,6 +47,29 @@ to `true`. This tells Auth.js to trust the `X-Forwarded-Host` header from the re
 
 This environment variable is mostly unnecessary with v5 as the host is inferred from the request headers. However, if you are using a different base path, you can set this environment variable as well. For example, `AUTH_URL=http://localhost:3000/web/auth` or `AUTH_URL=https://company.com/app1/auth`
 
+#### Using Next.js `basePath`
+
+If you have configured a `basePath` in your `next.config.js`, you must also set the `nextJsBasePath` option in your NextAuth.js configuration:
+
+```ts
+// next.config.js
+module.exports = { basePath: "/my-app" }
+
+// auth.ts
+export const { handlers, auth } = NextAuth({
+  basePath: "/my-app/api/auth",
+  nextJsBasePath: "/my-app", // Required when using Next.js basePath
+  // ...
+})
+```
+
+<Callout type="info">
+  The `nextJsBasePath` option is required because Next.js does not include the
+  basePath in the request URL when processing API routes. Without this
+  configuration, NextAuth.js will not be able to correctly handle authentication
+  requests.
+</Callout>
+
 ### `AUTH_REDIRECT_PROXY_URL`
 
 > **_NOTE:_** Some providers (eg Apple) do not support [redirect proxy](https://github.com/nextauthjs/next-auth/blob/3ec06842682a31e53fceabca701a362abda1e7dd/packages/core/src/lib/utils/providers.ts#L48) usage.
--- a/packages/next-auth/src/index.ts
+++ b/packages/next-auth/src/index.ts
@@ -68,7 +68,7 @@
  */
 
 import { Auth, customFetch } from "@auth/core"
-import { reqWithEnvURL, setEnvDefaults } from "./lib/env.js"
+import { reqWithBasePathURL, reqWithEnvURL, setEnvDefaults } from "./lib/env.js"
 import { initAuth } from "./lib/index.js"
 import { signIn, signOut, update } from "./lib/actions.js"
 
@@ -376,9 +376,10 @@ export default function NextAuth(
     | NextAuthConfig
     | ((request: NextRequest | undefined) => Awaitable<NextAuthConfig>)
 ): NextAuthResult {
+  let _config: NextAuthConfig
   if (typeof config === "function") {
     const httpHandler = async (req: NextRequest) => {
-      const _config = await config(req)
+      _config = await config(req)
       setEnvDefaults(_config)
       return Auth(reqWithEnvURL(req), _config)
     }
@@ -406,19 +407,21 @@ export default function NextAuth(
     }
   }
   setEnvDefaults(config)
-  const httpHandler = (req: NextRequest) => Auth(reqWithEnvURL(req), config)
+  _config = config
+  const httpHandler = (req: NextRequest) =>
+    Auth(reqWithBasePathURL(reqWithEnvURL(req), _config), _config)
   return {
     handlers: { GET: httpHandler, POST: httpHandler } as const,
     // @ts-expect-error
-    auth: initAuth(config),
+    auth: initAuth(_config),
     signIn: (provider, options, authorizationParams) => {
-      return signIn(provider, options, authorizationParams, config)
+      return signIn(provider, options, authorizationParams, _config)
     },
     signOut: (options) => {
-      return signOut(options, config)
+      return signOut(options, _config)
     },
     unstable_update: (data) => {
-      return update(data, config)
+      return update(data, _config)
     },
   }
 }
--- a/packages/next-auth/src/lib/env.ts
+++ b/packages/next-auth/src/lib/env.ts
@@ -3,13 +3,27 @@ import { NextRequest } from "next/server"
 import type { NextAuthConfig } from "./index.js"
 import { setEnvDefaults as coreSetEnvDefaults } from "@auth/core"
 
+/** Estimate the base path from the config and request pathname. */
+function estimateBasePath(configPathName: string, requestPathname: string): string {
+  const configSegments = configPathName.slice(1).split("/")
+  const requestSegments = requestPathname.slice(1).split("/")
+  if (configSegments[0] === requestSegments[0]) return ""
+
+  let i = 0
+  while (i < configSegments.length && i < requestSegments.length && configSegments[i] !== requestSegments[0]) {
+    i++
+  }
+  return "/" + configSegments.slice(0, i).join("/")
+}
+
 /** If `NEXTAUTH_URL` or `AUTH_URL` is defined, override the request's URL. */
 export function reqWithEnvURL(req: NextRequest): NextRequest {
   const url = process.env.AUTH_URL ?? process.env.NEXTAUTH_URL
   if (!url) return req
-  const { origin: envOrigin } = new URL(url)
-  const { href, origin } = req.nextUrl
-  return new NextRequest(href.replace(origin, envOrigin), req)
+  const { origin: envOrigin, pathname: envPathname } = new URL(url)
+  const { href, origin, pathname } = req.nextUrl
+  const basePath = estimateBasePath(envPathname, pathname)
+  return new NextRequest(href.replace(origin, envOrigin).replace(pathname, basePath + pathname), req)
 }
 
 /**
--- a/packages/next-auth/src/lib/env.ts
+++ b/packages/next-auth/src/lib/env.ts
@@ -4,10 +4,10 @@ import type { NextAuthConfig } from "./index.js"
 import { setEnvDefaults as coreSetEnvDefaults } from "@auth/core"
 
 /** Estimate the base path from the config and request pathname. */
-function estimateBasePath(configPathName: string, requestPathname: string): string {
+function estimateBasePath(configPathName: string, requestPathname: string): string | undefined {
   const configSegments = configPathName.slice(1).split("/")
   const requestSegments = requestPathname.slice(1).split("/")
-  if (configSegments[0] === requestSegments[0]) return ""
+  if (configSegments[0] === requestSegments[0]) return undefined
 
   let i = 0
   while (i < configSegments.length && i < requestSegments.length && configSegments[i] !== requestSegments[0]) {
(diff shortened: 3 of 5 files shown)


tunchunairarko:
Eagerly looking forward to see this PR into production, the basePath configuration is quite necessary.

forbesd7:
@ThangHuuVu Is there any way you can check / merge this PR? auth-js is currently unusable for me because I need to have my application behind a basePath.

PauloJSGG:
+1 on this, my app needs a basePath as well

atularora82:
even my app needs a basePath as well

Code-DJ:
When using SessionProvider, I was able to specify basePath in layout.tsx like so and get it to work. Had to provide hostname without which it won't work. Not sure of the implications, any thoughts on this?:

.env.local
```
BASE_PATH=/basepath
```

layout.tsx
```tsx
<SessionProvider basePath={process.env.BASE_PATH ? `http://localhost:3000${process.env.BASE_PATH}/api/auth` : 'http://localhost:3000/api/auth'}>
```

itsskofficial:
I have tried configuring a base path with NextAuth including putting the basePath in authConfig, in SessionProvider with the relative or the full URL, as the AUTH_URL. Sadly, none of them work, waiting for this fix to get merged

ejzeronimo:
Spent the past day trying all sorts of workarounds to get basePath to work then found this PR, really strange that there is no update on this.

PauloJSGG:
Yeah, I just hope the next-auth team takes a look at this. A lot of apps need a basePath. Our team actually tried different implementations without success. 
We are seriously thinking of removing next-auth all together due to this 🫤. We like this library, but a basePath is mandatory for us.

ThangHuuVu:
I understand the issue, thanks for putting together a thoughtful PR @k3k8 and everyone for flagging this. 
I actually ran into this issue a few months ago when I build the example to use next-auth deployed behind a proxy. At that time my workaround is basically appending the base path like this:
```ts
// app/api/auth/[...nextauth]/route.ts
import { handlers } from "auth";
import { NextRequest } from "next/server";

const reqWithBasePath = (req: NextRequest, basePath = "/authjs") => {
  const baseUrl = new URL(
    (req.headers.get("x-forwarded-proto") ?? "https") +
      "://" +
      (req.headers.get("x-forwarded-host") ?? req.nextUrl.host)
  );
  const urlWithBasePath = new URL(
    `${basePath}${req.nextUrl.pathname}${req.nextUrl.search}`,
    baseUrl
  );
  return new NextRequest(urlWithBasePath.toString(), req);
};

export const GET = (req: NextRequest) => {
  const nextReq = reqWithBasePath(req);
  return handlers.GET(nextReq);
};

export const POST = (req: NextRequest) => {
  const nextReq = reqWithBasePath(req);
  return handlers.POST(nextReq);
};
```
https://github.com/ThangHuuVu/next-auth-behind-proxy/blob/cde4368b46fbda57157d62eaf1b93b258f5fadde/app/api/auth/%5B...nextauth%5D/route.ts

I'm a bit hesitant to merge this PR, though. I would actually prefer an official next.js API to get the `basePath` value  (this isn't exist yet AFAIK) instead of calling the `estimateBasePath` function. The reason is that functions like `estimateBasePath` is quite complex and could easily introduce bugs, we need at least some tests to make sure it won't break any existing behaviors.

k3k8:
I understand the complexity of the estimateBasePath function. I created it while testing with the following test cases. These tests were created based on packages/core/test/url-parsing.test.ts.
Adding these test cases would ensure that future changes don't break existing behavior.

```
function estimateBasePath(configPathName: string, requestPathname: string): string | undefined {
  const configSegments = configPathName.slice(1).split("/")
  const requestSegments = requestPathname.slice(1).split("/")
  if (configSegments[0] === requestSegments[0]) return undefined

  let i = 0
  while (i < configSegments.length && i < requestSegments.length && configSegments[i] !== requestSegments[0]) {
    i++
  }
  return "/" + configSegments.slice(0, i).join("/")
}

const testCases = [
  { configPathName: '/base_path/api/auth', requestPathname: '/api/auth/session', expected: '/base_path' },
  { configPathName: '/api/auth', requestPathname: '/api/auth/session', expected: '' },
  { configPathName: '/test_path/api/auth', requestPathname: '/api/auth/signin/google', expected: '/test_path' },
  { configPathName: '/api/auth', requestPathname: '/api/auth/signin/google', expected: '' },
  { configPathName: '/test_path/test2/api/auth', requestPathname: '/api/auth/signout', expected: '/test_path/test2' },
  { configPathName: '/auth', requestPathname: '/auth/signin', expected: '' },
  { configPathName: '/sso/auth', requestPathname: '/auth/signin', expected: '/sso' },
];

testCases.forEach(({ configPathName, requestPathname, expected }) => {
  const result = estimateBasePath(configPathName, requestPathname);
  console.log(`Config: ${configPathName}, Request: ${requestPathname}, Expected: ${expected}, Result: ${result}`);
  console.assert(result === expected, 'Test case failed');
});
```

The solution of adding the base path in route.ts, as @ThangHuuVu's suggested, works well in my environment too and seems like a good approach.

Given the numerous comments on this issue, I'd like to implement one of the following:

1. Add information about the base path issue and @ThangHuuVu's code to the documentation
2. Add these test cases to the next-auth package to test estimateBasePath

I'm planning to commit to either of these options.

I'd appreciate your thoughts on these proposals. I want to contribute in the way that best suits the project.

jbogedahl-st:
Any update on getting this fix in? I'm hitting it as well... setting a nextjs base path breaks the oauth redirect_uri, or if you set the nextjs base path and set a authjs base path to match, then you get Unknown Action (because the pathname doesn't match the base path).  I can't find any way around this...

k3k8:
Can you please try the solution from ThangHuuVu's most recent reply?
If you want to resolve the issue sooner, this would be the most effective approach.

Alternatively, you can test my code using the following method:

```bash
git clone -b fix-base-path https://github.com/k3k8/next-auth.git
cd next-auth
corepack enable pnpm
pnpm install
pnpm build
cd packages/next-auth/
pnpm pack
```

This will generate a package file named something like `next-auth-5.0.0-beta.20.tgz`.

Move this file to your project directory and install it by updating your `package.json`:

```json
{
  "dependencies": {
    "next-auth": "./next-auth-5.0.0-beta.20.tgz"
  }
}
```

Then run `pnpm install` to update your dependencies.

andreynovikov:
Overwriting `NextRequest` is a **bad** idea as it breaks other non-auth middleware. The major feature of `basePath` in Next.js is that it is stripped deep in internals and is not used it routing. But with this change it appears in `pathname` and breaks routing in middleware.

calcavara:
> Can you please try the solution from ThangHuuVu's most recent reply? If you want to resolve the issue sooner, this would be the most effective approach.
> 
> Alternatively, you can test my code using the following method:
> 
> ```shell
> git clone -b fix-base-path https://github.com/k3k8/next-auth.git
> cd next-auth
> corepack enable pnpm
> pnpm install
> pnpm build
> cd packages/next-auth/
> pnpm pack
> ```
> 
> This will generate a package file named something like `next-auth-5.0.0-beta.20.tgz`.
> 
> Move this file to your project directory and install it by updating your `package.json`:
> 
> ```json
> {
>   "dependencies": {
>     "next-auth": "./next-auth-5.0.0-beta.20.tgz"
>   }
> }
> ```
> 
> Then run `pnpm install` to update your dependencies.

I'm having the same issue reported and worked here... What is ThangHuuVu's most recent reply? I'm sorry I can't find it. Thanks!

k3k8:
I apologize for making the content difficult to read due to my Git operations.
This one, replied on Jul 27

> I understand the issue, thanks for putting together a thoughtful PR @k3k8 and everyone for flagging this.
> I actually ran into this issue a few months ago when I build the example to use next-auth deployed behind a proxy. At that time my workaround is basically appending the base path like this:
> 
> // app/api/auth/[...nextauth]/route.ts
> import { handlers } from "auth";
> import { NextRequest } from "next/server";
> 
> const reqWithBasePath = (req: NextRequest, basePath = "/authjs") => {
>   const baseUrl = new URL(
>     (req.headers.get("x-forwarded-proto") ?? "https") +
>       "://" +
>       (req.headers.get("x-forwarded-host") ?? req.nextUrl.host)
>   );
>   const urlWithBasePath = new URL(
>     `${basePath}${req.nextUrl.pathname}${req.nextUrl.search}`,
>     baseUrl
>   );
>   return new NextRequest(urlWithBasePath.toString(), req);
> };
> 
> export const GET = (req: NextRequest) => {
>   const nextReq = reqWithBasePath(req);
>   return handlers.GET(nextReq);
> };
> 
> export const POST = (req: NextRequest) => {
>   const nextReq = reqWithBasePath(req);
>   return handlers.POST(nextReq);
> };
> https://github.com/ThangHuuVu/next-auth-behind-proxy/blob/cde4368b46fbda57157d62eaf1b93b258f5fadde/app/api/auth/%5B...nextauth%5D/route.ts
> 
> I'm a bit hesitant to merge this PR, though. I would actually prefer an official next.js API to get the basePath value (this isn't exist yet AFAIK) instead of calling the estimateBasePath function. The reason is that functions like estimateBasePath is quite complex and could easily introduce bugs, we need at least some tests to make sure it won't break any existing behaviors.

mdmahendri:
so this is why my next-auth does not work after adding `basePath`, thanks guys

uchar:
Still same problem with authjs 5 beta 25 (also auth.js 4)
What this library is very simple Redirect , take some shits back and save them in cookie or somewhere else , I had written this stuffs for my projects from scratch before and now feel like idiot removing them and using authjs instead couple of months ago.
Library that supposed to do authentication with third paraties cant do a simple redirect??? , there is only one variable here "BASEPATH" how hard it is to just use it ,  I dont understand why thousand of people need to waste hours for such a basic functionality

tunchunairarko:
> Still same problem with authjs 5 beta 25 (also auth.js 4)
> 
> What this library is very simple Redirect to a provider , take some shits back and save them in cookie or somewhere else , I had written this stuffs for my projects from scratch before and now feel like idiot removing them and using authjs instead couple of months ago.
> 
> Library that supposed to do authentication with third paraties cant do a simple redirect??? , there is only one variable here "BASEPATH" how hard it is to just use it ,  I dont understand why thousand of people need to waste hours for such a basic functionality

---
I think next-auth is permanently unusable as we go along if this issue is not resolved. I think the root cause lies in how next-auth was designed. I wasted 3 work weeks before I gave up and started using iron-session.

SrayasPMaatrum:
Hi All i found some work around  for this and working for me but this is only the work around , so basically  the basepath in next is working as expected for all the main issue is implementing in next-auth it is not taking the base path as how we declaring  so i have added these to methods in next config and it is working but appending duplicate basepath again in the auth api's thats the only draw back but code

this ins next.config.js
async rewrites() {
    return [
      {
        source: '/custom-base/api/:path*',
        destination: '/api/:path*',
      },
    ];
  },
  async redirects() {
    return [
      {
        source: '/api/:path*',
        destination: '/custom-base/api/:path*',
        permanent: true,
      },
    ];
  },

this  where you have use the session provider
 <><SessionProvider basePath="/custom-base/api/auth">

and add either basepath option in auth.ts or as env variable as NEXTAUTH_URL

then the api's will work as this  and we'll get the output http://localhost:3000/custom-base/custom-base/api/auth/session this is what i mentioned the base path will be duplicated in this but this is working

simone81:
> Hi All i found some work around for this and working for me but this is only the work around , so basically the basepath in next is working as expected for all the main issue is implementing in next-auth it is not taking the base path as how we declaring so i have added these to methods in next config and it is working but appending duplicate basepath again in the auth api's thats the only draw back but code
> 
> this ins next.config.js async rewrites() { return [ { source: '/custom-base/api/:path*', destination: '/api/:path*', }, ]; }, async redirects() { return [ { source: '/api/:path*', destination: '/custom-base/api/:path*', permanent: true, }, ]; },
> 
> this where you have use the session provider <>
> 
> and add either basepath option in auth.ts or as env variable as NEXTAUTH_URL
> 
> then the api's will work as this and we'll get the output http://localhost:3000/custom-base/custom-base/api/auth/session this is what i mentioned the base path will be duplicated in this but this is working

I'm tring your solution, but no luck till now.
Do you have a sample code somewhere?

I'm working on this issue since last week. It seems so horrible to fix this problem.

Let's recap all possible variables that could be engaged in the solution:

- next.config >>  basePath: '/[mycustompath]'
- next.config >> rewrites + redirects   (same as your solution)
- base path in auth.ts or alternatively in NEXTAUTH_URL env:
    basePath: 'http://localhost/[mycustompath]/api/auth'
- basePath in <SessionProvider basePath='??????????'>


Is ther a possibility to fix them all in some way?

uchar:
@simone81 Take a look at https://github.com/nextauthjs/next-auth/discussions/12160 , I had this problem couple of months ago and managed to solve it by that solution

k3k8:
@ThangHuuVu

I've refactored the PR to use an explicit configuration approach, removing the estimation logic entirely.

- Explicit Config: Added nextJsBasePath option to NextAuthConfig.
- Robust Logic: reqWithBasePathURL now uses the URL API and includes a safety check to prevent double-appending.
- Tests: Added comprehensive tests for the new logic.

This removes the "guesswork" and ensures deterministic behavior without breaking existing setups.

PR #10826, opened 2024-05-05
feat(providers): Add Riot Games RSO Provider

ArkShocer:
## ☕️ Reasoning

Adding Riot Games RSO Provider. Tested it with the current stable next-auth version `4.24.7` on nextjs `14.2.3`

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources


docs/pages/getting-started/providers/rso.mdx
```diff
@@ -0,0 +1,141 @@
...
+RSO_CLIENT_ID=<copy Application (client) ID here>
+RSO_CLIENT_SECRET=<copy generated client secret value here>
+```
+
+<Callout type="warning">
+  Riot Sign On currently does not allow local development so you need to have a
+  public domain to test it or create a proxy in your environment.
+</Callout>
+
+### Signed-JWT Method
```
ndom91, line 120: In this section, maybe we just take that callout warning out of the callout, i.e. just put that as regular text under this heading. And then remove the superfluous example at the bottom. It's not adding anything new that the above examples aren't, right? 

Also it uses the pages router path and doesn't use the env var inference, so it could be confusing to new users

packages/core/src/providers/rso.ts
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * If you think you found a bug in the default configuration, you can [open an issue](https://authjs.dev/new/provider-issue).
+ *
+ * Auth.js strictly adheres to the specification and it cannot take responsibility for any deviation from
+ * the spec by the provider. You can open an issue, but if the problem is non-compliance with the spec,
+ * we might not pursue a resolution. You can ask for more help in [Discussions](https://authjs.dev/new/github-discussions).
+ *
+ * :::
+ */
+export default function RSO<P extends RSOProfile>(
```
ndom91, line 77: Because this seems to be an openid-connect compatible provider, this can be significantly simplified. Can you give this version a shot and let me know if it still works for you? Also, please double check the `profile` return values. We need them to be `id`, `name`, `email`, `image`. Since that's what all next-auth users will have in their database via the default schemas, for example. We can't just return any fields we want here (by default).


```ts
export default function RSO<P extends RSOProfile>(
  options: OAuthUserConfig<P>
): OAuthConfig<P> {
  return {
    id: "rso",
    name: "Riot Sign On",
    type: "oidc",
    issuer: "https://auth.riotgames.com",
    authorization: {
      params: {
        scope: "openid offline_access access_tokens cpid",
      },
    },
    profile(profile) {
      return {
        id: profile.sub,
        name: profile.sub,
        email: profile.cpid,
        image: profile.jiti,
      }
    },
    style: {
      brandColor: "#f05b5e"
    },
    options: options,
  }
}
```

docs/pages/getting-started/providers/rso.mdx
ArkShocer: True that, moved/removed it now.

PR #10861, opened 2024-05-09
fix(core): update WebAuthn `authenticator` schemas and types

ndom91:
## ☕️ Reasoning

- Loosen up `authenticator.transports` type to also accept `null`
	- Removes the need for each adapter to have an extra method to just coerce `transports` to `undefined` if they're missing
	- This then led to the following cleanup in the pre-existing webauthn implementations in unstorage and prisma adapters
- Fix `prisma` adapter
	- Cleanup authenticator method implementations
	- Remove unnecessary `transports` coercion methods
	- Cleanup recommended db schemas in docs
- Fix `unstorage` adapter
	- `listAuthenticatorsByUserId` wasn't fully implemented, this is fixed here
	- All other authenticator-realted methods are cleaned up as well
	- Cleanup authenticator method implementations
	- Clean up multiple types
	- Enable almost all tests again

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


packages/adapter-prisma/src/index.ts
```diff
@@ -310,45 +309,47 @@ export function PrismaAdapter(
...
     },
     async updateAuthenticatorCounter(credentialID, counter) {
       return p.authenticator
         .update({
           where: { credentialID: credentialID },
           data: { counter },
         })
-        .then(fromDBAuthenticator)
+        .then((authenticator) => {
+          const { id, ...rest } = authenticator
```
juliusmarminge, line 348: Q: why does the id have to be omitted in the first place? Is this being shipped to the client so it's omitted for security reasons?

packages/adapter-prisma/src/index.ts
```diff
@@ -310,45 +309,47 @@ export function PrismaAdapter(
...
+          return {
+            ...rest,
+          }
```
juliusmarminge, line 351:
```suggestion
         return rest
```

packages/adapter-prisma/src/index.ts
ndom91: Good question, wasn't involved in the writing of this part, but I assume it's just because its internal info that isn't required externally. 

That `id` is just the Prisma / whatever DB primary key in the table.

Afaik there's no lookup by that ID that a user could execute or anythign like that.

The security here comes from the webauthn private key that the user has and the public key that auth.js / this table contain. Not anything related to this `id`, right.

Sooo long story short, _I think_ we can probably also drop this part and return the `id`. What do you think?

juliusmarminge: Ehm if it's not used, why is it required in the db schema? Can we not omit it and have a different PK similar to account model?

ndom91: So theoretically yeah. The author probably went with a separate `id` column just because that's the common first thing to do.

I like the idea of iterating here and removing the unnecessary column, I'm just worried about breaking changes for folks potentially already running this. Although that being said, webauthn support isn't officially even announced and labelled as experimental on top of an already "beta" `5.x` release, so I guess it wouldn't be that big of a deal

ndom91: Okay so I updated the `authenticator` schema to have the `credentialId + userId` as the primary, and updated docs/types/tests.

I'll double check it tomorrow, but should be okay. 

Migration-wise it actually shouldn't be tough for folks that may have already used Passkeys cause we weren't using the `id` field at all.

PR #10874, opened 2024-05-10
Add keycloak as default example provider.

falcowinkler:
## ☕️ Reasoning

Since we now have our own "dummy" provider, people can get started with the example project much easier. 
If we configure this dummy provider by default, you just have to run the app and use keycloak to log in. Then also the client-example will work right out of the box.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/apps/examples/nextjs/.env.local.example
+++ b/apps/examples/nextjs/.env.local.example
@@ -1,8 +1,9 @@
 AUTH_SECRET= # `npx auth secret` or `openssl rand -hex 32`
 
 # Hosted example provider. Credentials: bob/bob
+# This is for demonstration purposes. Never commit plain secrets to git.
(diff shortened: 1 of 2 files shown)


ndom91 (commented):
I'd prefer we don't add this. 

The user doesn't have access to the IdP, so they can't add users or anything anyway, so they'll just end up deleting it rather quickly. If we just want to show them a demo, we have the hosted example apps. Plus this way we're sharing the Keycloak secret (which we already are "leaking" through this PR :sweat_smile:). I get its not that important and they can't do much with it, but yeah still not great imo

falcowinkler:
hey, 
yes sorry, i forgot to close this 🙏 we will document the credentials differently

PR #10880, opened 2024-05-11
chore: version mismatch in sveltekit dev app

ThangHuuVu:
The mismatch version of vite causes Sveltekit dep app to throw 500 error when using form actions.


ndom91 (approved):
Fixes it for me too!

PR #10901, opened 2024-05-14
chore(examples): use correct http code in express config

uhidontkno:
## ☕️ Reasoning

The wrong HTTP response code is being used.
Instead of HTTP 400 (Bad Request) use HTTP 401 (Unauthorized).
<br><br>
*The other things in the pull request template have been removed as they are not applicable. It is a minor fix.*


ndom91 (approved):
Good catch, thanks!

PR #10907, opened 2024-05-14
fix(next-auth): signOut and middleware cookie error

jbpin:
## ☕️ Reasoning
Using `signOut()` in middleware cause the NextJS error: `Error: Cookies can only be modified in a Server Action or Route Handler. Read more: [https://nextjs.org/docs/app/api-reference/functions/cookies#cookiessetname-value-options`](https://nextjs.org/docs/app/api-reference/functions/cookies#cookiessetname-value-options%60)

We can make a redirect and settings cookie in a middleware using Next but it should be assign to the NextResponse object as 

```
const response = NextResponse.redirect(url)
response.cookies.set("myCookie", myJwt);
return response
```
This PR allow the signOut method to return the response object without setting the cookie using the next/headers cookie method. So doing, passing `redirect: false` in the signOut options, give the user the reponsibility/ability to set the cookie in the response of the redirect directly. 

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues
N/A

## 📌 Resources


SergioDep:
I have the same error when using unstable_update() from a server action inside a try catch statement, do you know if this solution could be applied to that?
Thanks 🙏

karolinakuzniewicz:
In general, the `NextAuth(config)` and its methods should be a Server Action. This would allow us to manage the cookies as they can only be modified in Server Action or Route Handler and this would also finally cover the Rotation of the Refresh Token fully on the server side for devs who would like to store auth tokens in HTTP cookies.

PR #10943, opened 2024-05-16
fix(core): add error `code` to `SignInResponse`

kaioduarte:
## ☕️ Reasoning

Makes `code` property from custom error available on the client side. See: 
- https://github.com/nextauthjs/next-auth/pull/9871#issuecomment-2003457866

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


jhoanborges:
+1

kaioduarte:
@ndom91 @ThangHuuVu can we have your review on this one?

ThangHuuVu (approved):

ThangHuuVu:
thanks @kaioduarte , merging this & release 🔜

dml0031:
Another update in the beta 19 may be conflicting with this change. After updating to `5.0.0-beta.19` the only error that gets returned from the `authorize` method is `error?error=Configuration` 

Tested this by using the exact same code in beta 18 and then beta 19, on 18 it will return `CredentialsSignIn` as error type and the return url from the credentials endpoint contains both `error` and `code` params and the `code` param just doesnt get parsed by the `next-auth/react` package and returned to the client.

In 19, the return url from credentials endpoint doesnt contain the `code` param and the only thing it has is `?error=Configuration` but the object returned from the `signIn` method does contain the code, its just always `null` no matter what error gets thrown in `authorize`

guimmarques:
> Another update in the beta 19 may be conflicting with this change. After updating to `5.0.0-beta.19` the only error that gets returned from the `authorize` method is `error?error=Configuration`
> 
> Tested this by using the exact same code in beta 18 and then beta 19, on 18 it will return `CredentialsSignIn` as error type and the return url from the credentials endpoint contains both `error` and `code` params and the `code` param just doesnt get parsed by the `next-auth/react` package and returned to the client.
> 
> In 19, the return url from credentials endpoint doesnt contain the `code` param and the only thing it has is `?error=Configuration` but the object returned from the `signIn` method does contain the code, its just always `null` no matter what error gets thrown in `authorize`
> 
> Edit: More Clarification
> 
> Example error: Screenshot 2024-06-02 at 12 05 41 PM
> 
> Beta 18 Return URL: `http://localhost:3000/login?error=CredentialsSignin&code=custom` Beta 19 Return URL: `http://localhost:3000/login?error=Configuration`

same issue here

dwalker93:
any updates on this issue?

kaioduarte:
It seems #11050 caused the regression

cc @ndom91

PR #10950, opened 2024-05-18
fix(next-auth): proper handling of env for createActionURL

michaellopez:
## ☕️ Reasoning

Next.js replaces environment variables at build time but they need to be reference directly and cannot be accessed dynamically for this to work. https://github.com/nextauthjs/next-auth/commit/3c035ec62f2f21d7cab65504ba83fb1a9a13be01 broke this contract.

This PR fixes this by changing the next-auth caller to extract the variables directly from `process.env` and putting them in an object compatible with `createActionURL` from core. Other frameworks might have other inner workings in regards to this env handling, so I only made the changes in next-auth rather than refactoring `createActionURL` which would've affected other frameworks I'm not familiar with.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

N/A

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/lib/index.ts
+++ b/packages/next-auth/src/lib/index.ts
@@ -60,12 +60,17 @@ export interface NextAuthConfig extends Omit<AuthConfig, "raw"> {
 }
 
 async function getSession(headers: Headers, config: NextAuthConfig) {
+  // Allows Next.js to replace the env variables with their values at build time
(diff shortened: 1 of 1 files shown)


michaellopez:
I still think this is relevant. I will rebase.

PR #11010, opened 2024-05-25
fix(next-auth): redirecting to `/undefined` when signIn method throws an error

### LINKED ISSUE #11008: [v5] Throwing error with custom sign in page leads to undefined route
### Environment

```
  System:
    OS: macOS 14.4.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 102.29 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.1.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.7.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 125.0.6422.77
    Safari: 17.4.1
  npmPackages:
    @auth/prisma-adapter: ^2.1.0 => 2.1.0
    next: 14.2.3 => 14.2.3
    next-auth: ^5.0.0-beta.18 => 5.0.0-beta.18
    react: ^18 => 18.3.1
```

### Reproduction URL

https://github.com/peguerosdc/nextauth-bug-error-undefined

### Describe the issue

When logging in via Magic Links using Resend, a custom `sendVerificationRequest` (following [this official guide](https://authjs.dev/getting-started/providers/resend#email-body)) and NOT using a custom sign in page, if I throw an error inside `sendVerificationRequest` like this:

```typescript
export const { handlers, signIn, signOut, auth } = NextAuth({
  debug: true,
  adapter: PrismaAdapter(prisma),
  providers: [
    Resend({
      from: "dummy@domain.com",
      sendVerificationRequest: async (params) => {
        throw new Error("Error");
      }
    })
  ],
})
```

I am redirected to a default Error page with a configuration error in the search params: `/api/auth/error?error=Configuration`.

But when I use a custom login page like this:

```typescript
export const { handlers, signIn, signOut, auth } = NextAuth({
  /* same settings as above */,
  pages: {
    signIn: "/auth/signin",
  },
})
```

And I throw the error, I am redirected to `/auth/undefined/`.

Moreover, when other kind of errors are thrown, I am redirected to my `/auth/signin?error=ErrorType` page WITH the `error` search param indicating what happened.

### How to reproduce

1. Clone and install the repository in the reproduction URL
2. Start the server
3. Click on the "Sign in" button
4. Enter an email and click on "Sign in"
5. See how you are redirected to `/auth/undefined/`.
6. Optionally, you can go to `/auth.ts`, remove the custom pages configuration and repeat steps 3-5 to see how you are redirected to `/api/auth/error?error=Configuration`

### Expected behavior

I would expect:

1. Consistency in how we should handle errors, whether we are using custom pages or not
2. To be able to gracefully handle errors when using a custom sign in page, instead of being redirected to `/auth/undefined/` (being the `/auth/signin?error=ErrorType` approach my preferred one as I don't have to create a whole new page just to show an error and a button that would redirect me again to the `sign in` page)

andresgutgon:
## ☕️ Reasoning
When using a custom signin page in a NextJS app and using a server action as explained in the docs:
https://authjs.dev/guides/pages/signin

When the auth fails. For example `error=ConfigurationError` it doesn't redirect to error page. It redirects to `/undefined`

### The problem
I think that `@auth/core`  `Auth` function can return an `InternalResponse` or a standard web `Response`. Internal response has a `redirect` attribute. The problem is that the logic inside `Auth` method when there is an error returns a Standard web Response that doesn't have `redirect` attribute but we can get a redirecting url by using `headers.get('Location')`.

## What changes are being made? What feature/bug is being fixed here?
We check if the returned response is a next auth `InternalResponse` or a standard web `Response`. This way we always get redirecting URL.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues
Fixes: https://github.com/nextauthjs/next-auth/issues/11008


packages/next-auth/src/lib/actions.ts
```diff
@@ -2,11 +2,9 @@ import { Auth, raw, skipCSRFCheck, createActionURL } from "@auth/core"
 import { headers as nextHeaders, cookies } from "next/headers"
 import { redirect } from "next/navigation"
 
-import type { AuthAction } from "@auth/core/types"
 import type { NextAuthConfig } from "./index.js"
 import type { NextAuthResult, Session } from "../index.js"
 import type { ProviderType } from "@auth/core/providers"
-import type { headers } from "next/headers"
```
andresgutgon, line 9: The linter was telling me these 2 imports are not used so I removed it. I hope is ok

ThangHuuVu:
thanks @andresgutgon , could you help to add a test for this so that we don't regress? the change LGTM otherwise.

andresgutgon:
Hi @ThangHuuVu where are tests? All I see is this one https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/test/index.test.ts

andresgutgon:
Ok I made 2 test for `signIn`. One successful and the other when mail provider fail which was my case. In that case I make sure that the code I changed works by checking redirect URL is there.

On caveat I have the way I did the tests is that in test logs debug error are shown

image

Tests pass but Vitest complain of unhandle error

image


I think is fine unless you can suggest a better way of testing it

andresgutgon:
What did I do to get mad at the bot? 😂

docs/pages/reference/_meta.js
```diff
@@ -41,4 +41,5 @@ export default {
   "unstorage-adapter": "@auth/unstorage-adapter",
   "upstash-redis-adapter": "@auth/upstash-redis-adapter",
   "xata-adapter": "@auth/xata-adapter",
+  "test-adapter": "@auth/test-adapter",
```
andresgutgon, line 44: In order to really test `signIn` method I made a dummy test adapter 
image

packages/next-auth/src/lib/actions.ts
```diff
@@ -73,8 +71,12 @@ export async function signIn(
 
   for (const c of res?.cookies ?? []) cookies().set(c.name, c.value, c.options)
 
-  if (shouldRedirect) return redirect(res.redirect!)
-  return res.redirect as any
+  const responseUrl =
+    res instanceof Response ? res.headers.get("Location") : res.redirect
```
andresgutgon, line 75: This is the fix

docs/pages/guides/pages/signin.mdx
```diff
@@ -107,7 +108,20 @@ export default async function SignInPage() {
...
+              // not existing, or the user not having the correct role.
+              // In some cases, you may want to redirect to a custom error
+              if (error instanceof AuthError) {
+                return redirect(`${SIGNIN_ERROR_URL}?error=${error.type}`)
+              }
+
+              // Otherwise if a redirects happens NextJS can handle it
+              // so you can just re-thrown the error and let NextJS handle it.
+              throw error
+            }
```
andresgutgon, line 124: One point of confusion to me that have work not much with NextJS was to find out that in next redirects throw an error. And `signIn` can fail. Also next auth can fail for legic reasons and is the user the responsible of managing those exceptions in their code

I tried to make this clear in the docs:
image

andresgutgon:
@ThangHuuVu I think I did tests and docs. I can see docs in my machine working and the tests I did are passing too. 

Thanks for this software and glad to help!

docs/pages/reference/_meta.js
andresgutgon: This can be used for other frameworks like Svelte or SolidJS to easy the testing

ThangHuuVu: great attempt, although I think no need for a new package, you could move this file: `packages/core/test/memory-adapter.ts` under `util` package & use it in `next-auth`, that should be enough

docs/pages/guides/pages/signin.mdx
ThangHuuVu: Could you link a next.js doc here about redirects throwing?

packages/next-auth/src/lib/actions.test.ts
```diff
@@ -0,0 +1,112 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest"
```
ThangHuuVu, line 1: I think we can move the tests in the `test` directory, on the same level with `src`, it's conventional for us

ThangHuuVu (changes requested):

ThangHuuVu:
hi @andresgutgon got some comments about the new changes, thanks again! for this comment:
> great attempt, although I think no need for a new package, you could move this file: packages/core/test/memory-adapter.ts under util package & use it in next-auth, that should be enough

I think you can copy and paste to util first, and I'll do clean up in core later, no need to move the file to avoid changing too many files

packages/next-auth/src/lib/actions.test.ts
andresgutgon: Ok I'll do that. Sorry, this is my way of placing tests collocated with their files so when you remove the file you remove the test. Anyway I'll move to tests 👍

packages/next-auth/src/lib/actions.ts
```diff
@@ -73,8 +71,11 @@ export async function signIn(
...
+
+  if (shouldRedirect) return redirect(responseUrl!)
```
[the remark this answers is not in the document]
andresgutgon: Agree but what should happen when `undefined`?

ThangHuuVu:
> Why do you think having a test adapter is not the best approach? To understand. Happy to revert and move to utils but would love to know your reasoning

well to me as long as it's implementing the Adapter interface, its the same thing right. And putting it in util means all frameworks can reuse, so reusability is not an issue. A new package is more overhead in terms of repo maintaining, if that makes sense 🙏 thanks for your understanding

andresgutgon:
>  A new package is more overhead in terms of repo maintaining, if that makes sense 🙏 thanks for your understanding

Fair point changing it : )

packages/next-auth/src/lib/actions.ts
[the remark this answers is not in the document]
andresgutgon: I mean, what should happen here? What should dev using next-auth expect when this is `undefined`.

To be clear. It always is possible in the future for this to be `undefined`. But as far as my understanding of the code with this fix is not possible.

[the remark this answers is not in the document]
ThangHuuVu: So we landed here because an error is thrown somewhere in `AuthInternal` - in this case its the `sendVerificationToken`. I think it'll be logical to redirect to the `signin` page. It's the same behavior of the signin action in core when missing the provider:
https://github.com/nextauthjs/next-auth/blob/770112c7a12c30f9a294bd8e0bbaef5274c4ad4f/packages/core/src/lib/actions/signin/index.ts#L18

I think we can do `return redirect(url.toString())`

docs/pages/reference/_meta.js
andresgutgon: I have a problem `core/test` is not exposed from `@auth/core`. How I can import. Indeed it was already a memory adapter 😂 

https://github.com/nextauthjs/next-auth/blob/main/packages/core/test/memory-adapter.ts

andresgutgon: This works but we're breaking the boundaries between packages. For that I though it would be cleaner to have a dedicated package that everyone can use. 

Anyway if this is fine I'll move on
image

docs/pages/guides/pages/signin.mdx
andresgutgon: Done

andresgutgon:
@ThangHuuVu I think I review all the comments. Let me know anything I need to change

packages/next-auth/test/actions.test.ts
```diff
@@ -0,0 +1,131 @@
...
+      const signIn = actionModule.signIn
+      let redirectTo: string | undefined | null
+      redirectTo = await signIn(
+        "nodemailer",
+        { ...options, redirect: false },
+        authorizationParams,
+        config
+      )
+      expect(redirectTo).toEqual(
+        "http://localhost/api/auth/signin/nodemailer?"
```
andresgutgon, line 127: @ThangHuuVu I made a test for that case. Is this what you meant?

andresgutgon:
@ThangHuuVu please let me know if there's anything else I can do to move forward with these changes?

ThangHuuVu:
thanks @andresgutgon , sorry I was absent for the week. I pushed a commit to simplify the tests. 
IMO, we don't have to mock the Auth response with an empty `location` headers - the core tests should already ensure this case isn't happening. I'll merge this once the CI passes 🤞

ThangHuuVu (approved):

PR #11020, opened 2024-05-27
docs[drizzle]: replace "next-auth/adapters" to "@auth/core/adapters"

guissalustiano:
## ☕️ Reasoning
I'm installing `authjs` with Svelte, so I would like to not install the `next-auth`.
[The `next-auth` only re-export it from core,](https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/adapters.ts) so this work for both.

## 🧢 Checklist

- [X] Ready to be merged

## 📌 Resources


ndom91:
Have you tried importing from `@auth/sveltekit/adapters`? 

The `next-auth` part is just an example. It shouldn't be necessary to install next-aurh or @auth/core in your case

guissalustiano:
Oh, that makes sense, thanks!

PR #11053, opened 2024-06-01
chore(docs) fix custom sign in redirect

N8-I:
## ☕️ Reasoning

I spent a lot of time figuring out how to get a custom login page to redirect to the calling route after authentication. Thought it might help others (and future me) to have it live in the guide. Im no expert so open to suggestions if I missed anything.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources


docs/pages/guides/pages/signin.mdx
```diff
@@ -100,14 +100,18 @@ We can now build our own custom sign in page.
...
+    }
+}) {
   return (
     <div className="flex flex-col gap-2">
       {Object.values(providerMap).map((provider) => (
         <form
           action={async () => {
             "use server"
-            await signIn(provider.id)
+            await signIn(provider.id, , { redirectTo: props.searchParams?.callbackUrl ?? "" })
```
ndom91, line 114: Two things:

1. Little typo here:

```suggestion
            await signIn(provider.id, { redirectTo: props.searchParams?.callbackUrl ?? "" })
```

2. While I get pain here, the `redirectTo` option is detailed in the API docs (See: https://authjs.dev/reference/nextjs#signin-2), and I think a plain beginner-level "signin" guide isn't the right place to surface these additional details.

N8-I: Hey @ndom91! Good catch on the comma 🤦‍♂️.

You are right the API documentation does do a good job of explaining the parameter but it is not immediately clear from the guide or the API document how it should be used in the context of the custom login page. 

I ended up finding what I think is the default [signin page](https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/lib/pages/signin.tsx) to try and figure out how it is handling the redirect. It looks to be submit it through the form and letting the provider page handle the redirect.

As a beginner myself I would have found it helpful to have it on the guide as a login page that does not redirect would not pass many tests.

I will concede that I know a lot more about how it works now but I am 3 hours in and 12 hours older lol.

Issue #10852, opened 2024-05-08
Custom Provider broken by V5


Nik-Novak:
### Environment

```
System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (16) x64 AMD Ryzen 9 5950X 16-Core Processor
    Memory: 5.12 GB / 15.75 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 20.12.0 - ~/.nvm/versions/node/v20.12.0/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v20.12.0/bin/yarn
    npm: 10.5.0 - ~/.nvm/versions/node/v20.12.0/bin/npm
  npmPackages:
    @auth/prisma-adapter: ^1.5.1 => 1.5.1 
    next: ^14.2.3 => 14.2.3 
    next-auth: ^5.0.0-beta.17 => 5.0.0-beta.17 
    react: ^18.3.1 => 18.3.1
```

### Reproduction URL

https://github.com/Nik-Novak/Mind-Knight

### Describe the issue

I use a custom provider (https://github.com/Nekonyx/next-auth-steam) to enable steam sign in using steam's OpenId 2.0 support (not OpenID connect)

After requiring an upgrade to V5, the provider no longer works with the following:

```
[auth][error] InvalidEndpoints: Provider "steam" is missing both `issuer` and `token` endpoint config. At least one of them is required.. Read more at https://errors.authjs.dev#invalidendpoints
    at assertConfig (webpack-internal:///(rsc)/./node_modules/next-auth/node_modules/@auth/core/lib/utils/assert.js:92:24)
    at Auth (webpack-internal:///(rsc)/./node_modules/next-auth/node_modules/@auth/core/index.js:88:95)
```

I've rewritten the provider to match any new formats but to no avail:

```
import { randomUUID } from "crypto"
import { NextApiRequest } from "next"
import { AUTHORIZATION_URL, EMAIL_DOMAIN, LOGO_URL, PROVIDER_ID, PROVIDER_NAME, SteamProfile } from "next-auth-steam"
import { SteamProviderOptions } from "next-auth-steam/lib/steam"
import { OAuthConfig, OAuthUserConfig } from "next-auth/providers"
import { NextRequest } from "next/server"
import { RelyingParty } from 'openid'
import { TokenSet } from 'openid-client'

export function Steam(
[...]

Issue #10732, opened 2024-04-26
Provider configs `token` and `userinfo` ignore `url` and `request` properties


joonhyungshin:
### Environment

```
  System:
    OS: macOS 14.4.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 2.24 GB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.12.2 - /usr/local/bin/node
    npm: 10.5.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 124.0.6367.92
    Safari: 17.4.1
  npmPackages:
    next: 14.2.3 => 14.2.3
    next-auth: ^5.0.0-beta.17 => 5.0.0-beta.17
    react: ^18 => 18.3.0
```

### Reproduction URL

https://github.com/joonhyungshin/next-auth-mre

### Describe the issue

I am working on a personal project with social account. I did not want Auth.js to call `userinfo/` endpoint, because in my case token verification and user info fetch are done in a separate backend server. So I only wanted Auth.js to receive an access token using the standard OAuth2 flow, so in my root `auth.ts` file I replaced `userinfo.request` with a no-op function.
```ts
export const { handlers, signIn, signOut, auth } = NextAuth({
  providers: [
    Twitter({
      userinfo: { request: () => {} }
    })
  ],
  // callbacks and events omitted
})
```
However, I realized that Auth.js still calls the `userinfo/` endpoint. It seems like the `url` and `request` properties are all ignored, since the following config worked with no error.
```ts
export const { handlers, signIn, signOut, auth } = NextAuth({
  providers: [
    Twitter({
      userinfo: { url: "asdf", request: () => {} }
    })
  ],
  // callbacks and events omitted
})
```
On the other hand, the following code errors as expected.
```ts
export const { handlers, signIn, signOut, auth } = NextAuth({
  providers: [
    Twitter({
      userinfo: "asdf",
    })
  ],
  // callbacks and events omitted
})
```
So I suspect that Auth.js just falls back to the default config if `userinfo` is not a string. The [code](https://github.com/nextauthjs/next-auth/blob/2159de2660ba33a3b1dafefa1a3b17d657d05b2a/packages/core/src/lib/utils/providers.ts#L132) doesn't seem to do so, so I don't understand why.

I also noticed that the `token` property has the same issue.

### How to reproduce

1. Start a new Next.js app with `npx create-next-app@latest`.
[...]

PR #11057, opened 2024-06-01
Re-enable ability for the token request to be overriden completely

PaulWild:
## ☕️ Reasoning

When dealing with a custom oauth provider that has a non compliant token endpoint this functionality allows this library to be used. This functionality was present in v4 of the library. 

In v5 this has been removed. A conform endpoint has been added for internal use to manipulate the response from token endpoints. However this does help if the request does not conform exactly to the standard. 

As an example the company I work for requires some extra parameters to be sent in the token request (including the state parameter) Other than that the solution is oauth compliant. It would be a shame if we couldn't use v5 because of this restraint.

Another example I can see in the v5 branch is the tiktok provider which I assume is currently broken by its requirement to have a custom token request. 

This PR implements the existing type contract that was carried over from the v4 branch but not implemented.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

There are quite a few issues and comments about not being able to control the request endpoint

https://github.com/nextauthjs/next-auth/issues/10852
https://github.com/nextauthjs/next-auth/issues/10732
https://github.com/nextauthjs/next-auth/discussions/10829

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/actions/callback/oauth/callback.ts
+++ b/packages/core/src/lib/actions/callback/oauth/callback.ts
@@ -6,8 +6,8 @@ import {
 } from "../../../../errors.js"
 
 import type {
-  Account,
   InternalOptions,
+  InternalProvider,
   LoggerInstance,
   Profile,
   RequestInternal,
@@ -104,6 +104,67 @@ export async function handleOAuth(
   if (!options.isOnRedirectProxy && provider.redirectProxyUrl) {
     redirect_uri = provider.redirectProxyUrl
   }
+
+  let tokens: TokenSet
+  if (provider.token?.request) {
+    const result = await provider.token.request({
+      params: {
+        state,
+        redirect_uri,
+        ...query,
+      },
+      checks,
+      provider,
+    })
+    if (!result) {
+      throw new Error("Custom token request did not return a valid tokenset")
+    }
+    tokens = result.tokens
+  } else {
+    tokens = await getTokens(
+      as,
+      client,
+      codeGrantParams,
+      redirect_uri,
+      codeVerifier,
+      provider,
+      cookies,
+      resCookies,
+      options
+    )
+  }
+
+  const profile =
+    provider.type === "oidc"
+      ? o.getValidatedIdTokenClaims(tokens)
+      : await getOauthProfile(provider, as, client, tokens)
+
+  if (tokens.expires_in) {
+    tokens.expires_at =
+      Math.floor(Date.now() / 1000) + Number(tokens.expires_in)
+  }
+
+  const profileResult = await getUserAndAccount(
+    profile,
+    provider,
+    tokens,
+    logger
+  )
+
+  return { ...profileResult, profile, cookies: resCookies }
+}
+
+async function getTokens(
+  as: o.AuthorizationServer,
+  client: o.Client,
+  codeGrantParams: URLSearchParams,
+  redirect_uri: string,
+  codeVerifier: string | undefined,
+  provider: InternalProvider<"oauth" | "oidc">,
+  cookies: Partial<Record<string, string>> | undefined,
+  resCookies: Cookie[],
+  options: InternalOptions<"oauth" | "oidc">
+) {
   let codeGrantResponse = await o.authorizationCodeGrantRequest(
     as,
     client,
@@ -137,9 +198,6 @@ export async function handleOAuth(
     throw new Error("TODO: Handle www-authenticate challenges as needed")
   }
 
-  let profile: Profile = {}
-  let tokens: TokenSet & Pick<Account, "expires_at">
-
   if (provider.type === "oidc") {
     const nonce = await checks.nonce.use(cookies, resCookies, options)
     const result = await o.processAuthorizationCodeOpenIDResponse(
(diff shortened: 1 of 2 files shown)


balazsorban44 (commented):
We might re-introduce something here, but it won't be the old format, it's too loose. Rather than patching up non-spec providers, we want to push them for correct implementation as much as possible.

We already have a [`conform` API](https://github.com/nextauthjs/next-auth/blob/38c5ee47b70f81d792fdde567599c7ad039a8975/packages/core/src/lib/actions/callback/oauth/callback.ts#L128) to correct the code grant responses, I want to go with a similar API for other cases as well.

I want to keep the TS type `@internal` to avoid relying on patching non-spec implementations too much.

In my experience, if we make it too easy to opt-out of conformance, we will keep maintaining legacy/broken providers.

PaulWild:
Thanks for the reply, I completely understand the rational and from the discussions being had in the issues linked it is not unexpected.

From a selfish perspective I think my works provider wont change any time soon so if there is room for some customization of the code grant request that would be fantastic.  I am more than happy to contribute towards a more acceptable solution if you have one in mind? Perhaps the ability to supply extra parameters to the request (with access to properties like state)?

milo-:
@balazsorban44 We really need some way to override this token method. 

We work with multiple vendors who are not going to change their non token APIs so we can use the latest version of Auth.js. This was not an issue with v4, but we are unable to use v5 with them as as we cannot override the token requests.

cmayoracurzio:
Adding to this discussion a use case that requires overriding the default `token.request`, and for which the provider is **within** the OAuth 2.0 specification:
- Wordpress OAuth requires sending the client credentials (`client_id` and `client_secret`) [in the request-body](https://developer.wordpress.com/docs/oauth2/#3-server-code-authentication).
- [RFC 6749, section 2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1) mentions that this is allowed (even if not recommended).
- However, the default `token.request` behavior only uses the Basic Authorization header scheme.

Note: The V5 `next-auth` WordpressProvider would anyways ignore a custom `token.request` because the merging approach for default and custom options is still in-progress. Although this is a different issue, I mention it here in case someone decides to test Wordpress OAuth using this built-in provider (use a custom one instead).

philwolstenholme:
I'm in the same boat, we have a provider managed by a separate team at work and the way things are set up there means we need to modify the token request. The separate team manage a whole hunch of different things and so are busy and unlikely to prioritise changing the provider just because of us. 

We'd like to use the latest NextAuth but can't without having this extra bit of control.

dpashk-figma:
+1. A number of big providers don't comply with the spec, and there's not much you can do to make them comply. Breaking compatibility with them only hurts developers using your library.

DewinU:
There's ways to use the new conform function for the providers that do not comply with the spec, like tiktok for example, but it requires to make another  or more requests after  NextAuth tries its owns, here is a code reference
```ts
let tkCode: string | undefined;
let tkCallback: string | undefined;
export const { handlers, auth, signIn, signOut } = NextAuth(req =>{
  if (req?.method === 'GET' && req?.headers.get('referer')?.includes('tiktok')) {
    const url = new URL(req?.url);
    tkCode = url.searchParams.get('code') as string;
    tkCallback = url.pathname
  }
  return {
    theme: { logo: "https://authjs.dev/img/logo-sm.png" },
    adapter: UnstorageAdapter(storage),
    providers: [
      GitHub,
      TikTok({
        clientId: process.env.AUTH_TIKTOK_KEY,
        clientSecret: process.env.AUTH_TIKTOK_SECRET,
        token: {
          async conform(response: Response) {
            const res = await fetch(response.url, {
              method: 'POST',
              headers: {
                'Cache-Control': 'no-cache',
                'Content-Type': 'application/x-www-form-urlencoded',
              },
              body: new URLSearchParams({
                client_key: process.env.AUTH_TIKTOK_KEY!,
                client_secret: process.env.AUTH_TIKTOK_SECRET!,
                code: tkCode!,
                grant_type: 'authorization_code',
                redirect_uri: process.env.AUTH_URL! + tkCallback ,
              }),
            });
            return res;
          },
        }
      })
    ],
    basePath: "/auth",
    callbacks: {
      authorized({ request, auth }) {
        const { pathname } = request.nextUrl
        if (pathname === "/middleware-example") return !!auth
        return true
      },
      jwt({ token, trigger, session, account }) {
        if (trigger === "update") token.name = session.user.name
        if (account?.provider === "keycloak") {
          return { ...token, accessToken: account.access_token }
        }
        return token
      },
      async session({ session, token }) {
        if (token?.accessToken) {
          session.accessToken = token.accessToken
        }
        return session
      },
    },
    experimental: {
      enableWebAuthn: true,
    },
    debug: process.env.NODE_ENV !== "production" ? true : false,
  } satisfies NextAuthConfig
})
```

PaulWild:
I closed this as the response was that this wasn't the approach that would be taken for this issue, which is totally understandable. 

However there is enough discussion going on that I am going to reopen it, so it gets a bit more visibility to encourage said discussion .

emulienfou:
Released a temporary package to add back this feature with latest release of package `@auth/core:0.35.1`:

https://www.npmjs.com/package/@emulienfou/auth-core

> Thanks to @PaulWild

DewinU:
Seems we have a solution, haven't tested it out yet tho

#11978

balazsorban44:
Superseded by https://github.com/nextauthjs/next-auth/pull/11975

This adds the ability to override any of the requests now. Use it sparingly!

Example usage, to conform the Microsoft Entra ID provider that returns the wrong `issuer`: https://github.com/nextauthjs/next-auth/pull/11980

PR #11061, opened 2024-06-02
chore(docs): improve edge compatibility

### LINKED ISSUE #11001: Split configuration is not working
### Environment

```
  System:
    OS: Linux 6.1 Debian GNU/Linux 12 (bookworm) 12 (bookworm)
    CPU: (12) x64 AMD Ryzen 5 5600H with Radeon Graphics
    Memory: 8.71 GB / 22.88 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 20.13.1 - /home/linuxbrew/.linuxbrew/opt/node@20/bin/node
    npm: 10.5.2 - /home/linuxbrew/.linuxbrew/opt/node@20/bin/npm
  Browsers:
    Brave Browser: 125.1.66.113
    Chromium: 125.0.6422.76
  npmPackages:
    @auth/drizzle-adapter: ^1.1.0 => 1.1.0 
    next: 14.2.3 => 14.2.3 
    next-auth: ^5.0.0-beta.18 => 5.0.0-beta.18 
    react: ^18.3.1 => 18.3.1 

```

### Reproduction URL

https://github.com/Chenalejandro/nextauth-bug

### Describe the issue

The provided documentation (https://authjs.dev/guides/edge-compatibility) for Split Config doesn't seem to work, I'm getting the error:
```
[auth][error] JWTSessionError: Read more at https://errors.authjs.dev#jwtsessionerror
[auth][cause]: JWEInvalid: Invalid Compact JWE
```
The error described in [https://errors.authjs.dev#jwtsessionerror](https://errors.authjs.dev/#jwtsessionerror) shouldn't be possible (It says that it is related to jwt strategy) since I am using the database strategy (having an adapter in auth.ts automatically uses that stategy).

Another side note: I'm using an edge compatible driver (@vercel/postgres) and deploying to vercel (edge middleware), so in theory I am able to access the database in the middleware.
The problem I faced was too much latency (each page takes more than 1200ms to load). This is because the middleware always runs in the edge, which is a problem when the database is in another region (for example, if the edge middleware is running in brasil and the database is in us-east). It only works when the edge middleware is being run in the same region. As soon as I remove the next-auth middleware, the page only takes 200ms to load.

Vercel does not allows you to change the edge middleware region, it always default to the region nearest to the user.

### How to reproduce

1. clone the repository and npm install
2. populate these env variables
```
AUTH_SECRET=putYourSecretHere
AUTH_GITHUB_ID=putYourSecretHere
AUTH_GITHUB_SECRET=putYourSecretHere
POSTGRES_URL=putYourSecretHere
```
3. npm run dev
4. Go to localhost:3000, click sign in and use github to sign in.
5. after the sign in is successful, you can see your username. But checking the server console logs, you can see the error. If you refresh the page, you can check that you are indeed signed out.

### Expected behavior

There should not be any console error and the user should not be signed out.

ThangHuuVu:
fixes #11001

### DIFF
--- a/docs/pages/getting-started/adapters/prisma.mdx
+++ b/docs/pages/getting-started/adapters/prisma.mdx
@@ -103,53 +103,11 @@ The original database edge-runtime workaround, to split your `auth.ts` configura
 At the moment, Prisma is still working on being fully compatible with edge runtimes like Vercel's. See the issue being tracked [here](https://github.com/prisma/prisma/issues/20560), and Prisma's announcement about early edge support in the `5.9.1` [changelog](https://github.com/prisma/prisma/releases/tag/5.9.0). There are two options to deal with this issue:
 
 - Use the Prisma's [Accelerate](https://pris.ly/d/accelerate) feature
-- Switch to the `jwt` session strategy
+- Follow our [Edge Compatibility](/guides/edge-compatibility) page as the workaround. This uses the `jwt` session strategy and separates the `auth.ts` configuration into two files.
 
 Using Prisma with the `jwt` session strategy and `@prisma/client@5.9.1` or above doesn't require any additional modifications, other than ensuring you don't do any database queries in your middleware.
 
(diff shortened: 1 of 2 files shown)


ndom91 (approved):
Perfect :ok_hand:

PR #11062, opened 2024-06-02
issue#10966: adding explicit session strategy note to credentials getting started guide

### LINKED ISSUE #10966: Unclear why Credentials Provider fails with Database Adapter

The documentation found here https://authjs.dev/getting-started/authentication/credentials and here https://authjs.dev/getting-started/providers/credentials

Does not include any warnings about trying to use credentials provider with a database adapter. 

The fact that it's not supported is totally okay, especially considering the stance against using the Credentials provider.

That being said, email/password login is required for my app and it took me several hours to understand why the session object was not being retrieved. 

Also next-auth fails setting the session token silently. The only way I discovered the source of my problems is when I turned on debug mode and found this error message:

`UnsupportedStrategy: Signing in with credentials only supported if JWT strategy is enabled. Read more at https://errors.authjs.dev#unsupportedstrategy`

Basically, for some reason credentials provider only works with JWT tokens and not database session tokens. 

Interestingly, this information is available here: https://next-auth.js.org/providers/credentials but, not on the authjs website. I honestly did not know they were connected. 

https://authjs.dev/getting-started/authentication/credentials

swordensen:
## ☕️ Reasoning

This just adds a small comment to the credentials getting started guide indicating that jwt is required for session management. Hopefully this will prevent other people from wasting time trying to implement database storage like I did!

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

fixes https://github.com/nextauthjs/next-auth/issues/10966

## 📌 Resources


ndom91:
Hmm, I think the wording is a bit strict. The credentials provider is useable without the jwt session strategy, but using the database session strategy doesn't auto generate accounts, users, sessions, etc. from the credentials provider 🤔

swordensen:
> Hmm, I think the wording is a bit strict. The credentials provider is useable without the jwt session strategy

That's a good point. I changed the verbage to express that while sessions are optional if you want to use sessions with credentials you should choose JWT because it is supported by the NextAuth > session > strategy option. 

> , but using the database session strategy doesn't auto generate accounts, users, sessions, etc. from the credentials provider 🤔

I am a little confused by this though. I assumed that the database strategy does not do anything with a credentials sign in provider.

PR #11068, opened 2024-06-03
feat: add Qwik integration

gioboa:
## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


packages/frameworks-qwik/src/index.ts
```diff
@@ -0,0 +1,388 @@
...
+const defaultCallbackURL = (req: RequestEventCommon) => {
+  req.url.searchParams.delete("qaction")
+  return req.url.href
+}
+
+async function getSessionData(
+  req: Request,
+  options: AuthConfig
+): GetSessionResult {
+  options.secret ??= process.env.AUTH_SECRET
```
ThangHuuVu, line 311: I think we should call `fillOptionsWithEnvVariables` here

ThangHuuVu (commented):
Thanks! Sorry I am behind with the PR review, I left a few comments. Let me know what you think!

packages/frameworks-qwik/src/index.ts
```diff
@@ -0,0 +1,388 @@
...
+ * };
+ *
+ * ```
+ *
+ * @module @auth/qwik
+ */
+
+import type { AuthConfig } from "@auth/core"
+import { Auth, skipCSRFCheck } from "@auth/core"
+import { AuthAction, Session } from "@auth/core/types"
```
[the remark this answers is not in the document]
gioboa: Sure, nice catch

packages/frameworks-qwik/src/index.ts
```diff
@@ -0,0 +1,388 @@
...
+    })
+  )
+
+  const useAuthSession = routeLoader$((req) => {
+    return req.sharedMap.get("session") as Session | null
+  })
+
+  const onRequest = async (req: RequestEvent) => {
+    if (isServer) {
+      const prefix: string = "/api/auth"
```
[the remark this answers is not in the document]
gioboa: The way we thought about this integration, the end user doesn't have to create folders or anything like that to make the library work. It's all automatic so I don't think it's necessary to change this parameter in the future.

packages/frameworks-qwik/src/index.ts
gioboa: options is already patched here. I changed the name of the variable and clean up the code. Thanks

packages/frameworks-qwik/src/index.ts
```diff
@@ -0,0 +1,388 @@
...
+  type RequestEvent,
+  type RequestEventCommon,
+} from "@builder.io/qwik-city"
+import { EnvGetter } from "@builder.io/qwik-city/middleware/request-handler"
+import { isServer } from "@builder.io/qwik/build"
+import { parseString, splitCookiesString } from "set-cookie-parser"
+import { GetSessionResult, QwikAuthConfig, qwikAuthActions } from "./types.js"
+
+export function qwikAuthQrl(
+  authOptions: QRL<(ev: RequestEventCommon) => QwikAuthConfig>
```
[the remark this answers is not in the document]
gioboa: yup, more or less the same. I added the documentation.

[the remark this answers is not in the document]
gioboa: It's more like this one 
```
import GitHub from "next-auth/providers/github"
export const { handlers, auth } = NextAuth({ providers: [ GitHub ] })
```

packages/frameworks-qwik/src/index.ts
```diff
@@ -0,0 +1,382 @@
...
+) {
+  const useAuthSignIn = globalAction$(
```
ThangHuuVu, line 128: IMO for simplicity, let's use `useSignIn`, `useSignOut`, and `useSession` as our exports

packages/frameworks-qwik/src/index.ts
```diff
@@ -0,0 +1,382 @@
...
+  authOptions: QRL<(ev: RequestEventCommon) => QwikAuthConfig>,
+  req: RequestEventCommon
+) => {
+  let options = await authOptions(req)
+  fillOptionsWithEnvVariables(req.env, options)
+
+  return options
+}
+
+export const fillOptionsWithEnvVariables = (
```
ThangHuuVu, line 339: can we name this function `setEnvDefaults` to align between framework packages?
normally, this function can do some basic assignments and  call `setEnvDefaults` from `@auth/core` - For example, here is how it's used in SvelteKit package:
https://github.com/nextauthjs/next-auth/blob/main/packages/frameworks-sveltekit/src/lib/env.ts

Let's simplify this function like the one in SvelteKit package

packages/frameworks-qwik/src/index.ts
```diff
@@ -0,0 +1,382 @@
...
+    return { data: null, cookie }
+  }
+  if (status === 200) {
+    return { data, cookie }
+  }
+
+  throw new Error(data.message)
+}
+
+const patchAuthOptions = async (
```
ThangHuuVu, line 329: nit: let's remove this function, I think it will be simpler to just call the two lines for each actions

packages/frameworks-qwik/src/index.ts
```diff
@@ -0,0 +1,382 @@
...
+      const body = new URLSearchParams({ callbackUrl })
+      await authAction(body, req, `/api/auth/signout`, auth)
+    },
+    zod$({
+      callbackUrl: z.string().optional(),
+    })
+  )
+
+  const useAuthSession = routeLoader$((req) => {
+    return req.sharedMap.get("session") as Session | null
```
ThangHuuVu, line 196: I think this has to be implemented similarly to signin/signout actions, with the call to `Auth` core - unless Im missing something here. the implementation of this function is currently in `getSessionData`.

packages/frameworks-qwik/src/index.ts
```diff
@@ -0,0 +1,382 @@
...
+  const useAuthSignIn = globalAction$(
+    async (
+      { providerId, callbackUrl: deprecated, options, authorizationParams },
+      req
+    ) => {
+      if (deprecated) {
+        console.warn(
+          "\x1b[33mWARNING: callbackUrl is deprecated - use options.callbackUrl instead\x1b[0m"
+        )
+      }
```
ThangHuuVu, line 137: this is the first version, I'm not sure I understand this deprecated warning, could you help to explain it a bit?

ThangHuuVu (changes requested):
OK I have a chance to rereview it more carefully and leave some change requests, let me know what you think.

For the doc: I think you just need to add `@auth/qwik#build` to https://github.com/nextauthjs/next-auth/blob/main/turbo.json#L75 and it should be enough

packages/frameworks-qwik/src/index.ts
gioboa: `onRequest` is reading the session from the Auth core and it's setting it the `sharedMap`.
onRequest is running before the useSession so we don't need to call the Auth core another time

packages/frameworks-qwik/src/index.ts
gioboa: Fixed

packages/frameworks-qwik/src/index.ts
ThangHuuVu: ah I see, the way the `onRequest` middleware works is that if we export it in `plugin@auth.ts`, it would be ran before anything.

I think this could not be a good default for `onRequest` - instead of calling the `getSessionData` function on every requests that are not auth,  how about we put the function in `useSession` and allow apps to lazily call it when needed? It would be similar to the SvelteKit approach, where we expose a `getSession` instead of the session object. cc @balazsorban44

balazsorban44 (approved):

PR #11075, opened 2024-06-04
fix: refetch session even if there is no current session

OrJDev:
## ☕️ Reasoning

I have an app that automatically sign in on the server side after signing up with credentials, trying to refetch the session (after calling the server function) without hard refreshing the page seems to be impossible, that is because no functions update the session directly except for `update`, however this function is only a thing when the session is already fetched and exists on the client side, that makes it so there is no possible way to refetch the session at any time i want it to. 

```ts
   const session = await getSession();
                        console.log({ session });
                        // const t = toast.success("Loggin in...");
                        const r = await update();
                        console.log({ r });
```

In this example session is an object and r is not defined

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources


OrJDev:
Bumping this

OrJDev:
@balazsorban44 really need this one its just 1 line of changes

OrJDev:
so currently having this:

```ts
  const s = await checkSignup.mutateAsync({
                        email,
                        password,
                        countryCode: country,
                        refUser,
                        oId: currentOffer?.id,
                      });
                      if (s) {
                        setLogin(true);
                        const s = await getSession({
                          broadcast: true,
                        });
                        console.log(s);
                        // window.location.href = "/earn";
                      } else {
                        toast.success("Account created successfully");
                        await router.push(`/auth?type=sign_in&email=${email}`);
                      }
                      ```
                      
                      so as you see it does print the user, however it doesn't update it client side so it looks like im signed out, that is because i signed in from the server side in order to allow automatic sign in for known users using my own logic, after refreshing the page ofc the session does seems to be correct
                      
                      
Screenshot 2024-08-01 at 12 30 04

ThangHuuVu (approved):
hi @OrJDev , I looked at the code change - I'm not 100% sure why the logic is here in the first place. IMO, it is OK to get this in, but if there's any regression I am afraid that we'll have to revert.

OrJDev:
Awesome thank you

Issue #11104, opened 2024-06-08
Custom SignIn example fix


angel-luis:
### What is the improvement or update you wish to see?

The Server Action example needs the `redirect` import:

`import { redirect } from "next/navigation";`

Also `SIGNIN_ERROR_URL` is undefined, I would add something like:

`const SIGNIN_ERROR_URL = "/error";`

And finally I would remove `auth` from the import as is not needed.

As extra, I would add some redirect when user is logged in, and another redirect when user is logged in and still tries to access the sign in page.

### Does the docs page already exist? Please link to it.

https://authjs.dev/guides/pages/signin

andresgutgon:
Fixed : )

Pixel-Panda:
@balazsorban44 not fixed per [issue](https://github.com/nextauthjs/next-auth/issues/12065)

PR #11107, opened 2024-06-08
chore(docs): add missing import in `SignInPage` code exmaple

andresgutgon:
## 🎫 Affected issues
https://github.com/nextauthjs/next-auth/issues/11104


ndom91 (approved):
Good catch, thanks!

PR #11111, opened 2024-06-08
Eveonline provider add Checks for State

Grasume:
## ☕️ Reasoning

According to Eveonline and the Oauth2 standard to help ensure the prevention of cross-site request forgery.

I have tested this change in my local development environment and have no issues. Should not cause anyone to make changes to there code moving forward.


Grasume:
Is this the right section of code for the New version 5.0 as well?

PR #11137, opened 2024-06-11
docs(solid-start): Make the example api handler work

IvanLisz:
## ☕️ Reasoning

Make the example api handler work

## 🧢 Checklist

- [X] Documentation
- N/A Tests
- [X] Ready to be merged

### DIFF
--- a/packages/frameworks-solid-start/README.MD
+++ b/packages/frameworks-solid-start/README.MD
@@ -44,6 +44,7 @@ export const authOpts: SolidAuthConfig = {
     }),
   ],
   debug: false,
+  basePath: "/api/auth",
 }
 
 export const { GET, POST } = SolidAuth(authOpts)


k-taro56:
`basePath` has defult value.
https://authjs.dev/reference/solid-start#default

IvanLisz:
> `basePath` has defult value. https://authjs.dev/reference/solid-start#default

I don't understand what you are trying to communicate. 

_______________

[@auth/solid-start](https://www.npmjs.com/package/@auth/solid-start) needs `@auth/core`.
`@auth/core`'s default is /auth

If I remember correctly I tracked it to this fn:
```ts
export async function toInternalRequest(
	req: Request,
	config: AuthConfig,
): Promise<RequestInternal | undefined> {
	try {
		if (req.method !== "GET" && req.method !== "POST")
			throw new UnknownAction("Only GET and POST requests are supported.");

		// Defaults are usually set in the `init` function, but this is needed below
		config.basePath ??= "/auth";
```

balazsorban44:
> docs(solid-start) suggest routes/api/auth/[...solidauth].ts

this should be fixed there, not adding a value that already has a default. :pray:

The intention is that `/auth` is the default path for all libraries, except `next-auth` which is kept as `/api/auth` for backwards compatibility only.

PR #11170, opened 2024-06-16
refactor: move as discovery to normalize providers

ThangHuuVu:
## ☕️ Reasoning

Addressing some TODOs in `authorization-url` and `providers`:
- `OAuthConfigInternal['authorization']` now returns a `URL` and `AuthorizationServer` if the provider type is `oidc`. 
- At the normalize step, throw early if `c.authorization` is empty for OAuth 2 providers

 
## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/init.ts
+++ b/packages/core/src/lib/init.ts
@@ -65,7 +65,7 @@ export async function init({
   options: InternalOptions
   cookies: cookie.Cookie[]
 }> {
-  const { providers, provider } = parseProviders({
+  const { providers, provider } = await parseProviders({
     providers: authOptions.providers,
     url,
     providerId,
--- a/packages/core/src/lib/init.ts
+++ b/packages/core/src/lib/init.ts
@@ -65,34 +65,15 @@ export async function init({
   options: InternalOptions
   cookies: cookie.Cookie[]
 }> {
-  const { providers, provider } = await parseProviders({
-    providers: authOptions.providers,
-    url,
-    providerId,
-    options: authOptions,
-  })
-
   const maxAge = 30 * 24 * 60 * 60 // Sessions expire after 30 days of being idle by default
-
-  let isOnRedirectProxy = false
-  if (
-    (provider?.type === "oauth" || provider?.type === "oidc") &&
-    provider.redirectProxyUrl
-  ) {
-    try {
-      isOnRedirectProxy =
-        new URL(provider.redirectProxyUrl).origin === url.origin
-    } catch {
-      throw new TypeError(
-        `redirectProxyUrl must be a valid URL. Received: ${provider.redirectProxyUrl}`
-      )
-    }
-  }
 
   // User provided options are overridden by other options,
   // except for the options with special handling above
   const options: InternalOptions = {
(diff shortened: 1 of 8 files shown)


packages/core/src/lib/init.ts
```diff
@@ -65,7 +65,7 @@ export async function init({
   options: InternalOptions
   cookies: cookie.Cookie[]
 }> {
-  const { providers, provider } = parseProviders({
+  const { providers, provider } = await parseProviders({
```
ThangHuuVu, line 68: This would mean we are hitting the `well-known` endpoints on every auth requests now, compare to the previous approach where we only hit it on getting the authorization URL.

balazsorban44: we could cache it in the module scope

packages/core/src/lib/utils/providers.ts
```diff
@@ -9,65 +10,85 @@ import type {
...
-function normalizeOAuth(
-  c: OAuthConfig<any> | OAuthUserConfig<any>
-): OAuthConfigInternal<any> | {} {
-  if (c.issuer) c.wellKnown ??= `${c.issuer}/.well-known/openid-configuration`
+async function normalizeOAuth(c: OAuthConfig<any> | OAuthUserConfig<any>) {
+  let authorization: OAuthConfigInternal<any>["authorization"] | undefined
+  if (c.issuer) {
+    const issuer = new URL(c.issuer)
+    try {
+      const discoveryResponse = await o.discoveryRequest(issuer)
```
balazsorban44, line 64: we can cache this in the module scope to potentially hit it less

packages/core/src/lib/utils/providers.ts
```diff
@@ -9,65 +10,85 @@ import type {
...
+    }
+  } else {
+    authorization = normalizeEndpoint(c.authorization)
 
-  const authorization = normalizeEndpoint(c.authorization, c.issuer)
-  if (authorization && !authorization.url?.searchParams.has("scope")) {
-    authorization.url.searchParams.set("scope", "openid profile email")
+    if (!authorization) {
+      throw new TypeError(
+        "The `authorization` options must be provided for this OAuth provider."
```
balazsorban44, line 84:
```suggestion
        "The `authorization` option must be provided for this OAuth provider."
```

balazsorban44 (commented):
Nice, thanks! Left a comment, Wil try to re-review more thoroughly later.

packages/core/src/lib/utils/providers.ts
ThangHuuVu: Not sure what do you mean by "module scope" but I implemented the caching mechanism with the cookies. My worry is that this could be considered bloat :/

ThangHuuVu: another solution is to make authorization become a lazy function for oidc providers. Technically, it's not much different with the original code, just move the discovery function call to providers, and parseProviders won't need to be async either. This is the first solution I've thought of.

PR #11176, opened 2024-06-16
fix: Use sub instead of id in Gitlab

tedzards509:
https://docs.gitlab.com/ee/integration/openid_connect_provider.html

The profile id is in the `sub` field not in the `id` field

## ☕️ Reasoning
https://docs.gitlab.com/ee/integration/openid_connect_provider.html
According to this documentation for Gitlab as an OIDC provider, "The ID of the user" is in the `sub` field. An `id` field is not provided. This change works with Gitlab Enterprise version 16.11 while editing nect-auth version 4.24.5 due to other issues that were fixed by downgrading.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests 
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/src/providers/gitlab.ts
+++ b/packages/core/src/providers/gitlab.ts
@@ -120,7 +120,7 @@ export default function GitLab<P extends GitLabProfile>(
     userinfo: "https://gitlab.com/api/v4/user",
     profile(profile) {
       return {
-        id: profile.id.toString(),
+        id: profile.sub.toString(),
         name: profile.name ?? profile.username,
         email: profile.email,
         image: profile.avatar_url,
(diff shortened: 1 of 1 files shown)


tedzards509:
When sub.toString() is called on an undefined sub, this prevents it from reaching the error handler for a non-existing id field leading to a more cryptic error. Hence the addition of the ?

balazsorban44 (changes requested):
I saw your comment, but based on the linked documentation, `sub` should never be `undefined` and is already of type `string` @tedzards509. Can you verify this?

tvh:
I don't think this PR is correct @tedzards509. As I read the code in https://github.com/nextauthjs/next-auth/blob/5e5a7fc5b41ea2e7e687f5c6e6d89c7967609dcb/packages/core/src/lib/actions/callback/oauth/callback.ts#L146, it will only use the token if the provider is marked as `oidc`. However, GitLab is marked as `oauth`. That means `profile` is called using the result of `userinfo` instead.

PR #11344, opened 2024-07-07
feat: add options object where you can specify a schema

jlvdh:
## ☕️ Reasoning

The @auth/kysely-adapter currently depends on the tables to be configured in postgres. But it's good practice to add the tables in their own schema (eg auth.User, auth.Session, etc.) This PR would enable the configuration of a schema that is used for all the necessary tables by using: ```KyselyAdapter(db, {schemaName: "auth"})```

Interested to hear if this type of feature would be eligible to merge in the project? 

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/docs/pages/getting-started/adapters/kysely.mdx
+++ b/docs/pages/getting-started/adapters/kysely.mdx
@@ -254,6 +254,12 @@ export async function down(db: Kysely<any>): Promise<void> {
 
 For more information about creating and running migrations with Kysely, refer to the [Kysely migrations documentation](https://kysely.dev/docs/migrations).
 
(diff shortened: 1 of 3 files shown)


ndom91:
Yeah so in general we do something similar in most other adapters. Ideally we wouldn't have an implicit link between what you name something in your DB schema manually adn the name we expect in the kysely adapter code..

I'm not familiar with Kysely, but can you explicitly define the schemas and pass them to the Kysely instance? Like we do in the [Drizzle adapter](https://github.com/nextauthjs/next-auth/blob/main/packages/adapter-drizzle/src/lib/pg.ts)

jlvdh:
Added tests and docs now too

PR #11377, opened 2024-07-12
Updated Descope Provider Docs

gaokevin1:
## ☕️ Reasoning

The links were changed to our OIDC docs, and also you will need to provide a client ID and secret from the environment variables in the definition of the provider. You will also no longer need the Issuer URL as an environment variable.

## 🧢 Checklist

- [X] Documentation
- [X] Tests
- [X] Ready to be merged

## 📌 Resources

### DIFF
--- a/docs/pages/getting-started/providers/descope.mdx
+++ b/docs/pages/getting-started/providers/descope.mdx
@@ -62,11 +62,7 @@ import NextAuth from "next-auth"
 import Descope from "next-auth/providers/descope"
 
 export const { handlers, auth, signIn, signOut } = NextAuth({
-  providers: [ Descope({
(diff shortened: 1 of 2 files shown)


ndom91:
Thanks for the update! However, we auto-detect the clientId and clientSecret env vars, so these could all be simplified to:

```ts
export const { handlers, auth, signIn, signOut } = NextAuth({
  providers: [ Descope({
      checks: ["pkce", "state"],
    })]
})
```

Actually, if it *always* requires those two checks to be passed, I think we'd be better off adding that as a default to the [provider](https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/providers/descope.ts#L99) directly.

PR #11381, opened 2024-07-13
fix(core): invalid url error when using custom error page

k0d13:
## ☕️ Reasoning

When a configuration error occurs and there is a custom error page (`pages.error` config option), `@auth/core` does not currently prepend the url origin to the path before passing it to `Response.redirect`, resulting in an invalid URL error.
image
Passing the full URL into the `pages.error` config option would fix this.
This PR copys the URL building code used to handle other error types, and adds it to the configuration error handling.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -153,7 +153,8 @@ export async function Auth(
       return toResponse(page)
     }
 
-    return Response.redirect(`${pages.error}?error=Configuration`)
+    const url = `${internalRequest.url.origin}${pages.error}?error=Configuration`
+    return Response.redirect(url)
   }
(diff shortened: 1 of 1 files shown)


ThangHuuVu:
@apteryxxyz Could you open an issue with a minimal reproduction first? 🙏

balazsorban44 (approved):

balazsorban44:
@ThangHuuVu I think this should be reasonable. On server-side, we don't have the origin url and relative redirects won't work, so I think OP is right.

PR #11409, opened 2024-07-19
fix(drizzle): update schemas & refactor

realmikesolo:
I have updated schemas and types with `email` being nullable in `users` table. Also, I have bumped Drizzle ORM & Drizzle Kit versions and refactored tests.

## ☕️ Reasoning

As already discussed in [discord channel](https://discord.com/channels/1200116961590399008/1231534836859146241), `email` column should be nullable.


realmikesolo:
I have added `id` to `fixtures` because NextAuth always pass value for `id`, so I think it should be in `fixtures` to provide correct test cases.

ndom91:
@realmikesolo why are there so many changes? :thinking:  That can't all be from the `pnpm-lock.yaml`, right?

realmikesolo:
Hello @ndom91! I have fixed `pnpm-lock.yaml`

ndom91:
Nice, looks much better! 

Out of curiosity, do the non-PK columns really all need that `isPrimaryKey: false`? 🤔

realmikesolo:
Hey @ndom91. I have updated type, so if we don't need `isPrimaryKey: false` for non-PK columns. But I have faced this issue `Cannot find module '@auth/core/adapters' or its corresponding type declarations` and it is a bit strange for me because I didn't change any dependencies related to `auth/core`. Could you please check it?

ndom91:
> Hey @ndom91. I have updated type, so if we don't need `isPrimaryKey: false` for non-PK columns. But I have faced this issue `Cannot find module '@auth/core/adapters' or its corresponding type declarations` and it is a bit strange for me because I didn't change any dependencies related to `auth/core`. Could you please check it?

Yeah I don't have time today, but this weekend I can definitely take a closer look and hopefully get this merged. The lock file definitely looks much better!

ndom91:
> Hey @ndom91. I have updated type, so if we don't need `isPrimaryKey: false` for non-PK columns. But I have faced this issue `Cannot find module '@auth/core/adapters' or its corresponding type declarations` and it is a bit strange for me because I didn't change any dependencies related to `auth/core`. Could you please check it?

Where are you running into this? I was abel to run all the tests successfully still

realmikesolo:
> > Hey @ndom91. I have updated type, so if we don't need `isPrimaryKey: false` for non-PK columns. But I have faced this issue `Cannot find module '@auth/core/adapters' or its corresponding type declarations` and it is a bit strange for me because I didn't change any dependencies related to `auth/core`. Could you please check it?
> 
> Where are you running into this `@auth/core/adapters` issue? I was abel to run all the tests successfully still

I was facing this issue during the build. It might be problem on my local setup then. Thank you for your time

PR #11463, opened 2024-07-26
feat(adapter-dynamodb): extend schema options

DylanLacey:
This change allows for more flexibility when customising the schema used by the DynamoDB adapter. Specifically, it adds the ability to customize:

* **EntityTagName**: The attribute name under which to store the *type* of the entity.
* **EntityTags**: The value to use for each Entity type for the above attribute.
* **EntitySlugs**: The slug which is prepended to each type's partition/sort key.

`index.test.ts` and `format.test.ts` were updated to reflect changes required to the `format.from` function.

`custom_config.test.ts` was added, which runs a full suite of the basic tests against both the existing customizations _and_ those included in this PR.

## 💻 Example

`EntityTags` and `EntitySlug` are both of type `DynamoDBEntityTypeOptions` and can take an (optional, string) value for each type of entity the adapter stores:

```
const entityTags: DynamoDBEntityTypeOptions = {
    user: "AUTHUSER",
    account: "ACT",
    session: "SES",
    vt: "VT"
}
```

`EntitySlug` does not format the provided slugs in any way; If users wish to include a separator such as `#`, they need to include it.

## ☕️ Reasoning
Firstly, to increase the ability to integrate Auth.js with other systems. For instance , [DynamoDBToolbox](https://www.dynamodbtoolbox.com/) uses an attribute `_et` by default to type records.

Secondly, to give users more control in general. They may wish to disambiguate AuthJS `User` items from some extant `User` entity, or shorten `ACCOUNT` to `ACNT`.

## 🧢 Checklist

- [X] Documentation
- [X] Tests
- [X] Ready to be merged


DylanLacey:
Hey folks, this is still relevant; Is there something I can do to help make it easier to get a trial deployment run, besides updating to remove the merge conflicts?

Issue #11428, opened 2024-07-22
[auth][error] CallbackRouteError


jelmd:
### Environment

```
  System:
    OS: Linux 5.4 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (8) x64 Intel(R) Xeon(R) CPU E3-1505M v5 @ 2.80GHz
    Memory: 49.06 GB / 62.40 GB
    Container: Yes
    Shell: 6.21.00 - /usr/bin/tcsh
  Binaries:
    Node: 20.5.1 - /usr/bin/node
    npm: 9.8.0 - /usr/bin/npm
    Watchman: 4.9.0 - /usr/bin/watchman
  npmPackages:
    next: ^15.0.0-rc.0 => 15.0.0-rc.0 
    next-auth: ^5.0.0-beta.19 => 5.0.0-beta.19 
    react: ^19.0.0-rc-512b09b2-20240718 => 19.0.0-rc-512b09b2-20240718
```

### Reproduction URL

https://github.com/jelmd/nextjs-bugs/tree/01_auth/01_auth

### Describe the issue

If a signIn fails e.g. because of a wrong password, nextAuth throws the following Error/stacktrace but does not explain, what the real problem is, or what it really expects. Documentation gives also no usable hint, what the problem is:
```
[auth][error] CallbackRouteError: Read more at https://errors.authjs.dev#callbackrouteerror
[auth][cause]: Error
    at Module.callback (webpack-internal:///(rsc)/./node_modules/@auth/core/lib/actions/callback/index.js:226:23)
    at async AuthInternal (webpack-internal:///(rsc)/./node_modules/@auth/core/lib/index.js:66:24)
    at async Auth (webpack-internal:///(rsc)/./node_modules/@auth/core/index.js:126:34)
    at async /home/elkner/work/nextjs-bugs/01_auth/node_modules/next/dist/compiled/next-server/app-route.runtime.dev.js:6:63352
    at async eP.execute (/home/elkner/work/nextjs-bugs/01_auth/node_modules/next/dist/compiled/next-server/app-route.runtime.dev.js:6:54549)
    at async eP.handle (/home/elkner/work/nextjs-bugs/01_auth/node_modules/next/dist/compiled/next-server/app-route.runtime.dev.js:6:64693)
    at async doRender (/home/elkner/work/nextjs-bugs/01_auth/node_modules/next/dist/server/base-server.js:1419:42)
[...]

Issue #11190, opened 2024-06-17
Cannot return custom error to the client side in the Next-Auth v5 Credentials Auth


dsmabulage:
### Provider type

Credentials

### Environment

```
 "next": "14.2.3",
 "next-auth": "^5.0.0-beta.19",
```

### Reproduction URL

https://github.com/dsmabulage/inventes

### Describe the issue

I'm trying to implement the email, password login in the next app using next-auth v5.

In the NextAuth function inside authorize, I tried various methods to throw a custom error message that can passed to the user to show what's going wrong with the credentials.

### How to reproduce

```
export const { auth, handlers, signIn, signOut } = NextAuth({
  session: {
    strategy: 'jwt',
  },
  pages: {
    signIn: '/auth/login',
  },
  providers: [
    CredentialsProvider({
      credentials: {
        email: {},
        password: {},
      },
      async authorize({ email, password }) {
        try {
          throw new Error(
            JSON.stringify({ errors: 'user.errors', status: false })
          );
        } catch (error) {
          throw new Error(
            JSON.stringify({ errors: ' catch user.errors', status: false })
          );
        }
      },
    }),
  ],
```

### Expected behavior

I want to get that error message to the client side and show that error message to the user.

hezean:
Hi, I'm facing the same problem. I found this temporary solution that may be helpful: https://github.com/nextauthjs/next-auth/issues/9900#issuecomment-2167681517.

Ali-Raza764:
The same issue is being discussed [here](https://github.com/nextauthjs/next-auth/issues/11074#issue-2334069653) in detail. I might also tell you that there is a way to check for some error but that also seems to broken currently In previous versions you could do something like this:
```
import { AuthError } from "next-auth";
import { signIn } from "next-auth/react";
    try {
      setLoading(true);
      setError("");
      const res = await signIn("credentials", {
        redirect: false,
        email,
        password,
      });

      //* The res here has no credentials data only error:true or error:null so we can manage the state based on that
      //* Next auth does not send the data due to security reasons
      if (res.error === null) {
        router.push("/protected");
      }
      if (res.error) {
        setError("Check Your Email Or Password");
      }
    } catch (error) {
      console.error(error.message);
      if (error instanceof AuthError) {
        switch (error.type) {
          case "CredentialsSignin":
            console.log("Invalid Password");
            setError("Please Check Your Password")
          default:
            console.log("Something went wrong");
            setError("Something went wrong")
        }
      }
[...]

PR #11469, opened 2024-07-27
fix(core): revert #11050

### LINKED ISSUE #11074: CallbackRouteError on Credentials when following docs
### Provider type

Credentials

### Environment

```
System:
    OS: Linux 5.15 Ubuntu 22.04.3 LTS 22.04.3 LTS (Jammy Jellyfish)
    CPU: (16) x64 13th Gen Intel(R) Core(TM) i7-1360P
    Memory: 10.10 GB / 15.47 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 20.14.0 - ~/.nvm/versions/node/v20.14.0/bin/node
    npm: 10.7.0 - ~/.nvm/versions/node/v20.14.0/bin/npm
    pnpm: 9.1.4 - ~/.nvm/versions/node/v20.14.0/bin/pnpm
  npmPackages:
    next: 14.2.3 => 14.2.3 
    next-auth: 5.0.0-beta.19 => 5.0.0-beta.19 
    react: ^18 => 18.3.1 

```

### Reproduction URL

https://github.com/theswampire/authjs-bug-reproduction

### Describe the issue

When using the Credentials Provider, submitting a wrong password throws an CallbackRouteError. I tested it on 5.0.0-beta.18, 17 and 16 too but they throw CredentialsSignIn instead. 
I followed the v5 docs for setting up Nextjs and the Credentials Provider step by step thus this should be unexpected behaviour.

This is the `auth.ts` file:
```typescript
// auth.ts
import Credentials from "next-auth/providers/credentials";

export const { handlers, signIn, signOut, auth } = NextAuth({
  providers: [
    Credentials({
      credentials: {
        password: {},
      },

      authorize: async (credentials) => {
        if (credentials.password !== "password") {
          return null;
        }
        return { id: "admin", name: "admin" };
      },
    }),
  ],

  pages: {
    signIn: "/signin",
  },
});
```

The SignIn page looks like this:
```tsx
import { signIn } from "@/auth";

export default function SignIn() {
  return (
    <form
      className="flex flex-col max-w-sm gap-2"
      action={async (formData) => {
        "use server";
        await signIn("credentials", formData);
      }}
    >
      <label>
        Password
        <input name="password" type="password" className="border" />
        <input name="redirectTo" type="hidden" value="/secure" />
      </label>
      <span>Password is &quot;password&quot;</span>
      <button className="bg-neutral-300">Sign In</button>
    </form>
  );
}
```
I tried to wrap the signIn call in a try-catch like in #11010 but that obviously didn't fix the underlying issue.

According to the documentation, the CallbackRouteError using the Credentials Provider means that either the `authorize` or another callback throws but I didn't overwrite any callbacks nor should `authorize` throw anything. Furthermore, if I remember correctly, when setting `redirect` to false, the `SignInResponse` contained `ok: true` and a Configuration error which I find odd.

### How to reproduce

Enter anything but the correct password "password" and then the exception should be thrown.
Using the correct password works just fine.

### Expected behavior

Don't throw an exception.

ThangHuuVu:
in #11050 we accidentally introduced a regression issue for the `authorize` function of the `CredentialProvider`. Returning `null` in this callback is now throwing a `ConfigurationError` which should not be the case. 

This PR reverts the changes in #11050 and adds some tests to avoid regressing in the future.

Fixes #11074 #11190 #11428

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -98,6 +98,7 @@
     "dev": "pnpm css && pnpm providers && tsc -w",
     "test:e2e": "playwright test -c ../utils/playwright.config.ts",
     "test": "vitest run -c ../utils/vitest.config.ts",
+    "test:watch": "vitest -c ../utils/vitest.config.ts",
     "providers": "node scripts/generate-providers"
   },
   "devDependencies": {
--- a/packages/core/src/errors.ts
+++ b/packages/core/src/errors.ts
@@ -193,7 +193,7 @@ export class InvalidCallbackUrl extends AuthError {
  * 1. The user is redirected to the signin page, with `error=CredentialsSignin&code=credentials` in the URL. `code` is configurable.
  * 2. If you throw this error in a framework that handles form actions server-side, this error is thrown, instead of redirecting the user, so you'll need to handle.
  */
-export class CredentialsSignin extends Error {
+export class CredentialsSignin extends SignInError {
   static type = "CredentialsSignin"
   /**
    * The error code that is set in the `code` query parameter of the redirect URL.
--- a/packages/core/src/lib/actions/callback/index.ts
+++ b/packages/core/src/lib/actions/callback/index.ts
@@ -324,12 +324,8 @@ export async function callback(
       )
       const user = userFromAuthorize
 
-      if (!user) {
-        console.error(
-          "Read more at https://errors.authjs.dev/#credentialssignin"
-        )
-        throw new CredentialsSignin()
-      } else user.id = user.id?.toString() ?? crypto.randomUUID()
+      if (!user) throw new CredentialsSignin()
+      else user.id = user.id?.toString() ?? crypto.randomUUID()
 
       const account = {
         providerAccountId: user.id,
--- a/packages/core/test/actions/callback.test.ts
+++ b/packages/core/test/actions/callback.test.ts
@@ -7,6 +7,7 @@ import Credentials from "../../src/providers/credentials.js"
 
 import { makeAuthRequest } from "../utils.js"
 import { skipCSRFCheck } from "../../src/index.js"
+import { CredentialsSignin } from "../../src/errors.js"
 
 describe("assert GET callback action", () => {
   beforeEach(() => {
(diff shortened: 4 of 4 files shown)


ndom91 (approved):

BertramYe:
but thre were problem, if we use the `try....catch....` to wrap the `signIn` function, even we can catch the error we defined, but it won't auto redirect the path we want, even we have login in with correct credentails !!!!!!

TruongNam0301:
i still cannot handle error message to return clientSide
`import { APP_ENV } from "@/constants/env.constant";
import { type NextAuthConfig } from "next-auth";
import Credentials from "next-auth/providers/credentials";
import Google from "next-auth/providers/google";

export default {
  providers: [
    Google({
      clientId: APP_ENV.GOOGLE_CLIENT_ID,
      clientSecret: APP_ENV.GOOGLE_CLIENT_SECRET,
    }),
    Credentials({
        const payload = {
          email: credentials.email as string,
          password: credentials.password as string,
        };
        throw new Error("Email not found");
        // const user = await signin(payload);
        // return user ?? null;
      },
    }),
  ],
  pages: {
    signIn: "/login",
    error: "/login",
  },
  secret: APP_ENV.AUTH_SECRET,
} satisfies NextAuthConfig;
`
in clientSide always return this
image

russssl:
@TruongNam0301 Same error here. No metter what error is thrown, I always get 'Configuration'

manishoctal:
`Configuration` is still there.

image

domkoder:
> `Configuration` is still there.
> 
> image

Same thing here, any luck @manishoctal @TruongNam0301

manishoctal:
Not yet!!

Zhavii:
Really?

iamdevlinph:
I'm in `"next-auth": "5.0.0-beta.25"` and it's still happening. This is using the solution [here](https://github.com/nextauthjs/next-auth/issues/11190#issuecomment-2241657575)

image

M-Ali-Haider:
@iamdevlinph I love you

MauriRojas:
@iamdevlinph thank you, this is the only solution that worked.

russssl:
using better-auth now no such problems happening

PR #11489, opened 2024-07-31
fix: the provider should supports request function

kenxx:
## ☕️ Reasoning

provider.token should support `request` function that mentions in documentation, but it looks not support now.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources


balazsorban44:
Hi, this is currently not planned.

We rather introduced the internal `conform()` method.

https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/providers/oauth.ts/#L268-L269

liuhuapiaoyuan:
If this is not in the plan, then the `token` should not provide the `request` configuration item! I think this PR is perfect

kenxx:
> Hi, this is currently not planned.
> 
> We rather introduced the internal `conform()` method.
> 
> https://github.com/nextauthjs/next-auth/blob/0dcd73786d0baf36ca29ef2b9a494a5d7d5ab1dc/packages/core/src/providers/oauth.ts#L268-L269

Thank you for your reply! 

I understand that you may not have intended to do this, but I hope we can find a solution for my situation.

The OAuth integration I need might not be a standard OAuth. I need to initiate another request before obtaining the token, and I can only continue to get the token after receiving the response. The conform function also may not be able to solve my problem.

PR #11519, opened 2024-08-05
Correct an error in the Entra docs regarding tenant configuration

thomasmost:
## ☕️ Reasoning

The current docs for Entra (and probably Azure too) are very confusing and essentially mislead people into setting up a configuration that will not work (and worse, will likely _appear_ to work for their own testing purposes while within their tenancy, and then will fail for external customers).

Since Azure is deprecated, it makes sense to simply make this correction going forward.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

This **DOES NOT FIX** but mitigates confusion around this issue: https://github.com/nextauthjs/next-auth/issues/8374

Note it would still be nice for some sort of compatibility workaround to be merged! **It's worth noting that in next@4.24.7 this worked if you excluded the AZURE_AD_TENANT_ID** 

```json
    "next-auth": "^4.24.7"
```

### DIFF
--- a/docs/pages/getting-started/providers/microsoft-entra-id.mdx
+++ b/docs/pages/getting-started/providers/microsoft-entra-id.mdx
@@ -164,6 +164,6 @@ AUTH_MICROSOFT_ENTRA_ID_SECRET=<copy generated client secret value here>
 AUTH_MICROSOFT_ENTRA_ID_TENANT_ID=<copy the tenant id here>
 ```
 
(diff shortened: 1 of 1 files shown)


balazsorban44:
Thanks, but the correct solution to this will be https://github.com/nextauthjs/next-auth/pull/9718#pullrequestreview-1842680299

We should not link to GitHub issues as documentation.

thomasmost:
@balazsorban44 what do you mean correct solution? The docs are incorrect, or at the very least extremely misleading.

Why don't I just add a commit removing the reference to the github issue but still clarifying that the app will be single-tenant by default?

thomasmost:
I believe this PR should be reopened

PR #11533, opened 2024-08-07
ar(fix) [DPCP-62]: Normalize Body Value if Token Hash is present

angeloreale:
## ☕️ Reasoning

Depending on different cookie policies / layer 6 proxying / cardinal factors, the bodyValue for the CSRFToken might include the token hash.
Less than regressing it in case it comes directly from the form data, we can normalize/coerce it to account for that extra possibility.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

## 🎫 Affected issues

I didn't create an issue because I've solved it locally.

## 📌 Resources


angeloreale:
Pushing still because I'm trying to use the syntax you like. My original solution was a bit nastier.

PR #11540, opened 2024-08-08
support bun.sh

tungv:
## ☕️ Reasoning

When the runtime is bun, the `Request` object has the key `req` in it. That would make this check to fails.
In bun, response.headers is a read-only header object, so no `append` method on it.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


balazsorban44:
Thanks, but these sound like non-spec compliant changes, so it is undesired. Maybe you can open an issue first, explaining the incompatibility with a reproduction first. :pray:

PR #11600, opened 2024-08-13
chore(docs): update usage instructions for `useSession` hook

halvaradop:
## ☕️ Reasoning
This pull request updates the `migration-to-v5` section with important information on using the `useSession` hook for client components. The updates include

- Guidance on wrapping the `useSession` hook with `<SessionProvider />` to avoid errors
- A recommendation to place the `<SessionProvider />` at the top level of your project to ensure proper functionality.

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## 📌 Resources


balazsorban44:
Thanks, but IMO these don't need to be part of the migration docs, as it was always the case that useSession needs to be wrapped.

It should even throw this error:

https://github.com/nextauthjs/next-auth/blob/a5fcdfd63fbb89228d770484619ae94a4e7cf711/packages/next-auth/src/react.tsx#L141

Furthermore, it might not be the case that you need to wrap your entire app with SessionProvider. In most cases in App Router you would barely need to use SessionProvider or useSession at all. For the Pages Router, nothing has changed.

halvaradop:
Thanks, you're right. Maybe this content was unnecessary or obvious, but I think the code clarifies the usage of this hook. I will delete this content and create a commit if you think it is a good contribution.

PR #11661, opened 2024-08-23
chore(docs): cleanup built-in pages docs page

halvaradop:
## ☕️ Reasoning
This pull request adds code examples to the `Built-in Pages` section for the frameworks supported by Auth.js, including Next.js, Qwik, and SvelteKit. The examples demonstrate how automatic page generation works when the user provides providers in their configuration file.

Additionally, to avoid code repetition across frameworks, the provider configurations are now written in a separate file.

On the other hand, the code for the `SignInButton` component in the Next.js example has been removed. The focus is on demonstrating how to set up built-in pages. If you believe the `SignInButton` component is relevant, please let me know, and I will make the necessary changes.

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## 📌 Resources


docs/pages/guides/pages/built-in-pages.mdx
```diff
@@ -3,22 +3,66 @@ import { Screenshot } from "@/components/Screenshot"
...
+      username: { label: "Username" },
+      password: { label: "Password", type: "password" }
+    },
+    async authorize(credentials) {
+      // mplement some logic to check if the user exists.
+      return credentials
+    }
+  })
+]
+```
```
ndom91, line 26: Not sure that this example is very helpful or relevant here.

docs/pages/guides/pages/built-in-pages.mdx
```diff
@@ -3,22 +3,66 @@ import { Screenshot } from "@/components/Screenshot"
...
 
 <Code>
```
ndom91, line 28: I think you're right, the `SignInButton` example wasn't very helpful here. But I dont think these "how to set providers" examples are very useful here either.

I think we could probably just remove all of these examples for better clarity on this page :ok_hand:

docs/pages/guides/pages/built-in-pages.mdx
halvaradop: My idea with this code is to show the setup of the providers needed to create the same UI as shown in the screenshot at the bottom of the section. With this in mind, I have added the properties for the credentials because, if these are not set, it will throw an error as it expects an object or formData. I believe it’s better to include this content in the same section to avoid repeating the code for each implementation across different frameworks.

docs/pages/guides/pages/built-in-pages.mdx
halvaradop: I hadn’t read this comment before, but I think you’re right. I will wait for your response to my previous comment and make the necessary changes.

docs/pages/guides/pages/built-in-pages.mdx
ndom91: Hmm okay I see where you're coming from. However, I think folks looking into the pages will have at least 1 provider already setup, it's kind of the minimum required thing to use auth.js, ya know what I mean. And I just don't see a lot of value in repeating it again here. I think getting it to look 100% like the screenshot is not necessary

halvaradop:
I have made the changes based on your suggestions. 🧙‍♂️

ndom91 (approved):
Great, thanks!

PR #11666, opened 2024-08-24
fix(tsconfig): update `extends` property in packages configuration

halvaradop:
## ☕️ Reasoning
This pull request resolves issues caused by the incorrect configuration of the extends property in the `tsconfig.json` files within the `packages` folder. The errors were due to invalid paths in the configuration. The following changes were made:

- Updated `extends` Property: The extends property in each tsconfig.json file within the packages has been updated to `../utils/tsconfig.json`. This change corrects the path errors that were causing issues.
tsconfig
image

- You can verify the fixes by running the following bash script before and after applying the changes
```bash
#!/bin/bash

for dir in */ ; do
    cd "$dir"
    tsc --showConfig
    cd ..
done
```

### Before
not-found

### After
image

### Note
The extends property in the framework-sveltekit folder has not been updated, as the correct reference file is currently unknown.

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


k-taro56:
Related priviously-raised Issue: #11665

ndom91 (approved):
Yeah thanks this makes sense.

Regarding SvelteKit, thats a tricky one since it extends the auto-generated `.svelte-kit/tsconfig.json` by default.

In previous SvelteKit project's I've copied elements out of that extended file in order to add things to them, for example. 

Maybe we can copy it's entire contents (that differ from our tsconfig in the sveltekit projects) into the tsconfig and then the `extends` key becomes available again :thinking: Anyway, not important for now. I'll approve this as is :+1:

halvaradop:
Thank you for explaining the error in SvelteKit, even though I don’t have knowledge in SvelteKit. 👌

PR #11673, opened 2024-08-26
chore: clean up custom error section and add repository links

halvaradop:
## ☕️ Reasoning
This pull request cleans up the `custom error` section in `guides/pages` by removing the example code for the frameworks (currently located below this content). It also adds links to the repositories supported by Auth.js, including: Next.js, SvelteKit, Express, Solid, and Qwik.

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [x] Ready to be merged

## 📌 Resources

### DIFF
--- a/docs/pages/guides/pages/error.mdx
+++ b/docs/pages/guides/pages/error.mdx
@@ -5,21 +5,13 @@ import { Screenshot } from "@/components/Screenshot"
 
 Auth.js can be configured to display a custom error page when something goes wrong during the user authentication flow (sign in, sign out, etc..).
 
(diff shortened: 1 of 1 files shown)


ndom91:
The existing examples don't contain an error page example at all. I think the existing "Custom Error Page' example is good as-is :pray:

halvaradop:
I’m not sure how useful that solution is because, as I understand it, this section is meant for creating a custom error page, and it does not contain the relevant content. However, at the bottom of the section, there is information on how to do this using Next.js, with code that includes the same component name and directory. For this reason, I have removed that code.

On the other hand, I think it would be beneficial to add links to other frameworks supported by Auth.js to help users build a custom error page using framework-specific templates.

PR #11680, opened 2024-08-26
chore(docs): adds deep link to D1 Adapter doc

mackenly:
Adds a deep link to a Pages application's environment variables for the D1 adapter rather than using the generic dashboard home page link. This link prompts the users to select their account, Pages application, and then redirects them to the correct page where environment variables can be added/edited.

This deep linking functionality is built into the Cloudflare dashboard and is a best practice to improve user experience.


ndom91 (approved):
Nice addition, thanks!

PR #11733, opened 2024-08-30
chore(docs): add caveat for stateless session when using db providers

shawnmclean:
Ref: https://discord.com/channels/1200116961590399008/1266171017584181389/1266306385885532262


ndom91 (approved):
Good point 👌

PR #11737, opened 2024-08-31
chore: e2e

ThangHuuVu:
- Removed the cache steps, as they are [not recommended](https://playwright.dev/docs/ci#caching-browsers) in the Playwright docs. 
- Move the tests under next-auth which I think makes more sense, since the e2e tests are ran against a next.js app
- Have to disable a test (Signout). The session is still shown in the UI, this could be an issue.


turbo.json
```diff
@@ -35,6 +35,15 @@
         "playwright-report/**",
         "playwright/.cache/**",
         "test-results/**"
+      ],
+      "env": [
```
ndom91, line 39: Minor tweak, but lets change these to `passThroughEnv` ([src](https://turbo.build/repo/docs/crafting-your-repository/using-environment-variables#passthrough-variables)) so changes to them don't potentially kick off fresh/uncached re-builds

.github/workflows/release.yml
```diff
@@ -113,20 +113,20 @@ jobs:
...
-        with:
-          path: |
-            ~/.cache/ms-playwright
-          key: ${{ runner.os }}-playwright-${{ env.PLAYWRIGHT_VERSION }}
       - name: Install Playwright
-        if: steps.playwright-cache.outputs.cache-hit != 'true' && github.repository == 'nextauthjs/next-auth'
+        if: github.repository == 'nextauthjs/next-auth'
         run: |
           pnpm exec playwright install --with-deps chromium
+      - name: Pass environment variables to Playwright
```
ndom91, line 120: So I think this might be the issue.. These env vars are in github secrets, which are accessed differently. Let's try adding the following to after like 78, so they apply to this whole `test` job. Something like:

```yml
  env:
    AUTH_SECRET: ${{ secrets.AUTH_SECRET }}
    TEST_KEYCLOAK_USERNAME: ${{ secrets.TEST_KEYCLOAK_USERNAME }}
    TEST_KEYCLOAK_PASSWORD: ${{ secrets.TEST_KEYCLOAK_PASSWORD }}
    AUTH_KEYCLOAK_ID: ${{ secrets.AUTH_KEYCLOAK_ID }}
    AUTH_KEYCLOAK_SECRET: ${{ secrets.AUTH_KEYCLOAK_SECRET }}
    AUTH_KEYCLOAK_ISSUER: ${{ secrets.AUTH_KEYCLOAK_ISSUER }}
    AUTH_TRUST_HOST: ${{ secrets.AUTH_TRUST_HOST }}
```

PR #11771, opened 2024-09-04
docs: remove duplicated search bar in mobile menu

halvaradop:
## ☕️ Reasoning
This pull request fixes the issue of the duplicated search bar in the mobile menu and adds a unique identifier for the search bar. The changes are limited to CSS styles to remove the duplicate HTML. However, I encountered a problem where the project breaks when attempting to format the CSS styles (which is quite unusual).

### Before
before-header
### After
after-header

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/docs/components/DocSearch/wrapper.tsx
+++ b/docs/components/DocSearch/wrapper.tsx
@@ -48,7 +48,7 @@ export default function () {
   }, [])
 
   return (
-    <div className="relative max-md:hidden">
+    <div className="relative md:hidden">
       <InstantSearch
         indexName="next-auth"
         // @ts-expect-error
--- a/docs/components/DocSearch/wrapper.tsx
+++ b/docs/components/DocSearch/wrapper.tsx
@@ -48,7 +48,7 @@ export default function () {
   }, [])
 
   return (
-    <div className="relative">
+    <div className="relative searchbar">
       <InstantSearch
         indexName="next-auth"
         // @ts-expect-error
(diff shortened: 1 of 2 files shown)


docs/components/DocSearch/wrapper.tsx
```diff
@@ -48,7 +48,7 @@ export default function () {
   }, [])
 
   return (
-    <div className="relative">
+    <div className="relative searchbar">
```
ndom91, line 51: This is only supposed to hide in mobile? Maybe a tailwind class like this would cover it then: 

``` 
max-md: hidden 
```

balazsorban44: agree

ndom91: Applied :+1:

halvaradop:
Maybe I did not explain this well in my previous comment, but if you add a class to the wrapper.tsx component, the class is applied to both instances of this component — one at the top of the navbar and the second within the menu. For that reason, I added this CSS class to specifically target the component rendered within the menu. In your latest changes, the components are not rendered correctly because the search bar is hidden in both the mobile and desktop versions. My goal is to remove the search bar only from the component within the menu. Please see the deployment of my changes.

ndom91:
Hey sorry I tried to push another commit to your branch but I didn't have the correct permissions to do so. So I pushed a new branch / PR - https://github.com/nextauthjs/next-auth/pull/11795

PR #11829, opened 2024-09-14
fix: avoid doing the same operation twice on Middleware and Route Handlers

ianvieira:
When the middleware changes the session, it transmits the changed value on the set-cookie. To avoid re-doing the same operation again on the route handlers, we need to merge the set-cookie with the cookie, to have the latest value. Not doing this operation is critical for refresh token, as they can only be refreshed once.

This MR aims to solve the issue on this discussion:
https://github.com/nextauthjs/next-auth/discussions/11317

## ☕️ Reasoning

Currently, when you try to refresh a token using the `jwt()` callback, it gets refreshed twice. This happens due to the middleware refreshing once and then the server refreshing again. In theory, only the middleware should try to refresh it, but when the request arrives to the server it has on the `cookie` header the old value for the session, forcing it to re-do the same operation again. In order to fix it, in this MR I'm merging the result from the middleware (available on the `set-cookie` header) into the `cookie`, this way, the server will have the latest transformed value and will not try to perform the same operation again.

## 🧢 Checklist

- [X] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


ianvieira:
@ThangHuuVu @balazsorban44 @wobsoriano @ndom91 @iaincollins can we have it reviewed? 

Without the fix, the refresh token example on the Next Auth page is unusable, as most of the providers do not allow double refresh of a token.

After at least having a positive indication on the direction I can cover it with tests.

packages/next-auth/src/lib/index.ts
```diff
@@ -59,6 +59,33 @@ export interface NextAuthConfig extends Omit<AuthConfig, "raw"> {
   }
 }
 
+function mergeCookiesAndSetCookiesHeaders(headers: Headers): string {
+  const setCookies = headers.get("set-cookie")
```
balazsorban44, line 63: This should use `.getSetCookie`, which is the standard way of retrieving potentially multiple set-cookie values.

packages/next-auth/src/lib/index.ts
```diff
@@ -59,6 +59,33 @@ export interface NextAuthConfig extends Omit<AuthConfig, "raw"> {
...
+  const cookies = headers.get("cookie");
+
+  if(!setCookies || (!cookies && !setCookies)) {
+    return cookies ?? "";
+  }
+
+  // We need to reparse the headers because the headers from Next returns [] on getSetCookie().
```
balazsorban44, line 70: If this is the case, it should be reported as a bug on Next.js.

Manually parsing cookies is error-prone, see below.

packages/next-auth/src/lib/index.ts
```diff
@@ -59,6 +59,33 @@ export interface NextAuthConfig extends Omit<AuthConfig, "raw"> {
...
+  const headersToParse = new Headers(headers);
+  const setCookiesHeaders = headersToParse.getSetCookie();
+  const cookiesArray = cookies?.split(";") ?? [];
```
balazsorban44, line 73: This won't be reliable, since dates can contain the `;` character, which will produce an  incorrect cookie array.

packages/next-auth/src/lib/index.ts
```diff
@@ -59,6 +59,33 @@ export interface NextAuthConfig extends Omit<AuthConfig, "raw"> {
...
+  const cookiesObject = cookiesArray.reduce((acc, cookie) => {
+    const [key, value] = cookie.split("=");
+    acc[key.trim()] = value;
+    return acc;
+  }, {} as Record<string, string>);
+
+  for(const setCookie of setCookiesHeaders) {
+    const [keyValue, ..._parts] = setCookie.split(";");
```
balazsorban44, line 81: As mentioned above, `;` is not safe to rely on for splitting.

balazsorban44 (changes requested):
_Note: First, thank you, I really appreciate the engagement! But I kindly ask you to refrain from mass-tagging random people online for attention, especially if you don't understand their involvement in the project. At least two of the people you tagged have not been involved for _years_. The others - including myself - work on the project in our free time. I check PRs whenever I can._

That said, I feel like this issue requires some more thorough work before we can merge. Likely, we need to introduce https://www.npmjs.com/package/set-cookie-parser as a dependency to make sure we correctly split cookies, as I mentioned in my comments.


I also think that we should make sure we only apply this behavior to our own prefixed cookies to not alter behavior for others unexpectedly.

Need to add tests to verify the correct (and incorrect) behavior before merging.

Lastly, I wonder if this is an issue/potential future issue for other integrations, not just NextAuth.js, and if we can address it fully/partially in `core` instead, or expose the necessary helper methods so the behavior is consistent across all of our framework integrations, not just Next.js. We can explore this in a separate PR though!

balazsorban44:
I am not 100%, but since `set-cookie` headers can influence cache behavior, this could potentially be a security issue, if we don't return them to the browser. :thinking:

We need to also add tests to make sure that HttpOnly and other options are preserved correctly, as that would also be a potential security issue.

balazsorban44:
I'm going to close this for now, as it needs more thinking. Merging `Cookie` and `Set-Cookie` does not sound like the right approach because of the aforementioned concerns.

PR #11848, opened 2024-09-15
chore: add hint about issuer in case provider is keycloak

### LINKED ISSUE #11802: Missing env variable in OIDC configurations that require issuer

When following the [getting started guide](https://authjs.dev/getting-started/authentication/oauth) and selecting keycloak, there is no mention of the "AUTH_KEYCLOAK_ISSUER" variable

image

It will result in the following error "[auth][error] InvalidEndpoints: Provider "keycloak" is missing both `issuer` and `authorization` endpoint config."

The information is correct in the dedicated [keycloak page](https://authjs.dev/getting-started/providers/keycloak) 
image

none

_No response_

falcowinkler:
closes https://github.com/nextauthjs/next-auth/issues/11802


balazsorban44:
We need a different approach, there are many other providers (mostly OIDC ones) that require an issuer

PR #11854, opened 2024-09-16
Forward cookies set in middleware to next part of request routing

### LINKED ISSUE #11853: Session cookie updated in middleware not accessible in server component
### Environment

```

 System:
   OS: macOS 14.6.1
   CPU: (10) arm64 Apple M1 Pro
   Memory: 47.84 MB / 32.00 GB
   Shell: 5.9 - /bin/zsh
 Binaries:
   Node: 20.13.0 - ~/.nvm/versions/node/v20.13.0/bin/node
   npm: 10.8.3 - ~/.nvm/versions/node/v20.13.0/bin/npm
   pnpm: 9.10.0 - ~/Library/pnpm/pnpm
   bun: 1.0.30 - /opt/homebrew/bin/bun
 Browsers:
   Chrome: 128.0.6613.138
   Safari: 17.6
 npmPackages:
   next: ^14.2.2 => 14.2.7
   next-auth: 5.0.0-beta.20 => 5.0.0-beta.20
   react: ^18 => 18.3.1
```

### Reproduction URL

https://github.com/nphmuller/repro-middleware-cookie/commit/d3082fdccb9eb7d18a1a5b817919823681b1ef8b

### Describe the issue

When update the session cookie in the middleware (for example when refreshing an access token) the updated session cookie is not available in a server component that gets rendered, resulting in the access token being refreshed twice

### How to reproduce

https://github.com/nphmuller/repro-middleware-cookie/commit/d3082fdccb9eb7d18a1a5b817919823681b1ef8b contains a full repro, but this is the gist:

update the token in the jwt callback like this: (this simulates a token that is valid for 10 seconds)

```
     if (
         !token.refreshedAt ||
         addSeconds(token.refreshedAt, 10) < new Date()
       ) {
         console.log("refreshing token....")
         token.refreshedAt = new Date()
       }
```

Make sure next-auth is called via a middleware (already the default in `next-auth-example`).

Now run `nom run dev` and open the main page.

### Expected behavior

The console output `refreshing token....` should be printed once, but it's printed multiple times.

This happens because the new session cookie with the refreshed token, created in the middleware, is not passed to the server component.

nphmuller:
## ☕️ Reasoning

When update the session cookie in the middleware (for example when refreshing an access token) the updated session cookie is not available in a server component that gets rendered, resulting in the access token being refreshed twice.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues

Fixes: https://github.com/nextauthjs/next-auth/issues/11853

## 📌 Resources

### DIFF
--- a/packages/next-auth/package.json
+++ b/packages/next-auth/package.json
@@ -86,7 +86,9 @@
   ],
   "license": "ISC",
   "dependencies": {
-    "@auth/core": "workspace:*"
+    "@auth/core": "workspace:*",
+    "@types/set-cookie-parser": "^2.4.10",
+    "set-cookie-parser": "^2.7.0"
   },
   "peerDependencies": {
     "@simplewebauthn/browser": "^9.0.1",
--- a/packages/next-auth/src/lib/index.ts
+++ b/packages/next-auth/src/lib/index.ts
@@ -1,4 +1,5 @@
 import { Auth, createActionURL, type AuthConfig } from "@auth/core"
+import { parse } from "set-cookie-parser"
 // @ts-expect-error Next.js does not yet correctly use the `package.json#exports` field
 import { headers } from "next/headers"
 // @ts-expect-error Next.js does not yet correctly use the `package.json#exports` field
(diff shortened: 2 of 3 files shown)


nphmuller:
The PR introduces a new dependency to parse the set-cookies header. Is this acceptable?

There are also no tests yet. If anyone can advice on the direction to test this, and if the general direction of the PR is fine, I'll work on them.

balazsorban44 (changes requested):
Thanks! Although I like this approach more than #11829, I share the same concerns: https://github.com/nextauthjs/next-auth/pull/11829#issuecomment-2371034628

Merging `cookie` and `set-cookie` headers might be too dangerous.

PR #11894, opened 2024-09-23
fix(getToken): make 'secret' optional when raw JWT is requested

### LINKED ISSUE #11889: Unnecessary requirement for secret and salt in getToken when raw is set to true
### Environment

```
 System:
    OS: Linux 6.8 Ubuntu 22.04.5 LTS 22.04.5 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i7-4810MQ CPU @ 2.80GHz
    Memory: 16.27 GB / 19.20 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 20.11.0 - /usr/bin/node
    Yarn: 1.22.21 - /usr/bin/yarn
    npm: 10.8.2 -
    pnpm: 9.6.0 - ~/.local/share/pnpm/pnpm
  Browsers:
    Brave Browser: 129.1.70.117
    Chrome: 129.0.6668.58
  npmPackages:
    next: ^14.0.3 => 14.2.5 
    next-auth: ^5.0.0-beta.20 => 5.0.0-beta.20 
    react: ^18.3.1 => 18.3.1 

```

### Reproduction URL

https://github.com/alimorgaan/next-auth-getToken-raw-issue/

### Describe the issue

In the `getToken `helper function from `next-auth/jwt`, setting `raw: true `should return the token without requiring the `secret` or `salt`. However, currently, the function throws an error asking for both `secret` and `salt` even when `raw` is set to `true`.

This seems unnecessary because when `raw: true `is used, the function should return the raw token directly without needing any decoding or verification, which usually requires the `secret` and `salt`.

image

What I’ve tried so far:

- i attempted passing any dummy string to avoid throwing error it worked fine 

image

### How to reproduce

1. Import the `getToken` function from `next-auth/jwt`.
2. Use the following code to call `getToken` with `raw: true`:
```js
import { getToken } from "next-auth/jwt";

    req: request,
    raw: true,
});
```

3. Do not provide `secret` or `salt` in the configuration.
4. Observe the error requesting the `secret` and `salt`.

### Expected behavior

When using `raw: true`, the `getToken` function should return the token as-is without requiring the `secret` or `salt`, since no verification or decoding should occur in this case. The current behavior of throwing an error requesting these fields is unexpected.

alimorgaan:
## ☕️ Reasoning
This PR resolves an issue in the `getToken` function where the `secret` and `salt` parameters were required even when the `raw` option was set to `true`, leading to unnecessary validation errors. 

### Changes:
- Updated the `GetTokenParamsBase` type to make `secret` and `salt` optional.
- Modified the logic in `getToken` to ensure `secret` is only required when `raw: false`, allowing raw JWTs to be returned without needing to provide a `secret`.

This fix allows for a more flexible `getToken` usage where raw tokens can be retrieved without unnecessary parameters.

## 🧢 Checklist
- [x]  Ready to be merged

## 🎫 Affected issues
This PR fixes issue #11889.

Fixes #11889

## 📌 Resources
* [Security guidelines](https://github.com/nextauthjs/.github/blob/main/SECURITY.md)
* [Contributing guidelines](https://github.com/nextauthjs/.github/blob/main/CONTRIBUTING.md)
* [Code of conduct](https://github.com/nextauthjs/.github/blob/main/CODE_OF_CONDUCT.md)
* [Contributing to Open Source](https://kcd.im/pull-request)

### DIFF
--- a/packages/core/src/jwt.ts
+++ b/packages/core/src/jwt.ts
@@ -108,7 +108,7 @@ export async function decode<Payload = JWT>(
 }
 
 type GetTokenParamsBase = {
-  secret: JWTDecodeParams["secret"]
+  secret?: JWTDecodeParams["secret"]
   salt?: JWTDecodeParams["salt"]
 }
 
@@ -155,8 +155,7 @@ export async function getToken(
   } = params
 
   if (!req) throw new Error("Must pass `req` to JWT getToken()")
-  if (!secret)
-    throw new MissingSecret("Must pass `secret` if not set to JWT getToken()")
+  
 
   const headers =
     req.headers instanceof Headers ? req.headers : new Headers(req.headers)
@@ -179,6 +178,9 @@ export async function getToken(
   if (!token) return null
 
   if (raw) return token
+
+  if (!secret)
+    throw new MissingSecret("Must pass `secret` if not set to JWT getToken()")
 
   try {
     return await _decode({ token, secret, salt })
--- a/packages/core/src/jwt.ts
+++ b/packages/core/src/jwt.ts
@@ -155,7 +155,6 @@ export async function getToken(
   } = params
 
   if (!req) throw new Error("Must pass `req` to JWT getToken()")
-  
 
   const headers =
     req.headers instanceof Headers ? req.headers : new Headers(req.headers)


balazsorban44 (approved):
Thanks! Makes sense! That said, note that `getToken` is not recommended anymore, we recommend using `auth()` instead. See https://authjs.dev/getting-started/migrating-to-v5#authenticating-server-side

PR #11978, opened 2024-10-05
feat(oauth): add `conformResponse` option for non-conform providers

### LINKED ISSUE #9635: AzureAD common tenant URLs aren't being checked properly

balazsorban44:
On pair with https://github.com/nextauthjs/next-auth/pull/11975, this PR adds a new option for making non-conform responses compliant before they are passed down to `oauth4webapi`.

Fixes https://github.com/nextauthjs/next-auth/issues/9635
Closes https://github.com/nextauthjs/next-auth/pull/9718


balazsorban44:
Scratch this. Unnecessary. [`customFetch`](https://github.com/nextauthjs/next-auth/pull/11975) can also be used to process the responses, and you can even match which request's response you are processing based on the URL for example.

Superseded by https://github.com/nextauthjs/next-auth/pull/11980

PR #11980, opened 2024-10-06
fix(providers): conform Entra ID when responding with the wrong `issuer`

### LINKED ISSUE #9635: AzureAD common tenant URLs aren't being checked properly

### LINKED ISSUE #12041: "response" is not a conform Authorization Server Metadata response
### Provider type

Azure Active Directory

### Environment

Docker container built with the Dockerfile present in the repo.

### Reproduction URL

https://github.com/lbenevento/authjs-response-not-conform

### Describe the issue

When I try to login AuthJS throw the following error: 
```
[auth][error] e2: "response" is not a conform Authorization Server Metadata response
    at tn (/app/.next/server/chunks/633.js:379:7136)
    at rP (/app/.next/server/chunks/633.js:379:43662)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async rC (/app/.next/server/chunks/633.js:379:46272)
    at async rj (/app/.next/server/chunks/633.js:379:48912)
    at async rL (/app/.next/server/chunks/633.js:379:53178)
    at async /app/node_modules/.pnpm/next@14.2.15_react-dom@18.3.1_react@18.3.1/node_modules/next/dist/compiled/next-server/app-route.runtime.prod.js:6:36957
    at async eC.execute (/app/node_modules/.pnpm/next@14.2.15_react-dom@18.3.1_react@18.3.1/node_modules/next/dist/compiled/next-server/app-route.runtime.prod.js:6:27552)
    at async eC.handle (/app/node_modules/.pnpm/next@14.2.15_react-dom@18.3.1_react@18.3.1/node_modules/next/dist/compiled/next-server/app-route.runtime.prod.js:6:38291)
    at async doRender (/app/node_modules/.pnpm/next@14.2.15_react-dom@18.3.1_react@18.3.1/node_modules/next/dist/server/base-server.js:1352:42)
```

The only notable thing I can share about it is the fact that the Next build runs fine on my local machine. The error only occurs when it is run from the docker container.

### How to reproduce

Rename .env.local.example to .env.local (this one is in .gitignore) and fill the environment variables.
Then run the docker container with the following commands:

`docker build -t authjs-response-not-conform`
`docker run --rm -p 3000:3000 --env-file .env.local authjs-response-not-conform`

### Expected behavior

No error I guess?

balazsorban44:
Fixes https://github.com/nextauthjs/next-auth/issues/9635
Closes https://github.com/nextauthjs/next-auth/pull/9718
Fixes #12041

Microsoft has a non-compliant provider and shows no signs of wanting to fix it 😮‍💨: https://github.com/MicrosoftDocs/azure-docs/issues/38427, https://github.com/MicrosoftDocs/azure-docs/issues/113944 

The current common-practice approach requires throwing the spec in the :wastebasket: and:

- For the discovery endpoint in general, fake the response by replacing `{tenantid}` with `common`
- Specifically for the token response, extract the `tid` claim from the `id_token` and manually override the `issuer` of the discovery metadata with a reconstructed `issuer`, this time replacing `{tenantid}` with the `tid` value. Technically another discovery request is then necessary to get the `issuer`.

In this PR, I am trying to find a way that would not require modifying the core library (https://github.com/panva/oauth4webapi/pull/131#issuecomment-2351806266) just for the sake of Microsoft, but as it seems, it is quite impossible with a generic API...


balazsorban44:
As of this comment, I managed to get things working, if the developer knows the `tenantId` when registering the provider in Auth.js libraries. Ideally, they would not need to do this though...

MJomaa:
From what I understand the PR adds a `conformInternal` boolean and if it is not conform, it tries to use a custom fetch depending on the non-conform provider.

Hmm what if you create a function like "createAuthorizationServer(...)` that has the non-conform fix in it for the issuer? Would that work? For example it could be applied here: https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/lib/actions/callback/oauth/callback.ts#L87C15-L87C23

edwardccg:
Hi @balazsorban44 

Sorry for replying in this particular PR, as this is relevant with this change it seems.
When trying to run the code inside Docker with a standalone build using the latest version of next-auth (`next-auth@5.0.0-beta.25`). I am getting following error.

```
[auth][error] TypeError: fetch failed
    at node:internal/deps/undici/undici:13178:13
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async [custom-fetch] (webpack-internal:///(rsc)/../../node_modules/@auth/core/providers/microsoft-entra-id.js:126:34)
    at async getAuthorizationUrl (webpack-internal:///(rsc)/../../node_modules/@auth/core/lib/actions/signin/authorization-url.js:25:35)
    at async Module.signIn (webpack-internal:///(rsc)/../../node_modules/@auth/core/lib/actions/signin/index.js:16:56)
    at async AuthInternal (webpack-internal:///(rsc)/../../node_modules/@auth/core/lib/index.js:76:24)
    at async Auth (webpack-internal:///(rsc)/../../node_modules/@auth/core/index.js:130:34)
```

This pointed to the `[custom-fetch]` code (line 126 on the microsoft-entra-id.js):
image

When trying to understand this logic, I found out the `tenantRe` might be incorrect? It matched a single word like `common`
image

However, in actual input, tenant id is a GUID. Following doesnt match, so `tenantId` are not able to extract
image

It's probably not the direct cause of the issue above as it work fine when i run the standalone code in my local machine but not on my Docker. Can you review and verify if the `tenantRe` are intended?

alamchrstn:
if I may add a dumb question here, am I correct to use `https://login.microsoftonline.com/{my-tenant-id` as the `issuer` property? not sure if this is related to @edwardccg 's issue above, but after bumping to `5.0.0-beta.25`, I got the following error
```
[auth][error] OperationProcessingError: "response" body "issuer" property does not match the expected value
```
previously I was using `5.0.0-beta.22` with `tenantId` config rather than `issuer`, and it worked fine. 🤔

packages/core/src/providers/azure-ad.ts
```diff
@@ -2,181 +2,29 @@
...
-    tenantId?: string
-  }
-): OAuthConfig<P> {
-  const { tenantId = "common", profilePhotoSize = 48, ...rest } = options
-  rest.issuer ??= `https://login.microsoftonline.com/${tenantId}/v2.0`
+export default function AzureAD(
+  config: Parameters<typeof MicrosoftEntraID>[0]
+): ReturnType<typeof MicrosoftEntraID> {
   return {
+    ...MicrosoftEntraID(config),
```
benhovinga, line 26: @balazsorban44 Azure AD should have been left alone and simply marked as deprecated. Turning Azure AD into a wrapper for Microsoft Entra ID breaks existing products that import the azure-ad module as the two have different parameters.

I am mainly referring to the `tenantID` vs `issuer` dilemma. The docs for both Azure AD and Microsoft Entra ID provide conflicting information for configuration.

PR #12017, opened 2024-10-09
Correct the type signature of `account` in callbacks.

w9:
## ☕️ Reasoning

The `account` argument can be `undefined` if it's not immediately after a sign-in. The current type signature does not reflect that.

## 🧢 Checklist

- [x] Documentation (Not applicable)
- [x] Tests
- [x] Ready to be merged

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -324,7 +324,7 @@ export interface AuthConfig {
      */
     signIn?: (params: {
       user: User | AdapterUser
-      account: Account | null | undefined
+      account?: Account | null
       /**
        * If OAuth provider is used, it contains the full
        * OAuth profile returned by your provider.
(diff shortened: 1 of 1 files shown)


copyleft21 (approved):

packages/core/src/index.ts
```diff
@@ -324,7 +324,7 @@ export interface AuthConfig {
      */
     signIn?: (params: {
       user: User | AdapterUser
-      account: Account | null
+      account: Account | null | undefined
```
balazsorban44, line 327:
```suggestion
      account?: Account | null
```

packages/core/src/index.ts
```diff
@@ -452,7 +452,7 @@ export interface AuthConfig {
        * Also includes {@link TokenSet}
        * @note available when `trigger` is `"signIn"` or `"signUp"`
        */
-      account: Account | null
+      account: Account | null | undefined
```
balazsorban44, line 455:
```suggestion
      account?: Account | null
```

packages/core/src/index.ts
```diff
@@ -497,7 +497,7 @@ export interface AuthConfig {
      */
     signIn?: (message: {
       user: User
-      account: Account | null
+      account: Account | null | undefined
```
balazsorban44, line 500:
```suggestion
      account?: Account | null
```

balazsorban44 (changes requested):

PR #12051, opened 2024-10-16
🐛: User id property is not honoured from the OAuth profile() callback

richbayliss:
- set the id property to come from the callback result OR fallback to a random UUID if it's missing.

## ☕️ Reasoning

I am unable to use the `id` property from the result of the OAuth `profile()` callback.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/actions/callback/oauth/callback.ts
+++ b/packages/core/src/lib/actions/callback/oauth/callback.ts
@@ -267,7 +267,7 @@ export async function getUserAndAccount(
     const userFromProfile = await provider.profile(OAuthProfile, tokens)
     const user = {
       ...userFromProfile,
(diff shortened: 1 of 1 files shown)


balazsorban44:
Thanks, but this is on purpose. The user id should remain automatically generated, instead of being a random providers' id. This is more prominent when you have multiple social providers set up.

However, we might be open to have a method that can override the ID method, as not everyone is happy with the UUID being forced here.

Feel free to open a PR for that!

balazsorban44:
Potentially it should be a method on the `Adapter` interface.

https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/adapters.ts/#L282

PR #12055, opened 2024-10-16
Description of error when using 'Credentials' Provider

ynng3:
## ☕️ Reasoning
If you use 'Credentials' Provider, you will get an error due to the security policy of auth.js. You must set the 'strategy' option, which is a sub-option of the 'session' option, to 'jwt'.

```ts
ExpressAuth({
  session: {
    strategy: "jwt"
  },
  providers: [...]
})
```

I wasted 5 hours due to the unfriendly error description and the Docs not describing the information in detail.

I thought it would be good to report this error at the stage of using Credentials, so I decided to make a Pull Request.

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/docs/pages/getting-started/providers/credentials.mdx
+++ b/docs/pages/getting-started/providers/credentials.mdx
@@ -110,6 +110,18 @@ app.use("/auth/*", ExpressAuth({
   ],
 });
 ```
+    <Callout type="warning">
+      If you use 'Credentials' Provider, you will get an error due to the security policy of auth.js. You must set the 'strategy' option, which is a sub-option of the 'session' option, to 'jwt'.
+```ts
+ExpressAuth({
+  session: {
+    strategy: "jwt"
+  },
+  providers: [...]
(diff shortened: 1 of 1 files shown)


balazsorban44:
@ynng3 could you please explain the intention a bit better? I'm not sure I understand fully. With the Credentials provider the session strategy being set to `database` should already throw the following error https://github.com/nextauthjs/next-auth/blob/a7491dcb9355ff2d01fb8e9236636605e2090145/packages/core/src/lib/utils/assert.ts#L194

If that's not the case, could you pleae open a bug report instead?

Or maybe if you get this error, but the action is unclear, maybe we could make the error message itself better instead.

We have no information related to the Credentials provider in our security policy https://authjs.dev/security

We do state however that you should be persisting passwords in a database with caution on our https://authjs.dev/getting-started/authentication/credentials

PR #12063, opened 2024-10-18
chore: optimize packages to reduce bundle size

halvaradop:
## ☕️ Reasoning
This pull request aims to reduce the bundle size of the `@auth/core` package. I noticed that the package size is larger than necessary due to the inclusion of `.ts` files in the build process. These files are compiled into `.d.ts`, `.js`, and `.d.ts.map` files, which already contain the codebase, leading to duplication and increased package size.

While including `.ts` files can help users understand the code structure when navigating through files, it can also make contributing to the package more cumbersome due to the increased size and clutter in the IDE.

Here are the key points regarding the changes:

### Current Bundle Size:
- Size: 1.65MB
- Disk Size: 2.50MB
### Bundle Size After Removing .ts Files:
- Size: 1.02MB
- Disk Size: 1.67MB

### Dry Run Comparison
Now, the difference in size executing `pnpm publish --dry-run --no-git-checks` is shown in the images below:

### Now
dry-run-unoptimized

### After
dry-run-optimized

These changes not only improve the size of the package but also enhance the organization of compiled files. Although this pull request primarily focuses on the @auth/core package, the intent is to implement similar optimizations across all packages supported by Auth.js.

I would like to discuss this approach further, as I believe it will benefit contributors and users alike.
 
> [!WARNING]
 I have noticed that these changes caused some errors, which are expected because the export files are now different from what is expected by other packages. I will be actively working on resolving these issues.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -26,11 +26,7 @@
   "type": "module",
   "types": "./index.d.ts",
   "files": [
-    "*.js",
-    "*.d.ts*",
-    "lib",
-    "providers",
-    "src"
+    "dist"
   ],
   "exports": {
     ".": {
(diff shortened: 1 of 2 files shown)


balazsorban44:
Thanks, but this is intentional! Together with the `.d.ts.map` files, this let's developers jump straight to the real source code from their editor.

See https://www.typescriptlang.org/tsconfig/#declarationMap

PR #11994, opened 2024-10-07
chore: bump to oauth4webapi@3

panva:
Bumps the oauth4webapi dependency to [v3.x](https://github.com/panva/oauth4webapi/releases/tag/v3.0.0) while 

- maintaining support for insecure HTTP requests (marking TODOs where applicable)
- keeping client_secret_basic as the default client auth method with the incomplete charset encoding of oauth4webapi@2 (marking TODOs where applicable)
- keeping multi-audience on private_key_jwt assertions (marking TODOs where applicable)

PR #12066, opened 2024-10-19
fix(core): adds 'none' token endpoint auth method for oauth and oidc

JonnyBGod:
https://github.com/nextauthjs/next-auth/pull/11994 introduced an undocumented breaking change by removing support for OAuth Token Endpoint Authentication Methods ```"none"```.

``` ts
{
  type: 'oidc',
  client: {
    token_endpoint_auth_method: 'none',
  }
}
```

This restores compatibility.

### DIFF
--- a/packages/core/src/lib/actions/callback/oauth/callback.ts
+++ b/packages/core/src/lib/actions/callback/oauth/callback.ts
@@ -118,6 +118,9 @@ export async function handleOAuth(
         },
       })
       break
+    case "none":
+      clientAuth = o.None()
+      break
     default:
       throw new Error("unsupported client authentication method")
   }


balazsorban44:
thanks @JonnyBGod. Out of curiosity. Are you working with a provider that only supports this method? 🤔 Sounds not ideal.

panva:
> thanks @JonnyBGod. Out of curiosity. Are you working with a provider that only supports this method? 🤔 Sounds not ideal.

Exactly the reason why "none" wasn't included in the case switch, there is no use case for public clients on a server-side framework.

JonnyBGod:
Some leagacy backend compatibility, inititally thought it would be a wast of time improving it at this point. But actually tooks some time last night to make it work with ```client_secret_basic``` in the end.

Was going to close the PR but figured that since it aparently is part of the spec and someone else might have the same problem it should be better to support it for now.

PR #12199, opened 2024-11-08
Allow setting custom cookie name instead of "authjs"

paramaggarwal:
## ☕️ Reasoning

Currently the cookie is always `authjs` by default - this exposes the implementation detail of the cookie and advanced usecases should allow customising the name of the cookie to prevent malicious users from introspecting implementation details of the authentication mechanism used by a particular website.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -583,6 +583,18 @@ export interface AuthConfig {
    */
   useSecureCookies?: boolean
   /**
+   * When set the cookie name will use this as it's name. For example, the session cookie will be 
+   * of the form `mycookiename.session-token` or `__Secure-mycookiename.session-token`.
+   * The exact full name of the cookie will change based on other cookie-related options.
+   *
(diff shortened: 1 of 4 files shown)


balazsorban44:
Thanks, but this is already possible via the `config.cookies` option, we could rather improve its documentation. See: https://authjs.dev/reference/core#cookies and https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/index.ts/#L585-L599

paramaggarwal:
Thanks, I'm aware of the `config.cookies` option - but it requires overriding every single cookie name separately and also handle the `__Secure` or `__Host` prefix addition when needed. My goal was to give a simple setting that allowed one to update just the main `authjs` part of the cookie name. This would be the most common usecase if you think about it.

We could also move my suggestion inside the `config.cookies` setting as `config.cookies.name`?

PR #12249, opened 2024-11-16
fix: improve client-side submodules

balazsorban44:
## ☕️ Reasoning

1. Improved provider id inference in `signIn` method.
 - Generate known email provider id types
 - simplify inference by dropping `LiteralUnion` type
3. Allow `redirect: false` for OAuth providers in client-side `signIn`. We already allowed this for the server-side version. In some cases, you might just want the URL before redirecting via `window.location.href`, eg. you want to use a popup window to initiate the signin.
4. Improved return type of `signIn` and `signOut` methods. Now, unless `redirect: false` is explicitly added, these methods will return `void`.
5. throw proper errors when WebAuthn signin is initiated from the wrong import. This gives early feedback to the developer
7. Unify API across framework integrations with client-side methods. (Looks like there is a lot of overlap, that we could consolidate via adding `@auth/core/client`)

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


balazsorban44:
| Test |
|------|
| <a href=example.com target=_blank>New tab</a> |
| <a href=example.com>Same tab</a> |

sinomsinom:
These changes seem to have changed the sveltkit signin and signout endpoints to `{basePath}/signout` and `{basePath}/signin` instead of the previous `{basePath}/auth/signout` and `{basePath}/auth/signin` however the handler was not changed to reflect this change. Because of this existing code using the @auth/sveltekit package doesn't work with version 1.8.0 that merged this pull request

packages/frameworks-sveltekit/src/lib/client.ts
```diff
@@ -39,79 +54,122 @@ export type SignInAuthorizationParams =
...
-  const basePath = base ?? ""
-  const signInUrl = `${basePath}/auth/${
-    isCredentials ? "callback" : "signin"
-  }/${providerId}`
+  const baseUrl = base ?? ""
 
-  const _signInUrl = `${signInUrl}?${new URLSearchParams(authorizationParams)}`
+  const signInUrl = `${baseUrl}/${
+    provider === "credentials" ? "callback" : "signin"
+  }/${provider}`
```
sillvva, line 91: You are missing the `/auth` part of the path. This breaks the SvelteKit hooks.

PR #12381, opened 2024-12-14
docs(pg-adapter): add unique and foreign key constraints to the PG adapter schema

baptiste00001:
## ☕️ Reasoning

The contraints in the PG adapter schema are non existant while they are defined in all other adapters. I added them to the schema.

## 🧢 Checklist

- [X] Documentation
- [X] Tests
- [X] Ready to be merged


ndom91 (approved):
Nice, thanks!

ndom91:
Ah wait, we may need to remove the unique constraint on email. Will that fail if we try to use `null` for multiple users `email`?

baptiste00001:
> Ah wait, we may need to remove the unique constraint on email. Will that fail if we try to use `null` for multiple users `email`?

It works fine with `null` value for multiple users `email` because there is no NULL NOT DISTINCT after the UNIQUE constraint. 
So two `null` values are considered distinct.
I just tested it.

ndom91:
Approved the test run from fork to ensure the pg schema still works. It ran into a few prettier issues. I know one isn't from a file in your PR, but would you mind running prettier over these here? :pray: 

See: https://github.com/nextauthjs/next-auth/actions/runs/12443333045/job/34755721579?pr=12381

baptiste00001:
> I approved the test run from fork to ensure the pg schema still works. It ran into a few prettier issues. ~I know one isn't from a file in your PR, but would you mind running prettier over these here? 🙏~
> 
> ~See: https://github.com/nextauthjs/next-auth/actions/runs/12443333045/job/34755721579?pr=12381~
> 
> EDIT: Neverind, fixed it [here](https://github.com/nextauthjs/next-auth/pull/12414)

Oh OK
I actually fixed it as well and commited.
It was a bunch of extra spaces that needed to be removed in two files.
I hope you can merge the PR now.
If you need anything else just ask me.

PR #12401, opened 2024-12-18
feat(prisma-adaper): update `PrismaAdapter` options to accept optional custom model names

ogroppo:
## ☕️ Reasoning

If there are more tables with the same name e.g. `User` there is no way to work around it.
This is super typical if there are 2 auth systems in the same DB example platform users and admin users...

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [x] Ready to be merged

## 📌 Resources


packages/adapter-prisma/src/index.ts
```diff
@@ -15,114 +15,130 @@
...
 
 export function PrismaAdapter(
-  prisma: PrismaClient | ReturnType<PrismaClient["$extends"]>
+  prisma: PrismaClient | ReturnType<PrismaClient["$extends"]>,
+  options?: {
+    accountModel?: string;
+    authenticatorModel?: string;
+    sessionModel?: string;
+    userModel?: string;
+    verificationTokenModel?: string;
```
ogroppo, line 33: just allow users to override the model name

packages/adapter-prisma/src/index.ts
```diff
@@ -15,114 +15,130 @@
...
+  }
 ): Adapter {
-  const p = prisma as PrismaClient
+  const p = prisma as PrismaClient;
+  const accountModel = p[options?.accountModel ?? "account"];
+  const authenticatorModel = p[options?.authenticatorModel ?? "authenticator"];
+  const sessionModel = p[options?.sessionModel ?? "session"];
+  const userModel = p[options?.userModel ?? "user"];
```
ndom91, line 40: Looks like there's some TS issues here. See: https://github.com/nextauthjs/next-auth/actions/runs/12449775424/job/34755845781?pr=12401

ogroppo: Not sure how to set up TS locally correctly, having a bunch of errors namely all args of methods have implicit any but I am not overriding the default tsconfig file afaik 🤔 
Screenshot 2024-12-22 at 23 50 04

ogroppo:
Can someone look into this please, it's a highly requested feature, and also necessary in most cases!

packages/adapter-prisma/src/index.ts
```diff
@@ -25,53 +25,75 @@ import type {
...
+  const authenticatorModel =
+    p[(options?.authenticatorModel ?? "authenticator") as keyof PrismaClient]
+  const sessionModel =
+    p[(options?.sessionModel ?? "session") as keyof PrismaClient]
+  const userModel = p[(options?.userModel ?? "user") as keyof PrismaClient]
+  const verificationTokenModel =
+    p[
+      (options?.verificationTokenModel ??
+        "verificationToken") as keyof PrismaClient
+    ]
```
WikiRik, line 49:
```suggestion
  const accountModel = p[
    (options?.accountModel ?? "account") as keyof PrismaClient
  ] as Prisma.AccountDelegate
  const authenticatorModel = p[
    (options?.authenticatorModel ?? "authenticator") as keyof PrismaClient
  ] as Prisma.AuthenticatorDelegate
  const sessionModel = p[
    (options?.sessionModel ?? "session") as keyof PrismaClient
  ] as Prisma.SessionDelegate
  const userModel = p[
    (options?.userModel ?? "user") as keyof PrismaClient
  ] as Prisma.UserDelegate
  const verificationTokenModel = p[
    (options?.verificationTokenModel ??
      "verificationToken") as keyof PrismaClient
  ] as Prisma.VerificationTokenDelegate
```

This would get rid of the TypeScript errors on the schema that is used for the build (and tests). But TypeScript will give errors if you actually use the options for different models. You need to find some kind of generic Delegate.

I would recommend starting with adding a custom schema + tests first and then try to fix the TypeScript issues after

packages/adapter-prisma/src/index.ts
```diff
@@ -25,53 +25,75 @@ import type {
...
+      userModel.update({
         where: { id },
         ...stripUndefined(data),
       }) as Promise<AdapterUser>,
     deleteUser: (id) =>
-      p.user.delete({ where: { id } }) as Promise<AdapterUser>,
+      userModel.delete({ where: { id } }) as Promise<AdapterUser>,
     linkAccount: (data) =>
-      p.account.create({ data }) as unknown as AdapterAccount,
+      accountModel.create({ data }) as unknown as AdapterAccount,
```
WikiRik, line 70:
```suggestion
      accountModel.create({ data }) as Promise<AdapterAccount>,
```
With the changes above, you can also do this change iirc

WikiRik (commented):
Maybe these suggestions and tips might help you improve this. Not sure how widely requested it is, but it's nice to have the ability to change model names

ogroppo:
Hi @WikiRik - thank you for reviewing this! I don't feel comfortable enough to be able to go through the checklist and adding a custom schema. Any chance you could take over, or know someone that could?
I keep seeing developers copy/pasting the whole script because of the missing options 😱 - I really think the original package should be used! 🙏🏻 Appreciate

PR #12421, opened 2024-12-25
Use host and proto headers to determine request origin if AUTH_URL is not set

bfichter:
## ☕️ Reasoning

The v5 docs state that `AUTH_URL` "is mostly unnecessary with v5 as the host is inferred from the request headers," but this is only the case in some areas of `@auth/core`, and the `url` of the request is often the fallback for determining URLs in the framework. But in the case that the app is running behind a reverse proxy or in a number of other networking setups, requests to `@auth/core` from `next-auth` won't have a relevant url (`0.0.0.0` observed in our use case) and authentication will fail in a number of ways without `AUTH_URL` set.

This change checks request headers (if they're set) to determine the url to send to `@auth/core`, in the case that `AUTH_URL` or `NEXTAUTH_URL` are not set. This properly cuts reliance on `AUTH_URL` (in our use case) and also as a bonus enables multiple authentication urls to be used in a given system.

Possibly this behavior should be behind some config if we want to make it opt in. If so I'm happy to add that.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/lib/env.ts
+++ b/packages/next-auth/src/lib/env.ts
@@ -3,9 +3,12 @@ import { NextRequest } from "next/server"
 import type { NextAuthConfig } from "./index.js"
 import { setEnvDefaults as coreSetEnvDefaults } from "@auth/core"
 
-/** If `NEXTAUTH_URL` or `AUTH_URL` is defined, override the request's URL. */
+/** If `NEXTAUTH_URL` or `AUTH_URL` is defined or host headers are set,
+ * override the request's URL.
+ */
 export function reqWithEnvURL(req: NextRequest): NextRequest {
-  const url = process.env.AUTH_URL ?? process.env.NEXTAUTH_URL
+  const url =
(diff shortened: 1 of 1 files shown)


balazsorban44:
Thanks, but I think this should already be the case, as this code is in `@auth/core`:

https://github.com/nextauthjs/next-auth/blob/7eb36c137ff964de4c2983dc1644388c1d329e5b/packages/core/src/lib/utils/env.ts#L94-L101

You might need to enable this behavior by adding `trustHost: true` though, as we don't want to do it by default, in case your hosting provider doesn't sanitize these headers.

See: https://authjs.dev/getting-started/deployment#self-hosted

bfichter:
@balazsorban44 we use `trustHost: true` but do not observe the url being constructed from the headers without this change.

balazsorban44:
Would love to have a look in a bug report, maybe with some details on the method of deploying as well. 🙏

PR #12434, opened 2024-12-28
fix(prisma): replace select with include for user relation query

mcorbelli:
This PR fixes an issue where Prisma's `select` statement was incorrectly used to query the `user` relation in the `Account` model. Since `select` is only applicable for scalar fields, this caused Prisma to throw an error:  
`Unknown field 'user' for select statement on model 'Account'`. 

The fix replaces `select: { user: true }` with `include: { user: true }`, which correctly fetches the `user` relation and resolves the issue.


ndom91 (approved):
Hmm yeah that's definitely not right. I guess we didn't require the result of that function for anything critical 😅 

Tests pass, so LGTM 🙏

mcorbelli:
When will it be included in the npm release? Is there a date yet?

mcorbelli:
When will it be included in the npm release? Is there a date yet? @ndom91

ndom91:
I don't have a date or even a guesstimate unfortunately. All I can say is that it'll be a part of the next release 😅

mcorbelli:
@balazsorban44 Are there any updates regarding the upcoming release with this fix?

mcorbelli:
news?

PR #12475, opened 2025-01-07
Feat/add gitee provider

wushide:
## ☕️ Reasoning

This will add a login method to Gitee. I have already resolved the type issue, as there was previously a type gap.

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources


packages/core/src/providers/gitee.ts
```diff
@@ -0,0 +1,94 @@
...
+  login: string;
+  email?: string;
+  avatar_url?: string;
+}
+
+export default function Gitee(
+  config: OAuthUserConfig<GiteeProfile>
+): OAuthConfig<GiteeProfile> {
+  const baseUrl = "https://gitee.com";
+  const apiBaseUrl = "https://gitee.com/api/v5";
```
halvaradop, line 25: It is not necessary to use these two variables, since `apiBaseUrl` only adds the path `/api/v5` to baseUrl

packages/core/src/providers/gitee.ts
```diff
@@ -0,0 +1,94 @@
...
+  return {
+    id: "gitee",
+    name: "Gitee",
+    type: "oauth",
+    authorization: {
+      url: `${baseUrl}/oauth/authorize`,
+      params: {
+        scope: "",
+      },
+    },
```
halvaradop, line 36: If no additional configuration is needed, there’s no need to use the object. Simply use the API route directly.

packages/core/src/providers/gitee.ts
```diff
@@ -0,0 +1,94 @@
...
+    token: {
+      url: `${baseUrl}/oauth/token`,
+      params: {
+        grant_type: "authorization_code",
+      },
+    },
```
halvaradop, line 42: The `grant_type` is set to `authorization_code` by default, so there’s no need to use the object.

halvaradop (commented):
Great work! Just a few minimal changes. Additionally, could you provide documentation about the introduced provider, such as the APIs used or more resources on how to use the provider correctly?

PR #12505, opened 2025-01-15
Add workos authkit provider

fierysolid:
## ☕️ Reasoning

WorkOS has added their own IDP in the form of AuthKit and I'd like to use it with next-auth

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources


packages/core/src/providers/authkit.ts
```diff
@@ -0,0 +1,83 @@
...
+        }
+        return Response.json({ ...data, token_type: "bearer" }, res)
+      },
+    },
+    client: {
+      token_endpoint_auth_method: "client_secret_post",
+    },
+    userinfo: {
+      url: "https://api.workos.com/user_management/users",
+      async request({ tokens, provider }) {
```
fierysolid, line 60: The `/user_management/authenticate` actually returns the user profile at the same time, so you could in theory just immediately return `tokens.user`, but I assumed there may be other places where the "userinfo" endpoint gets called to get an updated profile.

packages/core/src/providers/authkit.ts
```diff
@@ -0,0 +1,84 @@
...
+  object: string
+  id: string
+  email: string
+  email_verified: boolean
+  first_name: string
+  last_name: string
+  profile_picture_url: string | null
+  created_at: string
+  updated_at: string
+}
```
halvaradop, line 19: According to the WorkOS documentation, the properties `first_name`, `last_name`, and `profile_picture_url` are optional. Therefore, sometimes these attributes may not exist, which can cause errors in the `profile` function since these properties, [please read](https://workos.com/docs/reference/user-management/user)

https://github.com/nextauthjs/next-auth/blob/66fbe22bdf37bbd76a0cb5cfb249d722028ca80b/packages/core/src/providers/authkit.ts#L73-L80

halvaradop (commented):
I’ve read the documentation, but I’m confused about the focus of this provider. According to their explanation:

> If you are using `AuthKit`, set the provider parameter to `authkit`, which will generate an authorization URL for your AuthKit domain. AuthKit will take care...
Otherwise, to generate an authorization URL for a WorkOS SSO connection, you’ll need to specify the user’s connection, organization, or OAuth provider as a parameter.

Based on this, if the user selects an `authkitProvider` value other than `authkit`, do they need to set both the `connection_id` and `organization_id`? If so, these values don’t seem to be available in the provider options for configuration.

On the other hand, `Auth.js` already includes a provider called `WorkOs`. What is the difference between this and the one mentioned in the documentation? Do you know if this API is still functional?

Thanks for your time!

packages/core/src/providers/authkit.ts
halvaradop: Great reasoning. I think it’s best to wait for an official response from a maintainer regarding how the userinfo property/request works.

fierysolid:
> I’ve read the documentation, but I’m confused about the focus of this provider. According to their explanation:
> 
> > If you are using `AuthKit`, set the provider parameter to `authkit`, which will generate an authorization URL for your AuthKit domain. AuthKit will take care...
> > Otherwise, to generate an authorization URL for a WorkOS SSO connection, you’ll need to specify the user’s connection, organization, or OAuth provider as a parameter.
> 
> Based on this, if the user selects an `authkitProvider` value other than `authkit`, do they need to set both the `connection_id` and `organization_id`? If so, these values don’t seem to be available in the provider options for configuration.
> 
> On the other hand, `Auth.js` already includes a provider called `WorkOs`. What is the difference between this and the one mentioned in the documentation? Do you know if this API is still functional?
> 
> Thanks for your time!

WorkOS has an SSO passthrough and they also now have an IDP. The WorkOS provider is for the SSO only API, and this provider is for the IDP + SSO API which uses different URLs.

packages/core/src/providers/authkit.ts
fierysolid: Good catch, I'll update the types and profile.

halvaradop:
Thank you for taking the time to provide a response. The differences and improvements applied to the new provider compared to the older one make perfect sense. 🤠

packages/core/src/providers/authkit.ts
fierysolid: I went ahead and made those properties `string | null | undefined` because the docs say `string | undefined`, but the API returns `string | null`. This covers all bases. Additionally, in the profile function I use a waterfall approach to generate `name` from `first_name` and `last_name`.

JohnRSim:
Any updates with this - also notices {issuer} had been replaced with direct url.

fierysolid:
> Any updates with this - also notices {issuer} had been replaced with direct url.

Nope, just sitting here waiting...

JohnRSim:
@fierysolid Just tested on my workos instance 
- all great only 1 comment not sure if you checked in the svg for the auth signin page - 

https://authjs.dev/img/providers/authkit.svg

fierysolid:
@JohnRSim we should just reuse the WorkOS svg. I'll change it so it references it.

fierysolid:
Okay, SVG added.

PR #12608, opened 2025-02-05
Downgrade requested OAuth scope of TikTok provider

and-oli:
## ☕️ Reasoning

The provider is currently requesting a token with `user.info.profile` scope, which is more restrictive than `user.info.basic`. The latter is the one provided by default by the TikTok Login kit and as such, should be the one used for authentication purposes with TikTok. The user info requested by the provider is also covered by the `user.info.basic` scope. For this reason, this PR changes the requested scope to `user.info.basic`.

See [TikTok's documentation on scopes needed for user info](https://developers.tiktok.com/doc/tiktok-api-v2-get-user-info)
## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/src/providers/tiktok.ts
+++ b/packages/core/src/providers/tiktok.ts
@@ -293,7 +293,7 @@ export default function TikTok(
       url: "https://www.tiktok.com/v2/auth/authorize",
       params: {
         client_key: options.clientId,
-        scope: "user.info.profile",
+        scope: "user.info.basic",
       },
     },
 
--- a/packages/core/src/providers/tiktok.ts
+++ b/packages/core/src/providers/tiktok.ts
@@ -299,14 +299,15 @@ export default function TikTok(
 
     token: "https://open.tiktokapis.com/v2/oauth/token/",
     userinfo:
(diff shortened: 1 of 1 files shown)


balazsorban44 (approved):

DewinU:
Hey @and-oli 
Probably late to the party, should this be the default behavior tho, to not add the workaround (username field in email with scope `user.info.profile` for the email field?

Heres some [reference](https://github.com/better-auth/better-auth/pull/397) from another library where i implemented it the same way

and-oli:
> Hey @and-oli Probably late to the party, should this be the default behavior tho, to not add the workaround (username field in email with scope `user.info.profile` for the email field?
> 
> Heres some [reference](https://github.com/better-auth/better-auth/pull/397) from another library where i implemented it the same way
hey @DewinU ! not sure I'm following what you mean. The workaround also felt weird as storing a username in an email field can easily cause runtime breakages down the road. I think it's more appropriate to be explicit about the lack of email (hence the `null`). The `user.info.basic` scope provides a `display_name` which can be used by the user to identify the account in question.

Issue #9050, opened 2023-11-03
Fusion Auth with next-auth@beta invalid url error


cezarneaga:
### Provider type

FusionAuth

### Environment

System:
OS: macOS 14.0
CPU: (8) arm64 Apple M2
Memory: 112.03 MB / 16.00 GB
Shell: 5.9 - /bin/zsh
Binaries:
Node: 18.16.1 - ~/Library/Caches/fnm_multishells/43693_1698930429983/bin/node
Yarn: 1.22.19 - /opt/homebrew/bin/yarn
npm: 9.5.1 - ~/Library/Caches/fnm_multishells/43693_1698930429983/bin/npm
pnpm: 8.6.12 - ~/Library/pnpm/pnpm
bun: 1.0.0 - /opt/homebrew/bin/bun
Browsers:
Chrome: 119.0.6045.105
Safari: 17.0

### Reproduction URL

https://github.com/nextauthjs/next-auth-example

### Describe the issue

Hi there, 

```
[auth][warn][debug-enabled] Read more: [https://warnings.authjs.dev#debug-enabled](https://warnings.authjs.dev/#debug-enabled)
[auth][error][SignInError]: Read more at [https://errors.authjs.dev#signinerror](https://errors.authjs.dev/#signinerror)
[auth][cause]: TypeError [ERR_INVALID_URL]: Invalid URL
at new NodeError (node:internal/errors:399:5)
at new URL (node:internal/url:560:13)
at Module.processDiscoveryResponse (webpack-internal:///(rsc)/../../node_modules/oauth4webapi/build/index.js:272:9)
at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
at async getAuthorizationUrl (webpack-internal:///(rsc)/../../node_modules/@auth/core/lib/oauth/authorization-url.js:24:20)
at async Module.signin (webpack-internal:///(rsc)/../../node_modules/@auth/core/lib/routes/signin.js:22:20)
at async AuthInternal (webpack-internal:///(rsc)/../../node_modules/@auth/core/lib/index.js:160:36)
at async Auth (webpack-internal:///(rsc)/../../node_modules/@auth/core/index.js:121:30)
at async .../node_modules/next/dist/compiled/next-server/app-route.runtime.dev.js:6:61856
[auth][details]: {
"provider": "fusionauth"
}
```
tried to debug it but no success.
maybe it clicks for someone else here. 

thanks,
Cezar

### How to reproduce

[...]

PR #12611, opened 2025-02-05
fix: More descriptive error in case that discovery endpoint returns an invalid issuer

falcowinkler:
## ☕️ Reasoning

From https://www.rfc-editor.org/rfc/rfc8414.html#section-2:

> issuer
      REQUIRED.  The authorization server's issuer identifier, which is
      a URL that uses the "https" scheme and has no query or fragment
      components. 
      
However, some providers like fusion-auth allow a non-conformant issuer, e.g. without the "https" scheme.
If such an issuer is configured on the authorization server, we see the cryptic "Invalid URL" error, that some people get [confused about](https://github.com/nextauthjs/next-auth/issues/9050). 

The solution is to throw a more descriptive error here. Alternatively, we could open a PR to oauth4webAPI to throw a more descriptive error [here](https://github.com/panva/oauth4webapi/blob/6caa822b9edb5374524c6c3c93b44ff16e70e2da/src/index.ts#L1328).

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/actions/signin/authorization-url.ts
+++ b/packages/core/src/lib/actions/signin/authorization-url.ts
@@ -33,8 +33,11 @@ export async function getAuthorizationUrl(
     const as = await o
       .processDiscoveryResponse(issuer, discoveryResponse)
       .catch((error) => {
-        if (!(error instanceof TypeError) || error.message !== "Invalid URL") throw error
-        throw new TypeError(`Discovery request responded with an invalid issuer. expected: ${issuer}`)
+        if (!(error instanceof TypeError) || error.message !== "Invalid URL")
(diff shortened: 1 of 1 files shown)


balazsorban44 (approved):

balazsorban44:
I think the root cause is because we are swallowing the error properties here: 
https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/lib/utils/logger.ts/#L29-L48

ie. there should be an `input` property that shows the incorrect URL
image


So that's where we also need to make sure we log everything, but this is a great hint to the user!

PR #12634, opened 2025-02-09
fix(docs): svelte 5 compatibility updates 

nicolecomputer:
## ☕️ Reasoning

These were changes I needed to make while working through the docs with a new Svelte 5 app.

## 🧢 Checklist

- [X] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/docs/pages/getting-started/adapters/drizzle.mdx
+++ b/docs/pages/getting-started/adapters/drizzle.mdx
@@ -55,7 +55,7 @@ import {
 } from "drizzle-orm/pg-core"
 import postgres from "postgres"
 import { drizzle } from "drizzle-orm/postgres-js"
-import type { AdapterAccountType } from "next-auth/adapters"
+import type { AdapterAccountType } from "@auth/core/adapters";
 
 const connectionString = "postgres://postgres:postgres@localhost:5432/drizzle"
 const pool = postgres(connectionString, { max: 1 })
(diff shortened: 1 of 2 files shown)


docs/pages/getting-started/adapters/drizzle.mdx
```diff
@@ -55,7 +55,7 @@ import {
 } from "drizzle-orm/pg-core"
 import postgres from "postgres"
 import { drizzle } from "drizzle-orm/postgres-js"
-import type { AdapterAccountType } from "next-auth/adapters"
+import type { AdapterAccountType } from "@auth/core/adapters";
```
ndom91, line 59: You should never have to install / interact with `@auth/core` as a consumer of these libraries. 

Could you not get the `AdapterAccountType` type from `@auth/sveltekit/adapters`? If not, then that's a bug that should be fixed - let me know 🙏

ndom91 (changes requested):
One change requested, but otherwise looks good for Svelte 5 compat 🙏

docs/pages/getting-started/adapters/drizzle.mdx
nicolecomputer: Nope, it's not there. Here's the error message: `Cannot find module '@auth/sveltekit/adapters' or its corresponding type declarations.ts(2307)`

From my package.json:

```
    "@auth/sveltekit": "^1.7.4",
```

nicolecomputer: I filed this as a bug: https://github.com/nextauthjs/next-auth/issues/12649

nicolecomputer:
IT seems like isn't needed/wanted right now. I'm going to close soon unless someone pipes up to support this

PR #12653, opened 2025-02-14
fix(next-auth): provide request context in lazy initialization for `auth()` calls in RSCs

huboh:
- add createRequestFromHeaders utility for consistent request creation
- pass constructed Request to config function instead of undefined
- enable request-based config customization and runtime modifications for `auth()` calls in RSCs, matching other contexts
- support documented use cases like env-specific provider configuration
- allow header/request access during server-side `auth()` calls

## ☕️ Reasoning

Lazy initialization is documented to provide request context access for customizing configuration, with examples like:
- Accessing request data during server-side auth() calls
- Adding different providers based on staging/dev environments
- Customizing configuration and runtime modifications based on request headers 

However, RSCs were receiving undefined instead of the request context, making these customizations impossible in RSC contexts. This fix makes `auth()` call in RSCs work consistently with other contexts (middleware, route handlers) and enables the documented use cases.

Example use case that was broken but now works:
```typescript
const { auth } = NextAuth(
  async (req) => {
    if (req) {
      // This now works for `auth()` calls in RSCs too
      const headers = req?.headers
      const isStaging = headers?.get('host')?.includes('staging')
    }
  
    return {
      providers: [
        isStaging 
          ? StagingProvider() 
          : ProductionProvider()
      ]
    }
})
```

### Previous Behavior

```typescript
// RSCs received undefined
const _config = await config(undefined)
```

### New Behavior

```typescript
// RSCs now receive a proper Request object
const _request = createRequestFromHeaders(_headers)
const _config = await config(_request)
```

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [x] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/next-auth/src/index.ts
+++ b/packages/next-auth/src/index.ts
@@ -46,7 +46,7 @@
  * If you need to override the default values for a provider, you can still call it as a function `GitHub({...})` as before.
  *
  * ## Lazy initialization
- * You can also initialize NextAuth.js lazily (previously known as advanced intialization), which allows you to access the request context in the configuration in some cases, like Route Handlers, Middleware, API Routes or `getServerSideProps`.
+ * You can also initialize NextAuth.js lazily (previously known as advanced intialization), which allows you to access the request context in the configuration in some cases, like Route Handlers, React Server Components, Middleware, API Routes or `getServerSideProps`.
  * The above example becomes:
  *
  * ```ts title="auth.ts"
--- a/packages/next-auth/src/lib/index.ts
+++ b/packages/next-auth/src/lib/index.ts
@@ -117,6 +117,19 @@ function isReqWrapper(arg: any): arg is NextAuthMiddleware | AppRouteHandlerFn {
   return typeof arg === "function"
 }
 
+/**
+ * Creates a Request object from Headers, primarily for server-side contexts
+ * where we need to reconstruct request information.
+ */
+function createRequestFromHeaders(headers: Headers): Request {
+  const url = headers.get("x-url") ||
+    `${headers.get("x-forwarded-proto") || "http"}://${headers.get("host")}`;
+
+  return new Request(url, {
+    headers: new Headers(headers)
+  })
+}
+
 export function initAuth(
   config:
     | NextAuthConfig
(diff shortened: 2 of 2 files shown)


huboh:
Just wanted to share this while waiting for the merge, I had to implement a workaround in my company by forcing `auth()` calls in RSCs to go through the getServerSideProps/API route path instead of the RSC path. I do this by constructing and passing a fake context when `auth()` is called with no parameters:

```typescript
const { handlers, signIn, signOut, auth } = NextAuth(
  async (req) => {
    if (req) {
      // code here will have the request context for `auth()` calls inside RSCs
      // in addition with old behaviour inside route handlers, middleware and api routes or getServerSideProps
    } else {
      // no request context for `signIn()`, `signOut()` and `unstable_update()` calls...No need for that ? 
    }

    return {
      ...
    }
  }
);

const auth_ = async (...args: Parameters<typeof auth> | []) => {
  if (args.length) {
    return (
      auth(...args)
    );

  } else {
    // `auth()` was called inside rsc and lazy init function will now have request context
    const heads = new Headers(await headers());
    const reqUrl = (
      (heads.get("x-url")) ||
      (`${heads.get("x-forwarded-proto") || "http"}://${heads.get("host")}`)
    );
    const context: GetServerSidePropsContext = {
      resolvedUrl: reqUrl,
      query: {}, // not used by original `auth()`, so no need.
      req: (
        new Request(reqUrl, { headers: heads }) as unknown as GetServerSidePropsContext["req"]
      ),
      res: (
        NextResponse.next({ request: { headers: heads } }) as unknown as GetServerSidePropsContext["res"]
      )
    };

    // Force RSC calls to use getServerSideProps path
    return (
      auth(context)
    );
  }
};

export {
  handlers,
  signOut,
  signIn,
  auth_ as auth
};
```

ThangHuuVu (commented):
Thanks for the detailed PR and description. In my opinion, this is not the direction that we want though. I would prefer to alter the function's parameter to support passing `headers`, but not constructing the request object again without all the other `context` since the RSC world discourages having a global `context` from my understanding.

PR #12733, opened 2025-03-05
chore(docs): Fix syntax error in RBAC guide

### LINKED ISSUE #12462: Improve Role Based Access Control Page in Guides Section

The code must be explained more detail, in the Getting a role part. 

The code: 
return { role: profile.role ?? "user", ... } 
(The ... part shows an error said "Expression Expected)

Give more details explanation about Role Based Access Control, especially using with database (ex: MongoDb)

https://authjs.dev/guides/role-based-access-control

benhovinga:
## ☕️ Reasoning

Moved the Ellipsis ( ... ) onto a separate line as a comment. The Ellipsis was throwing "Expression Expected" error. I assume this was used to indicate to the developer that more properties can be added here. Moving it inside of a comment performs the same function without throwing any errors.

**Before**

```typescript
export const { handlers, auth } = NextAuth({
  providers: [
    Google({
      profile(profile) {
        return { role: profile.role ?? "user", ... }
      },
    })
  ],
})
```

**After**

```typescript
export const { handlers, auth } = NextAuth({
  providers: [
    Google({
      profile(profile) {
        return {
          role: profile.role ?? "user",
          // ...
        }
      },
    })
  ],
})
```

This fix has been applied to all frameworks.

## 🧢 Checklist

- [X] Documentation
- [x] Tests - N/A
- [X] Ready to be merged

## 🎫 Affected issues

Fixes: #12462

## 📌 Resources


tgeorge73:
I believe the proposed documentation update is incorrect. The `...` is not an ellipse, but the `spread operator`. (The error occurs in typescript because the spread operator is missing the object to spread, in this case `profile`).

So removing the ", ..." as proposed causes the user to be created in the database with ONLY the properties in the function.  changing "..." to "...profile" causes the properties from the profile to be merged with the values set manually in the function.

One other note, specific to the spread operator, is the order of statements with respect to the spread operator.  When the spread operator is last, the properties of the profile (object being spread) from the IdP will override and matching property statement manually set -- this seems safest as a pattern such that the IdP identity information is most accurate.  If the spread operator is first, then any manually set properties will override matching properties returned from the IdP profile.  (This would only be appropriate if there was a valid reason to change an IdP provided value -- which arguably should not be done.)  While not applicable to the role property because it does not exist in the profile, this is an important note in my opinion as it can be a source of confusion to newer devs (maybe link out to spread operator documentation here?)

benhovinga:
Ah, this makes more sense. Usually when I see `...` on it's own I assume this is a only placeholder. It is clear now that `profile` was missing after it to make it a spread operator.

So this should be the correct version

```typescript
export const { handlers, auth } = NextAuth({
  providers: [
    Google({
      profile(profile) {
        return { role: profile.role ?? "user", ...profile }
      },
    })
  ],
})
```

When I get on my dev machine I will update the PR.

ndom91 (approved):
Thanks! 

**EDIT**: Ah just read the comments. Not sure if it was meant as a spread operator where the author just forgot the `profile` part. But that should ideally be there, so let's go that route. 

If you update it I can get this merged. Sorry for the delay!

benhovinga:
Corrected the spread operator to `...profile`

PR #12818, opened 2025-03-26
Update Apple provider documentation

mvarrieur:
## ☕️ Reasoning

Update Apple provider documentation to recommend using Auth CLI for setup to make the process easier.

## 🧢 Checklist

- [x] Documentation
- N/A Tests
- [x] Ready to be merged

## 📌 Resources

### DIFF
--- a/docs/pages/getting-started/providers/apple.mdx
+++ b/docs/pages/getting-started/providers/apple.mdx
@@ -14,6 +14,9 @@ import { Code } from "@/components/Code"
 
 ## Setup
 
+### Auth CLI
+Using `npx auth add apple` makes setting up the Apple provider a much simpler process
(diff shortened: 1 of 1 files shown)


ndom91:
Thanks, useful tip. But maybe let's put it somewhere higher-level in the docs as it applies to every provider, right.

mvarrieur:
@ndom91 I am not sure about other providers but I think with most of them you just need to provide the normal secrets in the .env, with Apple it is signing and creating a JWT that needs to be stored in your secrets which the CLI tool helps with. 

I'm good to close this if you'd like but we were just smashing our heads against the wall too many times before discovering the CLI helps immensely with Apple specifically.

Issue #12767, opened 2025-03-12
Possible issue in the migration-to-v5 guide


kossidts:
### What is the improvement or update you wish to see?

Hello, maybe I am missing something but shouldn't a (context) provider be at least one layer/level above the consumer?

Looking at [migrating-to-v5](https://authjs.dev/getting-started/migrating-to-v5) the section (Step 4 Details -> Client component (App)) does not look right. 

```ts filename="components/clientComponent.tsx"
 'use client';
 
 import { useSession, SessionProvider } from 'next-auth/react';
 
 const ClientComponent = () => {
   const session = useSession();
 
   return (
     <SessionProvider>
       <p>Welcome {session?.user?.name}</p>
     </SessionProvider>
   )
 }
 ```

### Is there any context that might help us understand?

I think it should be something like the following instead. 

```ts
import { SessionProvider } from 'next-auth/react';
import { ClientComponent } from '@/components';
 
 const ParentComponent = () => {
 
   return (
     <SessionProvider>
       <ClientComponent />
     </SessionProvider>
   )
 }
```

```ts
'use client';
 
 import { useSession } from 'next-auth/react';
 
 const ClientComponent = () => {
   const session = useSession();
 
   return (
       <p>Welcome {session?.user?.name}</p>
   )
 }```

### Does the docs page already exist? Please link to it.

https://authjs.dev/getting-started/migrating-to-v5

PR #12872, opened 2025-04-14
chore(docs): update `migrating-to-v5` docs

halvaradop:
## ☕️ Reasoning
This pull request updates the documentation in the `migrating-to-v5` section by expanding and improving the existing content. A new `Route Handlers` tab was added to the details section, providing guidance on how to use NextAuth.js v5 with Route Handlers. Additionally, the `Client Component` section was revised to address multiple issues raised by the community regarding unclear documentation.

The `format:write` script was also run to ensure consistent formatting across the documentation files.

It’s been a while since my last contribution to the project, so I’m glad to be back and help improve the docs.

### Key Changes

- Updated the `Client Component` section
- Added a new `Route Handlers` section
- Ran the `format:write` script

## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 🎫 Affected issues
#12767

## 📌 Resources


ThangHuuVu (approved):

docs/pages/getting-started/migrating-to-v5.mdx
```diff
@@ -170,6 +180,31 @@ Check out additional [Middleware docs](/getting-started/session-management/prote
 </Tabs.Tab>
 <Tabs.Tab>
 
+As of `NextAuth.js v5`, you can use the `auth()` function inside [Route Handlers](https://nextjs.org/docs/app/building-your-application/routing/route-handlers) to access the session.
```
ThangHuuVu, line 183: in Route handler, we can just use the `req.auth` like the other part of the docs here: https://authjs.dev/reference/nextjs#in-route-handlers
I think we can just link to this page here, HDYT?

ThangHuuVu (changes requested):

docs/pages/getting-started/migrating-to-v5.mdx
halvaradop: You're right, however, this method of accessing the session isn't working as expected. It's been reported by users that `req.auth` often returns `null`. Because of this, I’ve chosen to access the session directly using the `auth()` function instead of the wrapper.

## Test Case

I created a demo using the `Credentials` provider. After logging in, I set up an endpoint to access the session via the wrapper. During testing, all values passed through the `Credentials` provider are authenticated correctly. The session is properly stored, and cookies are created with the `authjs` prefix.

When accessing the session using the `auth()` function directly—or via `useSession()`—everything works as expected. However, accessing the session through `request.auth` consistently returns `null`.

Tell me what you think, and I’ll make the necessary changes.


```ts
export const GET = auth((request) => {
    const session = request.auth
    if(!session) {
        return NextResponse.json(
            { message: "No session" }, 
            { status: 401 }
        )
    }
    return NextResponse.json({ 
        auth: session, 
        message: "GET /api/session"
    }, { status: 200 }) 
})
```

PR #12903, opened 2025-04-24
Add last.fm provider

michaelg11821:
## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


packages/core/src/providers/lastfm.ts
```diff
@@ -0,0 +1,189 @@
...
+ * ### Setup
+ *
+ * #### Callback URL
+ * ```
+ * https://example.com/api/auth/callback/lastfm
+ * ```
+ *
+ * #### Configuration
+ *```ts
+ * import { Auth } from "@auth/core"
```
ndom91, line 83: Quick note here - as a user you should never have to interact with or install `@auth/core`. 

So these imports should both come from the client lib the user is using, i.e. `next-auth` for next.js or `@auth/sveltekit` for sveltekit, etc.

ndom91 (changes requested):
Small nit, but otherwise looks pretty good overall 🥳

PR #12919, opened 2025-04-28
Ensure trigger is set in JWT callback when update() is called

mbdevntc:
## ☕️ Reasoning
When update() from the useSession() hook is called without any parameters, a GET request is sent (making its behavior similar to getSession()), instead of a POST request.
This prevents the trigger property (in JWT callback) from being set and used to determine whether the user session data needs to be re-fetched.

This is important when session data updates are handled inside the JWT callback and rely on the trigger value to determine whether user data should be re-fetched.

### DIFF
--- a/packages/next-auth/src/react.tsx
+++ b/packages/next-auth/src/react.tsx
@@ -507,13 +507,16 @@ export function SessionProvider(props: SessionProviderProps) {
       async update(data: any) {
         if (loading) return
         setLoading(true)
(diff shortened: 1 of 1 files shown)


dBianchii:
Would really love to see this merged. It's affecting our team today

dBianchii:
I applied this exact patch in my 5.0.0-beta.29 repo, and it worked for me.

PR #12947, opened 2025-05-09
Fix undefined providerId

Regloom:
## ☕️ Reasoning

Client-side auth with no `providerId` param results in error. Sample code: 
```js
import { signIn, signOut } from "@auth/sveltekit/client"
<button onclick={() => signIn()}>Sign In</button>
```

It's caused by [parseActionAndProviderId](https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/lib/utils/web.ts#L136) function which returns providerId = "undefined" (as a string). This breaks the following check [auth/core/src/lib/utils/providers.ts](https://github.com/nextauthjs/next-auth/blob/main/packages/core/src/lib/utils/providers.ts#L66)
```js
export default function parseProviders(params) {
...
    if (providerId && !provider) {
        const availableProviders = providers.map((p) => p.id).join(", ");
        throw new Error(`Provider with id "${providerId}" not found. Available providers: [${availableProviders}].`);
    }
    return { providers, provider };
}
```

## 🧢 Checklist

- [x] Documentation
- [x] Tests
- [X] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/utils/web.ts
+++ b/packages/core/src/lib/utils/web.ts
@@ -144,5 +144,8 @@ export function parseActionAndProviderId(
   )
     throw new UnknownAction(`Cannot parse action at ${pathname}`)
 
-  return { action, providerId }
+  return { 
+    action, 
+    providerId: providerId == "undefined" ? undefined : providerId 
+  }
 }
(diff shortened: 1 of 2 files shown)


ThangHuuVu:
thanks for the fix, could you add a test for it?

Regloom:
Ok, tests added.

ThangHuuVu (approved):

PR #12949, opened 2025-05-10
add missing step for credntials provider

elbasel-404:
- [x ] Documentation
- [ ] Tests
- [ ] Ready to be merged

Credntials provider needs the `strategy` inside the `session` parameter to be set to `jwt', however this is not mentioned in the docs.


halvaradop:
There's no need to manually set the session strategy when using the Credentials provider—auth.js sets it by default.
You can find more details in the official documentation: [Session Strategies](https://authjs.dev/concepts/session-strategies)

elbasel-404:
@halvaradop my testing summarized that it doesn't work without setting the strategy, can you please double check?

PR #12616, opened 2025-02-06
fix(provider): Microsoft Entra ID

benhovinga:
## ☕️ Reasoning

This PR resolves several issues with the Microsoft Entra ID provider.

### Fix Documentation

Inconsistencies between the use of `issuer` and `tenantId`. This implementation of the provider doesn't directly use `tenantId` however it is used within part of the `issuer` URI string. This PR clarifies how to use the `issuer` parameter in both the [Provider Docs](https://authjs.dev/getting-started/providers/microsoft-entra-id) and [API Reference](https://authjs.dev/reference/core/providers/microsoft-entra-id).

Also fixed some inconsistencies between frameworks

### Fix `MicrosoftEntraIDProfile` interface

The `MicrosoftEntraIDProfile` interface was missing a lot of the claims returned from the provider server. This PR adds all `id_token` claims.

### Fix default issuer

If the environment variable `AUTH_MICROSOFT_ENTRA_ID_ISSUER` is configured but the `issuer` parameter is not set, the default issuer overwrites the environment variable. This PR checks if the environment variable is configured first before falling back to the default issuer. 

Now the provider "can" be used without any configuration, like this.

```typescript:name=auth.ts
import NextAuth from 'next-auth';
import MicrosoftEntraID from 'next-auth/providers/microsoft-entra-id';

export const { handlers, auth, signIn, signOut } = NextAuth({
  providers: [MicrosoftEntraID]
});
```

```env:.env.local
AUTH_MICROSOFT_ENTRA_ID_ID="<client id>"
AUTH_MICROSOFT_ENTRA_ID_SECRET="<client secret>"
AUTH_MICROSOFT_ENTRA_ID_ISSUER="https://login.microsoftonline.com/<tenant id>/v2.0"
```

### Finish Provider Implementation

Add the provider to the [OAuth](https://authjs.dev/getting-started/authentication/oauth) providers search and the [Issues Template](https://github.com/nextauthjs/next-auth/blob/main/.github/ISSUE_TEMPLATE/2_bug_provider.yml) providers dropdown.

## 🧢 Checklist

- [X] Documentation
- [x] Tests
- [X] Ready to be merged

## 🎫 Affected issues

Fixes: #12314, #12193, #12195 
Corrects: #12612 

## 📌 Resources

- [Security guidelines](https://github.com/nextauthjs/.github/blob/main/SECURITY.md)
- [Contributing guidelines](https://github.com/nextauthjs/.github/blob/main/CONTRIBUTING.md)
[...]

PR #12967, opened 2025-05-13
feat(providers): add Feishu OAuth provider integration

James-Lam:
## 🧢 Checklist

- [x] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources


ThangHuuVu:
Ref to https://github.com/nextauthjs/next-auth/pull/12616/files, could you help to add a few more changes on the same files with that PR?

packages/core/src/providers/feishu.ts
```diff
@@ -0,0 +1,269 @@
...
+    id: "feishu",
+    name: "Feishu",
+    type: "oauth",
+    authorization: {
+      url: "https://accounts.feishu.cn/open-apis/authen/v1/authorize",
+      params: {
+        client_id: options.clientId,
+        response_type: "code",
+        state: "RANDOMSTRING",
+        scope: "",
```
halvaradop, line 170: If `state` and `scope` aren’t being set and your implementation doesn’t require them, feel free to remove them.

packages/core/src/providers/feishu.ts
```diff
@@ -0,0 +1,269 @@
...
+          avatar_big: data.data.avatar_big,
+          open_id: data.data.open_id,
+          union_id: data.data.union_id,
+          email: data.data.email,
+          enterprise_email: data.data.enterprise_email,
+          user_id: data.data.user_id,
+          mobile: data.data.mobile,
+          tenant_key: data.data.tenant_key,
+          employee_no: data.data.employee_no,
+        } satisfies FeishuProfile;
```
halvaradop, line 246: There's no need to pass all of these properties manually—you can simply return `data.data`, since its structure matches the expected response format.
The same applies to the `profile` method, where `data.data` already contains the necessary fields

halvaradop (commented):
Great work

James-Lam:
> Ref to https://github.com/nextauthjs/next-auth/pull/12616/files, could you help to add a few more changes on the same files with that PR?

@ThangHuuVu  Done! ✅ I've updated the same files. Ready for review!

PR #13166, opened 2025-08-09
Changed basePath in auth.ts to match the official documentation

jeremyko:
(Next.js v15.4.6)
The official documentation (https://authjs.dev/guides/configuring-github) suggests setting the Callback URL to "http://localhost:3000/api/auth/callback/github", but the official example (apps/examples/nextjs) doesn't include an api folder. need to change the basePath in auth.ts to match the documentation (/api/auth). Also, when running the prettier command, if the path contains "[...]", the following error occurs:

    [error] No files matching the pattern were found: "'apps/examples/nextjs/app/api/auth/[...nextauth]/route.ts'".

So, I added a path exception to lefthook.yml. There may be another way.

## ☕️ Reasoning

Modified the Next.js example to match the official documentation.

## 🧢 Checklist

- [X] Documentation
- [X] Tests
- [X] Ready to be merged

## 🎫 Affected issues
N/A

## 📌 Resources


apps/examples/nextjs/auth.ts
```diff
@@ -100,7 +100,7 @@ export const { handlers, auth, signIn, signOut } = NextAuth({
     WorkOS({ connection: process.env.AUTH_WORKOS_CONNECTION! }),
     Zoom,
   ],
-  basePath: "/auth",
+  basePath: "/api/auth",
```
halvaradop, line 103: The basePath option was configured to point to auth/[...nextauth] instead of api/auth/[...nextauth]. As you noted, the configuration is in auth/[...nextauth]/route.ts, not api/auth/[...nextauth].

https://github.com/nextauthjs/next-auth/blob/39dd3b92de194c1a835f2d87631f4deb9d9fdf65/apps/examples/nextjs/app/auth/%5B...nextauth%5D/route.ts#L1-L2

​The NextAuth.js documentation recommends using the api/auth structure. However, this isn't a strict requirement, as the basePath option was introduced specifically to allow for custom authentication routes. The user isn't obligated to configure their app with /api/auth.

jeremyko: Oh, then I think the official documentation (https://authjs.dev/guides/configuring-github) needs to be revised. To avoid confusion, the documentation should explicitly specify "http://localhost:3000/auth/callback/github." We develop our code by referencing both the official documentation and the examples. It would be best if one of the two were revised to eliminate any inconsistencies.

PR #13253, opened 2025-09-27
(Fix): broken link of middleware setup

coderrshyam:
This pull request makes a small documentation update to the Next.js middleware guide link in the migration guide. The change ensures that the link points to the correct external URL.

* Updated the Next.js middleware guide link in `migrate-to-better-auth.mdx` to correct the url.


coderrshyam:
@Bekacru would you merge it, as it create problems for user who want to migrate to better auth.

PR #13265, opened 2025-10-01
chore(qwik): add peerDeps for qwik

wmertens:
## ☕️ Reasoning

without this, it isn't guaranteed to have access to the qwik libararies, which is certainly the case when using the overrides for qwik v2.

## 🧢 Checklist

- [x] Ready to be merged


himself65 (approved):
LGTM

PR #13376, opened 2026-02-06
Move local variables inside `assertConfig`

edoardocavazza:
## ☕️ Reasoning

Moved local variables inside `assertConfig` in order to prevent conflicts between two providers.

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

## 📌 Resources

### DIFF
--- a/packages/core/src/lib/utils/assert.ts
+++ b/packages/core/src/lib/utils/assert.ts
@@ -44,10 +44,6 @@ function isSemverString(version: string): version is SemverString {
   return /^v\d+(?:\.\d+){0,2}$/.test(version)
 }
 
-let hasCredentials = false
-let hasEmail = false
-let hasWebAuthn = false
-
 const emailMethods: (keyof Adapter)[] = [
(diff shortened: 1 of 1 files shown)


HardikDewra (commented):
Clean fix. Moving `hasCredentials`, `hasEmail`, and `hasWebAuthn` from module-level `let` declarations into the `assertConfig` function body eliminates stale state between calls.

With the variables at module level, the first call to `assertConfig` that encounters a credentials provider sets `hasCredentials = true`, and that value persists across all subsequent calls. In a serverless environment where the module might be reused across requests, this means:

- A request that configures only email providers would still see `hasCredentials = true` from a previous request's config
- The assertion logic would produce incorrect results based on ghost state from prior invocations

Moving them inside the function ensures each call starts with a clean slate. Simple, correct, no behavioral change for single-call scenarios.

No concerns here - this is a straightforward bug fix for a real state leakage issue.

PR #13385, opened 2026-02-23
Migrations fail silently on Cloudflare Workers

alakazamn:
## ☕️ Reasoning

Each migration statement is not being properly awaited, so awaiting the up() method does nothing.

Not sure how exactly it's desired to handle errors.. but the existing does not work and fails silently, leaving users to either manually migrate or do as I have done. 

## 🧢 Checklist

- [ ] Documentation
- [ ] Tests
- [ ] Ready to be merged

### DIFF
--- a/packages/adapter-d1/src/migrations.ts
+++ b/packages/adapter-d1/src/migrations.ts
@@ -49,13 +49,7 @@ export const down = [
 ]
 
 async function up(db: D1Database) {
-  upSQLStatements.forEach(async (sql) => {
-    try {
-      await db.prepare(sql).run()
-    } catch (e: any) {
-      console.error(e.cause?.message, e.message)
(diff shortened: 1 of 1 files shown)


ndom91:
Thanks let me take a look

alakazamn:
Thanks-- I believe the issue comes down to:

In a normal environment you might not technically need to await every migration; though you lose the ability to wait until they're all done, they'd all execute. 

CF workers heavily optimizes for responses, so if you move on and deliver a response to a user before the async statements complete execution, it ends execution prematurely. 

One question is-- if execution of these migrations is not idempotent or unable to run concurrently, then instead of Promise.all, we'd want to use a for loop, but not`.forEach`. However, the commit seemed to work fine for me.

Issue #12354, opened 2024-12-06
client session keeps authenticated even if user signout by signout methods.


takakikasuga:
### Environment

```
(base) ➜  web git:(development) ✗ npx envinfo --system --binaries --browsers --npmPackages "{next,react,next-auth,@auth/*}"
Need to install the following packages:
envinfo@7.14.0
Ok to proceed? (y) y


  System:
    OS: macOS 14.6.1
    CPU: (10) x64 Apple M1 Max
    Memory: 12.75 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.17.0 - ~/.volta/tools/image/node/20.17.0/bin/node
    Yarn: 1.22.22 - ~/.volta/tools/image/yarn/1.22.22/bin/yarn
    npm: 10.8.2 - ~/.volta/tools/image/node/20.17.0/bin/npm
    pnpm: 9.14.1 - ~/.volta/bin/pnpm
  Browsers:
    Chrome: 131.0.6778.109
    Safari: 17.6
```

### Reproduction URL

https://github.com/nextauthjs/next-auth-example

### Describe the issue

client session keeps authenticated even if user signout by signout methods.
https://next-auth-example.vercel.app/client-example

https://github.com/user-attachments/assets/94873312-94dd-4060-97e0-5c6c8a8d8fc6

### How to reproduce

1. user login by next-auth
2. user logout by signout methods provided by next-auth

### Expected behavior

after signout, value of useSession hooks and auth methods is the same!!

Issue #3995, opened 2022-02-17
signOut doesn't clear session


hyounoo:
### Environment

System:
    OS: macOS 11.3.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 4.29 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.3.0 - /usr/local/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 8.3.0 - /usr/local/bin/npm
    Watchman: 2021.06.07.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 98.0.4758.102
    Safari: 14.1
    Safari Technology Preview: 14.2
  npmPackages:
    next: 12.0.8 => 12.0.8 
    next-auth: ^4.2.1 => 4.2.1 
    react: 17.0.2 => 17.0.2 

### Reproduction URL

none

### Describe the issue

Using federatedlogout code from #3938, I was able to signout from my idp and it redirects me to /signedOut page as below.

`
import React, { useEffect } from 'react'
import { signOut } from 'next-auth/react'
import { useRouter } from 'next/router'

const signedOut = () => {
  const router = useRouter()

  console.log('visiting signedOut')
  useEffect(() => {
    signOut({ redirect: false, callbackUrl: `${process.env.NEXTAUTH_URL}` })
      .then((data) => {
        console.log('data: ', data)
        router.push(data.url)
      })
      .catch((error: any) => {
        console.debug('error: ', error)
      })
  }, [])

  return <div className="m-10">Signing you out...</div>
}

export default signedOut

`

But whatever I tried, it occationally doesn't clear the session and the use is still signed in.
It sometimes clear the session but most of time it doesn't.

I've tried signOut() with or without redirect: false but the result were the same.

is there any way to clear out the session manually? 

### How to reproduce

Using below federatedlogout.ts code and signedOut.tsx

federated-logout.ts
`
import { getToken } from 'next-auth/jwt'
import type { NextApiRequest, NextApiResponse } from 'next'

const secret = process.env.NEXTAUTH_SECRET

export default async function federatedLogout(req: NextApiRequest, res: NextApiResponse) {
  try {
    const token = await getToken({ req, secret })
    console.info('token: ', token)
    if (!token) {
      console.warn('No JWT token found when calling /federated-logout endpoint')
[...]

Issue #3991, opened 2022-02-17
Session still present in API route check after client side signOut


kramer99:
### Question 💬

I expect there's an option somewhere I'm missing, but I haven't managed to find it after scouring the documentation and closed issues for the past two hours.

I have client side signOut working fine via:
```
import { signOut, useSession } from 'next-auth/react';

const Header = () => {
  const { data: session } = useSession();

  const onClick: any = signOut;

  return (
    <Navbar bg="dark" variant="dark" className="page-header sticky-top">
...
      <Button className="btn btn-secondary btn-sm" onClick={onClick}>
        Logout
      </Button>
      <Notification />
    </Navbar>
  );
};
```
I'm also trying to protect an API route using the guidance here: https://next-auth.js.org/tutorials/securing-pages-and-api-routes#using-getsession

I'm using a wrapper function, but it's the same result if I put the session check directly in the API route:

```
export default function withAuth(handler: NextApiHandler) {
  return async (req: NextApiRequest, res: NextApiResponse) => {
    const session = await getSession({ req });
    // const s = await getServerSession({ req, res }, options);
    // console.log(s);
    if (session) {
      // Signed in
      console.log('Session', JSON.stringify(session, null, 2));
      await handler(req, res);
    } else {
      // Not Signed in
      res.status(401);
    }
    res.end();
  };
}

```

Now, what I'm seeing is that on signing out, my pages are protected and redirect to the signin url, as expected, however, calling the API route via Postman - without passing any header cookies - the API route still sees an active session.

I'd expect that after signout, the session being logged above should be null.

Probably not important, but for completeness, here is my `[...nextauth].ts`
```
import type { NextApiRequest, NextApiResponse } from 'next';
import NextAuth from 'next-auth';
import OktaProvider from 'next-auth/providers/okta';

const options = {
  providers: [
    OktaProvider({
      clientId: process.env.OKTA_CLIENTID || '',
      clientSecret: '', // not required with PKCE
      issuer: process.env.OKTA_ISSUER,
      checks: ['pkce', 'state'],
      client: {
        token_endpoint_auth_method: 'none'
      }
    })
  ],
  secret: ...
};

export default (req: NextApiRequest, res: NextApiResponse) =>
  NextAuth(req, res, options);
```

Thank you very much for any help.

### How to reproduce ☕️

See above.

### Contributing 🙌🏽

No, I am afraid I cannot help regarding this

balazsorban44:
[...]

PR #13465, opened 2026-07-18
fix(next-auth): prevent stale session fetches from resurrecting the session after signOut

ytoshiki:
## ☕️ Reasoning

While looking into why a user can stay logged in after `signOut()`, I found a
race between sign-out and the refetches `SessionProvider` fires on its own
(mount, window focus, polling, cross-tab broadcasts):

1. a `GET /api/auth/session` starts (poll, focus, ...)
2. `signOut()` completes and clears the client state
3. the GET from step 1 resolves last and puts the old session back

With the default JWT strategy it's worse than a UI glitch: the session
endpoint re-signs the token and sets a fresh rolling cookie on every valid
GET, so the late response can leave the browser holding a valid session
cookie after sign-out. Reload and you're still logged in. This matches
long-standing reports like #4612.

The fix keeps an AbortController on `__NEXTAUTH`:

- `signIn`/`signOut` (and the WebAuthn `signIn`) abort any in-flight session
  fetch before sending their POST, and once more after it returns
- a session response is only applied if its fetch wasn't aborted or
  superseded by a newer one (fresh controller per fetch, newest wins)
- `update()` goes through the same guard
- a failed sign-in replays the refetch it aborted, and an aborted initial
  fetch keeps the status at "loading" instead of flashing "unauthenticated"
  (that flash could trigger `useSession({ required: true })` redirects under
  StrictMode)

Two behavior changes worth calling out: a failed non-redirect `signIn` now
does one extra session GET (the replay), and `update()` resolves `null` when
it was superseded by an auth-state change.

What this can't fix from the client side: another tab's concurrent GET can
still get a rolling cookie, and JWT sessions can't be invalidated
server-side. Both are noted in code comments.

The tests reproduce the races deterministically by controlling fetch
resolution order — they fail without the fix. They're also the first React
client tests in this package, which is why the diff includes jsdom /
@testing-library/react / react-dom dev deps and a package-local vitest
config (the shared config aliases react to preact/compat, which breaks
@testing-library/react).

## 🧢 Checklist

- [ ] Documentation — not needed, JSDoc updated inline
- [x] Tests
- [x] Ready to be merged

## 🎫 Affected issues

Related: #4612, #12354, #3995, #3991

### DIFF
--- a/packages/next-auth/package.json
+++ b/packages/next-auth/package.json
@@ -72,8 +72,8 @@
     "build": "pnpm clean && pnpm providers && tsc",
     "clean": "rm -rf *.js *.d.ts* lib providers",
     "dev": "pnpm providers && tsc -w",
-    "test": "vitest run -c ../utils/vitest.config.ts",
-    "test:watch": "vitest -c ../utils/vitest.config.ts",
+    "test": "vitest run -c ./vitest.config.ts",
+    "test:watch": "vitest -c ./vitest.config.ts",
     "test:e2e": "playwright test",
     "providers": "node ../utils/scripts/providers"
   },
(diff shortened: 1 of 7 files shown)


goransh-walia:
Clean and minimal — appreciate that it stays scoped to the actual fix.

PR #13485, opened 2026-08-17
fix(deps): allow nodemailer 9 as a peer dependency

AdzerKI:
## ☕️ Reasoning

Every `nodemailer@8` release is covered by published advisories (SMTP command injection via `envelope.size` and via CRLF in the transport name, `List-*` header injection, `jsonTransport` bypassing `disableFileAccess`/`disableUrlAccess`, OAuth2 token fetch without TLS verification, message-level `raw` bypass). They are fixed in `nodemailer@9.0.1+`, and `npm audit` reports `nodemailer <=9.0.0` as high severity.

The peer range in `@auth/core`, `next-auth` and `@auth/sveltekit` is `^7.0.7 || ^8.0.5`, so an app that also uses nodemailer directly (a contact form, transactional mail) cannot install a fixed release without breaking peer resolution: `npm ci` fails with ERESOLVE and the only workarounds are `--legacy-peer-deps` or an override — i.e. keeping a vulnerable dependency or silencing the resolver.

Nodemailer 9's only breaking change is TLS certificate validation for remotely fetched content (attachment URLs, OAuth2 token endpoints, proxy CONNECT). The provider here calls `createTransport` and `sendMail` with inline `text`/`html` and reads `rejected`/`pending` from the result, and the transport type imports (`nodemailer/lib/*-transport`) are unchanged in 9.x, so nothing in this repo depends on the old behaviour.

## 🧢 Checklist

- [x] Documentation — not applicable, dependency range only
- [x] Tests — existing suite covers the nodemailer provider
- [ ] Ready to be merged

## 🎫 Affected issues

None open for this.

## 📌 Resources

- https://github.com/advisories/GHSA-c7w3-x93f-qmm8
- https://github.com/advisories/GHSA-vvjj-xcjg-gr5g
- https://github.com/advisories/GHSA-268h-hp4c-crq3
- https://github.com/advisories/GHSA-wqvq-jvpq-h66f
- https://github.com/advisories/GHSA-r7g4-qg5f-qqm2
- https://github.com/advisories/GHSA-p6gq-j5cr-w38f


AdzerKI:
Closing: we migrated the affected project off next-auth, so we no longer have a way to validate this change in practice. The underlying issue stands for other users — every nodemailer 8.x release is covered by published advisories, and the peer range keeps apps that also use nodemailer directly on a vulnerable line. Feel free to reopen or reuse the patch.

